PR #1, opened 2015-08-11
Add step to install babel globally to have access to babel-node

brentvatne:
I didn't have babel installed globally on my machine so `babel-node` couldn't be found when I tried to follow the example instructions, this just adds that step to the README

I used the verbose cli flags for npm so it's more clear to people who aren't so familiar with it what is going on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ The repository comes with an implementation of [TodoMVC](http://todomvc.com/). T
 ```
 git clone https://github.com/facebook/relay.git
 cd relay/examples/todo && npm install
+npm install --global babel
 npm start
 ```
 


alinz:
I don't like to have the babel install on global. It would be nice to have the code loads all its dependencies from local node_modules folder instead. I found this way more reliable and less error prone because of people might have different version of babel install.

vjeux:
Sorry about this, should be fixed now: https://github.com/facebook/relay/commits/master

brentvatne:
great :smile: thanks @vjeux

voideanvalue:
Fixed in #2. Thanks!

PR #2, opened 2015-08-11
add babel to devDependencies for examples

voideanvalue:
The examples use `babel-node` which is provided by the `babel` npm package.


steveluscher:
Thanks! https://github.com/facebook/relay-starter-kit will need this too.

voideanvalue:
facebook/relay-starter-kit#1 :)

PR #4, opened 2015-08-11
Typo in snippet?

bbirand:
The `<Image>` tag seems to use `user.profilePhoto.uri`, which is missing int he schema?


yungsters:
Thanks for fixing this! I think you have to sign the Facebook CLA, though.

voideanvalue:
Thank you!

PR #14, opened 2015-08-12
Prevent crash on browsers which String.prototype.repeat is not supported yet

sairion:
There are a few browsers which does not support `String.prototype.repeat`: http://kangax.github.io/compat-table/es6/#String.prototype_methods_String.prototype.repeat

My latest Safari crashes on example. I'm not entirely sure about directly importing babel-runtime in `src`, but at least it is cleaner than using for-loop :) (I hoped fbjs have something like this)

(edit: I think directly requiring `babel-runtime` did not work, so I've just desugared it.)

### DIFF
--- a/src/network-layer/default/RelayDefaultNetworkLayer.js
+++ b/src/network-layer/default/RelayDefaultNetworkLayer.js
@@ -157,6 +157,16 @@ function throwOnServerError(response: any): any  {
 }
 
 /**
+ * Repeats given string, mostly equivalent of `String.prototype.repeat`
+ */
+function repeat(str: string, times: number): string {
+  for (var result = '', i = 0; i < times; i++) {
+    result += str;
+  };
+  return result;
+}
+
+/**
  * Formats an error response from GraphQL server request.
  */
 function formatRequestErrors(
@@ -169,7 +179,8 @@ function formatRequestErrors(
   var queryLines = request.getQueryString().split('\n');
   return errors.map(({locations, message}, ii) => {
     var prefix = (ii + 1) + '. ';
-    var indent = ' '.repeat(prefix.length);
+
+    var indent = repeat(' ', prefix.length);
     return (
       prefix + message + '\n' +
       locations.map(({column, line}) => {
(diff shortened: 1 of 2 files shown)


yungsters:
Relay makes pervasive use of ES2015 functionality. This might be the result of an issue with how we're using `babel-runtime`.

steveluscher:
I'll look into this to see if there's a bug with the polyfill step.

steveluscher:
As @yungsters mentioned, we're using ES2015 quite heavily. I know now why `babel-runtime` isn't polyfilling `String::repeat()`. @sebmck explains:

> It doesn't alias prototype methods. You can never be sure that something is always the right type. So in `foo.repeat()` you can't tell that `foo` is always going to be a string (you could with flow types but that's not implemented), so it's an unsafe transformation. Although in that case you can easily tell since it's just a string literal, but it'd be weird to support one and not the other.

This has helped to open up a wider discussion: should Relay attempt to bring its own polyfills, or should we empower (and require) you to bring the polyfills appropriate for your project. [React does this](https://facebook.github.io/react/docs/working-with-the-browser.html#polyfills-needed-to-support-older-browsers), to a certain extent. cc/ @zpao

yungsters:
In the interest of moving fast and promoting ES2015 (after all, it is 2015 now), I personally think we should just do a better job of documenting that polyfills are necessary, and we should include them in our examples. Thoughts?

sairion:
(Sorry, I mistakenly closed and reopened)

I see the point. But in respect of shipping with shim, shimming ES5 is rather easy, and es5-shim/es5-sham do most of the job. Shimming ES2015+ requires more than just including some files in `<head />` (which is you people already know well), and the border line of transpiling and shimming is a bit blur IMHO (i.e. Transpiled and bundled `lib/*`, they already get shipped with `babel-runtime/core-js/[counterparts]`). Including whole `babel/polyfill` solves this issue, it would be overkill and problematic to some extent (it is just a personal opinion over core-js). I think dealing with Babel, like making Babel plugin, is the nicest option but surely will take some time.

Well, whether it is documenting 'polyfill needed here' or another option, I think it is fine. I just wanted to make people not disappointed right after running the examples :)

yungsters:
I totally agree about the fact that shipping with shims might be too imposing. For now, we'll 1) fix the examples to include shims, and 2) update our documentation with information about shimming and recommended shimming options.

sairion:
@yungsters  OK, then I'm going to close this and will send separate PR. Thanks!

PR #18, opened 2015-08-12
fix typo

mjw56:
I believe the provided parameters text field and the response text field should be matching. This PR matches those two text fields.


yungsters:
Thanks!

PR #22, opened 2015-08-12
Don't hard code the version on the website

zpao:
This makes it slightly easier to update the site when new versions are release.


zpao:
Note: I did it this way because I couldn't require the package data from within a file in website.

And here's what it looks like - definitely prevents moving past.
screen shot 2015-08-11 at 9 31 14 pm

yungsters:
Nice!

PR #23, opened 2015-08-12
Ensure cookies are passed on in fetch

devknoll:
By default, no cookies are passed along with `fetch`. It seems unintentional that we would want everyone to implement their own `Relay.NetworkLayer` to support authentication, so I think we should send them by default.


yungsters:
Sounds good to me. Thanks!

PR #25, opened 2015-08-12
Build index.js into lib

devknoll:
`index.js` needs to be built by gulp because it contains ES6 spreads.

c.c. @yungsters


devknoll:
Solved by c495d68eb6 instead

steveluscher:
…and inspired by your feedback. Thanks!

PR #33, opened 2015-08-12
Fix missing container and some missing parens in the docs

fson:
The LikeButton component in the Mutations guide seemed to be missing
the RelayContainer: the fragments were directly specified on the
component even though this doesn't seem to be supported.
Add the container for LikeButton.

Also add a bunch of missing closing parens for containers in the docs.


steveluscher:
Right you are; thanks!

PR #34, opened 2015-08-12
Remove double JSON.stringify calls around variables

mhart:
The POST bodies with these escaped strings are currently overly verbose, harder to debug and require extra parsing on both ends.

These bodies already have Content-Type application/json and express-graphql will only parse the variables param if it needs to (for example in the case of FormData): https://github.com/graphql/express-graphql/blob/9e6765f3c9e9e2f14da3c5fd2616afe73239f62b/src/index.js#L138-L146

I've signed the CLA and all tests pass, thanks!

### DIFF
--- a/src/network-layer/default/__tests__/RelayDefaultNetworkLayer-test.js
+++ b/src/network-layer/default/__tests__/RelayDefaultNetworkLayer-test.js
@@ -111,7 +111,7 @@ describe('RelayDefaultNetworkLayer', () => {
       });
       expect(body).toEqual(JSON.stringify({
         query: request.getQueryString(),
-        variables: JSON.stringify(variables),
+        variables: variables,
       }));
     });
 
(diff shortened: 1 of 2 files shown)


src/network-layer/default/__tests__/RelayDefaultNetworkLayer-test.js
```diff
@@ -111,7 +111,7 @@ describe('RelayDefaultNetworkLayer', () => {
       });
       expect(body).toEqual(JSON.stringify({
         query: request.getQueryString(),
-        variables: JSON.stringify(variables),
+        variables: variables,
```
yungsters: Could also have used short notation, but not a big deal. :)

```
expect(body).toEqual(JSON.stringify({
  query: request.getQueryString(),
  variables,
}));
```

yungsters:
Thanks again for discovering this! And thanks for making sure tests pass. :)

PR #38, opened 2015-08-12
Move React to peerDependencies

### LINKED ISSUE #13: react should be in peerDependencies
https://github.com/facebook/relay/blob/0f546869f549bd3939f7fe5744f233f1f7574401/package.json#L35 should be in peerDependencies, or else you'll get multiple copies of React in some cases and it's important that that doesn't happen.

zpao:
Fixes #13.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,8 +33,10 @@
     "babel-runtime": "5.8.20",
     "crc32": "^0.2.2",
     "fbjs": "0.1.0-alpha.7",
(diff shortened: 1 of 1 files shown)


zpao:
Note, we probably want to update docs a bit as well to make it clear that the app should be installing React.

steveluscher:
Documented in 9612c90.

PR #46, opened 2015-08-13
Update Guides-Containers.md

willpiers:


willpiers:
hadn't finished reading the docs

PR #49, opened 2015-08-13
Require matching query/fragment names

### LINKED ISSUE #20: "TODO: Fix this. It relies on the query and fragment names matching."
Well, that was a pretty annoying problem to try to hunt down... ;-)

If your queries don't seem to be running, it's because the query in your `Relay.Route` and the query in your `Relay.Container` need to have the same name, otherwise it isn't run.

devknoll:
Partially fixes #20 by warning unsuspecting users.

### DIFF
--- a/src/container/getRelayQueries.js
+++ b/src/container/getRelayQueries.js
@@ -46,11 +46,15 @@ function getRelayQueries(
     return cache[cacheKey];
   }
   var querySet = {};
+  var unmatchedNameMap = {};
+  Object.keys(route.queries).forEach(queryName => {
+    unmatchedNameMap[queryName] = true;
+  });
   Component.getFragmentNames().forEach(fragmentName => {
-    // TODO: Fix this. It relies on the query and fragment names matching.
     var queryName = fragmentName;
     var queryBuilder = route.queries[queryName];
     if (queryBuilder) {
+      delete unmatchedNameMap[queryName];
       var concreteQuery = buildRQL.Query(
         queryBuilder,
         Component,
@@ -75,9 +79,20 @@ function getRelayQueries(
           return;
         }
       }
+    } else {
+      unmatchedNameMap[queryName] = true;
     }
     querySet[fragmentName] = null;
   });
+  var unmatchedNames = Object.keys(unmatchedNameMap);
+  invariant(
+    unmatchedNames.length === 0,
+    'getRelayQueries: %s.fragments and %s.queries have unmatched fragment ' +
+    'names. Make sure that both define: %s',
+    Component.displayName,
+    route.name,
+    unmatchedNames.join(', ')
+  );
   cache[cacheKey] = querySet;
   return querySet;
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## master
 
 * Upgraded jest to 0.5 and switched Relay to use iojs v2+ only.
+* All fragments in a Relay.Container must now have a query with a matching
+  name in any Relay.Route that uses it and vice versa.
 
 ## 0.1.1 (August 14, 2015)
 
--- a/src/container/__tests__/getRelayQueries-test.js
+++ b/src/container/__tests__/getRelayQueries-test.js
@@ -20,6 +20,7 @@ describe('getRelayQueries', () => {
 
   var getRelayQueries;
 
+  var MockPageComponent;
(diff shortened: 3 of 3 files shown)


devknoll:
@yungsters @josephsavona any thoughts on this? Happy to add the corresponding warning to the route itself too.

devknoll:
Actually, strike that. It's easier to just handle both situations. I'll do that and rebase.

devknoll:
This now addresses https://github.com/facebook/relay/issues/20#issuecomment-130770234, raising an error if either a route or the container have any unmatched queries/fragments.

cc @yungsters @josephsavona

devknoll:
I'm not actually super convinced that a XOR is the way to go here.

In my case, I have a `<Header>` component and various page components (like `<Dashboard>`). Then I have a `DashboardRoute` that defines the data requirements to render `<Header>` AND `<Dashboard>`. Then I render both of them, pointing to the same route.

Using a XOR would mean that the `<Header>` component needs to know about all of the possible root queries for every other Route.

I've done it this way, because there isn't a way to compose fragments dynamically, like:

```js
var AppContainer = Relay.createContainer(App, {
  fragments: {
    header: Header.getFragment('header'),
    content: (Component) => Component.getFragment('content')
  }
});
```

And even if there was, this would limit my inner container to a single fragment/root query.

Thoughts?

devknoll:
Oops. I've removed the ES6 dependency and pushed a new version. Would love any thoughts + feedback on this!

yungsters:
How about doing this in `getRelayQueries` so that the warning is closer to where the dependent code exists?

src/container/RelayRootContainer.js
```diff
@@ -141,6 +141,28 @@ class RelayRootContainer extends React.Component {
...
+      fragmentNames.forEach(fragmentName => {
+        // XOR both sets to find unmatched names
+        if (unmatchedNameSet[fragmentName]) {
+          delete unmatchedNameSet[fragmentName];
+        } else {
+          unmatchedNameSet[fragmentName] = 1;
+        }
+      });
+      var unmatchedNames = Object.keys(unmatchedNameSet);
+      invariant(
```
yungsters: If we move this into `getRelayQueries` and can make this check cheaper, we can drop the `__DEV__` check. (In practice, there are not that many fragment names.)

Otherwise, we should convert this into a `warning`. Behavior of code should not be significantly altered by `__DEV__`, in order to guarantee that bugs that happen in production (where `__DEV__` is false) can be reproduced in development (where `__DEV__` is true). I'll add this to `CONTRIBUTING.md`.

src/container/RelayRootContainer.js
```diff
@@ -141,6 +141,28 @@ class RelayRootContainer extends React.Component {
...
+        unmatchedNames.length === 0,
+        'RelayRootContainer: %s.fragments and %s.queries have unmatched ' +
+        'field names. Make sure that both objects define: %s',
```
yungsters: `s/field/fragment/`

yungsters:
If you stick with `invariant`, add an entry under **master** in [`CHANGELOG.md`](https://github.com/facebook/relay/blob/master/CHANGELOG.md).

src/container/RelayRootContainer.js
devknoll: Are we dead set on requiring a Relay.Route to include all of a container's fragments, and not just vice versa? It seems a little overly strict to me, and not dealing with that would make the change simpler.

devknoll:
Ha, this was actually originally in `getRelayQueries`. I'll move it back.

devknoll:
Ok, should be up to your specs now :smile:

src/container/getRelayQueries.js
```diff
@@ -46,11 +46,15 @@ function getRelayQueries(
     return cache[cacheKey];
   }
   var querySet = {};
+  var unmatchedNameMap = {};
+  Object.keys(route.queries).forEach(queryName => {
+    unmatchedNameMap[queryName] = 1;
```
yungsters: Super-nit, assign to `true`?

devknoll:
:heavy_check_mark:

src/container/getRelayQueries.js
devknoll: :heavy_check_mark:

yungsters:
This is great! Thanks.

Issue #42, opened 2015-08-12
TypeError: undefined is not a function (evaluating ''\\ '.repeat(prefix.length)')


eventualbuddha:
The TodoMVC example from the README does not work in browsers that do not support `String#repeat`. At time of writing [the browsers that support it are Chrome and Firefox](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat#Browser_compatibility). This should either be done manually or the babel polyfill core-js should be included in the build that bundles Relay.

### Steps to Reproduce

1. Follow the [directions in the README](https://github.com/facebook/relay/blob/58b4e27bcfd3da53ef9917a912b2fffd92994246/README.md#example) for running an example.
2. Load [http://localhost:3000/](http://localhost:3000/) in a browser without support for `String#repeat`, i.e. Safari.

### Expected

The page should load the TodoMVC app.

### Actual

A runtime error occurs before the page can be properly loaded (`TypeError: undefined is not a function (evaluating ''\\ '.repeat(prefix.length)')`).

voideanvalue:
Thanks for reporting! @steveluscher is looking into this in #14.

josephsavona:
#50 added the babel polyfill to all the examples, which should resolve this issue. When using Relay in your own projects, be sure to include a polyfill.

PR #50, opened 2015-08-13
Include Babel polyfill to prevent crash

sairion:
This is discussd in #14 and reported in #42. /cc @yungsters


josephsavona:
@sairion thanks!

PR #52, opened 2015-08-13
Move express from devDependencies to dependencies

chentsulin:
`express` should be a `dependency` instead of devDependency because it's used by server.


josephsavona:
awesome, thanks!

PR #54, opened 2015-08-13
update engines to >= 0.10

chentsulin:
for node 0.11 up and iojs removing WARN message on npm install

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "test": "NODE_ENV=test jest"
   },
   "engines": {
-    "node": "0.10.x"
+    "node": ">= 0.10.0"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


wincent:
Thanks for the PR! Although I think we might want to keep the dependency pegged at 0.10.x until Jest's issues with Node are sorted out (see https://github.com/facebook/jest/issues/243).

zpao:
@wincent I think we still have this in React but we should probably remove it there too. Unfortunately that field is checked on end users. Ideally we could just check it on developers machines (actually, we probably could write a simple prepublish hook…).

It's not an enforced dependency and with iojs shuffling along it might be time to give it up.

cpojer:
We are now only compatible with iojs v2, see: #74

Issue #57, opened 2015-08-13
Add eslint to the project


knowbody:

josephsavona:
Let's discuss on the #58

PR #58, opened 2015-08-13
add eslint to the project

knowbody:
This is going to solve #57 

I thought it would be good to add eslint to the project at this stage. I have borrowed the eslint rules from react repo. I have fixed most of the linting errors, however in those I was not sure how you guys want to deal with refactoring I left there and added `/* eslint ... */` on top of the file to just generate the warning for now.

The most common issues are:  
- [`no-shadow`](http://eslint.org/docs/rules/no-shadow.html)
- [`no-unused-expressions`](http://eslint.org/docs/rules/no-unused-expressions.html)
- and also sometimes `warnings` and `invariants` are missing a error message

Please let me know how would you like to fix the above and I do those changes.


wincent:
This is awesome. I think we'll want to do something better than copy-pasta the React files as-is though.

On the subject of all the `shadow` warnings, it's such a common idiom in the codebase that I wonder whether we should just disable that lint (in some cases the readability actually suffers due to the awkward variable names).

josephsavona:
@wincent Agreed about disabling `shadow` warnings.

knowbody:
@wincent @josephsavona cool I'll disable it and fix the issues. I copied react's one just to start this off, thought that you might have some kind of guidelines at Facebook. Any other suggestions?

josephsavona:
Thinking about the implications of this more: all of the code passes our internal lint checks, and it will complicate syncing if the ESLint config here is more restrictive than our internal one. Can you relax the rules just enough to not require any changes to files in `src/`? E.g., allow for missing trailing commas, disable `no-unused-expressions` and `no-shadow`. 

Thanks again for collaborating on this!

zpao:
Next on my list for fbjs is to get our lint config in a good place to be shared and reused. For React we opted to be more restrictive than the rest of FB because we could be (we sync in, not out so it doesn't matter).

FWIW, we actually might not be passing internally, we should force a full check :)

`no-unused-expressions` is off internally (there were too many uses, but is one of those cases where we opted to be stricter in React) so we shouldn't change those here. Not sure of `no-shadow` off the top of my head.

knowbody:
@josephsavona ok, I'll do that and update the PR

knowbody:
@josephsavona updated the code, let me know what you think.  
  
travis is still failing while trying to run `npm install`

package.json
```diff
@@ -22,6 +22,7 @@
   "scripts": {
     "build": "[ $(ulimit -n) -lt 4096 ] && ulimit -n 4096; gulp",
     "prepublish": "npm run build",
+    "lint": "`npm bin`/eslint .",
```
zpao: This should just be `eslint .` (npm adds `node_modules/.bin` to the PATH)

package.json
```diff
@@ -38,10 +39,15 @@
...
   "devDependencies": {
     "babel-core": "5.8.21",
+    "babel-eslint": "^4.0.5",
     "babel-loader": "5.3.2",
     "babel-relay-plugin": "^0.1.2",
     "del": "^1.2.0",
     "envify": "^3.4.0",
+    "eslint": "^1.1.0",
+    "eslint-plugin-react": "^3.2.2",
+    "eslint-plugin-react-internal": "file:eslint-rules",
```
zpao: This actually requires npm v2+ so we'll need a `npm install -g npm@2` in the `.travis.yml` `before_install` hook.

josephsavona:
@knowbody I'm sorry if my previous comment wasn't clear. The addition of ESLint here should require no changes whatsoever to the `src/` directory. This will ensure that future syncs between OSS and our internal repository can proceed smoothly. Let's disable the `comma-dangle` rule (also it appears some variables were renamed - leftover fixes from `no-shadow`, maybe?

Regarding style: we do prefer trailing commas, but this is the one preference for which we don't also have an internal lint rule (and so there are some inconsistencies in the existing code).

knowbody:
@zpao thanks for the hints :+1: 

@josephsavona check it now please. I think I fixed all you asked for

knowbody:
is this one okay? @josephsavona

knowbody:
merged recent changes, not sure why it fails on the flow check

knowbody:
@josephsavona @wincent is anything blocking this to be merged?

src/__forks__/interface/RelayConnectionInterface.js
```diff
@@ -80,4 +80,4 @@ var RelayConnectionInterface = {
   },
 };
 
-module.exports = RelayConnectionInterface;
\ No newline at end of file
+module.exports = RelayConnectionInterface;
```
josephsavona: We cannot accept this PR if there are changes to the `src/` directory - rules should lenient enough to allow the code to pass as-is.

src/container/__tests__/RelayContainer-test.js
```diff
@@ -25,12 +25,9 @@ describe('RelayContainer', function() {
   var React;
   var ReactTestUtils;
   var Relay;
-  var RelayMetaRoute;
```
josephsavona: Again, we sincerely appreciate your effort in making these fixes, but we can't accept the PR unless there are zero changes to files in `src/`.

josephsavona:
@knowbody See comments above. Again, we *sincerely* appreciate all the effort you've put into this, but to simplify syncing code for now the configuration should consider the code valid as-is, and there should be no changes to the `src/` directory.

knowbody:
@josephsavona sure, sorry fixing all that now

knowbody:
@josephsavona should be good to go now

zpao:
You'll be able to drop a lot of this, apart from some finer tuning, once we pull in https://github.com/facebook/fbjs/pull/49 (might end up being better to end up abandoning this PR entirely and starting clean).

knowbody:
@zpao just seen the merge you've done with react. agree, let's close and I start the new PR with the rules from fbjs repo. How does it sound? @josephsavona

josephsavona:
closing and moving discussion to #202

PR #60, opened 2015-08-13
expose relay internals on devtools global

jaredly:
These are the things the relay devtools will need.

because who doesn't want this goodness?
image


josephsavona:
@jaredly awesome - can you add a link to the code for that screenshot? Those are internal modules and the API is likely to change.

jaredly:
getting that together now

jaredly:
here's the relevant code https://github.com/jaredly/react-devtools/blob/relay/plugins/Relay/backend.js
that interfaces w/ relay internals

PR #62, opened 2015-08-13
Update babel relay plugin

zpao:
Little bit of cleanup, little bit of fixing things.

I tested this by running the starter kit with `npm install --production` (well for this, `npm install --production` and then link it)


yungsters:
@wincent I think that was just me being n00b when I initially created the repository. (I was testing with `npm pack` and forgot to remove the file before checking in.)

PR #64, opened 2015-08-14
Configure babel-relay-plugin as a dependency for Jest

yungsters:
See discussion in f0ea07a7e87bfbfae26776cbb476d222802a3e67.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,6 @@
   "repository": "facebook/relay",
   "scripts": {
     "build": "[ $(ulimit -n) -lt 4096 ] && ulimit -n 4096; gulp",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -40,6 +39,7 @@
   "devDependencies": {
     "babel-core": "5.8.21",
     "babel-loader": "5.3.2",
+    "babel-relay-plugin": "^0.1.2",
```
zpao: Oh yea, I was going to do this a while ago (or rather, I was going to make this a `file:` reference since it lives here and we weren't public).

This is good and will survive if we move that out of here at some point.

zpao:
:shipit:

PR #67, opened 2015-08-14
Update Star Wars example schema to mimic real database API

cletusw:
A back-end implementation with a real database wouldn't have direct access to the full dataset. The relay-treasurehunt and todo examples are already written this way.

See graphql/graphql-js#120

### DIFF
--- a/examples/star-wars/data/starWarsDatabase.js
+++ b/examples/star-wars/data/starWarsDatabase.js
@@ -55,24 +55,19 @@ var executor = {
   name: 'Executor',
 };
 
-var nextShip = 9;
-export function getNewShipId() {
-  return '' + (nextShip++);
-}
-
-export var rebels = {
+var rebels = {
   id: '1',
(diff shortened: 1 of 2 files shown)


cletusw:
@dschafer @steveluscher

steveluscher:
I checked this out, and can confirm that it works! I'll leave it to @dschafer to push the button, since he might be interested in something like this for his [original example](https://github.com/graphql/graphql-relay-js/blob/master/src/__tests__/starWarsData.js).

cletusw:
@steveluscher Just curious -- how did you test the IntroduceShip mutation? I feel like the UI for that is missing...

cletusw:
Also, I've submitted graphql/graphql-relay-js#7 to do the same thing to the example in graphql-relay-js. It's a shame it has to be duplicated (and slightly different) here and there.

dschafer:
Thanks!

Issue #48, opened 2015-08-13
Q: beforeSend hooks in DefaultNetworkLayer


keithpitt:
Hey, so I'm playing around with Relay and I need to add an authentication token to all outgoing HTTP calls made by Relay. Is my only option to implement my own NetworkLayer?

Do you think it'd be useful to add a `beforeSend` type hook inside the layer so we can make changes to the requests before sending them off? In my case, I need to add a HTTP header.

Let me know if there's another place where I should be asking this question! Thanks!

Tlvenn:
More broadly the default network layer should be capable of handling interceptors that can manipulate both the request and the response.

yungsters:
The default network layer was intended to be a simple example for how to create your own. (In fact, I thought that not everyone would want to be using a `fetch` polyfill.) It really does not do all that much, so while I'm not opposed to making it more flexible, is there any particular reason you wouldn't want to simply create your own?

If there is something valuable in the default network layer that you really would not want to re-implement, maybe we should be exposing it in a way that custom network layers can re-use them.

If we think the best path is still to make the network layer more flexible, would it be sufficient to replace the second and third constructor arguments (`timeout` and `retryDelays`) with an fetch `init` argument that gets spread into the actual `init` that we pass (to either `fetch` or `fetchWithRetries`)?

devknoll:
I'm not so sure about interceptors for request/response (though these may be good for dev tools - but a dev tool could also probably just implement its own proxying network layer, right?) but I think that some way to add headers is probably the last thing the default network layer needs to truly become a default. It seems like a common enough use case.

`init` seems like it might be a good compromise.

yungsters:
> though these may be good for dev tools - but a dev tool could also probably just implement its own proxying network layer, right?

Yup, see #60. Alternatively, a man-in-the-middle network layer could be used for this:

```
Relay.injectNetworkLayer({
  ...defaultNetworkLayer,
  sendQueries(queryRequests) {
    // Sniff, sniff.
    return defaultNetworkLayer.sendQueries(queryRequests);
  },
  sendMutation(mutationRequest) {
    // Sniff, sniff.
    return defaultNetworkLayer.sendMutation(mutationRequest);
  },
});
```

yungsters:
We're trying to maintain good GitHub hygiene, so I'm going to close this out in favor of #70. Feel free to continue discussion there.

keithpitt:
That's great! That PR solves my problem. Thanks for that!! :heart:

PR #70, opened 2015-08-14
More Flexible Default Network Layer

yungsters:
Per discussion in #48, replaces the second and third arguments to `DefaultNetworkLayer` with a `fetch`-like `init` object. This will allow a higher degree of configuration (e.g. it allows setting custom headers).

### DIFF
--- a/src/network-layer/default/RelayDefaultNetworkLayer.js
+++ b/src/network-layer/default/RelayDefaultNetworkLayer.js
@@ -32,7 +32,7 @@ type GraphQLErrorLocation = {
 
 class RelayDefaultNetworkLayer {
   _uri: string;
-  _init: InitWithRetries;
+  _init: $FlowIssue; // InitWithRetries
 
   constructor(uri: string, init?: ?InitWithRetries) {
     this._uri = uri;
--- a/src/network-layer/default/RelayDefaultNetworkLayer.js
+++ b/src/network-layer/default/RelayDefaultNetworkLayer.js
@@ -19,6 +19,7 @@ import type RelayQueryRequest from 'RelayQueryRequest';
 
 var fetch = require('fetch');
 var fetchWithRetries = require('fetchWithRetries');
+import type {InitWithRetries} from 'fetchWithRetries';
 
 type GraphQLError = {
   message: string;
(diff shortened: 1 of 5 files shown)


yungsters:
I originally wanted to make the network layer throw if `body`, `headers['Content-Type']`, or `method` were supplied, but I figured I'd just keep things simple.

src/network-layer/default/RelayDefaultNetworkLayer.js
```diff
@@ -121,17 +121,22 @@ class RelayDefaultNetworkLayer {
         }
       }
       init = {
+        ...this._init,
         body: formData,
         method: 'POST',
       };
     } else {
       init = {
+        credentials: 'same-origin',
```
yungsters: Does this still make sense as a default? Or should it instead be pre-configured in [`src/__forks__/Relay.js`](https://github.com/facebook/relay/blob/master/src/__forks__/Relay.js#L20)?

src/network-layer/default/RelayDefaultNetworkLayer.js
```diff
@@ -143,15 +148,17 @@ class RelayDefaultNetworkLayer {
...
+      ...this._init,
       body: JSON.stringify({
         query: request.getQueryString(),
         variables: request.getVariables(),
       }),
-      credentials: 'same-origin',
-      fetchTimeout: this._timeout,
-      headers: {'Content-Type': 'application/json'},
+      headers: {
+        ...this._init,
```
devknoll: Should this be `...this._init.headers`?

yungsters: Yes, and I should have tests.

src/network-layer/default/RelayDefaultNetworkLayer.js
devknoll: I like it since it makes `fetch` act more like a regular `XMLHttpRequest`. That being said, since it's easily configurable now, I could swing either way.

devknoll: I guess the better question would be: when **shouldn't** cookies be sent? :wink:

yungsters: Hmm... maybe if you don't trust the server from which you're fetching data? But in that case, you would have had to knowingly set `mode: 'cors'` anyway, so you could just as easily set `credentials: null`.

It's a bit more dangerous of a default, but I could go either way.

src/network-layer/default/RelayDefaultNetworkLayer.js
```diff
@@ -34,13 +35,12 @@ var DEFAULT_RETRY_DELAYS = [1000, 3000];
 
 class RelayDefaultNetworkLayer {
   _uri: string;
-  _timeout: number;
+  _init: InitWithRetries;
   _retryDelays: Array<number>;
```
josephsavona: now unused?

josephsavona: `_retryDelays` is now used - remove?

src/network-layer/default/RelayDefaultNetworkLayer.js
devknoll: Yeah, I could go either way too. I *do* like safe defaults, though.

yungsters:
Note that this PR depends on changes in fbjs. I'll publish a new version of fbjs to go with this PR when this is ready.

@devknoll Per our discussion, I've removed `credentials: 'same-origin'` as a default.

PR #79, opened 2015-08-15
Correct profilePhoto fragment name in Guides-Containers

drabinowitz:
In Guides-Containers.md the fragment is named `profilePicture` but the key the React component accesses is at `user.profilePhoto`

profilePhoto is more consistently used in the doc so I went with that name.


wincent:
Thanks, @drabinowitz!

wincent:
I went ahead and merged this, as I think it's a step in the direction of consistency, but we could go further, as the component itself is still called `ProfilePicture` so we still have some mixed usage. If you want to go all the way I'd suggest making everything "picture".

PR #80, opened 2015-08-15
Remove reference to text in relay-treasurehunt mutate and get payload

drabinowitz:
The `CheckHidingSpotForTreasureMutation` contains a reference to `text` in the `mutateAndGetPayload` method. However this variable is not used and is `undefined`. It looks like it might be a relic from the todo example.


josephsavona:
cc @steveluscher

steveluscher:
:blush:

PR #83, opened 2015-08-15
Update QuickStart-Tutorial.md

h1nr1ch:
Without adapting js/components/App.js and js/routes/AppHomeRoute.js server start throws errors

### DIFF
--- a/docs/QuickStart-Tutorial.md
+++ b/docs/QuickStart-Tutorial.md
@@ -248,7 +248,6 @@ Before going any further, we need to serialize our executable schema to JSON for
 
 ```
 ./scripts/updateSchema.js
-npm start
 ```
 
 ## Writing the game


josephsavona:
It's necessary to call `npm start` at some point to run the application, so this will require some changes. We'd prefer either:
- documenting in the tutorial that errors are expected and they will be resolved in the succeeding steps
- or moving the `npm start` command further down in the tutorial to the point where no errors would be thrown if previous steps were done correctly.

My preference is probably the second option - cc @steveluscher

steveluscher:
Would you like to update this PR according to option 2, @h1nr1ch? That should be a good stopgap solution until I get a chance to rewrite the tutorial later this month (#6).

josephsavona:
I'm closing this - feel free to reopen, we'd be happy to accept a PR for option 2.

PR #85, opened 2015-08-16
export Game and HidingSpot class

tomatau:
Currently the nodeDefinitions object function checks the object instance against undefined as Game and HidingSpot are not exported.

This doesn't seem to make any difference the example apps functionality -- but should export these anyway to be thorough.


tomatau:
Have signed

josephsavona:
good find, thanks!

PR #90, opened 2015-08-17
Update variable invariant to newer syntax

devknoll:
`<width>` was the old variable syntax, `$width` is the new one. Lets make the invariant a little less confusing!


josephsavona:
@yungsters @yuzhi Perhaps we should add a `RelayGraphQLInterface` with a `formatVariable(name: string): string` method, so that we can format these strings appropriately?

yungsters:
@josephsavona We can just deal with the new syntax internally. The eventual plan is to move to the new syntax anyway, so we might as well get everyone used to it. :)

PR #113, opened 2015-08-18
Allow other strings as cursors

### LINKED ISSUE #89: Non-string cursors
We'd like to use a custom scalar type for cursors, but a spec says that it should be a String. I wonder if spec can be changed so that any scalar can be used as cursor.

dschafer:
In #89 itwas pointed out that using other scalars as cursors would be useful,
since clients might want to provide a custom scalar the serializes as a String.
Update the spec to ensure that is possible.

This change keeps the requirement that cursors serialize as strings, which
be strongenough to allow clients to handle different servers. It no longer
mandates a type, so servers can just use String if they prefer, or define
a custom scalar if they prefer.

Fixes #89.

### DIFF
--- a/website/graphql/Connections.md
+++ b/website/graphql/Connections.md
@@ -166,7 +166,11 @@ certain optimizations if this field returns an object that implements
 ### Cursor
 
 An "Edge Type" must contain a field called `cursor`. This field must return
-either a String or a Non-Null wrapper around a String.
+either a type that serializes as a String; this may be a String, a Non-Null
+wrapper around a String, or a custom scalar that serializes as a String.
+
+Whatever type this field returns will be referred to as the *cursor type*
+in the rest of this spec.
 
 The result of this field is considered opaque by Relay, but will be passed
 back to the server as described in the "Arguments" section below.
@@ -214,6 +218,7 @@ returns
         {
           "name": "cursor",
           "type": {
+            // This shows the cursor type as String!, other types are possible
             "name": null,
             "kind": "NON_NULL",
             "ofType": {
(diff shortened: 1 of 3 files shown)


website/graphql/Connections.md
```diff
@@ -166,7 +166,11 @@ certain optimizations if this field returns an object that implements
 ### Cursor
 
 An "Edge Type" must contain a field called `cursor`. This field must return
-either a String or a Non-Null wrapper around a String.
+either a type that serializes as a String; this may be a String, a Non-Null
+wrapper around a String, or a custom scalar that serializes as a String.
```
fson: I think it might not be clear from this wording whether Non-Null wrapper is allowed for a custom scalar, but I believe this is the intent anyway? I.e. it may be a String, a custom scalar that serializes as a String or a Non-Null wrapper around either.

dschafer: > I.e. it may be a String, a custom scalar that serializes as a String or a Non-Null wrapper around either.

That's correct, I'll update that to make that explicit.

fson: Thanks, it's very clear now.

josephsavona:
@dschafer this requires no client changes, correct?

website/graphql/Connections.md
```diff
@@ -166,7 +166,12 @@ certain optimizations if this field returns an object that implements
 ### Cursor
 
 An "Edge Type" must contain a field called `cursor`. This field must return
-either a String or a Non-Null wrapper around a String.
+either a type that serializes as a String; this may be a String, a Non-Null
```
devknoll: Unnecessary "either" now?

dschafer:
> @dschafer this requires no client changes, correct?

Correct.

website/graphql/Connections.md
dschafer: Yep, will fix.

dschafer:
Will fix the inline, then merge!

Issue #112, opened 2015-08-18
Support Root Fields w/o Node Mapping


yungsters:
## Problem

Relay currently only knows how to handle these three types of root fields:

- Root field with no arguments and queries a single node.
 - e.g. `empire` queries `{id: "123", ...}`
- Root field with one argument and queries a single node.
 - e.g. `ship(id: "456")` queries `{id: "456", ...}`
- Root field with one array argument and queries an array of nodes.
 - e.g. `ships(ids: ["456", "789"])` queries `[{id: "456", ...}, {id: "789", ...}]`

However, it has become clear that Relay needs to support any kind of root field. For example:

- Root field with multiple arguments.
- Root field with no arguments and queries an array of nodes.
- Root field that is a connection.

## Workaround

For now, the unsupported use cases can be implemented by creating a "global node", commonly called the `viewer`. You can then add arbitrary fields to `viewer`.

```
static fragments = {
  viewer: () => Relay.QL`
    fragment on Viewer {
      # Can return an array of users.
      users
    }
  `
};
```

## Rationale

Historically, GraphQL (as used internally at Facebook) did not have a root type or root fields. Instead, it had special "root calls" such as `node`, `nodes`, `me`, and `usernames`. Much of Relay was built on top of this assumption that the "root calls" return nodes.

For example, when we fetch `me` and get `{id: "123", ...}`, we record the association between the `me` root field and the node with ID of 123. Now, if we ever encounter another query for `me`, we can check our store for the node with ID of 123 and resolve the query without having to potentially re-fetch all of the fields we already have for `me`.

Another example, when we fetch `nodes(ids: ["123", "456"])`, we record the association between each argument and their respective response nodes. This allows us to fulfill queries for `node(id: "123")` and `node(id: "456")` even though we may never have independently queried for either before. (We would also be able to fulfill `me` if the association from above was established.)

## Next Steps

- [x] Support literal enum and input object values (#894)
- [x] Support arbitrary values in root calls (#895)
- [ ] Define a consistent method for annotating a root argument as "identifying". Identifying arguments have a 1:1 correspondence between the argument value and the `id` of the response. Currently all root arguments are assumed to be identifying.
- [ ] Allow arbitrary root calls with or without identifying arguments.

devknoll:
@yungsters @josephsavona is it possible that this change would also allow us to remove the Relay-specific Input Objects on queries? It seems super strange that GraphQL supports multiple arguments but then Relay forces them to all be wrapped up ;-)

[...]

Issue #104, opened 2015-08-18
Make route queries more like mutations


devknoll:
See discussion in #20.

I'll try to get this in the next day or so if nobody beats me to it :smile:

devknoll:
@yungsters Are there any internal users of `buildRQL.Query`? Seems like it's only being used for routes. I'll probably make this change there if not.

yungsters:
@devknoll Nope, `getRelayQueries` is the only caller of `buildRQL.Query`.

devknoll:
#115 with the changes, so I'm going to close this issue out.

PR #115, opened 2015-08-18
Streamline route queries

devknoll:
See #104 and #20. This might also conflict a little with #112

Just  wanted to throw this up before I go too far down any rabbit holes. The goal of this PR is to allow route queries to be defined like:
```js
viewer: () => Relay.QL`
  query { viewer }
`,
```
without having to specify `${Component.getFragment('viewer')}` mechanically.

The current change works (for some definition of working) but lacks tests and I'm still wrapping my head around the Relay internals. That being said:
* Does this seem like the correct approach? Any gotchas with variables I'm missing?
* How to we want to handle the two old formats? Should we add some deprecation notices now?
* Any other questions/comments?

.

༼ つ ◕_◕ ༽つ

### DIFF
--- a/src/container/__tests__/getRelayQueries-test.js
+++ b/src/container/__tests__/getRelayQueries-test.js
@@ -135,4 +135,16 @@ describe('getRelayQueries', () => {
       'fragment `Relay(MockPageComponent).fragments.last` to be defined.'
     );
   });
+
+  it('includes route parameters when building component fragment', () => {
+    var MockRoute = makeRoute();
+    var params = {id: '123'};
+    var route = new MockRoute(params);
+    MockPageContainer.getFragment = jest.genMockFunction();
+
+    getRelayQueries(MockPageContainer, route);
+
+    expect(MockPageContainer.getFragment.mock.calls.length).toBe(4);
+    expect(MockPageContainer.getFragment.mock.calls[0][1]).toBe(params);
+  });
 });
--- a/src/container/getRelayQueries.js
+++ b/src/container/getRelayQueries.js
@@ -61,7 +61,8 @@ function getRelayQueries(
       var concreteQuery = buildRQL.Query(
         queryBuilder,
         Component,
-        Object.keys(route.params)
+        Object.keys(route.params),
+        Component.getFragment(queryName, route.params)
       );
       invariant(
         concreteQuery !== undefined,
--- a/src/query/__tests__/buildRQL-test.js
+++ b/src/query/__tests__/buildRQL-test.js
@@ -36,11 +36,10 @@ describe('buildRQL', () => {
     });
     MockComponent = React.createClass({render});
     MockContainer = Relay.createContainer(MockComponent, {
-      fragments: {},
+      fragments: {
+        foo: () => Relay.QL`fragment on Node { name }`
+      },
     });
-    MockContainer.getQuery = jest.genMockFunction().mockReturnValue(
-      Relay.QL`fragment on Node { name }`
-    );
 
     jest.addMatchers(RelayTestUtils.matchers);
   });
@@ -127,7 +126,7 @@ describe('buildRQL', () => {
         query {
           node(id:$id) {
             id,
-            ${Component.getQuery('foo')}
+            ${Component.getFragment('foo')}
           }
         }
       `;
@@ -144,7 +143,7 @@ describe('buildRQL', () => {
         value: '123',
       });
       expect(query.getChildren()[1].equals(
-        getNode(MockContainer.getQuery(), variables)
+        getNode(MockContainer.getFragment('foo'), variables)
       )).toBe(true);
     });
 
@@ -152,7 +151,7 @@ describe('buildRQL', () => {
       var builder = (Component, variables) => Relay.QL`
         query {
           node(id:$id) {
-            ${Component.getQuery('foo')}
+            ${Component.getFragment('foo')}
           }
         }
       `;
@@ -163,22 +162,71 @@ describe('buildRQL', () => {
 
     it('returns different queries for different components', () => {
       var MockContainer2 = Relay.createContainer(MockComponent, {
-        fragments: {},
+        fragments: {
+          foo: () => Relay.QL`fragment on Node { name }`
+        },
       });
-      MockContainer2.getQuery = jest.genMockFunction().mockReturnValue(
-        Relay.QL`fragment on Node { name }`
-      );
 
       var builder = (Component, variables) => Relay.QL`
         query {
           node(id:$id) {
-            ${Component.getQuery('foo')}
+            ${Component.getFragment('foo')}
           }
         }
       `;
       var node1 = buildRQL.Query(builder, MockContainer, ['id']);
       var node2 = buildRQL.Query(builder, MockContainer2, ['id']);
       expect(node1 === node2).toBe(false);
+    });
+
+    it('implicitly adds component fragments if not provided', () => {
+      var builder = () => Relay.QL`
+        query {
+          node(id:$id)
+        }
+      `;
+      var node = buildRQL.Query(
+        builder,
+        MockContainer,
+        ['id'],
+        MockContainer.getFragment('foo')
+      );
+      expect(GraphQL.isQuery(node)).toBe(true);
+
+      // Confirm that `${variables.id}` is a variable by applying
+      // variable values using `RelayQuery`:
+      var variables = {id: '123'};
+      var query = getNode(node, variables);
+      expect(query instanceof RelayQuery.Root).toBe(true);
+      expect(query.getRootCall()).toEqual({
+        name: 'node',
+        value: '123',
+      });
+      expect(query.getChildren()[1].equals(
+        getNode(MockContainer.getFragment('foo'), variables)
+      )).toBe(true);
+    });
+
+    it('throws if non-scalar fields are given', () => {
+      var builder = () => Relay.QL`
+        query {
+          viewer {
+            actor
+          }
+        }
+      `;
+
+      expect(() => {
+        buildRQL.Query(
+          builder,
+          MockContainer,
+          [],
+          MockContainer.getFragment('foo')
+        );
+      }).toFailInvariant(
+        'Relay.QL: Expected query `viewer` to be empty. For example, use ' +
+        '`node(id: $id)`, not `node(id: $id) { ... }`.'
+      )
     });
   });
 });
--- a/src/query/buildRQL.js
+++ b/src/query/buildRQL.js
@@ -16,9 +16,12 @@
 var GraphQL = require('GraphQL');
 var Map = require('Map');
 var RelayQL = require('RelayQL');
+var RelayFragmentReference = require('RelayFragmentReference');
 import type {RelayConcreteNode} from 'RelayQL';
 import type {Params} from 'RelayRoute';
 import type {RelayContainer, Variables} from 'RelayTypes';
+
+var invariant = require('invariant');
 
 export type FragmentBuilder = (variables: Variables) => RelayConcreteNode;
 export type QueryBuilder =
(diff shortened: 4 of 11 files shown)


src/container/__tests__/getRelayQueries-test.js
```diff
@@ -112,7 +112,7 @@ describe('getRelayQueries', () => {
     class BadRoute extends Relay.Route {}
     BadRoute.routeName = 'BadRoute';
     BadRoute.queries = {
-      first: () => Relay.QL`fragment on Node{id}`,
+      first: (Component) => Relay.QL`fragment on Node{id}`,
```
josephsavona: Curious why this test has to change?

src/query/buildRQL.js
```diff
@@ -96,9 +98,21 @@ var buildRQL = {
...
+        var query = queryBuilder();
+        var rootCall = query.calls[0];
+        node = new GraphQL.Query(
+          rootCall.name,
+          rootCall.value,
+          query.fields,
+          [fragment],
+          query.metadata,
+          query.name
+        );
```
josephsavona: This looks good. FYI the plan is to replace the `GraphQL.*` classes with plain objects and we try to avoid using `new GraphQL.*`, but there's no alternative here.

josephsavona:
Be sure to add a `babel-relay-plugin` fixture to ensure that this new syntax will compile without validation errors. We had to add an exception to skip one rule for mutations, and I suspect the same rule may cause issues with the below, because input variables are undefined. See [here](https://github.com/facebook/relay/blob/master/scripts/babel-relay-plugin/src/GraphQLDocumentTransformer.js#L106-L112)

```
// mail fail (current) validation bc no input variables defined
foo: () => Relay.QL`query { node($id) }
```

devknoll:
Thanks @josephsavona -- actually, I'm pleased that none of my changes are done in a _total_ vacuum :wink: they all at least get tested on my production app too, since that's where I usually run into an issue first :smile: So I would have caught that! 

Any thoughts with regards to the now two old deprecated formats? Does it still make sense for the signature here to be `() => Relay.QL` (maybe variables require it?) rather than the regular return like mutations?

josephsavona:
We have code internally that uses the deprecated variant, and in general functions are preferred for lazy evaluation. The `() => Relay.QL...` pattern feels appropriate here; mutations only use regular functions because of `class` syntax.

src/query/__tests__/buildRQL-test.js
```diff
@@ -188,5 +188,32 @@ describe('buildRQL', () => {
...
     });
+
+    it('implicitly adds component fragments if not provided', () => {
+      var builder = () => Relay.QL`
+        query {
+          node(id:$id)
+        }
+      `;
+      var node = buildRQL.Query(
+        builder, MockContainer,
```
josephsavona: separate lines

devknoll:
Whoops, thanks @josephsavona. #58 can't come soon enough ;-) Let me know if this is likely to be accepted and I'll add doc updates to the PR

src/container/getRelayQueries.js
```diff
@@ -54,7 +54,8 @@ function getRelayQueries(
       var concreteQuery = buildRQL.Query(
         queryBuilder,
         Component,
-        Object.keys(route.params)
+        Object.keys(route.params),
+        Component.getFragment(fragmentName)
```
josephsavona: It's important to pass the route params by default to the fragment, as the user can't manually do so otherwise.

```
Component.getFragment(fragmentName, route.params)
```

josephsavona: We should also add a test case in `getRelayQueries-test` that ensures route params are passed through properly.

src/query/buildRQL.js
```diff
@@ -98,6 +99,17 @@ var buildRQL = {
         node = queryBuilder(Component, variables, RelayQL);
       } else {
```
josephsavona: Let's be more explicit about when the new logic will take effect in order to avoid surprising behavior and have stronger invariants about the expected outcome:
- Only auto-inject the fragment when `queryBuilder` is a function with exactly zero arguments (anything else could be the user doing `(Component) => Relay.QL...`) - so make this a 3rd `else if` clause.
- When the above case is true, check the following conditions:

```javascript
if (GraphQL.isQuery(node) && node.fragments.length === 0) {
  if (!node.fields.every(field => field.isScalar())) {
    invariant(
      false,
      'RelayQL: Expected query `%s` to be empty. For example, use `node(id: $id)`, ' +
      'not `node(id: $id) { ... }`.'
      query.name
    );
  }
  var rootCall = ...
  ...
}
```

This prevent

josephsavona:
My apologies for not getting to a more thorough review of this earlier. Please see comments, but this is a good direction. :+1:

devknoll:
No worries, thanks for the feedback!

devknoll:
> Let's be more explicit about when the new logic will take effect in order to avoid surprising behavior 

Ahhh, had something like that originally. Fixed.

> `field.isScalar()`

`GraphQLFieldNode`s don't have an `isScalar`, so I'm checking for # of child fields directly instead.

> RelayQL: Expected query

I changed this to `buildRQL` since it seems to follow the conventions more as far as I can tell.

> `query.name`

Wasn't sure if you wanted the name from the route/fragment or the root call. Went with the latter, let me know.

> We should also add a test case in getRelayQueries-test that ensures route params are passed through properly.

Let me know if the new test addresses your concerns!

@josephsavona Thanks, and sorry for the delay! I'll get the doc changes in soon too.

src/query/__tests__/buildRQL-test.js
```diff
@@ -188,5 +188,55 @@ describe('buildRQL', () => {
...
+      var builder = () => Relay.QL`
+        query {
+          node(id:$id)
+        }
+      `;
+      var node = buildRQL.Query(
+        builder,
+        MockContainer,
+        ['id'],
+        MockContainer.getQuery()
```
josephsavona: `MockContainer.getFragment('<name>')`

src/query/__tests__/buildRQL-test.js
```diff
@@ -188,5 +188,55 @@ describe('buildRQL', () => {
...
+          }
+        }
+      `;
+
+      expect(() => {
+        buildRQL.Query(
+          builder,
+          MockContainer,
+          [],
+          MockContainer.getQuery()
```
josephsavona: same as above, `getFragment(...)`

src/query/buildRQL.js
```diff
@@ -81,7 +83,8 @@ var buildRQL = {
   Query(
     queryBuilder: QueryBuilder,
     Component: any,
-    variableNames: VariableNames
+    variableNames: VariableNames,
+    fragment: GraphQL.QueryFragment
```
josephsavona: `fragmentReference: RelayFragmentReference`

src/query/buildRQL.js
```diff
@@ -96,8 +99,29 @@ var buildRQL = {
...
+              'buildRQL: Expected query `%s` to be empty. For example, use ' +
+              '`node(id: $id)`, not `node(id: $id) { ... }`.',
+              rootCall.name
+            );
+          }
+          node = new GraphQL.Query(
+            rootCall.name,
+            rootCall.value,
+            node.fields,
+            [fragment],
```
josephsavona: `[fragmentReference],`

devknoll:
:white_check_mark: `getQuery` -> `getFragment`
:white_check_mark: `fragment` -> `fragmentReference`
:white_check_mark: `GraphQL.QueryFragment` -> `RelayFragmentReference`

I also went ahead and cleaned up the rest of the calls to `getQuery` in buildRQL-test.js, and added a line to the changelog.

cc @josephsavona Thanks!

josephsavona:
@devknoll awesome! thanks again for working on this. i'm working on syncing this and your other change.

src/container/__tests__/getRelayQueries-test.js
```diff
@@ -122,4 +122,16 @@ describe('getRelayQueries', () => {
...
+    var MockRoute = makeRoute();
+    var params = {id: '123'};
+    var route = new MockRoute(params);
+    MockPageContainer.getFragment = jest.genMockFunction();
+
+    getRelayQueries(MockPageContainer, route);
+
+    expect(MockPageContainer.getFragment.mock.calls.length).toBe(4);
+    expect(MockPageContainer.getFragment.mock.calls[0][1]).toBe(params);
+  })
```
josephsavona: semicolon

devknoll: Shoot, sorry!

src/container/getRelayQueries.js
```diff
@@ -54,7 +54,8 @@ function getRelayQueries(
       var concreteQuery = buildRQL.Query(
         queryBuilder,
         Component,
-        Object.keys(route.params)
+        Object.keys(route.params),
+        Component.getFragment(fragmentName, route.params)
```
josephsavona: `s/fragmentName/queryName/g` - to match the name change in #181

src/query/buildRQL.js
```diff
@@ -96,8 +100,29 @@ var buildRQL = {
...
+      } else if (isDeprecatedCallWithArgCountGreaterThan(queryBuilder, 0)) {
+        node = queryBuilder(Component, variables);
       } else {
         node = queryBuilder(Component, variables);
+        if (GraphQL.isQuery(node) && node.fragments.length === 0) {
+          var rootCall = node.calls[0];
+          if (!node.fields.every(field => field.fields.length === 0)) {
+            invariant(
+              false,
+              'buildRQL: Expected query `%s` to be empty. For example, use ' +
```
josephsavona: Let's use `'Relay.QL: ...` for consistency with other messages.

src/query/__tests__/buildRQL-test.js
```diff
@@ -180,5 +178,55 @@ describe('buildRQL', () => {
...
+
+      expect(() => {
+        buildRQL.Query(
+          builder,
+          MockContainer,
+          [],
+          MockContainer.getFragment('foo')
+        );
+      }).toFailInvariant(
+        'buildRQL: Expected query `viewer` to be empty. For example, use ' +
```
josephsavona: `'Relay.QL:...`

devknoll:
Should be good to go now @josephsavona, thanks!

josephsavona:
@devknoll this is awesome. thanks!

taion:
When using this syntax, the query parameters will not get forwarded to the fragment as variables, right?

devknoll:
`route.params` is always passed to `Component.getFragment` with this syntax

taion:
Ah, I see. Cool! Is the old syntax to be considered deprecated?

devknoll:
I don't think so. Passing different params to `getFragment` seems like a valid thing to want to do, but @josephsavona could comment on longer term plans.

taion:
Ah, cool. On closer inspection it also looks like https://github.com/facebook/relay/pull/115/files#diff-b64b5d2eefab1f19d7cf0a21d6546db7R107 gives me pretty good backward compatibility anyway. Very nice.

devknoll:
Glad to hear that. Let me know. The more I think about it, the more I'm /facepalming for not just doing this in the babel plugin :smile:

taion:
Well, if you change it... let me know.

The one thing I lose out on here is that, when I proxy the fragment for the React Router integration using the current trick, I can restrict the variables passed into the fragment to just the ones that are relevant to the current route (i.e. only the URL parameters for the current route and its ancestors, and the query parameters for the current route): https://github.com/relay-tools/react-router-relay/blob/v0.4.2/src/RouteAggregator.js#L54-L55

With the shorthand, all fragments will receive as variables all parameters on the Relay route. Probably not a big deal in practice. I'm still going to use the shorthand, anyway.

PR #128, opened 2015-08-20
add chat example (transformed from flux-chat to relay)

transedward:
I have been working on this chat example and now it's working as flux-chat
Feature Highlight:

1.  use with `relayNestedRoutes` and `React Router` to let multiple routes and route params as query params work
https://github.com/transedward/relay/blob/master/examples/chat/js/app.js#L39-L48

2. using `route params` as global state variable (ex: currentThreadID), not ideal for now, but it shows a 
different possibility
3. and a more complicated model structure as `user->threads->messages`

Currently the code maybe a little bit messy, but I would hear some feedback and advice from anyone
Thanks in advance!


yuzhi:
I haven't had the chance to look through all the code yet, but thank you for contributing!

examples/chat/js/components/MessageSection.js
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+}
+
+export default Relay.createContainer(MessageSection, {
+  fragments: {
+    thread: () => Relay.QL`
+      fragment on Thread {
+        name
+        messages(first: 9007199254740991) {
```
yuzhi: If you are feeling adventurous, it would be very cool to add an example of paging. You can increase the number of messages fetched when "load more" is clicked.

examples/chat/js/mutations/AddMessageMutation.js
```diff
@@ -0,0 +1,109 @@
...
+      }
+    `;
+  }
+  getConfigs() {
+    return [{
+      // use FIELDS_CHANGE here to make unreadCount and thread order
+      // changeed as adding new message
+      type: 'FIELDS_CHANGE',
+      fieldIDs: {
+        thread: this.props.thread.id,
```
yuzhi: `thread` might already be updated by being a parentID in the `RANGE_ADD`. Can you try with just `viewer` to see if everything is still updated?

examples/chat/js/mutations/AddMessageMutation.js
```diff
@@ -0,0 +1,109 @@
...
+          : 0;
+  // make sure no double decrementing on same thread and no minus unreadCount
+  // copied from MarkThreadAsReadMutation, see if we can do something better
+      }
+    }
+    let timestamp = Date.now();
+    return {
+      messageEdge: {
+        node: {
+          id: 'm_' + timestamp,
```
yuzhi: `id` field is not recommended for new nodes optimistic payload, since there might a chance that it could collide with another node from the server.

examples/chat/js/mutations/AddMessageMutation.js
transedward: thanks, interesting, really works here.

examples/chat/js/mutations/AddMessageMutation.js
transedward: fixed!

examples/chat/js/components/MessageSection.js
transedward: thanks, @yuzhi
I'm half way there, I'm done with the MessageSection part
but the problem I face is that when adding new message 
`(last: $number)`  number won't get changed, and I wanna do it in AddMessageMutation
I referenced here : https://facebook.github.io/relay/docs/guides-mutations.html#content
But I can't reference `this.props` in `prepareVariables`, so I can't get the current number of messages, so I can't add 1 here then pass to `fragments` and `getFatQuery`
Is there any way I can get around here, or what is the recommended way to so?

transedward: I get it done!

transedward:
I have added some pagination example for pagination [here](https://github.com/transedward/relay-chat)
and not sure if I should add it in this PR.
I hope to get some feedback.

examples/chat/js/components/MessageSection.js
yuzhi: `Relay.Store.update` takes a set of onSuccess and onFailure callbacks. Can you tried `setVariables` for the number of messages rendered using a callback?

transedward: Yes, get it done [here!](https://github.com/transedward/relay-chat/blob/add-pagination/js/components/MessageComposer.js#L53) as a suggestion by @steveluscher 
not sure if the pagination should be added in this PR. any thoughts? @yuzhi

yuzhi:
The Relay code looks pretty good to me. Can you include the pagination code in this PR too so I can take a look at the whole thing at the same time?
After discussing with my teammates, we decided that in order to ensure the long-term maintainability of everything in the Relay repo, we are not planning on adding any more examples (we might move the existing examples to a different repo in the future). However, we’ll be happy to take a look at and review code for new examples. We definitely want to encourage you to start a new repo with this awesome example.
Thank you so much for working on this. We are excited that you found Relay interesting and implemented an chat example. I think a lot of people will find it useful.

transedward:
thanks, @yuzhi nice to hear. I had one already. and below is how I do pagination on messages
https://github.com/transedward/relay-chat/blob/add-pagination/js/components/MessageComposer.js#L53-L60 how I add on `onFailure `

https://github.com/transedward/relay-chat/blob/add-pagination/js/components/MessageSection.js#L66-L74 `the handle callback passed from upper component`

I would keep developing this example for future feature like server-side rendering...
and any one interested can follow [here](https://github.com/transedward/relay-chat).

I am also looking forward for future development of Relay, hoping it will be more mature and bringing benefit for everyone using it.

PR #131, opened 2015-08-20
Revert "Require matching query/fragment names"

steveluscher:
We were a bit too hasty with this change; our bad. It's conceivable that a Relay Container could specify a fragment that contains fields or not, depending on the active route. In fact, this is how some of our internal apps are built; some routes contain a subset of a container's available fragments, with the knowledge that those fragments won't apply when that route is active.

I'll reopen #20 so we can continue talking about this.

Reverts facebook/relay#49

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,6 @@
 ## master
 
 * Upgraded jest to 0.5 and switched Relay to use iojs v2+ only.
-* All fragments in a Relay.Container must now have a query with a matching
-  name in any Relay.Route that uses it and vice versa.
 
 ## 0.1.1 (August 14, 2015)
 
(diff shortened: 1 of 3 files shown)


devknoll:
@steveluscher Yeah, I was worried that it would be a little too strict. Shall I open a PR that just requires that the route provides everything the container wants instead?

cc @josephsavona

steveluscher:
Unfortunately, that's exactly the thing that's breaking some of our apps. A container might have `foo` and `bar` fragments, but when `FooRoute` is active there's no need for the `bar` fragment. The `FooRoute` in this case doesn't have a `bar` fragment, and this invariant gets tripped.

devknoll:
Ahhh, right, that makes total sense, sorry about that.

PR #152, opened 2015-08-22
Tutorial: export Game and HidingSpot classes

lencioni:
In the following step of this tutorial, there is some code that imports
these classes from this file.

```
import {
  Game,
  HidingSpot,
  ...
} from './database';
```

However, since they are not exported, they will be undefined when
imported. This comes into play in the `nodeDefinitions` object type
resolver when we try to run `obj instanceof Game`. Since `Game` is
undefined, this will result in an error such as:

> Expecting a function in instanceof check, but got undefined

For the scope of this tutorial, this doesn't seem to make much
difference. However, for folks wanting to move beyond this tutorial into
something a little more complicated, such as pagination, this oddity
could be a stumbling block to learning.

### DIFF
--- a/docs/QuickStart-Tutorial.md
+++ b/docs/QuickStart-Tutorial.md
@@ -29,8 +29,8 @@ We need a place to hide our treasure, a way to check hiding spots for treasure, 
  */
 
 // Model types
-class Game extends Object {}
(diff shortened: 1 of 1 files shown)


wincent:
lg2m. Thanks, @lencioni!

PR #157, opened 2015-08-24
remove path from example routes

charlieschwabacher:
I found this `path` variable on the example routes confusing - I couldn't find any documentation on it, and changing or removing it seems to have no effect on the examples.  Is it safe to remove?


devknoll:
Can you PR the same change over at https://github.com/relayjs/relay-starter-kit ? :smile:

charlieschwabacher:
done! https://github.com/relayjs/relay-starter-kit/pull/18

PR #167, opened 2015-08-25
Update Interfaces-NetworkLayer.md

pasviegas:
queryRequest and mutationRequest fulfilment expect the data to be wrapped in a obj with shape {response: <data>},


josephsavona:
@pasviegas thanks for fixing!

pasviegas:
You're welcome! :)

PR #175, opened 2015-08-26
Change `npm test` to run jest even if flow fails

yungsters:


yungsters:
Hrrrm... this breaks `npm test [jest-args]`.

zpao:
What's the advantage of keeping these both in the `test` command? If it's just for travis, would it make flow and tests 2 separate runs that can fail independently? (that's what we did for fbjs)

yungsters:
I think it's so that `npm test` runs both Flow and Jest (to discourage people from submitting pull requests that break Flow types). @steveluscher would know the original intent.

steveluscher:
It was indeed to reduce the incidence of PRs that pass tests but fail typechecking.

It's somewhat wasteful to run flow on the whole project when doing something like `npm test RelayContainer`, but it's also Pretty Fast, So Meh™

PR #176, opened 2015-08-26
Build docs, unless there's a static file available to serve

steveluscher:
This prevents the docs from being rebuilt when the browser requests, for instance, relay.css.


zpao:
:thumbsup:

PR #181, opened 2015-08-27
Throw if container is missing fragments

devknoll:
Another crack at #20 / #131

Rather than building the query set from the set of all fragments that a component has, it builds them from the set **required** by the route, and throws if the component doesn't provide one.

I'm on the fence about keeping `hasFragment`.

This should be compatible with FB code, but let me know if it's not @steveluscher @josephsavona

### DIFF
--- a/src/container/__tests__/RelayContainer-test.js
+++ b/src/container/__tests__/RelayContainer-test.js
@@ -195,6 +195,12 @@ describe('RelayContainer', function() {
         ])
       ));
     });
+
+    it('returns whether a named fragment is defined', () => {
+      expect(MockContainer.hasFragment('foo')).toBe(true);
+      expect(MockContainer.hasFragment('bar')).toBe(true);
+      expect(MockContainer.hasFragment('baz')).toBe(false);
+    });
   });
 
   describe('conditional fragments', () => {
--- a/src/container/__tests__/getRelayQueries-test.js
+++ b/src/container/__tests__/getRelayQueries-test.js
@@ -20,6 +20,7 @@ var getRelayQueries = require('getRelayQueries');
 
 describe('getRelayQueries', () => {
   var MockPageContainer;
+  var MockPageComponent;
 
   var makeRoute;
 
@@ -28,7 +29,7 @@ describe('getRelayQueries', () => {
   beforeEach(() => {
     jest.resetModuleRegistry();
 
-    class MockPageComponent extends React.Component {
+    MockPageComponent = class MockPageComponent extends React.Component {
       render() {
         return <div/>;
       }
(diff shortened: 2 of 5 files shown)


src/container/getRelayQueries.js
```diff
@@ -46,9 +46,16 @@ function getRelayQueries(
...
+  Object.keys(route.queries).forEach(queryName => {
+    var fragmentName = queryName;
+    invariant(
+      Component.hasFragment(fragmentName),
+      'getRelayQueries: Route `%s` expected `%s` to have a fragment named ' +
+      '`%s`.',
+      route.name,
+      Component.displayName,
+      fragmentName
+    );
```
josephsavona: Let's keep the message consistent with the below:

```javascript
'Relay.QL defined on route `%s` named `%s` is not a valid query, ' +
'a matching fragment was not found on component `%s`.',
route.name,
fragmentName,
Component.displayName
```

src/container/__tests__/getRelayQueries-test.js
```diff
@@ -122,4 +123,22 @@ describe('getRelayQueries', () => {
       'query. A typical query is defined using: Relay.QL`query {...}`'
     );
   });
+
+  it('throws if a container doesnt include a required fragment', () => {
```
josephsavona: s/doesnt/does not/ (grammar)

src/container/getRelayQueries.js
josephsavona: GH won't let me add a comment to a non-updated line, so can you also change line 68-69 to:

```javascript
invariant(
  ...,
  'Relay.QL: query `%s.queries.%s` is invalid, a typical query is defined using: ' +
  '() => Relay.QL`query { ... }`.',
  route.name,
  queryName
```

src/container/getRelayQueries.js
```diff
@@ -46,9 +46,16 @@ function getRelayQueries(
     return cache[cacheKey];
   }
   var querySet = {};
-  Component.getFragmentNames().forEach(fragmentName => {
-    // TODO: Fix this. It relies on the query and fragment names matching.
-    var queryName = fragmentName;
+  Object.keys(route.queries).forEach(queryName => {
+    var fragmentName = queryName;
```
josephsavona: just use `queryName`

devknoll:
Fixed. @josephsavona do you have any thoughts about `hasFragment`? I don't really like bloating the RelayContainer interface...

src/container/__tests__/RelayContainer-test.js
```diff
@@ -210,6 +210,12 @@ describe('RelayContainer', function() {
         ])
       ));
     });
+
+    it('can tell if a fragment was given', () => {
```
josephsavona: `it('returns whether a named fragment is defined', ...`

josephsavona:
@devknoll looks like you made updates before i finished adding all my comments. Can you double-check all of the feedback?

`hasFragment()` seems reasonable.

devknoll:
:white_check_mark: Updated missing fragment invariant
:white_check_mark: Updated invalid query invariant
:white_check_mark: Changed to 'does not'
:white_check_mark: Nixed fragmentName
:white_check_mark: Fixed test name

josephsavona:
@devknoll we're basically all set to merge this once the merge conflicts are resolved

devknoll:
Whoops, thanks for the heads up @josephsavona

josephsavona:
@devknoll awesome job on this.

PR #186, opened 2015-08-28
Bring back the new NetworkLayer docs

devknoll:
Uh oh. @yungsters it seems like the NetworkLayer changes from #70 made it in, but the doc changes did not. This brings them back!


josephsavona:
thanks!

PR #189, opened 2015-08-29
Update tests to use top level requires.

cpojer:


clentfort:
Just out of curiosity: What is the advantage of doing this?

wincent:
@clentfort: consistency with other, non-test modules, which all use top-level requires.

We can do this now thanks to an "online requires" transform, which means that things like `resetModuleRegistry` still work.

fson:
If anyone else is curious about the "inline requires" transform, it's defined [here](https://github.com/facebook/fbjs/blob/41cd335d5afb17eb7b188f3ecba76ee595ad92f5/scripts/babel/inline-requires.js).

Looks like it transforms code like this:
```js
var printRelayQuery = require('printRelayQuery');
  // ...
  printRelayQuery(...);
  printRelayQuery(...);
```
to this:
```js
  // ...
  require('printRelayQuery')(...);
  require('printRelayQuery')(...);
```

cpojer:
From the internal document I shared about this change:

### Pros
* Modules work the same way as in regular JS files.
* no more `var A`, `A = require('A')`; making it easier to add modules to tests. This will also make it much easier for automated tooling (codemods) to modify and understand require's – because they are static instead of dynamic.
* If the test doesn't reset the module registry (because the tested module doesn't have internal state), this is just the much nicer and friendlier way to do things. Many modules actually already do it this way – if they add module internal state, they'll have to rewrite the require-calls to be in beforeEach.
* In a perfect future, when a full test run at FB takes only 10 seconds to complete and our mocking is 10x faster, we can get rid of jest.resetModuleFactory and automatically reset state for every tests. Because we'll have a better guarantee that there are no inter-test dependencies we can experiment with parallelizing individual test runs (ie. every call to `it`) within a single test file. Open source jest already resets the module registry on every call to `it` by default.
* Will allow us to move to ES2015 import/export syntax for tests: not doing this actively blocks us from moving forward on this when we decide to do so. If we switch for regular JS module and can't do it for tests, we'll maintain two ways to import modules forever.

### Cons
* Tests are a little bit more magical:
    * When the module registry is reset, A in one test is not equal to A in another test because the variable A actually doesn't even exist in a test any more and is replaced with require('A') calls transparently.
    *  *require* in tests is already magic I don't think making it slightly more magical is a big deal. require can require the actual module, the mock file or the mock file that requires the actual module. It works differently depending on dontMock/mock calls.
    * It will be different for people who are used to the current system but I expect people new to jest will understand that jest.resetModuleRegistry() (or whatever I will probably rename it to) reset state and that everyone is ok with throwing a little bit of JS knowledge overboard for tests. Personally, if I try to forget everything I know about tests, this is how I would expect my tests to work.
    * I think communication and documentation will take care of this issue.
    * I think all the items in the *Pros* section outweigh this point.

There were two more cons about deploying this to open source. This has, as shown in this and my previous PR, already been resolved. The last remaining thing is to find a way to deploy this to all users of jest (or babel-jest).

PR #190, opened 2015-08-30
Minor doc fixes

### LINKED ISSUE #182: Fix named anchors in the documentation
Not all `<href>` point to named anchors that exist, and on some pages there exist non-unique named anchors. Fixing this should make it easier for people to link to a specific point in the documentation.

devknoll:
Partially fixes #182 and another small issue inconsistency I found. This doesn't fix nested anchors yet.

### DIFF
--- a/docs/Interfaces-NetworkLayer.md
+++ b/docs/Interfaces-NetworkLayer.md
@@ -16,7 +16,7 @@ Custom network layers that must conform to the `RelayNetworkLayer` interface.
 <ul class="apiIndex">
   <li>
     <a href="#sendmutation">
-      <pre>sendMutations(mutationRequest)</pre>
+      <pre>sendMutation(mutationRequest)</pre>
     </a>
   </li>
   <li>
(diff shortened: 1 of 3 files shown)


docs/Interfaces-NetworkLayer.md
```diff
@@ -15,7 +15,7 @@ Custom network layers that must conform to the `RelayNetworkLayer` interface.
 
 <ul class="apiIndex">
   <li>
-    <a href="#sendmutations">
+    <a href="#sendmutation">
       <pre>sendMutations(mutationRequest)</pre>
```
fson: The text of the link is still wrong?

devknoll: Fixed, thanks!

devknoll:
@steveluscher For the separate problem of duplicate/nested anchors, I'm currently thinking of just modifying https://github.com/facebook/relay/blob/master/website/core/Header.js to keep the slug in the context and nesting them on render.

Sound good?

steveluscher:
:+1:

devknoll:
Went a slightly different direction with this. I noticed that the anchor links break down at around H4, everything above that is unique per page. So this will start generating nested anchors at H4. The benefit of this is that all of the existing anchors will continue to work.

WTB tests or anchor testing software :wink: 

cc @steveluscher

PR #200, opened 2015-09-01
Fix HomeRoute queries for all examples

drabinowitz:
The examples are missing the fragment queries in their home routes


devknoll:
This is intentional, see #115

drabinowitz:
Thanks for the quick update my bad. I looked for an issue on the subject
but I must have missed that one.

PR #202, opened 2015-09-01
add eslint rules from fbjs

knowbody:
As @zpao suggested (reference #58) here is the new PR with the rules from [facebook/fbjs#49](https://github.com/facebook/fbjs/pull/49).

I'll still need some guidance on what should be Relay specific.

At the moment with the current fbjs rules there is a lot of errors on [no-undef](http://eslint.org/docs/rules/no-undef.html) (example: $FlowIssue, $FixMe, $Enum and also when defining Flow types, this is related to [babel-eslint/known-issues](https://github.com/babel/babel-eslint#known-issues) - [babel-eslint#130](https://github.com/babel/babel-eslint/issues/130) and [babel-eslint#132](https://github.com/babel/babel-eslint/issues/132))

@josephsavona @zpao what are your thoughts?


.eslintignore
```diff
@@ -0,0 +1,9 @@
+dist/
+docs/
+examples/
+node_modules/
+lib/
+scripts/
+# for disable website but should be done as well
```
yungsters: Do you mean something like this?

```
# TODO: Enable ESLint for website.
```

.travis.yml
```diff
@@ -5,4 +5,4 @@ node_js:
 env:
 - TEST_DIR=.
 - TEST_DIR=scripts/babel-relay-plugin
-script: cd $TEST_DIR && npm install && npm test
+script: cd $TEST_DIR && npm install && npm run lint && npm test
```
zpao: This isn't going to work since `babel-relay-plugin` doesn't have that script defined. I think we'll probably actually want a better way to do this (maybe a `travis-test` task in each that gets run, looks sort of like relay's `test` to run multiple things)

yungsters:
Here are the "undefined" identifiers:

- `$Enum`
- `$FixMe`
- `$FlowFixMe`
- `$FlowIssue`
- `Iterator`
- `IteratorResult`
- `ReactClass`
- `ReactElement`

I think we should just define these as globals in `.eslintrc` for now.

zpao:
Yea, you can define them here for now. They should make their way back to fbjs (and there are probably other ones that we should just copy over from flow). We have them defined internally, I just went for the minimal sync.

.eslintignore
knowbody: @yungsters haha! yeah that's what I meant, not sure my comment is even in English lol

knowbody: oh it was supposed to be "for *now*... "

.travis.yml
knowbody: @zpao couldn't it be just something like:

```yml
script:
  - cd $TEST_DIR && npm install && npm test
  - npm run lint
```

`npm run lint` runs `eslint .` so it will check everything anyway (we would need to remove `/scripts` from `.eslintignore`)

zpao: `npm run lint` will look for a `lint` script in the `$TEST_DIR/package.json`. When `$TEST_DIR` is scripts/babel-relay-plugin` this will fail because there is not `lint` script there. That's *exactly* what's happening: https://travis-ci.org/facebook/relay/jobs/78187439.

I think even doing what you have there will be run from `$TEST_DIR` (should work if you `pushd $TEST_DIR && commands && popd` but then you'll be running top level lint twice)

The other piece of this is that we might want lint to be an allowable failure, which isn't possible like this. In React we put it as a separate test so could (if we needed, which we did for a while) to make it so that lint could fail but jest couldn't. In Relay's case, if there are other things we need to sync out, it might make sense to have the same capability.

knowbody: okay, thanks for the clarification, I understand now

knowbody: so should it look like the one you use in React? with the `TEST_TYPE`s?

zpao: I would probably do this in 2 totally different PRs.

1. Make lint work
2. Add it to travis

knowbody: ok so to satisfy 1. I can just remove `npm run lint` from travis config

and then submit another PR with the proper config?

zpao: Yup. You'll also want to revert the fbjs-scripts inclusion in package.json (I just landed #196) and then rebase

knowbody:
@zpao should be good

.eslintignore
```diff
@@ -0,0 +1,9 @@
...
+website/
+*.md
```
yungsters: When you update the comment, can you also add a newline at the end of the file? Makes life easier (internally, we have lints that try to automatically add these).

knowbody:
@yungsters done, the tests are failing but not sure why, haven't done any changes to `src/`

zpao:
:thumbsup: thanks! We'll go through the land internally and merge out process, should be all wrapped up soon.

PR #212, opened 2015-09-01
Expose printRelayQuery and flattenRelayQuery for devtools

kassens:


src/RelayPublic.js
```diff
@@ -30,6 +30,8 @@ if (typeof global.__REACT_DEVTOOLS_GLOBAL_HOOK__ !== 'undefined') {
   global.__REACT_DEVTOOLS_GLOBAL_HOOK__._relayInternals = {
     NetworkLayer: require('RelayNetworkLayer'),
     DefaultStoreData: require('RelayStoreData').getDefaultInstance(),
+    flattenRelayQuery: require('flattenRelayQuery'),
+    printRelayQuery: require('printRelayQuery'),
   };
 }
```
wincent: I wonder if we should move this block into a separate file, so that we're not leaking the names of internal modules into this file (which is effectively a manifest of the public bits of Relay). What do you think?

kassens:
Moved the exposed internals now to a new module RelayInternals.js

PR #215, opened 2015-09-02
Clarify description of globally unique IDs

wincent:
As written, this sentence was a little hard to read:

> A system without globally unique IDs can usually create one

Create one what? An ID, or a system?

I've reworded this to make it clear that the system can create IDs by
switching to the plural, and I've added a missing hyphen in
"type-specific".


kassens:
Thanks!

PR #217, opened 2015-09-02
Fix object shorthands in babel-relay-plugin

SimonDegraeve:
This ES6 features break the plugin if the code is not transpiled before publishing to NPM.


yungsters:
@josephsavona Looks like this made it into the published 0.2.1. We may want to publish a 0.2.2 with this fix after it's merged.

SimonDegraeve:
Indeed, version 0.2.1 is broken for me, that's why I made this pull request.

yungsters:
I'll merge this and push a new package first thing in the morning.

PR #225, opened 2015-09-02
Fix grammar in mutation docs

devknoll:


wincent:
Thanks @devknoll! I'll get this merged.

PR #226, opened 2015-09-02
Fix fragment in getFatQuery example

devknoll:


wincent:
Thanks again, @devknoll. Will merge.

PR #243, opened 2015-09-05
Add missing key attribute in treasure hunt example

l-urence:
With react and react-relay bundled into app.js bundle react is complaining about a missing key attribute. For some reason the warning is omitted if react is loaded with a script tag.

### DIFF
--- a/examples/relay-treasurehunt/js/components/App.js
+++ b/examples/relay-treasurehunt/js/components/App.js
@@ -48,6 +48,7 @@ class App extends React.Component {
         <div
           onClick={this._handleHidingSpotClick.bind(this, edge.node)}
           style={this._getHidingSpotStyle(edge.node)}
+          key={edge.node.id}
         />
       );
     });


josephsavona:
:+1:

wincent:
Thanks @l-urence!

wincent:
> For some reason the warning is omitted if react is loaded with a script tag.

Perhaps you are using a minified build when you include React via the script tag. AFAIK, the key warning only gets issued when using a `__DEV__` build.

In any case, the key warning should be fixed. I've merged this internally and it should sync back out here shortly.

Thanks once again for the PR @l-urence!

l-urence:
@wincent yes the ```__DEV__``` build makes sense. I noticed a second deprecated warning with ```React.render(/*....*/)``` in the example. It seems that ```React.render(/*...*/)``` should be replaced with ```ReactDOM.render(/*....*/)``` in React v0.14.0.

wincent:
> It seems that `React.render(/*...*/)` should be replaced with `ReactDOM.render(/*....*/)` in React v0.14.0.

Yep! Want to submit a PR for that too? I'll make a separate issue in any case so that we don't forget.

Issue #254, opened 2015-09-08
Can't run any examples


adriantoine:
I cloned the repo and tried:
`cd relay/`
`npm install` (this is needed to avoid issue stated here: https://github.com/facebook/relay/issues/248)
`cd examples/todo`
`npm install`
`npm start`

And I get this error on `npm start`:
```
$ npm start

> @ start /home/adrien/projects/relay/examples/todo
> babel-node ./server.js

GraphQL Server is now running on http://localhost:8080
Relay TodoMVC is now running on http://localhost:3000
Hash: 9a56cc72acac2de6f40c
Version: webpack 1.12.1
Time: 2095ms

Module build failed: /home/adrien/projects/relay/examples/todo/node_modules/babel-relay-plugin/src/GraphQLPrinter.js:701
    cursorType,
              ^
SyntaxError: Unexpected token ,
    at exports.runInThisContext (vm.js:73:16)
    at Module._compile (module.js:443:25)
    at Module._extensions..js (module.js:478:10)
    at Object.require.extensions.(anonymous function) [as .js] (/home/adrien/projects/relay/examples/todo/node_modules/babel/node_modules/babel-core/lib/api/register/node.js:214:7)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/home/adrien/projects/relay/examples/todo/node_modules/babel-relay-plugin/src/GraphQLDocumentTransformer.js:6:22)
    at Module._compile (module.js:460:26)
```

Note that I don't have any errors on install, only on `npm start`, I have tried `npm cache clean` and I get the same error with the other examples: `relay-treasurehunt/` and `star-wars/`.

adriantoine:
I found the issue and fixed it, see PR.

Issue #248, opened 2015-09-06
Local gulp error running TodoMVC example from README


mjw56:
I tried running the TodoMVC example from the README file but got the below error. 

I'm running node 0.12.3 and npm 3.2.2. I also have gulp CLI 3.9.0 installed globally.

Running npm install from ```relay/``` first solved the issue.

```bash
$ cd relay/examples/todo && npm install                                                                            [1:24:44]

> react-relay@0.2.1 prepublish /Users/mikewilcox/dev/relay
> npm run build


> react-relay@0.2.1 build /Users/mikewilcox/dev/relay
> [ $(ulimit -n) -lt 4096 ] && ulimit -n 4096; gulp

[01:24:59] Local gulp not found in ~/dev/relay
[01:24:59] Try running: npm install gulp
```

steveluscher:
Thanks for bringing this up!

The examples used to depend on the npm `react-relay` module, but they've since been changed to depend on `"../../"` to allow local development.

We either need to:

1. Document this.
2. Automate the building of `react-relay` when you try to install/run the examples.

PR #255, opened 2015-09-08
Fixed examples to use local version of babel-relay-plugin, like react-relay

adriantoine:
This is fixing an issue I had which is detailed here: https://github.com/facebook/relay/issues/254.

In examples, `react-relay` is using the local version and `babel-relay-plugin` needs to use the local one as well or the examples won't start.

### DIFF
--- a/examples/relay-treasurehunt/package.json
+++ b/examples/relay-treasurehunt/package.json
@@ -7,7 +7,7 @@
   "dependencies": {
     "babel": "5.8.23",
     "babel-loader": "5.3.2",
-    "babel-relay-plugin": "0.2.1",
+    "babel-relay-plugin": "file:../../scripts/babel-relay-plugin/",
     "classnames": "^2.1.3",
     "express": "^4.13.1",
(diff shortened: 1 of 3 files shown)


adriantoine:
Hi @facebook-github-bot, I just signed up!

devknoll:
We should probably address the counterpart issue (see #248) here too...

Maybe add a `preinstall` for each of the examples that runs `npm run build` on both `../../` and `../../scripts/babel-relay-plugin/`?

wincent:
Thanks, @adriantoine. This looks reasonable to me. I also see #260, which fixes this this in a different way. Let me check out both of these.

devknoll:
@wincent Not sure, but I think we might need both? #260 was intended to fix a slightly different problem

josephsavona:
:+1:

josephsavona:
This makes sense. We keep features in sync between the main & plugin packages so the examples should use the "same" version.

wincent:
Merging both through our internal tool. They should sync back out here shortly.

PR #269, opened 2015-09-09
Fix learn.json in TodoMVC example

taion:
I hit this while setting up my own "Relay TodoMVC with routing" example.

You need to set up `data-framework` and the framework key on `learn.json` to get the sidebar to show up (otherwise `learn.json` doesn't do anything). The `source_path` key from the `learn.json` template also doesn't do anything, and you need to drop in the `templates` from their example as well.

The `"url": ""` bit in the example is just to keep the "demo" link from showing up, since as far as I'm aware this isn't running publicly on the internet anywhere.

### DIFF
--- a/examples/todo/public/index.html
+++ b/examples/todo/public/index.html
@@ -1,5 +1,5 @@
 <!doctype html>
-<html lang="en">
+<html lang="en" data-framework="relay">
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1">
--- a/examples/todo/public/learn.json
+++ b/examples/todo/public/learn.json
@@ -1,11 +1,13 @@
 {
-	"framework": {
+	"relay": {
 		"name": "Relay",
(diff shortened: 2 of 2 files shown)


taion:
BTW, per https://github.com/facebook/relay/blob/v0.2.1/examples/todo/js/components/TodoListFooter.js#L20 I do also have a working TodoMVC example that includes routing.

However, it depends on React Router, so I don't know if it'd be in scope here. It's also set up more as a demonstration of routing than of an optimal Relay TodoMVC implementation.

yungsters:
Very nice! Thanks for fixing this, and I am excited for your React Router implementation of the TodoMVC.

For posterity, here is what gets rendered:

screen shot 2015-09-09 at 10 23 49 am

taion:
@yungsters 

The React Router based implementation is here - https://github.com/taion/relay-todomvc. I'd be happy to PR it if you think it fits.

Aside from the caveat above with how the schema is set up, it has a few other external dependencies and uses webpack a bit differently from the existing examples, though.

PR #286, opened 2015-09-10
Set flow version in .flowconfig

### LINKED ISSUE #283: Update to flow 0.15.0

kassens:
Closes #283


yungsters:
I think this is already fixed by 8ab8ff5ee9318abe5e5aaa7dec41e15b31449a72. Is there any reason we still need this in `.flowconfig`?

kassens:
I for example use flow that's installed with homebrew. This forces the expected flow version or will error that you don't have the latest flow version installed. (It has a few versions of backwards compatibility, so if you have a newer version, it would still use the older version for the checking)

PR #290, opened 2015-09-10
Bring the examples into the era of 0.14.0-rc and ReactDOM

steveluscher:


examples/relay-treasurehunt/public/index.html
```diff
@@ -23,6 +23,7 @@
     </script>
     <script src="http://localhost:3000/webpack-dev-server.js"></script>
     <script src="node_modules/react/dist/react.min.js"></script>
+    <script src="node_modules/react/dist/react-dom.min.js"></script>
```
zpao: Total tangent, but it might make sense to use the dev builds of these

PR #298, opened 2015-09-11
Update CHANGELOG and version number for v0.3.0 release

wincent:


wincent:
Note: I'm going to publish and bump the babel-relay-plugin dependency before shipping this.

wincent:
Looks like Travis wants the plugin published separately and previously, so I'll split that bit off into a separate PR.

PR #300, opened 2015-09-11
Update the website-building docs

steveluscher:


wincent:
Did you mean to create a `website-building-docs` branch directly on the https://github.com/facebook/relay repo?

steveluscher:
I did.

I don't have my own fork of Relay, and I prefer to keep feature branches as close to `facebook/relay#master` as possible. That, and I _always_ clean up merged branches.

I realize that this is an unpopular way of working, and an exclusive privilege of those with push access to this repository. Though I find this workflow to be simple and effective, I'm open to being compelled to change.

wincent:
> Though I find this workflow to be simple and effective, I'm open to being compelled to change.

I think it's fine, so I wouldn't try to compel you. FWIW, the two reasons I push to my fork rather than here directly are that:

1. There are a lot of eyes on this repo, so I prefer to avoid any confusion about such a branch enjoying "official" status when it's actually just a work-in-progress that's going to get deleted.
2. I like the layer of indirection that prevents me from doing something silly or destructive like pushing the wrong branch, or overwriting a branch etc in a highly visible place.

But I think you should keep doing whatever works for you.

Issue #301, opened 2015-09-12
Cannot find module "./filterObject" in relay 0.3.0


mhart:
I'm getting:
```
Uncaught Error: Cannot find module "./filterObject"
webpackMissingModule @ relay.js:5306
(anonymous function) @ relay.js:5306
__webpack_require__@ relay.js:33
(anonymous function)@ relay.js:11364
__webpack_require__@ relay.js:33
(anonymous function)@ relay.js:13790
(anonymous function)@ relay.js:13828
__webpack_require__@ relay.js:33
(anonymous function)@ relay.js:101
__webpack_require__@ relay.js:33
(anonymous function)@ relay.js:53
(anonymous function)@ relay.js:56
webpackUniversalModuleDefinition@ relay.js:12
(anonymous function)@ relay.js:13
```
When trying to use `dist/relay.js` (or `dist/relay.min.js`) just included in a script tag on the page:

```html
<script src=//fb.me/react-0.14.0-rc1.min.js></script>
<script src=//fb.me/react-dom-0.14.0-rc1.min.js></script>
<script src=/relay/relay.js></script>
```
(`/relay` is serving statically from `node_modules/react-relay/dist`)

This does not occur in 0.2.1

mhart:
FWIW I also have a browserified bundle.js included as a script tag further down – it does not include any of React or Relay in the bundle though, only my components. React and Relay are referenced in the following manner:
```js
var Relay = (typeof window !== "undefined" ? window['Relay'] : typeof global !== "undefined" ? global['Relay'] : null);
```
Again, this worked fine in 0.2.1

mhart:
Actually, scratch that, the error occurs even if I comment out the browserified bundle, so we can remove that from the equation.

Literally these three lines can reproduce the error:
```html
<script src=//fb.me/react-0.14.0-rc1.min.js></script>
<script src=//fb.me/react-dom-0.14.0-rc1.min.js></script>
<script src=/relay/relay.js></script>
```

wincent:
Thanks for the heads up. You need a later version of fbjs to get `filterObject` (see https://github.com/facebook/relay/commit/3f39600969b90294eeffd7a8bd3c1a2586fef387). Not at a computer now, but that hunk looks like it somehow got reverted. I'll push a fix presently.

wincent:
> but that hunk looks like it somehow got reverted

Er, actually, no. I just suck at reading diffs on my phone, the dependency is still intact in the `package.json`.

[...]

PR #302, opened 2015-09-12
Bump version to 0.3.1

wincent:
We need to rebuild the just-released (as v0.3.0) NPM package, which had
faulty `dist/` contents.

Fixes https://github.com/facebook/relay/issues/301

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## master
 
+## 0.3.1 (September 11, 2015)
+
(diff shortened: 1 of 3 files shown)


mhart:
(any way to avoid this in the future?)

wincent:
> (any way to avoid this in the future?)

I think so. I think there a failure in our `prepublish` task isn't correctly aborting the `publish`. I'll file a separate issue for that.

Issue #258, opened 2015-09-08
clientMutationId NonNull


freiksenet:
Hi!

Currently in the spec the `clientMutationId` in mutations has to be non-null. The requirement seems overly strict - in GraphiQL usage or while using other clients but Relay, one doesn't *have* to use `clientMutationId` to actually make a mutation.

I propose to remove the requirement on non-nullness from clientMutationId, both in input and in payload. Instead one can say that clientMutationId *must* be returned if it was provided.

Pinging stakeholders - @dschafer.

dschafer:
Yeah, I'm fine with this; let's change it to state that clientMutationId *may* be non-null, and it is non-null on the input if and only if it's non-null on the payload. Either way, the value of clientMutationId in the payload must be the value provided in the input, or null if the value was not provided.

freiksenet:
Great! Do you want me to do pull request to the spec?

dschafer:
Yep, go for it!

steveluscher:
Closed by #308.

PR #308, opened 2015-09-14
Allow clientMutationId to be null

freiksenet:
As per #258


wincent:
I've imported this into our internal Phabricator instance so we can get some other eyes on it. Thanks for the PR, @freiksenet.

josephsavona:
Relay currently assumes that `clientMutationId` is non-null; we should add the requisite invariant checks before updating the spec.

freiksenet:
Does it assume that in the schema or in result? Current spec enforces resulting `clientMutationId` to be non-null if it was passed in query, so relay will always get it and not getting it is a valid error.

josephsavona:
I see, I was probably being overly cautious. :+1:

Issue #303, opened 2015-09-12
Make publish process more fool-proof


wincent:
So v0.3.0 was a broken build (see #301) that I had to redo and supersede with v0.3.1.

This was the first time I'd done the release, so I inspected the `package.json` to see what would happen on running `npm publish`. Based on that reading, I expected this to happen:

- `prepublish` would run, which would...
- run `build`, which would...
- kick off `gulp`, which would...
- build things such as the files under `dist`, then...
- actually publish to NPM

What actually happened:

- as above, but the `gulp` build threw an error, and ...
- the module got published anyway

I blew away my state, reinstalled and ran `gulp` again to be sure it could run to completion without error, but I didn't realize that the built package itself was faulty until #301 was reported.

So, this is unfortunate and I am sorry it happened, but we should try to make the whole thing more resilient such that this particular failure mode can't happen again.

Open questions:

- did the `gulp` build actually exit with a non-zero exit status on hitting the error?
- if so, does that failure affect the overall return status of the `build` subtask?
- and correspondingly, is a `build` failure enough to make `prepublish` abort?

zpao:
I don't think gulp did exit with non-zero. And yes, I'm pretty sure a non-zero `prepublish` would have aborted `publish`.

There are 2 pieces here.

1. The very dumb module rewriting, which in order to be fast (and simple), doesn't actually resolve modules, just works purely in inputs (though will throw on duplicates). We could make it actually do some resolution (eg, just in map or relative), would just slow it down a little bit.
2. In this case it was the webpack step which "failed" and I would have expected to error with non-zero but perhaps this doesn't do what we want (https://github.com/facebook/relay/blob/master/gulpfile.js#L98).

wincent:
I'll see if I can repro what actually happened so that we're not shooting in the dark here.

zpao:
`npm install fbjs@0.2.0` and then you should see exactly what happened (perhaps with a minor difference of something saying how you got the dependency installed - `fbjs@^0.2.0` vs `fbjs@0.2.0`)

wincent:
I think ab2afb30b7c53bca75638758acc4a3a4f0542309 is mitigation enough for now. Thanks, @zpao.

PR #311, opened 2015-09-14
Throw on webpack compilation error

zpao:
Instead of just logging, actually treat webpack compilation errors as
a failure so that we don't continue.

Test Plan: `gulp dist && echo 'foo'` (with `fbjs@0.2.0`), don't see `foo` echoed.

cc @steveluscher @wincent 

This is the shortest path to prevent the specific issue in #303. Longer term, we might want to something a bit more versatile that also checks at the npm dependency level (might be some packages that do it already, but also pretty easy to do with some `npm outdated --json` and some semver checking or `npm ls --depth=0 | grep 'ERR!'`)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -100,7 +100,7 @@ var buildDist = function(opts) {
       throw new gulpUtil.PluginError('webpack', err);
     }
     if (stats.compilation.errors.length) {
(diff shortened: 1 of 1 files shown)


wincent:
Thanks, @zpao. Want to go ahead and import this on the FB side so we can get this in?

PR #317, opened 2015-09-15
Fix LikeStoryMutation example optimistic response

jshanson7:
Fixes `story.likers.count` of `getOptimisticResponse` in the `LikeStoryMutation` example for the case where the current viewer already likes the story (this should unlike the story, thus decrementing the optimistic `likers.count` by 1).


wincent:
Well caught! Thanks.

steveluscher:
Nice one. Thanks!

PR #320, opened 2015-09-15
RFC: re-jig node-building API

wincent:
Just a straight rename and relocating some static methods:

- `RelayQuery.Node.buildField` -> `RelayQuery.Field.build`
- `RelayQuery.Node.buildFragment` -> `RelayQuery.Fragment.build`
- `RelayQuery.Node.buildMutation` -> `RelayQuery.Mutation.build`
- `RelayQuery.Node.buildRoot` -> `RelayQuery.Root.build`

This shaves a few characters off each API, and moves both the code and
the concepts closer to the code for the constituent subclass
implementations, which feels like an improvement to me.

If people agree that this is a good idea, I'll apply the same treatment
to the `RelayQuery.create*` methods as well.


josephsavona:
:+1:

PR #329, opened 2015-09-16
Documented conditional fields

steveluscher:
screen shot 2015-09-16 at 4 18 18 pm


josephsavona:
Should we encourage the `@include` and `@skip` directives instead?

steveluscher:
Done!

Issue #293, opened 2015-09-11
Let server tell client where new item should be inserted in range


KyleAMathews:
Currently when a mutation returns a new item, in `getConfig` you can tell Relay to `append` or `prepend` the item to a list. This covers most cases as most lists are sorted by the created or updated date but there are times when using custom sorts that a new item would end up somewhere in the middle of a range. In these cases, it'd be nice if we could instruct the server to tell the client where a new item should be inserted.

devknoll:
Some ideas...

### Server Side

The "best" way to achieve this would probably be to somehow query the server with the different sets of field args being used locally on a connection call, with a result being e.g. before/after cursors to help place it.

The problem is that you'll need to implement some sorting/filtering logic on the client regardless if you want to support optimistic updates.

### Client Side

Another solution would be to allow `rangeBehaviors` to be set to a function. It would receive the newEdge, currentEdges, and any field arguments, and return a new index for the edge:

```js
rangeBehaviors: (newEdge, currentEdges, args) => {
  // Filtering
  if (args.startsWith && !newEdge.name.startsWith(args.startsWith))
    return null;

  // Sorting
  if (args.orderBy) {
    return findIndexForOrderBy(newEdge, currentEdges, args.orderBy);
  }

  // Append/prepend
  return currentEdges.length;
},
```

josephsavona:
Another option is to provide a way for the schema to define how field arguments correlate to sorting logic. For example, in the query `friends(first: 10, orderBy: "name")` the schema might indicate that edges should be sorted by the `name` field. Relay could use this metadata to automatically insert new edges at the correct location (and ensure that the `name` field was always fetched when this sort criteria was applied, so that every edge can be sorted correctly).

There are some cases where this may not be possible - in particular, algorithmically sorted connections such as the Facebook News Feed. In these cases, the schema could simply not define "order by" semantics for any arguments, and the client would have to specify append/prepend logic when adding edges.

These type of semantic schema annotations could apply more generally around connections, for example allowing for arbitrary names for `edges` and `node` fields and `first`/`after` arguments.

cc @dschafer @leebyron

devknoll:
But one of the features of GraphQL that I really like is that it _doesn't_ have notion or semantics for sorting built in, and I'd love for it to stay that way :wink: 

Couldn't renaming connection args just be Relay specific configuration? 

My original client side proposal also had an option for `rangeBehaviors ` that looked like:

```js
rangeBehaviors: {
[...]

PR #340, opened 2015-09-18
Add routing for TodoMVC

### LINKED ISSUE #118: Complete TodoMVC example with routing
There's still some work to be done to complete the TodoMVC example: add routing [as per the spec](https://github.com/tastejs/todomvc/blob/master/app-spec.md#routing).

taion:
Closes #118


examples/todo/js/mutations/AddTodoMutation.js
```diff
@@ -33,6 +29,8 @@ export default class AddTodoMutation extends Relay.Mutation {
       edgeName: 'todoEdge',
       rangeBehaviors: {
         '': 'append',
+        'complete()': 'append',
```
taion: @steveluscher @josephsavona

I don't understand this part and whether it's correct at all.

The `orderby:newest` example on https://facebook.github.io/relay/docs/guides-mutations.html#range-add-example seems to be actively misleading w/r/t the form of calls... that's easy enough to figure out in a debugger.

However, I can't square what I'm seeing with https://github.com/facebook/relay/issues/293#issuecomment-140782479. If I take off the explicit `append` behaviors, I don't see anything happen at all to the `todos` connection. However, even in the `complete(true)` case, I don't see the `todos` connection get re-fetched in its entirety.

If I add the naked `todos` field to the fat query, I instead see the entire `todos` connection get sent down from the server no matter what I set up in my `rangeBehaviors`.

What am I not understanding here? In general, is there a way to conditionally re-fetch the entire connection with a `RANGE_ADD` mutation?

taion: Never mind, I think I figured it out.

taion:
I broke out that last commit just in case I'm doing something wrong in there - would prefer to squash it into the original commit before merging.

taion:
Last update is mostly as mentioned, except:

- I'm using `completed` as the last status to match the route
- I don't think I can correctly simplify the range behaviors specified in the mutation:
  - Even when I specify the default value of `"any"` for `status`, any query on the `todos` field that omits explicitly specifying status gets a range behavior key of `''` (instead of `'status(any)'`)
  - If I require the `status` argument to be non-null, then all the mutations that just specify an unterminated `todos` fail GraphQL validation for omitting status in the fat queries, and I don't know how to correctly set those up
  - As mentioned above, if I don't explicitly specify the falsy range behavior for `status(completed)`, Relay re-fetches all the edges when I add a new todo, which seems undesirable

taion:
Although strictly speaking the `'': 'append'` range behavior doesn't do anything any more in this example, because I don't have any more unqualified queries on `todos`, but it feels more correct than not to still have it there.

steveluscher:
> As mentioned above, if I don't explicitly specify the falsy range behavior for status(completed), Relay re-fetches all the edges when I add a new todo, which seems undesirable

That sounds legit. cc/ @yuzhi

steveluscher:
@facebook-github-bot import

steveluscher:
Super excited to get this landed. Thanks @taion!

taion:
Understanding this range behavior thing is going to be really useful for me.

I squashed the commits, in case you're looking for something without the git autosquash commits.

steveluscher:
It's cool. Our robots autosquash them when they land.

examples/todo/data/database.js
```diff
@@ -39,8 +39,12 @@ export function getTodo(id) {
   return todosById[id];
 }
 
-export function getTodos() {
-  return todoIdsByUser[VIEWER_ID].map((id) => todosById[id]);
+export function getTodos(status = 'any') {
+  var todos = todoIdsByUser[VIEWER_ID].map((id) => todosById[id]);
```
wincent: For consistency, can you drop the parens around `id` here (like below, where you're using an arrow function without parens around the param)?

taion: I can do that. Do you want me to change these two: https://github.com/facebook/relay/blob/58e45dbb9fd08215884597171ce072551f344c06/examples/todo/data/database.js#L74-L75 as well?

examples/todo/js/components/TodoList.js
```diff
@@ -44,23 +44,36 @@ class TodoList extends React.Component {
...
-      fragment on TodoConnection {
+    viewer: () => Relay.QL`
+      fragment on User {
         completedCount,
-        edges {
-          node {
-            complete,
-            id,
-            ${Todo.getFragment('todo')},
+        todos(status: $status, first: 9007199254740991) {
```
wincent: This isn't very clear. Can we use `Number.MAX_SAFE_INTEGER` instead?

Or even:

```
const MAX_SAFE_NUMBER = Number.MAX_SAFE_NUMBER || 9007199254740991;
```

(If you are worried about IE and Safari, where it might not be defined.)

examples/todo/js/components/TodoList.js
```diff
@@ -44,23 +44,36 @@ class TodoList extends React.Component {
 }
 
 export default Relay.createContainer(TodoList, {
+  initialVariables: {
+    status: null,
+  },
+
+  prepareVariables({status}) {
+    let nextStatus;
```
wincent: As these files use `var` throughout, we should probably use it here too for consistency.

Separately (beyond the scope of this commit), we should switch to `const`/`let`. We might build a codemod to do most of the conversion for us automatically.

examples/todo/js/mutations/ChangeTodoStatusMutation.js
```diff
@@ -48,14 +45,11 @@ export default class ChangeTodoStatusMutation extends Relay.Mutation {
...
-      if (this.props.viewer.todos.completedCount != null) {
-        viewerPayload.todos.completedCount = this.props.complete
-          ? this.props.viewer.todos.completedCount + 1
-          : this.props.viewer.todos.completedCount - 1;
-      }
+    var viewerPayload = {id: this.props.viewer.id};
+    if (this.props.viewer.completedCount != null) {
+      viewerPayload.completedCount = this.props.complete
+        ? this.props.viewer.completedCount + 1
+        : this.props.viewer.completedCount - 1;
```
wincent: Style nit: we put these operators (`?` and `:`) in trailing position at the end of the line.

examples/todo/js/components/TodoList.js
taion: Sure - I actually do that in all of my own code - just did it like this to have a minimal change v. https://github.com/facebook/relay/blob/2a86be3e71cdc6511fa994e3de539f72070da1b4/examples/todo/js/components/TodoApp.js#L63

examples/todo/js/components/TodoList.js
taion: Oops. Unintentional oversight on my part.

examples/todo/data/database.js
wincent: Feel free to.

examples/todo/js/mutations/ChangeTodoStatusMutation.js
taion: Huh? I literally didn't change anything but whitespace with the ternary. Do you want all of the cases where this shows up to look like e.g.

```js
      viewerPayload.completedCount = this.props.complete ?
        this.props.viewer.completedCount + 1 :
        this.props.viewer.completedCount - 1;
```

examples/todo/js/components/TodoList.js
wincent: Yeah, may as well take the opportunity to leave things a little better on the way.

taion:
PR updated.

wincent:
Thanks @taion.

Now let's see if this machine wants to obey me:

@facebook-github-bot import

PR #345, opened 2015-09-18
Switched out the Relay playground's bespoke network layer for relay-local-schema

steveluscher:
Note: No relation to “local schema.” It's just a network layer that operates against an in-memory schema.


steveluscher:
@facebook-github-bot import

PR #346, opened 2015-09-18
Fix website/README.md

yungsters:
Do not assume `..` exists as an alias.

### DIFF
--- a/website/README.md
+++ b/website/README.md
@@ -3,7 +3,7 @@
 Install the supporting infrastructure.
 
 ```
-(../; npm install) && (../website-prototyping-tools; npm install) && npm install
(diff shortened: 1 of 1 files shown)


zpao:
Doesn't look like I have the right permissions on Travis but I fixed the failure - jest got an update that assumed it was running in node v4 (specifically it it was using an arrow function) and pushed as a patch update.

steveluscher:
@facebook-github-bot import

yungsters:
asdas

yungsters:
@facebook-github-bot import

yungsters:
@facebook-github-bot import

PR #348, opened 2015-09-18
Configure Travis CI to use node@4.0.0

yungsters:


zpao:
If you just say `4` it'll keep updating across the 4 version family (eg, 4.1 is already out and it would install that). Your call.

yungsters:
Oh, nice!

yungsters:
@facebook-github-bot shipit

PR #350, opened 2015-09-19
Apparently, hsl() hues can't be fractional

steveluscher:


steveluscher:
@facebook-github-bot import

PR #352, opened 2015-09-19
The Webpack build for the prototyping tools now outputs to the console when you publsh the site

steveluscher:


steveluscher:
@facebook-github-bot import

PR #353, opened 2015-09-19
Isolate the homepage examples to their own localStorage

steveluscher:
@facebook-github-bot import


steveluscher:
On second thought. I think I need to add an option that lets you opt *out* of local storage for examples such as these, so that:

1. We don't have to cache bust them when they change
2. You can get yourself out of trouble by refreshing

PR #355, opened 2015-09-19
Prevent hash changes from jumping the page to the top

steveluscher:
When the playground controller sets the hash to `''`, it caused the browser to scroll the page around. This prevents that.


steveluscher:
@facebook-github-bot import

PR #363, opened 2015-09-21
Change getQuery to getFragment in RelayRoute test

kassens:
`getQuery` is deprecated and we're trying to get rid of it.


kassens:
@facebook-github-bot import

PR #364, opened 2015-09-21
Remove unused envify require

vjeux:


vjeux:
cc @steveluscher

kassens:
@facebook-github-bot import

PR #373, opened 2015-09-23
Fix key in rangeBehaviors example

taion:
I don't know if you're looking for a PR to further elaborate on the use of `rangeBehaviors`, especially around when the connection is refreshed, but it seems like it'd be an improvement to fix the form of the range behavior key used in the existing example.


wincent:
Thanks!

@facebook-github-bot import

Issue #376, opened 2015-09-23
dataID.substring called on potentially numerical values


Aweary:
```js
  isClientID: function(dataID: string): boolean {
    return dataID.substring(0, 7) === 'client:';
  },
```

I had an instance where `dataID` was a being returned as a number from the server and this caused Relay to throw an error.

It should either check that it's a string or coerce. 

```js
 return typeof dataID === 'string' && dataID.substring(0, 7) === 'client:';
/* or */
 return (dataId + '').substring(0, 7) === 'client:';
```

Or if a numerical ID is something Relay doesn't want to support it should throw a more descriptive error.
I had a pull request ready using the `typeof` check, but figured I'd check with the contributors first.

josephsavona:
Yes, we should have a better error message to make it clear that `id` values are required to be strings (per the [spec](http://facebook.github.io/relay/docs/graphql-object-identification.html#content)). The place to check would be in the response processing phase - `RelayQueryWriter`, `RelayNodeInterface`, maybe `forEachRootCallArg` (anywhere we grab `result.id` from the raw payload) - and add an `invariant` that `id` is a string.

josephsavona:
@Aweary thanks for filing this, good catch!

Aweary:
Awesome, I'll get on this and get a PR going :+1:

josephsavona:
This was addressed in c54d49d62ac05095fe3fea8325aa2d92c7ca41e8

PR #379, opened 2015-09-24
Verify recordID is a string

Aweary:
Adds a check that `recordID` is a string, fixes https://github.com/facebook/relay/issues/376

### DIFF
--- a/src/store/RelayQueryWriter.js
+++ b/src/store/RelayQueryWriter.js
@@ -84,6 +84,15 @@ class RelayQueryWriter extends RelayQueryVisitor<WriterState> {
     responseData: mixed,
     path: RelayQueryPath
   ): void {
+
+
+    invariant(
+      typeof recordID === 'string',
+      'RelayQueryWriter: Expected id `%s` to be string' +
+      ', instead received a %s',
+      recordID, typeof recordID
+    );
+
     var state = {
       nodeID: null,
       recordID,
--- a/src/store/RelayQueryWriter.js
+++ b/src/store/RelayQueryWriter.js
@@ -84,15 +84,6 @@ class RelayQueryWriter extends RelayQueryVisitor<WriterState> {
     responseData: mixed,
     path: RelayQueryPath
   ): void {
-
-
-    invariant(
-      typeof recordID === 'string',
-      'RelayQueryWriter: Expected id `%s` to be string' +
-      ', instead received a %s',
-      recordID, typeof recordID
-    );
-
     var state = {
       nodeID: null,
       recordID,
--- a/src/traversal/writeRelayQueryPayload.js
+++ b/src/traversal/writeRelayQueryPayload.js
@@ -19,6 +19,8 @@ import type RelayQuery from 'RelayQuery';
 var RelayQueryPath = require('RelayQueryPath');
 import type RelayQueryWriter from 'RelayQueryWriter';
 
+var invariant = require('invariant');
+
 /**
  * @internal
  *
@@ -33,8 +35,17 @@ function writeRelayQueryPayload(
   var store = writer.getRecordStore();
   var path = new RelayQueryPath(query);
 
+
   RelayNodeInterface.getResultsFromPayload(store, query, payload)
     .forEach(({dataID, result}) => {
+      invariant(
+        typeof dataID === 'string',
+        'writeRelayQueryPayload: expected ID on the payload for query `%s` to be string' +
+        ', instead received the `%s` `%s`.',
+        path.getName(),
+        typeof dataID,
+        dataID
+      );
       writer.writePayload(query, dataID, result, path);
     });
 }
--- a/src/traversal/writeRelayUpdatePayload.js
+++ b/src/traversal/writeRelayUpdatePayload.js
@@ -76,6 +76,15 @@ function writeRelayUpdatePayload(
   payload: {[key: string]: mixed},
   {configs, isOptimisticUpdate}: UpdateOptions
 ): void {
+  invariant(
+    typeof payload[CLIENT_MUTATION_ID] === 'string',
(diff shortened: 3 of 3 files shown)


src/store/RelayQueryWriter.js
```diff
@@ -84,6 +84,15 @@ class RelayQueryWriter extends RelayQueryVisitor<WriterState> {
...
   ): void {
+
+
+    invariant(
+      typeof recordID === 'string',
+      'RelayQueryWriter: Expected id `%s` to be string' +
+      ', instead received a %s',
+      recordID, typeof recordID
+    );
+
```
josephsavona: `recordID` is typed here as `DataID` which is a string - can you move this invariant to the point where we read the ID out from the actual payload? This will also help to provide contextual error messages (e.g. that the id on the payload for query foo is a number, or that the id for the deleted field on a mutation is incorrect, etc.). `writeRelayQueryPayload` and `writeRelayUpdatePayload` are the two callers of this class.

josephsavona:
Awesome - thanks for working on this! Per comments, the `invariant` check should be wherever the `id` value is pulled from the payload object.

Aweary:
Sorry about that, still wrapping my head around the Relay source! Looking at https://github.com/facebook/relay/blob/2a86be3e71cdc6511fa994e3de539f72070da1b4/src/traversal/writeRelayQueryPayload.js#L36-L39, which is where I think you're recommending the `invariant` check be placed for queries, where can I pull the query name from?

Aweary:
Sorry about that @josephsavona, still wrapping my head around the Relay source. I added the checks to `writeRelayUpdatePayload` and `writeRelayQueryPayload`. Let me know if anything else looks off!

src/traversal/writeRelayQueryPayload.js
```diff
@@ -33,8 +34,15 @@ function writeRelayQueryPayload(
...
   var path = new RelayQueryPath(query);
 
+
   RelayNodeInterface.getResultsFromPayload(store, query, payload)
     .forEach(({dataID, result}) => {
+      invariant(
+        typeof dataID === 'string',
+        'writeRelayQueryPayload: expected ID on the payload for query %s to be string' +
+        ', instead received the %s %s',
+        path.getName(), typeof dataID, dataID
```
josephsavona: A few nits on formatting:
- all args on their own line
- in the template string, wrap `%s` with backticks so `` `%s` ``
- ending period on the sentence

src/traversal/writeRelayUpdatePayload.js
```diff
@@ -76,6 +76,13 @@ function writeRelayUpdatePayload(
   payload: {[key: string]: mixed},
   {configs, isOptimisticUpdate}: UpdateOptions
 ): void {
+  invariant(
+  typeof payload.clientMutationID === 'string',
+  'writeRelayUpdatePayload: expected clientMutationID on the payload for mutation %s to be string' +
+  ', instead received the %s %s',
+  operation.getCall().name, typeof payload.clientMutationID, payload.clientMutationID
+);
+
```
josephsavona: the same applies here, but use `payload[CLIENT_MUTATION_ID]` - internally at FB the key is `clientMutationId`, in OSS it's `client_mutation_id`, and this property allows us to use the correct one for the environment

josephsavona:
Nice! Bear with me on the back & forth, I just want to help you make this PR even better. There are a few other places in `writeRelayUpdatePayload` that access IDs - for example `deleteRecord` should check that its input `recordID` is a string, and everywhere else in that file where we have the form `payload[ID]`.

Aweary:
@josephsavona I really appreciate, this is helping me get a better handle how Relay works! I added some additional `invariant` checks for the ID in `writeRelayUpdatePayload` and reformatted per your recommendations.

src/traversal/writeRelayUpdatePayload.js
```diff
@@ -76,6 +76,15 @@ function writeRelayUpdatePayload(
   payload: {[key: string]: mixed},
   {configs, isOptimisticUpdate}: UpdateOptions
 ): void {
+  invariant(
+    typeof payload[CLIENT_MUTATION_ID] === 'string',
+    'writeRelayUpdatePayload: expected clientMutationID on the payload for mutation `%s` to be string' +
+    ', instead received the `%s` `%s`.',
+    operation.getCall().name,
```
Aweary: Is there a better way to get the mutations's name? This seems to be causing some tests to fail.

src/traversal/writeRelayQueryPayload.js
```diff
@@ -18,6 +18,7 @@ var RelayProfiler = require('RelayProfiler');
 import type RelayQuery from 'RelayQuery';
 var RelayQueryPath = require('RelayQueryPath');
 import type RelayQueryWriter from 'RelayQueryWriter';
+var invariant = require('invariant');
```
josephsavona: nit: add a newline between the class and function imports

src/traversal/writeRelayUpdatePayload.js
josephsavona: you can use `operation.getName()` instead

Aweary: I tried that and was getting the route name I believe, I'll have to double check

josephsavona:
@Aweary I got sidetracked but will pull this in first thing next week - thanks again for working on this!

Aweary:
@josephsavona no problem, thanks for bearing with me on it!

josephsavona:
@facebook-github-bot import

josephsavona:
@Aweary Just a heads up that I'm making a slightly more extensive version of this change, using Flow types to find places that need invariants. So be patient!

Aweary:
@josephsavona sounds great, thanks for the update!

PR #380, opened 2015-09-24
Change from error to warning when a request fails

### LINKED ISSUE #366: Using `console.error()` for Query Failures causes red screen of death on React Native
https://github.com/facebook/relay/commit/a0d80d5f4c9802b5eb6bda4a5d3d2855fb301197

This commit changed to using `console.error()` to report query failures. In a browser, hidden away in dev tools, this doesn't really matter. On React Native however, in development specifically, using `console.error()` rather than `console.log()` causes a red screen of death. This seems like incorrect behavior for relay, as it prevents the developer from handling query errors (such as an unavailable server) gracefully.

kassens:
Since React Native shows a big red box for `console.error` it can get in the way
of development when working on error handling.

Fixes #366.

### DIFF
--- a/src/store/RelayPendingQueryTracker.js
+++ b/src/store/RelayPendingQueryTracker.js
@@ -207,7 +207,7 @@ class PendingFetch {
     var queryID = this.getQuery().getID();
     delete pendingFetchMap[queryID];
 
-    console.error(error.message);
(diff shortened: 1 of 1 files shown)


kassens:
@facebook-github-bot import

PR #382, opened 2015-09-24
Fixed example projects building Relay twice on npm install

### LINKED ISSUE #367: `npm install` on the examples should only build Relay once
If you `npm install` from any of the projects in `examples/`, Relay will be built (with Gulp/Webpack) twice. It should only be built once.

michaelchum:
Fixes #367 where `npm install` on the projects in `examples/` would build twice. This is due to the nature of `npm install` which recursively executes `npm install --production` on parent directories, including `relay/`. 

Since the `preinstall` script of the projects already executes `npm install` on `relay/`, `npm install` runs twice on `relay/` therefore running the `build` script of `relay/` twice.

Adding `--ignore-scripts` to the `preinstall` script of the projects fixes this by not executing any scripts in the first `npm install` on `relay/`.

### DIFF
--- a/examples/relay-treasurehunt/package.json
+++ b/examples/relay-treasurehunt/package.json
@@ -3,7 +3,7 @@
   "scripts": {
     "start": "babel-node ./server.js",
     "update-schema": "babel-node ./scripts/updateSchema.js",
-    "preinstall": "cd ../.. && npm install"
+    "preinstall": "cd ../.. && npm install --ignore-scripts"
   },
   "dependencies": {
     "babel": "5.8.23",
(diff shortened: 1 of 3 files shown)


wincent:
Thanks!

@facebook-github-bot import

kassens:
@michaelchum According to the summary, wouldn't it be enough if we just removed the preinstall line (since install already installs parent directories?)

michaelchum:
@kassens I've done a bit of research before the commit :) Because the automatic `npm install --production` on parent directory `relay/` only installs `dependencies` and ignores `devDependencies`, it will not build since `gulp` (and some other tools required for build) are only in `devDependencies`. 

We need to manually go to `relay/` and run `npm install` at least once in order have `gulp` installed so that any subsequent `npm install` or `npm run build` commands would build Relay.

If the `preinstall` script is removed, we'd have to tell the user to `npm install` in `relay` first before installing projects in `example/`.

kassens:
Thanks for doing the research, I don't have much experience with npm.

@wincent ship it

Issue #381, opened 2015-09-24
pageInfo does not get pulled in when edges are only included via fragment


taion:
This looks like a bug in babel-relay-plugin. The logic here: https://github.com/facebook/relay/blob/615787490db3bd62b15e8bf22d9c46b2d59bbf73/scripts/babel-relay-plugin/src/GraphQLPrinter.js#L503 for pulling in `pageInfo` on connections doesn't seem to work when the connection `edges` are only pulled in via a fragment.

For example, in the current (non-routing) TodoMVC example, if you remove these lines: https://github.com/facebook/relay/blob/58e45dbb9fd08215884597171ce072551f344c06/examples/todo/js/components/TodoApp.js#L64-L68 (the data fetched there isn't actually used by the `TodoApp` component at all, so naively that part of the GraphQL query seems vestigial), you can see that the GraphQL query sent to the server does not request `pageInfo`, and correspondingly does not request `hasNextPage` or `hasPreviousPage`, and consequently, Relay assumes that those fields are false, and pagination doesn't work (or wouldn't work, in more elaborate cases that involve pagination).

josephsavona:
Good find! Fix it? ;-)

josephsavona:
Re the fix: it seems reasonable to always add `pageInfo` on a connection field if it isn't already present **and** one of the following applies:
- the `edges` field is present
- a fragment reference is present (which is likely enough to contain `edges` that we should fetch `pageInfo`).

Put another way, the only time we shouldn't add `pageInfo` is if only connection-level fields are being fetched (ex: `storiesConnection { totalCount }` doesn't need page info)

taion:
I have literally never looked at a Babel plugin before. But fine, challenge accepted.

I would think the right way to handle this would just be to extend the `pageInfo` check to cases where a fragment is on a connection type, though, rather than when a field is on a connection. Is there some reason why this would not be a good approach?

kassens:
We're adding the `pageInfo` during the static transform of the code [here in GraphQLPrinter](https://github.com/facebook/relay/blob/58e45dbb9fd08215884597171ce072551f344c06/scripts/babel-relay-plugin/src/GraphQLPrinter.js#L502-L504). During the transform, we can only look at the current file / GraphQL fragment we're transforming. This means we can't inspect what's inside fragments.

[...]

PR #385, opened 2015-09-24
Move adding pageInfo for edges up one level

taion:
Potential fix for #381.

### DIFF
--- a/scripts/babel-relay-plugin/src/GraphQLPrinter.js
+++ b/scripts/babel-relay-plugin/src/GraphQLPrinter.js
@@ -292,9 +292,9 @@ function printFieldsAndFragments(
         fields.push(selection);
 
         if (
-          isConnectionType(type) &&
+          getConnectionMetadataForType(type) &&
           getName(selection) === 'edges' &&
-          !!type.getFields()['pageInfo']
+          type.getFields()['pageInfo']
         ) {
           requisiteFields.pageInfo = true;
         }
@@ -455,7 +455,7 @@ function printField(
     metadata.pk = 'id';
   }
 
-  var connectionMetadata = getConnectionMetadata(options.schema, fieldDecl);
+  var connectionMetadata = getConnectionMetadata(fieldDecl);
   if (connectionMetadata) {
     metadata.connection = true;
 
@@ -785,15 +785,21 @@ function getArgNamed(field, name) {
   return remaining.length === 1 ? remaining[0] : null;
 }
 
-function getConnectionMetadata(schema, fieldDecl) {
-  if (!isConnectionType(fieldDecl.type)) {
-    return null;
-  }
+function getConnectionMetadata(fieldDecl) {
   // Connections must be limitable.
   if (!getArgNamed(fieldDecl, 'first') && !getArgNamed(fieldDecl, 'last')) {
     return null;
   }
-  var fieldType = types.getNamedType(fieldDecl.type);
+
+  return getConnectionMetadataForType(fieldDecl.type);
+}
+
+function getConnectionMetadataForType(fieldDeclType) {
+  if (!isConnectionType(fieldDeclType)) {
+    return null;
+  }
+
+  var fieldType = types.getNamedType(fieldDeclType);
 
   // Connections must have a non-scalar `edges` field.
   var edgesField = fieldType.getFields()['edges'];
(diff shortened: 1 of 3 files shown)


taion:
This is what I meant with https://github.com/facebook/relay/issues/381#issuecomment-142979970.

I believe this solution is more correct in when it requests `pageInfo`. Two things worth noting:
- Compared to https://github.com/facebook/relay/issues/381#issuecomment-142967061, this can potentially generate larger queries on the wire, in the event that there are multiple fragments on a given connection type that each request `pageInfo`, since each fragment will embed the `pageInfo` query
- I switched the check for the type being a connection for adding `pageInfo` from `getConnectionMetadata` to just `isConnectionType` - not sure if this would be a problem

scripts/babel-relay-plugin/src/GraphQLPrinter.js
```diff
@@ -499,9 +506,6 @@ function printField(
         ));
       }
     });
-    if (hasEdgesSelection && !!fieldDecl.type.getFields()['pageInfo']) {
-      subRequisiteFields.pageInfo = true;
-    }
   } else if (types.getNamedType(fieldDecl.type).name === 'PageInfo') {
```
josephsavona: Not a blocker for this PR, but I wonder if these other checks (that add `subRequisiteFields`) should also be moved up a level.

scripts/babel-relay-plugin/src/GraphQLPrinter.js
```diff
@@ -290,6 +290,14 @@ function printFieldsAndFragments(
         fragments.push(printQueryFragment(selection, options));
       } else if (selection.kind === kinds.FIELD) {
         fields.push(selection);
+
+        if (
+          isConnectionType(type) &&
```
josephsavona: this should probably be `!!getConnectionMetadata(...)` in order to avoid adding `pageInfo` on fields that look like connections but don't follow the Relay connection spec.

scripts/babel-relay-plugin/src/GraphQLPrinter.js
taion: I thought about this a little. I believe it's unnecessary.

If you have a `fragment on PageInfo`, you must have a fragment or field on the connection above it, which already would have included the `pageInfo` field, and then pulled in `hasPreviousPage` and `hasNextPage` via this logic. (Also, whatever component has the `fragment on PageInfo` is unlikely to itself care anyway. Same holds with the cursor stuff.

The `edges`/`pageInfo` thing is special because it's a sibling requisite field rather than a child; that's what caused my initial issue.

scripts/babel-relay-plugin/src/GraphQLPrinter.js
taion: I can do that if you think this is a reasonable direction for the fix. BTW, in this context, what is the `!!` for?

scripts/babel-relay-plugin/src/GraphQLPrinter.js
josephsavona: Good point!

scripts/babel-relay-plugin/src/GraphQLPrinter.js
kassens: `!!` is a common pattern for "cast to boolean". It's nice to have boolean expressions in conditionals, but not required.

```javascript
null && true // null
!!null && true // false

josephsavona: Yes, let's use `getConnectionMetadata()` (and no need for `!!var`, I've been over-using it to ensure that a value is a boolean - useful with Flow)

taion: On closer inspection, I don't think this is possible. For a fragment, I don't actually have access to the field, just the type, so I can't make these assertions because I don't have access to the field args when looking at just the fragment: https://github.com/facebook/relay/blob/d2c821efb6f9620c7e3597ad92b3e1e1a838829e/scripts/babel-relay-plugin/src/GraphQLPrinter.js#L793-L795.

What I can do is to lift these into a separate function: https://github.com/facebook/relay/blob/d2c821efb6f9620c7e3597ad92b3e1e1a838829e/scripts/babel-relay-plugin/src/GraphQLPrinter.js#L796-L826 and use that instead. Does that sound like a workable approach?

taion:
Latest commit pulls out the type-only portions of `getConnectionMetadata` and uses that for adding `pageInfo` as a requisite field.

taion:
(the `schema` arg to `getConnectionMetadata` wasn't used at all, so I removed it)

josephsavona:
@facebook-github-bot import

josephsavona:
@taion thanks for working on this!

PR #393, opened 2015-09-26
Build 2x faster

zpao:
I was testing some fbjs changes and got tired of waiting ~23s for each build so I made it faster.

1. Removed duplicate `modules` task. We already run it for `dist` and `dist:min` (where it gets deduped) so we don't need to run it before then.
  
  *(saves ~5s)*
2. Removed the babel webpack loader. We've already transformed our code. `fbjs` is already transformed. We shouldn't need to transform again. There are some differences (https://gist.github.com/zpao/16e13b4a85010c6fbdff) as a result of this - mostly because we don't reprint, though there are also a couple `'use strict'` statements missing but that's probably ok.
  
  *(saves ~6-7s for each `dist` and `dist:min` but they run in parallel so only see the win once)*

I ran a few times before and after to get some stable numbers with a warm disk cache. I dropped from 23s to 12s.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -63,11 +63,6 @@ var buildDist = function(opts) {
       'react': 'React',
       'react-dom': 'ReactDOM'
     },
-    module: {
-      loaders: [
-        {test: /\.js$/, loader: 'babel'}
-      ],
-    },
     output: {
       filename: opts.output,
       libraryTarget: 'umd',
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -63,11 +63,6 @@ var buildDist = function(opts) {
       'react': 'React',
       'react-dom': 'ReactDOM'
     },
-    module: {
-      loaders: [
-        {test: /\.js$/, loader: 'babel'}
-      ],
-    },
```
zpao: Ok, I'll come clean…

The real reason I found this was because I couldn't build relay after making changes to fbjs because babel didn't like transforming the code it generated and I got this:

```
ERROR in Loader /Users/poshannessy/FB/code/fbjs/node_modules/babel/index.js didn't return a function
 @ ../fbjs/lib/fetchWithRetries.js 26:27-60
```

kassens:
Thanks!

@facebook-github-bot import

PR #401, opened 2015-09-28
Remove last item in GraphQLList does not broadcast update

quazzie:
If you remove the last item in a GraphQLList it does not broadcast the update and the local version is not changed.

### DIFF
--- a/src/store/RelayQueryWriter.js
+++ b/src/store/RelayQueryWriter.js
@@ -552,7 +552,10 @@ class RelayQueryWriter extends RelayQueryVisitor<WriterState> {
     this._store.putLinkedRecordIDs(recordID, storageKey, nextLinkedIDs);
 
     // Check if length has changed
-    isUpdate = isUpdate || (prevLinkedIDs && prevLinkedIDs.length || 0) !== nextLinkedIDs.length;
+    isUpdate = (
+      isUpdate || 
+      (prevLinkedIDs && prevLinkedIDs.length || 0) !== nextLinkedIDs.length
+    );
 
     // Only broadcast a list-level change if a record was changed/added
     if (isUpdate) {
--- a/src/store/RelayQueryWriter.js
+++ b/src/store/RelayQueryWriter.js
@@ -552,7 +552,7 @@ class RelayQueryWriter extends RelayQueryVisitor<WriterState> {
     this._store.putLinkedRecordIDs(recordID, storageKey, nextLinkedIDs);
 
     // Only broadcast a list-level change if a record was changed/added
-    if (isUpdate) {
+    if (isUpdate || prevLinkedIDs.length !== nextLinkedIDs.length) {
       this.recordUpdate(recordID);
     }
   }
--- a/src/store/RelayQueryWriter.js
+++ b/src/store/RelayQueryWriter.js
@@ -551,8 +551,11 @@ class RelayQueryWriter extends RelayQueryVisitor<WriterState> {
 
     this._store.putLinkedRecordIDs(recordID, storageKey, nextLinkedIDs);
 
+    // Check if length has changed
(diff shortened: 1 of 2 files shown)


quazzie:
cla signed

yuzhi:
Can you add a unit test that tests this to ensure futures changes won't break this?

quazzie:
Sorry, I've never written a unit test and i can't even figure out where you test the function i changed.

yuzhi:
No worries! We can work through that if you don't mind. There is a first time for everything. 
The best place to add a unit test for what you added would be here: 
https://github.com/quazzie/relay/blob/patch-2/src/traversal/__tests__/writeRelayQueryPayload_pluralLinkedField-test.js

There are already a few examples for tests that writes to a plural linked field. You can use them as template. All you have to do is write to the field with some value and then write to it again with a shorter list of results and make sure the node with the update list shows up in the list of the updated nodes (something like line 62-67 of the above file).

To test it, you can do `npm test <path to test>`
Hope that helps! Let me know if you need more pointers.

quazzie:
After lots of trial and error i figured out i could not use windows as dev environment... 
But i think this is a correct test. Fails without fix and completes with it.

src/store/RelayQueryWriter.js
```diff
@@ -552,7 +552,7 @@ class RelayQueryWriter extends RelayQueryVisitor<WriterState> {
     this._store.putLinkedRecordIDs(recordID, storageKey, nextLinkedIDs);
```
yuzhi: Can you move the logic here so that it's more clear what is stored into the `isUpdated` variable and we only need to check one thing?
`isUpdate = isUpdate ||  prevLinkedIDs.length !== nextLinkedIDs.length`

src/traversal/__tests__/writeRelayQueryPayload_pluralLinkedField-test.js
```diff
@@ -204,6 +204,60 @@ describe('writeRelayQueryPayload()', () => {
         .toBe(newPhone.phoneNumber.countryCode);
     });
 
+    it('updates if length changes', () => {
```
yuzhi: Looks good!

src/traversal/__tests__/writeRelayQueryPayload_pluralLinkedField-test.js
```diff
@@ -204,6 +204,60 @@ describe('writeRelayQueryPayload()', () => {
...
+      var results = writePayload(store, query, payload);
+      expect(results).toEqual({
+        created: {},
+        updated: {
+          '123': true
+        }
+      });
+      var phoneIDs = store.getLinkedRecordIDs('123', 'allPhones');
+      expect(phoneIDs).toEqual(['client:1']);
+    });
```
yuzhi: Woot!

josephsavona:
nice!

quazzie:
When i broke out the logic flow complained about possibly undefined value.
Flow did not find this when logic was in ```if (isUpdate || ...)```

Added a check if prevLinkedIDs is defined.

src/store/RelayQueryWriter.js
```diff
@@ -551,6 +551,9 @@ class RelayQueryWriter extends RelayQueryVisitor<WriterState> {
 
     this._store.putLinkedRecordIDs(recordID, storageKey, nextLinkedIDs);
 
+    // Check if length has changed
+    isUpdate = isUpdate || (prevLinkedIDs && prevLinkedIDs.length || 0) !== nextLinkedIDs.length;
```
yuzhi: Sorry, one more super small nit: can you reformat this to fit in than 80 chars?
```
isUpdate = isUpdate ||
  (prevLinkedIDs && prevLinkedIDs.length || 0) !== nextLinkedIDs.length;
```

I'll merge it after that tweak.

PR #404, opened 2015-09-29
Synchronize examples with starter kit, using explicit imports

wincent:
Brings across:

- https://github.com/relayjs/relay-starter-kit/pull/36
- https://github.com/wincent/relay-starter-kit/pull/1

from the starter kit. Benefit here is explicitness and robustness, but the cost is that the webpack bundle may take longer to build (I'm going to look at tweaking the config though to not reparse React/Relay when bundling, if that can be done without [causing problems](http://stackoverflow.com/questions/32708927/how-do-you-move-from-stock-relay-to-relay-with-a-router)).

Let me know if this seems silly.


zpao:
I think it's a great idea - maps better to what people would be more likely to do if they are using webpack.

kassens:
This is great!

wincent:
@facebook-github-bot import

PR #405, opened 2015-09-30
Compile babel-relay-plugin using Babel

yungsters:
Changes `babel-relay-plugin` to be compiled via Babel. This will let us start using ES2015 features without having to require all users to use node v4+.

Tested by removing all `node_modules` and `lib` directories and running installation steps on the `README.md`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,9 @@
-dist/
-lib/
+.nvmrc
 node_modules
 npm-debug.log
-website/build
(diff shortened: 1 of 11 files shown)


yungsters:
Due to limitations internally (cc @cpojer @amasad), I am going to change this to instead check in the build products — `lib/`.

josephsavona:
:+1:

Issue #408, opened 2015-10-01
Invariant Violation: RelayQueryField() in todo example


reginedanica:
I tried to remove items in the todo list and I got this error 

'Uncaught Error: Invariant Violation: RelayQueryField(): Expected field `id` to be annotated with the type of the parent field.'

wincent:
Thanks for the report, @reginedanica. What version of the code are you running? I can't repro this on the current `HEAD` of master.

Maybe you are running an old version of the babel-relay-plugin. Try running `npm install && npm run update-schema && npm run start` in the todo example and see if you can repro.

zigomir:
@wincent I have same problem. I'm running 1844dc940b46021befd7703b574ddbbcee8656f8 atm. 
Running `npm run update-schema` will fill few `description` fields inside `schema.json`.

N0hbdy:
Also, at fe3e78f7fd4f9f16c883136a05efbb83a82d82b9 I'm unable to add a new todo, probably for the same reason?

pcarion:
i have also this same error on the current HEAD - I *did* run:

```bash
npm install && npm run update-schema && npm run start
```

The error is due to an invariant at:
https://github.com/facebook/relay/blob/master/src/query/RelayQuery.js#L890

under debugger, it appears that 
`this.__concreteNode__` is:
```js
GraphQLFieldNode {
  kind: "Field", 
__proto__: GraphQLFieldNode,
  fieldName: "id"…}
```
but `this.__concreteNode__.metadata` is:
```
edgesID: undefined
inferredPrimaryKey: undefined
inferredRootCallName: undefined
isConnection: false
isFindable: false
isGenerated: false
isPlural: false
isRequisite: false
isUnionOrInterface: false
parentType: undefined
```
and so that fails the invariant in:
```javascript
getParentType(): string {
    var parentType = this.__concreteNode__.metadata.parentType;
    invariant(
      parentType,
      'RelayQueryField(): Expected field `%s` to be annotated with the ' +
      'type of the parent field.',
      this.getSchemaName()
    );
    return parentType;
  }
```

steveluscher:
Yep! We broke this. @josephsavona is on it!

josephsavona:
Oops! Looking at it.

pcarion:
I can confirm that the TodoMVC app is working now. Thanks!

josephsavona:
@pcarion thanks for double-checking ;-)

reginedanica:
Thanks a lot guys!

PR #409, opened 2015-10-01
Don't require babel to be installed globally

wincent:
While investigating an issue reported for the todomvc example
(https://github.com/facebook/relay/issues/408), I tried to repro as follows:

```
git checkout master
git fetch upstream
git merge upstream/master
npm install && npm run update-schema && npm run start
```

But it went boom during the `npm install` because I didn't have `babel`
installed globally:

```
sh: babel: command not found
```

(Full output at https://gist.github.com/wincent/3221694ee3fdc46084f4)

What's happening here:

- example's `preinstall` tries to run `install` from the top-level
- top-level `install` tries to install babel-relay-plugin dependency
- babel-relay-plugin `build` step runs and tries to shell out to `babel`
- `babel` call will blow up if not globally installed, even though babel
  is listed as a dev dependency

Workaround:

```
npm install -g babel
```

Fix employed here: we add `npm install --ignore-scripts` to the
babel-relay-plugin `build` recipe, causing it to become available in
`node_modules` and therefore the `$PATH` when the `babel` command later
runs. Note that without the `--ignore-scripts` we get into an infinite
loop, although I am not sure of the exact chain of cause and effect
because all we see on the console at that point is this, endlessly
repeated:

```
> babel-relay-plugin@0.2.6 build /Users/glh/code/relay/scripts/babel-relay-plugin
> rm -rf lib; npm install; npm run babel
```

Also note the use of `;` instead of `&&` in the interest of preserving
potential future compatibility with Windows.

I think this change should be ok for end users or the
babel-relay-plugin, as they don't end up running the `build` script and
instead just get whatever we built into `lib/`.

Test Plan: Blow away global babel (`npm uninstall -g babel`) and example
`node_modules` and then `npm install` from inside example or top-level,
see it no longer blows from.

### DIFF
--- a/examples/relay-treasurehunt/package.json
+++ b/examples/relay-treasurehunt/package.json
@@ -3,7 +3,7 @@
   "scripts": {
     "start": "babel-node ./server.js",
     "update-schema": "babel-node ./scripts/updateSchema.js",
-    "preinstall": "cd ../.. && npm install --ignore-scripts"
+    "preinstall": "cd ../.. && npm install --ignore-scripts && cd scripts/babel-relay-plugin && npm install --ignore-scripts"
   },
   "dependencies": {
     "babel": "5.8.23",
--- a/examples/star-wars/package.json
+++ b/examples/star-wars/package.json
@@ -3,7 +3,7 @@
   "scripts": {
     "start": "babel-node ./server.js",
     "update-schema": "babel-node ./scripts/updateSchema.js",
-    "preinstall": "cd ../.. && npm install --ignore-scripts"
+    "preinstall": "cd ../.. && npm install --ignore-scripts && cd scripts/babel-relay-plugin && npm install --ignore-scripts"
   },
   "dependencies": {
     "babel": "5.8.23",
--- a/examples/todo/package.json
+++ b/examples/todo/package.json
@@ -3,7 +3,7 @@
   "scripts": {
     "start": "babel-node ./server.js",
     "update-schema": "babel-node ./scripts/updateSchema.js",
-    "preinstall": "cd ../.. && npm install --ignore-scripts"
+    "preinstall": "cd ../.. && npm install --ignore-scripts && cd scripts/babel-relay-plugin && npm install --ignore-scripts"
   },
   "dependencies": {
     "babel": "5.8.23",


wincent:
CC @yungsters who knows this stuff; I might be unaware of some constraints.

yungsters:
>  Note that without the --ignore-scripts we get into an infinite loop, although I am not sure of the exact chain of cause and effect [...]

According to [npm-build(1)](https://docs.npmjs.com/cli/build), the `build` script is a plumbing command called by `npm install` and `npm link`. So if the `build` script calls `npm install`, you have an infinite loop.

----

My understanding is that if you have a `file://...` dependency in your `package.json` (like our examples do), it won't install the dependency's dependencies. This is why `babel-relay-plugin`'s dependency on `babel` is not installed, which is what causes the problem raised by this pull request.

If you take a look at the example's `preinstall` script, you'll notice it works around this by installing the root directory's dependencies:

```
"preinstall": "cd ../.. && npm install --ignore-scripts",
```

The `--ignore-scripts` flag means `npm install` should install the dependencies without running the `build` script. Why's that? Well, it appears that running `npm install` *does* run the `build` script of `file://...` dependencies. If we left out `--ignore-scripts`, we would end up building the `react-relay` package twice (which is not wrong, just unnecessarily time consuming).

> **NOTE:**
> In my opinion, npm's behavior around `file://...` dependencies is kind of messed up.

----

So back to the problem at hand — `babel` is a dependency of `babel-relay-plugin`, which is a `file://...` dependency of the examples. The reason that `babel` has an error when installing an example is that the `babel-relay-plugin`'s `build` script is being executed without first installing its dependencies.

I don't think that changing `babel-relay-plugin`'s `build` script is the right fix. This would be the equivalent of replacing the example's existing `preinstall` script by adding `npm install --ignore-scripts` to `react-relay`'s `build` script, which is definitely wrong.

Instead, it seems that any package that has a `file://...` dependency needs to ensure that dependency's package dependencies are installed using `npm install --ignore-scripts`. So instead, we can just change the example's preinstall script to *also* correctly handle `babel-relay-plugin`:

```
"preinstall": "cd ../.. && npm install --ignore-scripts && cd scripts/babel-relay-plugin && npm install --ignore-scripts",
```

wincent:
Whoops, I accidentally closed this. I'll fix it up and resubmit.

wincent:
Ok, I think this should be good to go now.

@facebook-github-bot import

PR #412, opened 2015-10-01
Fix linting issues

enaqx:
Also suggestions:
* move eslint rules from dependency to relay own .eslintrc like in [graphql-js](https://github.com/graphql/graphql-js/blob/master/.eslintrc)
* include lint step in test

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -16,7 +16,3 @@ globals:
   IteratorResult: true
   ReactClass: true
   ReactElement: true
-
-rules:
-  # allow use constructors for side-effects
-  no-new: 0
--- a/.eslintrc
+++ b/.eslintrc
@@ -16,3 +16,7 @@ globals:
   IteratorResult: true
   ReactClass: true
   ReactElement: true
+
+rules:
+  # allow use constructors for side-effects
+  no-new: 0
--- a/src/traversal/__tests__/diffRelayQuery_connection-test.js
+++ b/src/traversal/__tests__/diffRelayQuery_connection-test.js
@@ -9,6 +9,8 @@
  * @emails oncall+relay
  */
 
+/* eslint-disable indent */
+
 'use strict';
 
 var RelayTestUtils = require('RelayTestUtils');
--- a/src/traversal/diffRelayQuery.js
+++ b/src/traversal/diffRelayQuery.js
@@ -11,8 +11,6 @@
  * @typechecks
  */
 
-/* eslint-disable consistent-return */
-
 'use strict';
 
 var GraphQLStoreDataHandler = require('GraphQLStoreDataHandler');
@@ -621,10 +619,12 @@ class RelayDiffQueryBuilder {
     // The returned `trackedNode` is never tracked directly: instead it serves
     // as an indicator to `diffConnection` that the entire connection field must
     // be tracked.
+    /* eslint-disable consistent-return */
     return {
       diffNode: null,
       trackedNode: hasSplitQueries ? edgeField : trackedNode,
     };
+    /* eslint-enable consistent-return */
   }
 }
 
--- a/src/traversal/diffRelayQuery.js
+++ b/src/traversal/diffRelayQuery.js
@@ -11,6 +11,8 @@
  * @typechecks
  */
 
+/* eslint-disable consistent-return */
+
 'use strict';
 
 var GraphQLStoreDataHandler = require('GraphQLStoreDataHandler');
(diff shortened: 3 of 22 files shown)


zpao:
> move eslint rules from dependency to relay own .eslintrc

No, the point of the shared config is that we can share that across our projects and enforce the same rules :)

> include lint step in test

Yea, that should be done (though personally I don't know that `npm test` should fail but having it be part of the matrix on travis ci makes sense)

src/traversal/printRelayOSSQuery.js
```diff
@@ -246,7 +246,7 @@ function printDirectives(node) {
   return ' ' + directiveStrings.join(' ');
 }
 
-function printDirective({name,value}) {
+function printDirective({ name, value }) {
```
zpao: We actually tend to prefer this without the space padding the braces (there's another case where you added that)

src/traversal/diffRelayQuery.js
```diff
@@ -11,6 +11,8 @@
  * @typechecks
  */
 
+/* eslint-disable consistent-return */
```
zpao: Thoughts on just fixing the 1 instance instead? I'm not completely sold that it's the most valuable rule but I'd like to make that decision on a broader level and try to avoid file-scope disables.

src/traversal/__tests__/diffRelayQuery_connection-test.js
```diff
@@ -9,6 +9,8 @@
  * @emails oncall+relay
  */
 
+/* eslint-disable indent */
```
zpao: Again, I'd really like to avoid disables like this. Fixing this file is a bit noisy but if we can time it with making lint fail tests then it's worth it.

.eslintrc
```diff
@@ -16,3 +16,7 @@ globals:
   IteratorResult: true
   ReactClass: true
   ReactElement: true
+
+rules:
+  # allow use constructors for side-effects
+  no-new: 0
```
zpao: Disabling like this won't work unless we do it internally at FB as well (since changes are still made internally first and synced out). And I'm not sure we want to disable it - it has value in real code (and it unfortunate we can't disable selectively in tests here). We could do this one on the per file basis perhaps…

src/traversal/printRelayOSSQuery.js
enaqx: Ok. Got it. But it's kinda of exception from proposed [airbnb codestyle](https://github.com/airbnb/javascript#5.1).

zpao: We aren't airbnb 😉

enaqx:
@zpao Thanks for detailed instructions

> move eslint rules from dependency to relay own .eslintrc
>> No, the point of the shared config is that we can share that across our projects and enforce the same rules :)

Ok. Then vice versa. graphql-js should depend on general rules rather than implementing its own, but it seems have quite different codestyle. :)

src/traversal/printRelayOSSQuery.js
enaqx: Ok. Thanks for clarification. :star:

src/traversal/diffRelayQuery.js
enaqx: I will wrap return in eslint exception rule.

src/traversal/__tests__/diffRelayQuery_connection-test.js
enaqx: Ok. Got it. I'm worrying about global disabling also. But this rule disabling can't produce real holes for bugs.

.eslintrc
enaqx: Ok. Got it.

enaqx:
Updated. Hope I've understood instructions correctly. :)

src/traversal/diffRelayQuery.js
```diff
@@ -619,10 +619,12 @@ class RelayDiffQueryBuilder {
     // The returned `trackedNode` is never tracked directly: instead it serves
     // as an indicator to `diffConnection` that the entire connection field must
     // be tracked.
+    /* eslint-disable consistent-return */
```
zpao: Let's just `return undefined` to make it explicit at the other site instead of disabling the rule

yungsters:
@facebook-github-bot import

yungsters:
@enaqx I'm going to make the following changes as I merge this upstream:

1. Change `diffRelayQuery.js` to return `null` to have a consistent return.
2. Remove disabling `no-unused-vars` from `RelayQueryTransform`. (This warning is a bug with ESLint and Flow, and not with the code itself.)

Let me know if either of these will be a problem!

enaqx:
@yungsters No problem, thanks!

Although if you remove `no-unused-vars` from `RelayQueryTransform` it will impact tests with warning if lint step will be included.

PR #413, opened 2015-10-02
Fix JSConf link

enaqx:


wincent:
Thanks, @enaqx!

@facebook-github-bot shipit

Issue #233, opened 2015-09-03
Implement RelayStore.reset()


wincent:
The motivating use case here is clearing the cache on logout. We already have an internal task for this (t6995165) but I'm creating this issue for external visibility. 

For now, the workaround is to do a full page refresh on logout. The long-term plan here is #559 - instead of resetting the store, applications will be create new instances of `RelayEnvironment` when necessary.

steveluscher:
This will be useful for the prototyping tools (#240). Right now, it's possible to paint yourself into a corner where you have cached results for a given field/call combination, but you've changed the underlying resolve method in the schema tab.

quazzie:
Great, need this :)

clentfort:
Things that needed to be touched just from a quick peek at the code: In `RelayStoreData`: 

- In `RelayStoreData.js`: `_records`, `_queuedRecords` must be reset to `{}`, `_recordsStore` and `_queuedStore` must be reinitialized with the empty record-sets. `_cachedRecords` should stay as is, I assume. `_queryTracker` must be reinitialized. `RelayStoreGarbageCollector` should also be reinitialized if needed.
- In `RelayStore.js`: `queuedStore` can no longer be cached. 

The more interesting part: How should `RelayStoreData` announce this change to the rest of the system? Should Relay go back into a `fetching`-state?

josephsavona:
@clentfort Yeah, the tricky part isn't so much the resetting as it is deciding what to do if components are still mounted.

clentfort:
@josephsavona: If we change some stuff we could use the infrastructure that is available for the garbage collector and simply error-out if any data is still subscribed too. We could change the `RelayStoreQueryResolver` to emit events when new subscriptions are created/disposed, this would enable us to count the number of active subscriptions and not allow `reset` if there are any.
This would even allow us to decouple some things (i.e. we could make `RelayStoreQueryResolver` unaware of the garbage collector)!

skevy:
Is there any work around for this that could be used on react-native? A page refresh isn't really an option there. :)

josephsavona:
@skevy The workaround for now in React Native is the brute-force equivalent of a page refresh - tear down the current JS context and start a new one.

josephsavona:
Here's what needs to change to make this happen:
- Implement an `invariant` check that there are no rendered Relay containers, no pending queries, and no mutations - only allow reset if this check passes.
[...]

PR #422, opened 2015-10-04
Move subscription count handling to `RelayStoreData`.

clentfort:
Currently `GraphQLStoreQueryResolver` is aware of the
`RelayStoreGarbageCollection` and updates the subscription-count directly. This
unnecessarily makes the `GraphQLStoreQueryResolver` aware of garbage collection.
Using `RelayStoreData` as a layer between decouples the components. Furthermore
this will help implementing `RelayStoreData#reset` since we can in the future
keep track of active subscriptions in `RelayStoreData` itself.

### DIFF
--- a/src/legacy/store/GraphQLStoreQueryResolver.js
+++ b/src/legacy/store/GraphQLStoreQueryResolver.js
@@ -17,7 +17,6 @@ import type {ChangeSubscription} from 'GraphQLStoreChangeEmitter';
 import type GraphQLFragmentPointer from 'GraphQLFragmentPointer';
 var GraphQLStoreChangeEmitter = require('GraphQLStoreChangeEmitter');
 var GraphQLStoreRangeUtils = require('GraphQLStoreRangeUtils');
-import type RelayStoreGarbageCollector from 'RelayStoreGarbageCollector';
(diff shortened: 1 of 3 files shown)


josephsavona:
Hmm. This doesn't reduce coupling - it only shifts the coupling such that `RelayStoreData` needs to understand more of how GC works. Can you describe more of how this helps implement `reset()`?

clentfort:
Well `RelayStoreData` is coupled with the `RelayStoreGarbageCollector` anyway, so I don't see any harm done. 
If I understood the discussion in #233 we need to keep track of the number of active subscriptions, so `{de,in}creaseSubscriptionFor` can be used to do the bookkeeping.

josephsavona:
Although `RelayStoreData` holds a reference to a garbage collector instance, it isn't currently coupled to its implementation. Adding increment/decrement methods actually increases coupling & complexity. 

An alternate approach (which I've started in #370) is to pass an instance of `RelayStoreData` to the query resolver, from which it can access the GC instance.

josephsavona:
Closing in favor of #370 - thanks @clentfort for exploring this idea!

PR #424, opened 2015-10-04
Improve Babel transform script

yungsters:
Makes a few improvements to the transform script in `babel-relay-plugin`:

 - Trims off `.js` extension and makes it executable.
 - More precise error messages for invalid input paths.
 - Accept source to transform via stdin.
 - Fix bug with path to dependent module.


josephsavona:
:+1:

PR #429, opened 2015-10-06
Minor doc nits

yungsters:
Fixed a newline issue that caused a list to not render properly. Also, got rid of an unnecessary newline and replaced `-` with `—`.


yungsters:
@facebook-github-bot shipit

PR #430, opened 2015-10-06
Fixed licence

akashnimare:
Fixed licence issue.


josephsavona:
@akashnimare The headers need to be changed for *all* the .js files in `examples/`. Can you confirm that lint passes after the change?

akashnimare:
No I did not check the lint test. I will. I think it's better to make a new branch for this. I will send a new PR tomorrow after checking the lint test.

PR #432, opened 2015-10-06
Fix build-lib script for BSD sed

kassens:
GNU and BSD `sed` have some differences. Make the script compatible with both by
using neither of them.


kassens:
@facebook-github-bot import

zpao:
Next up: write it in JS :wink:

PR #436, opened 2015-10-07
Update the TodoMVC example

taion:
Since @josephsavona improved the API for react-router-relay, we can now make the index route be a lot less hacky/weird.


josephsavona:
@facebook-github-bot import

PR #442, opened 2015-10-08
Relay: Simplify RQLTransformer Construction

yungsters:
Summary:
Instead of mutating `state.opts.extra`, this constructs a single `RQLTransformer` per plugin. (Yay, less global mutable state.)

Test Plan:
```
npm test
```
Also, verify that Relay Playground still works.


yungsters:
@facebook-github-bot import

yungsters:
@steveluscher FYI, this is why I was asking about the plugin w.r.t. Relay Playground.

Issue #445, opened 2015-10-09
Option to specify custom mutation query name


chandu0101:
Hi,
      Scala.js-relay mutation queries not working because of this reason https://github.com/facebook/graphql/issues/102

i don't see any activity from graphql side may be it requires more work!, In my understanding graphql should allow javascript valid identifiers as its not the case now and scala.js can't do anything either so is it possible to allow option( static queryName) to define mutation query names ? or can we modify query name to match graphql ASCII set before sending  ? 

please let me know if either of cases can possible ?  , i am happy to send a PR!

josephsavona:
Mutations can specify a custom name with ``Relay.QL`mutation YourNameHere { ... }` `` - does this help?

xsistens:
@josephsavona is the code you referring to the one from the ```getMutation``` method like in this example? 
```
class LikeStoryMutation extends Relay.Mutation {
  getMutation() {
    return this.props.story.viewerDoesLike
      return Relay.QL`mutation MutationName {likeStory}`;
  }
}
```
if yes, then this does not solve the problem. 
@chandu0101 is talking about the mutation name that is actually used in the query that is send to the graphql server. 

As an example from scala.js relay version we developed, the query looks like this, even after applying the change you suggested:
```
mutation $c_Lcom_scalajs_relay_examples_treasure_mutations_CheckHidingSpotForTreasureMutation($input_0:CheckHidingSpotForTreasureInput!){
  checkHidingSpotForTreasure(input:$input_0){
    clientMutationId,
    ...
  }
}
```

as you can see in scala.js it still has a weird name, that indicates me that the name that is used in the query is generated of the name from the actual mutation instance in the global namespace (just a guess). In scala.js this must be unique over the whole javascript libraries that are used, thats why the package name is used. If your suggestion would affect the name used in the query i guess it should be reflected somewhere in this path even in scala.js. but i'm not 100% certain.

josephsavona:
@xsistens Thanks for providing more detail, I see what's happening.

[...]

PR #448, opened 2015-10-09
use mutationNode name instead of constructor name for mutation queries

chandu0101:
fixes https://github.com/facebook/relay/issues/445


josephsavona:
@facebook-github-bot import

PR #453, opened 2015-10-12
Fix missing Relay import in examples

kassens:
Test Plan:
For every example `npm install` and `npm start` then open it in a browser and
see it render.

### DIFF
--- a/examples/relay-treasurehunt/js/app.js
+++ b/examples/relay-treasurehunt/js/app.js
@@ -13,6 +13,7 @@
 import 'babel/polyfill';
 import App from './components/App';
 import AppHomeRoute from './routes/AppHomeRoute';
+import React from 'react';
 import ReactDOM from 'react-dom';
 import Relay from 'react-relay';
 
--- a/examples/star-wars/js/app.js
+++ b/examples/star-wars/js/app.js
@@ -11,6 +11,7 @@
  */
 
 import 'babel/polyfill';
+import React from 'react';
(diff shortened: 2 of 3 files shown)


kassens:
@facebook-github-bot import

akashnimare:
Dude I removed those relay import line because it is unused var. Lint was showing that var relay is not used so I removed it.

yungsters:
cc @zpao @spicyj 

I assume this is an implicit dependency due to JSX? Is this something we could fix in `eslint-plugin-react`?

spicyj:
Yeah, you're supposed to import React. I guess that could be a nice lint rule, though in some cases React will be a global so you wouldn't want to warn.

yungsters:
> though in some cases React will be a global so you wouldn't want to warn.

But in that case, we should use the existing ESLint mechanisms for declaring `React` as a global, right?

----

Regardless, @kassens :+1:

zpao:
Yep. You'll want to enable https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-uses-react.md

spicyj:
https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/react-in-jsx-scope.md too.

yungsters:
Ahh, nice.

@kassens Can you incorporate @zpao's suggestion so that this pull request does not re-introduce the lint errors?

PR #460, opened 2015-10-13
Fix lint issues and use single .eslintrc

kassens:
This fixes a few lint issues and also uses a single `.eslintrc` to prevent duplication and divergence of lint rules.

Follow up for concerns raised in #453.


zpao:
👍

kassens:
@facebook-github-bot import

PR #467, opened 2015-10-14
Update TodoMVC example

taion:
This does 2 real things:
- Update `react-router-relay` per https://github.com/relay-tools/react-router-relay/pull/63
- Use the top-level exported `Link` components that are now available from React Router


kassens:
Great, I like that we no longer need to load a "private" module.

@facebook-github-bot import

Issue #469, opened 2015-10-15
If possible keep the examples up to date


xsistens:
Hey guys. It would be amazing if you could keep the examples up to date, when you release a new version. After updating the examples to the latest version i get a lot of warnings and i don't know if its because of a mistake by me or the example being outdated. 
For example they are still using old react-rc1 versions

Thanks a lot in advance!

wincent:
Thanks for the reminder!

I'm going to update our internal release checklist so that we don't forget this in future releases, and push an update presently to update the examples.

PR #472, opened 2015-10-15
Update example app deps

wincent:
Should eliminate warnings seen with the example apps.

Tested with `npm install && npm run start` in each app.

Fixes: https://github.com/facebook/relay/issues/469


wincent:
@facebook-github-bot import

Issue #478, opened 2015-10-16
trackedNodes for split edge queries returned in random order causing issues re-building edge fields


HurricaneJames:
Summary: Relay does not merge query fragments when calling setVariables. Instead it uses a unique id for each query. This works fine, until a mutation means everything must be refetched. Then, it tries to rebuild edges from those previous queries, but the edges are sorted in alpha order on the unique ids. `putRangeEdges` tosses out a warning (cannot find segment) when it reaches a segment that has a cursor that has not been processed yet, then quietly abandons the data and keeps going.

Full Description:
Lets say we have a simple list component.

````
class ASimpleList extends React.Component {
  onClick = (e) => {
    e.preventDefault();
    this.props.relay.setVariables({
      count: this.props.relay.variables.count + 4
    });
  }
  render() {
    return (
      <div>
        <ul>
        {
          this.props.viewer.listItems.edges.map((itemEdge) =>
            <li key={itemEdge.node.id}>
              <ItemView item={itemEdge.node} />
            </li>
          )
        }
        </ul>
        {
          this.props.viewer.listItems.pageInfo.hasNextPage &&
          <button onClick={this.onClick}>Load More</button>
        }
      </div>
    );
  }
}
````

This Simple component is wrapped in a Relay container.

````
export default Relay.createContainer(ASimpleList, {
  initialVariables: { count: 2 },
  fragments: {
    viewer: () => Relay.QL`
      fragment on RelayViewer {
        id
        listItems(first: $count) {
          edges {
            node {
              id
              ${ItemView.getFragment('item')},
            }
          }
          pageInfo { hasNextPage }
        }
      }
    `,
  }
})
````

Clicking on the "Load More" button works as expected and loads more items. Each time it generates an optimized query fragment to only load the extra. The first fetch, `items0000`, grabs 2 items. The second fetch, `items0002`, only grabs the next 4. The third fetch, `items0001`, requests the next 4 and returns 2 from my sample data.

The fragment names are not the actual names used. Relay uses random keys, `itemsemfrel`, `itemsywxi0w`, and `itemsx3ao8b`. These keys, in an alpha sort may as well be items0000, items0002, and items0001.

Now, lets assume there is another component on the page somewhere else that calls a mutation.

````
import Relay from 'react-relay';

export default class MoveItemMutation extends Relay.Mutation {
  getMutation()     { return Relay.QL`mutation { moveItem }`; }
  getVariables()    {
    return {
      itemId:  this.props.itemId,
[...]

Issue #479, opened 2015-10-16
Purging unneeded data


HurricaneJames:
Sorry if this is has been discussed somewhere, but I cannot find it. Is there a way to make Relay purge old queries. For example, I want to infinite scroll through a list of items. At some point while scrolling down the list I want to purge the old queries and just re-fetch them when scrolling back up.

wincent:
At the moment, the only way to force a refetch is using a mechanism like [`forceFetch`](http://facebook.github.io/relay/docs/api-reference-relay-container.html#forcefetch), but in the example that you describe that would refetch everything in the connection, which may be an intractably large amount of data.

Note that we already have some code in place ([RelayStoreGarbageCollector](https://github.com/facebook/relay/blob/master/src/store/RelayStoreGarbageCollector.js) and related classes) to do a simple form of garbage collection, but it is based on the notion of subscribed components (ie. which components are mounted; it has no notion of what is visible within a viewport nor how long since a given piece of data was accessed) and it is not active by default.

The use case you describe is one that we've talked about internally, although I don't know if we have a GitHub issue for it yet. If not, this one can be it!

wincent:
Added the "discussion" label as there are a number of ways this problem could be approached.

HurricaneJames:
Even with forceFetch, it does not appear to actually throw out everything and fetch all the data again. I'm working on issue #478 where tracking sub-query fragments (`trackedNodes`) returns those fragments with random ordering causing issues rebuilding connection edges. I will admit that I have not gone too deep into forceFetch, but it appears to still run the `diffRelayQuery` which builds up all those sub-fragments instead of tossing out all trackedNodes and refetching everything. 

As for my example, when I said refetch them, I didn't mean all at once. If I extend my infinite scroll by 20 items when I'm scrolling down, then I would probably do the same when scrolling back up. Though in practice I would try to avoid this type of infinite scroll. I implemented it once and the view position issues were a nightmare.

saikat:
Does `untrackNodesForID` in RelayQueryTracker.js help you?  It's not a documented method and I have not used it.

josephsavona:
@HurricaneJames Currently there is no way to avoid the buildup of multiple distinct tracked queries for e.g. ranges. We typically recommend specifying the result of connection changes via configs such as `RANGE_ADD`, which tell Relay how a connection will change as the result of a modification. This can help to avoid refetching the entire range - have you tried that? 

Note that we're actively working on changing tracked queries in #485.

[...]

PR #490, opened 2015-10-19
Fix issue where queries with multiple fields on the same Connection get sorted out of order.

HurricaneJames:
This is related to #478, where having multiple fields that query for different ranges on a Connection come back in random order causing the RelayQueryWriter to get confused and fail to build all the edges.


wincent:
Thanks for the PR, @HurricaneJames.

We originally added the type-first sorting — somewhat reluctantly — to satisfy an internal requirement (integration with another cache that relied on type-first behavior; D2022689 for any FBers that might be reading along). I'm going to check around and see whether this is still a requirement (I expect it is) and if we can fork this particular detail so that the open source world doesn't have to deal with this.

HurricaneJames:
If it is a requirement, then it might work to modify the sort algorithm to only sort type first and not alphanumeric.

I switched to a different task for the next couple hours, but something like

```
function sortTypeFirst(a: string, b: string): number {
  if (a === b) { return 0; }
  if (a === TYPE) { return -1; }
  if (b === TYPE) { return 1;  }
  return 0;
}
```

should still sort type first, but leave the sequencing alone otherwise.

HurricaneJames:
Alright, so I was too curious to try and see if it would work. It does seem to fix the problem I found, but it also causes tests to fail. Fortunately, it looks like just tests for `sortTypeFirst` which expected the previous behavior of sorting for type first and name second. I do not see anywhere else that uses the sortTypeFirst algorithm in the code, so perhaps this change is acceptable.

HurricaneJames:
@winent did we ever find out if this could be merged or if we needed to use the alternate sort algorithm?

wincent:
I don't think we want to drop the sorting entirely, as we want printed queries to preserve the ordering of the original underlying query (other than the noted constraint that types have to come first) for consistency, determinism and debuggability.

It seems that once #479 is resolved, your issue hopefully will be too. @josephsavona is actively working on addressing that, although it is a multi-step process with no quick fix.

HurricaneJames:
Git sometimes sucks. I tried to change the PR to the new sort method and merge that change, but it got all messed up. It would take longer to figure out than just to refork. So I'm just doing that.

New PR: #523

Issue #396, opened 2015-09-27
Flow problems and jest fails when using npm3


clentfort:
By default npm3 flattens the `node_modules` directory. This results in the nested path defined in [`.flowconfig`](https://github.com/facebook/relay/blob/master/src/.flowconfig#L9) not being valid. Furthermore there are problems running `jest`. All tests fail with a runtime error (see attached output of `npm t`). I assume this has something to do how jest is configured to transform the source before running the tests.

```
npm t

> react-relay@0.3.2 test /home/christian/js/relay
> f() { EXIT=0; npm run typecheck || EXIT=$?; NODE_ENV=test jest $@ || EXIT=$?; exit $EXIT; }; f


> react-relay@0.3.2 typecheck /home/christian/js/relay
> flow check src/

find: "/home/christian/js/relay/node_modules/fbjs/node_modules/promise": No such file or directory
find: "/home/christian/js/relay/node_modules/fbjs/node_modules/promise": No such file or directory

Found 0 errors
Using Jest CLI v0.5.8
 FAIL  src/legacy/store/__tests__/GraphQLRange-test.js 
● Runtime Error
TypeError: /home/christian/js/relay/node_modules/fbjs-scripts/jest/environment.js: /home/christian/js/relay/node_modules/core-js/es6/index.js: /home/christian/js/relay/node_modules/core-js/modules/es6.symbol.js: Cannot read property 'type' of null
  at makeComponent (/home/christian/js/relay/node_modules/jest-cli/src/lib/moduleMocker.js:83:19)
  at generateMock (/home/christian/js/relay/node_modules/jest-cli/src/lib/moduleMocker.js:232:16)
  at Object.generateFromMetadata (/home/christian/js/relay/node_modules/jest-cli/src/lib/moduleMocker.js:265:14)
  at Loader._generateMock (/home/christian/js/relay/node_modules/jest-cli/src/HasteModuleLoader/HasteModuleLoader.js:278:23)
  at Loader.requireMock (/home/christian/js/relay/node_modules/jest-cli/src/HasteModuleLoader/HasteModuleLoader.js:795:43)
  at Loader.requireModuleOrMock (/home/christian/js/relay/node_modules/jest-cli/src/HasteModuleLoader/HasteModuleLoader.js:918:17)
  at Object.<anonymous> (/home/christian/js/relay/node_modules/core-js/modules/es6.symbol.js:4:22)
[...]

PR #495, opened 2015-10-20
Prefer node 4.x, npm 2.x in development

wincent:
At the moment, our test suite won't run properly under npm 3.x, so let's
specify our requirement in the package.json.

Once npm 3.x goes "final" we'll make sure that we support both. In the
meantime, this will prevent developers running the tests from seeing
spurious failures like the ones reported here when using npm 3:

https://github.com/facebook/relay/issues/396

Tested by installing npm 3, then running `npm install`, and seeing the
expected error from the `check-dev-engines.js` script:

    AssertionError: Current npm version is not supported for
    development, expected "3.0.0" to satisfy "2.x".

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
   "repository": "facebook/relay",
   "scripts": {
     "build": "gulp",
+    "postinstall": "node node_modules/fbjs-scripts/node/check-dev-engines.js package.json",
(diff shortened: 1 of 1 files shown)


zpao:
👍

wincent:
@facebook-github-bot import

Issue #484, opened 2015-10-17
New for Relay (npm install problem)?


thearabbit:
I am a new for `Relay`, from `Meteor JS`.
I have problem with install
```js
Theara@THEARA-PC D:\relay\examples\todo                                 
$ npm install                                                           
                                                                        
> babel-relay-plugin@0.3.0 prepublish D:\relay\scripts\babel-relay-plugi
n                                                                       
> npm run build                                                         
                                                                        
                                                                        
> babel-relay-plugin@0.3.0 build D:\relay\scripts\babel-relay-plugin    
> scripts/build-lib                                                     
                                                                        
'scripts' is not recognized as an internal or external command,         
operable program or batch file.                                         
                                                                        
npm ERR! Windows_NT 6.1.7600                                            
npm ERR! argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Users\\Theara\\
AppData\\Roaming\\npm\\node_modules\\npm\\bin\\npm-cli.js" "run" "build"
                                                                        
npm ERR! node v4.1.2                                                    
npm ERR! npm  v3.3.5                                                    
npm ERR! code ELIFECYCLE                                                
npm ERR! babel-relay-plugin@0.3.0 build: `scripts/build-lib`            
npm ERR! Exit status 1                                                  
npm ERR!                                                                
npm ERR! Failed at the babel-relay-plugin@0.3.0 build script 'scripts/bu
ild-lib'.                                                               
npm ERR! This is most likely a problem with the babel-relay-plugin packa
ge,                                                                     
npm ERR! not with npm itself.                                           
npm ERR! Tell the author that this fails on your system:                
npm ERR!     scripts/build-lib                                          
npm ERR! You can get their info via:                                    
npm ERR!     npm owner ls babel-relay-plugin                            
npm ERR! There is likely additional logging output above.               
                                                                        
npm ERR! Please include the following file with any support request:    
npm ERR!     D:\relay\scripts\babel-relay-plugin\npm-debug.log          
npm ERR! addLocal Could not install D:\relay\scripts\babel-relay-plugin 
                                                                        
> react-relay@0.4.0 prepublish D:\relay                                 
> npm run build                                                         
                                                                        
                                                                        
> react-relay@0.4.0 build D:\relay                                      
> gulp                                                                  
```
I base on `Win 7`.

PharaohMaster:
Have a same problem when trying to do `npm install` inside treasure hunt sample.

wincent:
[...]

PR #499, opened 2015-10-22
Replace babel-relay-plugin build script with node script

wincent:
We've [had reports](https://github.com/facebook/relay/issues/484) of `npm install` not working on Windows because our `build-lib` script makes platform-specific assumptions,

This commit rewrites it in JS, which in theory should run everywhere `node` runs. (Note, I haven't actually tested this on Windows yet, although I have confirmed that it continues to do the right thing on OS X.)

Thing is, JS isn't actually that awesome for "shell" scripting, so I am relying on a couple of third-party modules -- rimraf for an equivalent to `rm -rf` and node-find-files to stand in for `find` -- which I chose based on the fact that we've previously used them at FB (at least, copies exist under `scripts/third_party/node_modules`).

### DIFF
--- a/scripts/babel-relay-plugin/package.json
+++ b/scripts/babel-relay-plugin/package.json
@@ -24,7 +24,9 @@
     "babel-jest": "^5.3.0",
     "flow-bin": "0.17.0",
     "jest-cli": "^0.5.8",
-    "minimist": "^1.1.3"
+    "minimist": "^1.1.3",
+    "node-find-files": "0.0.4",
+    "rimraf": "2.4.3"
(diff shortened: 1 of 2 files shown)


wincent:
@facebook-github-bot import

zpao:
Thoughts on simplifying this to avoid spawning and the whole read-write-read-write (making it just read-write). I can press the PR button on https://github.com/zpao/relay/commit/0efb66d4220774ac773991aea6f794dcd037b588 if you want.

wincent:
Go for it!

PR #506, opened 2015-10-23
Simplify plugin build script

zpao:
This simplifies the build process to avoid excessive i/o and process spawning, instead doing the work in a single pass.

cc @wincent


wincent:
lg2m

zpao:
@facebook-github-bot import

PR #508, opened 2015-10-24
Moved postinstall => prepublish in relay/package.json

### LINKED ISSUE #505: npm install fails in examples
`npm install` fails in `examples/star-wars`, `examples/todo` and `examples/relay-treasurehunt` with missing module error.
screenshot 2015-10-23 00 40 39 I'm on the latest `master` branch with `node 4.2.1` and` npm 2.14.7` on OS X Yosemite. 

`npm install` in `relay/` works fine.

michaelchum:
Fixes #505 Moved postinstall => prepublish to fix broken dependency when using relay pointing to a local path. This fixes `npm install` in all three examples.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,8 +22,7 @@
   "scripts": {
     "build": "gulp",
     "lint": "eslint .",
-    "postinstall": "node node_modules/fbjs-scripts/node/check-dev-engines.js package.json",
-    "prepublish": "npm run build",
+    "prepublish": "node node_modules/fbjs-scripts/node/check-dev-engines.js package.json && npm run build",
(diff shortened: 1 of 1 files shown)


michaelchum:
Tests are failing on the latest `master` branch :(

package.json
```diff
@@ -22,8 +22,7 @@
   "scripts": {
     "build": "gulp",
     "lint": "eslint .",
-    "postinstall": "node node_modules/fbjs-scripts/node/check-dev-engines.js package.json",
-    "prepublish": "npm run build",
+    "prepublish": "npm run build && node node_modules/fbjs-scripts/node/check-dev-engines.js package.json",
```
wincent: What do you think about doing the `check-dev-engines` check _before_ the `build` (to fail fast)?

michaelchum: Definitely! It makes more sense.

wincent:
Thanks!

@facebook-github-bot import

PR #509, opened 2015-10-24
Fix tests on master

michaelchum:
Tests were broken due to a minor bug in [this recent commit](https://github.com/facebook/relay/commit/cc7c0e5b16999e045937a5f75a4ef0fe05b4695e) where the `setImmediate()` module is imported. However, there is no need to import `setImmediate()`, it seems to be a native function in Node.

### DIFF
--- a/src/store/__tests__/readRelayDiskCache-test.js
+++ b/src/store/__tests__/readRelayDiskCache-test.js
@@ -20,7 +20,6 @@ var Relay = require('Relay');
 var RelayRecordStore = require('RelayRecordStore');
 
 var readRelayDiskCache = require('readRelayDiskCache');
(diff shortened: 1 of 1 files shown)


wincent:
Ah, thanks for pointing this out. We use `require('setImmediate')` because internally at Facebook there is a polyfill that we use for this.

I think that we might actually have to add a forked version of the module for OSS, that just calls the native implementation.

cc @yuzhi

wincent:
Internal fix for this is D2579879 — we're using `resolveImmediate` (from fbjs) elsewhere instead of `setImmediate`, so we should use that here too; I'll get that synced out soon, so I'll close this one out. In the meantime though, thanks for noticing the breakage and submitting the PR!

PR #515, opened 2015-10-26
Fix Relay website build

michaelchum:
Bumped relay-local-schema to the latest version.
screenshot 2015-10-25 15 45 22

Also added local `npm install` of `babel-relay plugin` to the website's installation command because otherwise `babel-relay-plugin` would not build due to some dependencies of its build script only available as `devDependencies`.
screenshot 2015-10-25 17 15 23


steveluscher:
@facebook-github-bot import

PR #517, opened 2015-10-27
Simplify import createHashHistory

dnshi:


wincent:
Thanks!

@facebook-github-bot import

PR #518, opened 2015-10-27
Update APIReference-Store.md

raineroviir:
update is a method on `Store`


wincent:
Thanks!

@facebook-github-bot import

PR #523, opened 2015-10-28
Fix issue where queries with multiple fields on the same Connection get sorted out of order

HurricaneJames:
This is related to #478, where having multiple fields that query for different ranges on a Connection come back in random order causing the RelayQueryWriter to get confused and fail to build all the edges.

It is also related to #490. The difference between the PR and #490 is that this request uses a type first sorting that does not otherwise change the order of the elements in the array.

### DIFF
--- a/src/traversal/flattenRelayQuery.js
+++ b/src/traversal/flattenRelayQuery.js
@@ -19,37 +19,10 @@ var RelayQueryVisitor = require('RelayQueryVisitor');
 
 var sortTypeFirst = require('sortTypeFirst');
 
-/* @flow */
-class FlattenedMap {
-  hash:  { [key: string]: Object };
-  array: [ string ];
-
-  constructor() {
-    this.hash = {};
-    this.array = [];
-  }
-
-  get(key): FlattenedQuery {
-    var hashValue = this.hash[key];
-    return hashValue && hashValue.value || undefined;
-  }
-
-  set(key, value) {
-    var old = this.hash[key];
-    if (old) { this.array.splice(old.index, 1); }
-    this.hash[key] = { value: value, index: this.array.length };
-    this.array.push(key);
-  }
-
-  keys(): [ string ] {
-    return this.array;
-  }
-}
-
 type FlattenedQuery = {
   node: RelayQuery.Node;
   // flattenedFieldMap: {[key: string]: FlattenedQuery};
-  flattenedFieldMap: FlattenedMap;
+  flattenedFieldMap: Map;
 };
 
 /**
@@ -63,16 +36,16 @@ type FlattenedQuery = {
  */
 function flattenRelayQuery<Tn: RelayQuery.Node>(node: Tn): ?Tn {
   var flattener = new RelayQueryFlattener();
-  var flattenedFieldMap = new FlattenedMap();
+  var flattenedFieldMap = new Map();
   flattener.traverse(node, {node, flattenedFieldMap});
   return toQuery(node, flattenedFieldMap);
 }
 
 function toQuery<Tn: RelayQuery.Node>(
   node: Tn,
-  flattenedFieldMap: FlattenedMap
+  flattenedFieldMap: Map
 ): ?Tn {
-  var keys = flattenedFieldMap.keys().sort(sortTypeFirst);
+  var keys = Array.from(flattenedFieldMap.keys()).sort(sortTypeFirst);
   return node.clone(
     keys.map(alias => {
       var field = flattenedFieldMap.get(alias);
@@ -93,7 +66,7 @@ class RelayQueryFlattener extends RelayQueryVisitor<FlattenedQuery> {
     if (!flattenedField) {
       flattenedField = {
         node,
-        flattenedFieldMap: new FlattenedMap()
+        flattenedFieldMap: new Map()
       };
       state.flattenedFieldMap.set(serializationKey, flattenedField);
     }
--- a/src/traversal/flattenRelayQuery.js
+++ b/src/traversal/flattenRelayQuery.js
@@ -19,9 +19,37 @@ var RelayQueryVisitor = require('RelayQueryVisitor');
 
 var sortTypeFirst = require('sortTypeFirst');
 
+/* @flow */
+class FlattenedMap {
+  hash:  { [key: string]: Object };
+  array: [ string ];
+
+  constructor() {
+    this.hash = {};
+    this.array = [];
+  }
+
+  get(key): FlattenedQuery {
+    var hashValue = this.hash[key];
+    return hashValue && hashValue.value || undefined;
+  }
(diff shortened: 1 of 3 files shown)


wincent:
This is equivalent to dropping the underlying sort, with the exception of keeping types at the top, and [I don't think we want to drop the sort](https://github.com/facebook/relay/pull/490#issuecomment-151617620) (because then we rely on implementation-specific semantics of key enumeration order, which are not guaranteed to be stable/consistent).

HurricaneJames:
The sort is reordering an array that is intentionally sequenced in a specific order (the order of the original queries). This is not a case where it is sorting the keys from some object. If it was I would absolutely agree with you. However, the keys for an array are supposed to have an enumeration order.

The only other way to avoid the bug is to rebuild that ordered array when the payload returns and use it to choose which fragment to use. However, due to the recursive nature of the query writer, that proved an exceptionally difficult problem in the general case.

#479 may fix this issue, but only if it is able to toss out split queries for subsequent calls. My guess is it cannot do that in the general case because there are good reasons to have split queries.

HurricaneJames:
Forgot to include the updated test in the last push. Pushed it up. Travis CI should pass now.

josephsavona:
> The sort is reordering an array that is intentionally sequenced in a specific order 

The sort is applied to the list of field names which is the result of calling `Object.keys`. The order is implementation-specific and there is no guarantee that field names will be returned in the same order they were created in the original query. 

Ultimately, the sort needs to be aware of the values of calls such that range segments are ordered, e.g. `first: 10` before `first: 10, after: 10th` before `first: 10, after: 20th`.

HurricaneJames:
See, this is why I should never respond to a comment after midnight, especially when I'm relying on week old knowledge. I would have sworn it was based on an array, but I must have been thinking about another rabbit hole I went down while working on this problem.

I'm not sure that making sort aware of the values of the calls would help, even if you limit the calls to just first/after. Maybe if you made after/before use a format like `10th` instead of a cursor, but that opens a whole other can of worms.

You are right, the order is implementation specific, though currently all implementations maintain the order for everything but numbers. The best solution is probably to use some kind of OrderedMap that allows lookups based on a key and iterates based on the order the items were inserted (the way the current implementation works). Would you be open to such a solution?

HurricaneJames:
I took some time during lunch to prototype up an OrderedMap style data structure for the flattenedFieldMap. That way the algorithm no longer relies on an implementation detail. Sort does not change order other than keeping types at top to meet the internal requirement.

steveluscher:
Can I ask you to take a look at this solution, @josephsavona?

josephsavona:
Thanks for working on this. Introducing a custom map implementation could affect performance and before we go down this path we'll need to understand by how much. I'll import this internally and try benchmarking.

josephsavona:
@facebook-github-bot import

src/traversal/flattenRelayQuery.js
```diff
@@ -19,9 +19,37 @@ var RelayQueryVisitor = require('RelayQueryVisitor');
...
+    var old = this.hash[key];
+    if (old) { this.array.splice(old.index, 1); }
+    this.hash[key] = { value: value, index: this.array.length };
+    this.array.push(key);
+  }
+
+  keys(): [ string ] {
+    return this.array;
+  }
+}
```
josephsavona: replace with `const Map = require('Map');`

src/traversal/flattenRelayQuery.js
```diff
@@ -19,9 +19,37 @@ var RelayQueryVisitor = require('RelayQueryVisitor');
...
+
 type FlattenedQuery = {
   node: RelayQuery.Node;
-  flattenedFieldMap: {[key: string]: FlattenedQuery};
+  // flattenedFieldMap: {[key: string]: FlattenedQuery};
+  flattenedFieldMap: FlattenedMap;
```
josephsavona: `flattenedFieldMap: Map;`

josephsavona:
@HurricaneJames After double-checking the spec we realized that ES2015 `Map`s preserve insertion order, so it should be sufficient to use those. Can you update to just use `Map`?

HurricaneJames:
Sure, I can update it to use ES2015 `Map`s. However, I suspect that ES2015 `Map` might have worse performance than my implementation for this specific problem.

My implementation is fairly streamlined for this specific case. The slowest method is the `array.splice` when updating a key, and we might be able to remove even that. In reviewing the ES2015 docs, I do not believe it changes the key position when updating an existing key, but I'll verify. If that is the case, then we can drop the `splice` call. Also, I'm not sure that we would ever hit that bit of code because we are just walking a tree, so there should be no overwrites.

ES2015 `Map` does a whole lot more "stuff" (tm). Being much more generic I suspect it will be slower, especially since many environments will need a Babel transpiled version. If we do not need all the power of a full `Map`, it does not make sense to pay the performance penalty.

I will work on an implementation using Map, then try it against my `FlattenedMap` implementation in jsperf so we can make an informed decision.

HurricaneJames:
Well, looks like I was wrong, at least in the case where the browser implements ES2015. ES2015 native Map is 35% faster.

[JSPerf: FlattenedMap vs ES2015 Map](http://jsperf.com/flattenedmap-vs-es2015-map)

josephsavona:
Yup! Let's use `Map`.

HurricaneJames:
I was just looking back over your diff comments and I noticed that you `require('Map')`. Is that needed? I thought Babel runtime handled that for us.

src/traversal/flattenRelayQuery.js
```diff
@@ -21,7 +21,8 @@ var sortTypeFirst = require('sortTypeFirst');
```
josephsavona: var Map = require('Map');

josephsavona:
> I noticed that you require('Map'). Is that needed?

Yes.

HurricaneJames:
You want it, so I will add it. Can you explain why it is required? I know the babel runtime automatically includes it. Browsers that support `Map` have it by default, just like `Array`. I cannot find anywhere that says it needs to be `require`d first. I am always interested in expanding my knowledge, so I am curious what you ran into that taught you to `require('Map')`.

josephsavona:
Using `require` makes the dependency explicit, allowing us to a) use our own version of a polyfill if necessary and b) not even send a polyfill to the user if no code relies upon it.

josephsavona:
@facebook-github-bot import

HurricaneJames:
I know that `require('Map')` in any of my repos blows up. Is the code coming from the `fbjs` package?

josephsavona:
@HurricaneJames yup, it's provided by `fbjs`. I believe the OSS version uses the same polyfill as babel.

josephsavona:
btw, thanks again for this!

PR #525, opened 2015-10-30
Fixes #511 - Unable to use Relay with stateless components

### LINKED ISSUE #511: Unable to use Relay with stateless components
I'm using Relay with [recompose](https://github.com/acdlite/recompose).
Recompose using stateless components inside.
When I wrap any Recompose decorator with Relay Container then I get this error:

```
Warning: Stateless function components cannot be given refs
(See ref "component" in doOnReceiveProps(mapProps(Theme(ProjectsList)))
created by Relay(doOnReceiveProps(mapProps(Theme(ProjectsList))))).
Attempts to access this ref will fail.
```

joshthecoder:
Resolves the React warnings seen when passing a function component to Relay.createContainer().
Adds a check to skip adding a "ref" to the container component if it is a function component.

This does mean setVariables() & forceUpdate() callbacks will not be provided the component
if you passed createContainer() a function component. 

Please let me know if I am on the right track here or what needs more work. :)

### DIFF
--- a/src/container/RelayContainer.js
+++ b/src/container/RelayContainer.js
@@ -52,6 +52,7 @@ var nullthrows = require('nullthrows');
 var prepareRelayContainerProps = require('prepareRelayContainerProps');
 var shallowEqual = require('shallowEqual');
 var warning = require('warning');
+var isReactComponent = require('isReactComponent');
 
 export type RelayContainerSpec = {
   initialVariables?: Variables;
@@ -808,7 +809,7 @@ function createContainerComponent(
           {...this.props}
           {...this.state.queryData}
           {...prepareRelayContainerProps(relayProps, this)}
-          ref="component"
+          ref={isReactComponent(Component) ? 'component' : null}
         />
       );
     }
--- a/src/container/RelayContainer.js
+++ b/src/container/RelayContainer.js
@@ -301,7 +301,7 @@ function createContainerComponent(
               }
             });
             if (callback) {
-              callback.call(this.refs.component, {...readyState, mounted});
+              callback.call(this.refs.component || null, {...readyState, mounted});
             }
           });
         } else {
--- a/src/container/isReactComponent.js
+++ b/src/container/isReactComponent.js
@@ -1,0 +1,30 @@
+/**
+ * Copyright 2013-2015, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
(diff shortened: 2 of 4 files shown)


steveluscher:
Thanks for this @joshthecoder!

Since we can't be guaranteed a ref, we should get rid of this `this` binding too: https://github.com/facebook/relay/blob/master/src/container/RelayContainer.js#L302-L304

Can you send an update?

joshthecoder:
@steveluscher Wouldn't this break the current behavior of onReadyStateChange callback you provide when calling setVariables() or forceUpdate()? I'm not sure what use case there is for having access to your component instance in that callback. Maybe for invoking setState()?

josephsavona:
@joshthecoder We should only pass `null` as the instance *if* its a stateless functional component, otherwise RelayContainer should continue to pass the instance.

joshthecoder:
@josephsavona These changes should maintain the current behavior for non-stateless functional components. `this` should still be the component instance.

Also I'm not sure if there's a better way to detect that besides looking for `isReactComponent` on the prototype. Seen the React code use it in a few spots for similar use cases.

src/container/RelayContainer.js
```diff
@@ -300,7 +301,7 @@ function createContainerComponent(
               }
             });
             if (callback) {
-              callback.call(this.refs.component, {...readyState, mounted});
+              callback.call(this.refs.component || null, {...readyState, mounted});
```
joshthecoder: Pass null to callback if we have no reference to a stateless functional component.

src/container/isReactComponent.js
```diff
@@ -0,0 +1,30 @@
...
+
+'use strict';
+
+/**
+ * @internal
+ *
+ * Helper for checking if this is a React Component
+ * created with React.Component or React.createClass().
+ */
+function isReactComponent(component: any): boolean {
```
josephsavona: @zpao Is this the preferred way to test if something is a React component?

josephsavona: sounds like this is the best way to test for now until React has better support for HOCs setting refs

josephsavona:
:+1:

src/container/isReactComponent.js
```diff
@@ -0,0 +1,30 @@
...
+  return (
```
kassens: Nit: the '!!' should be on this line. Otherwise this function might return something other falsy value. 

I think flow would see this, if you changed the input to 'mixed' instead of 'any'.

joshthecoder:
@kassens Addressed your comment. Good catch!

Rodeoclash:
Question, does this make it now possible to shallow rendering Relay wrapped components? At the moment I'm unable to test the rendered output of the shallow render as the shallow rendering does not support refs.

PR #526, opened 2015-10-30
Updating npm dependency to support newer npm versions in the 3.x range

joshblack:
Currently, when cloning from master and running `npm install` while running any version of `npm@3.x` you'll receive an Assertion Error during the prepublish step saying: `Current npm version is not supported for development, expected "3.3.9" to satisfy "2.x".`

This PR just updates the semver range of the npm devEngine to support `npm@3.x` versions in addition to `npm@2.x` ranges.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
   },
   "devEngines": {
     "node": "4.x",
-    "npm": "2.x"
+    "npm": "2.x || 3.x"
   },
   "jest": {
     "rootDir": "",


zpao:
This is actually intentional, mostly due to the way we have to use Flow and modules. See #495 & #396 for some previous discussions. So in order to have a reliable development environment, we have to restrict to npm@2

josephsavona:
@joshblack Thanks for filing this, but due to the restrictions that @zpao mentioned we can't support npm3 yet. I'll close this for now.

joshblack:
Sounds good!

PR #527, opened 2015-10-30
Removing extra renameTodo import from the Todo example schema file

joshblack:
Just a removal of an extra `renameTodo` in the imports from `./database`.

### DIFF
--- a/examples/todo/data/schema.js
+++ b/examples/todo/data/schema.js
@@ -44,7 +44,6 @@ import {
   getViewer,
   markAllTodos,
   removeCompletedTodos,
-  removeTodo,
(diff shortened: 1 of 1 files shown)


steveluscher:
Thanks for the PR! We still need this though, since `removeTodo` is used in this mutation: https://github.com/joshblack/relay/blob/31cc5204cd5dd2bd955052050e3af39ad49be2db/examples/todo/data/schema.js#L241

joshblack:
Thanks for the response! And you're totally right, I just happened to misread it and think it was a duplicate `import` declaration. Sorry!

PR #531, opened 2015-11-01
Cut lead-ins to non-existent formal definitions.

timoxley:
"Forward pagination arguments" & "Backward pagination arguments" both end with "More formally:", yet nothing follows.

image


kassens:
@facebook-github-bot shipit

PR #543, opened 2015-11-02
Start moving readRelayQueryData tests over to use writePayload

wincent:
Part of the effort to enable us to change the internal storage format without having to manually update hundreds of test cases. More to come after this commit.

### DIFF
--- a/src/store/__tests__/readRelayQueryData-test.js
+++ b/src/store/__tests__/readRelayQueryData-test.js
@@ -27,7 +27,10 @@ var readRelayQueryData = require('readRelayQueryData');
 describe('readRelayQueryData', () => {
   var RelayRecordStore;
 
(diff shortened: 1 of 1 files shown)


steveluscher:
Sorry we left this so long, and now it's a rebase-tastrophe. :(

wincent:
Yeah, I wanted to import it but then parenting. Probably easiest to just reroll it, which I'll do.

/literally sent from phone while laboring over a hot stove

josephsavona:
@wincent we didn't end up changing the store format (yet) - should we just close this and continue with it later if necessary?

wincent:
@josephsavona: @yungsters was asking me about this yesterday. I was going to reroll it, but it didn't seem like the highest impact thing I could to so I just let it sit. Closing is the right call for now!

PR #545, opened 2015-11-03
Exclude nodes that should be skipped because of directives in getChildren

### LINKED ISSUE #536: Full support for @include and @skip directives
Relay currently passes through directives to the server, but does not directly implement support for `@include` or `@skip`. The outward effects of this are that skipped fields (either `include: false` or `skip: true`) are not sent to the server, and that data for skipped fields is not returned to components in `props`.

The simplest implementation of the above is to make `RelayQueryNode.getChildren()` not return child fields that are skipped. 

Contributions welcome, otherwise we'll get to this as time permits.

xuorig:
Fixes #536 

I've been working with graphql-ruby and relay and been wanting to use directives. I found #536 yesterday night and tried to work on a solution today.

Following what @josephsavona said in the issue, I've modified`RelayQueryNode.getChildren` so it only return children that are not skipped (meaning `skip if: true` and `include if: false`). I've implemented the function `RelayQueryNode.shouldBeIncluded` which
  - Checks if a skip directive is present ( It has priority over include)
     - If an if argument is present, return the negated value
  - Checks if an include directive is present
     - if an if argument, return the value
  - return true in all other cases

I've tried to take into consideration that other directives and arguments other than `if:` might be added, which might make the code a little bit more complicated, let me know if I should just assume that the argument is if right now.

Let me know if the solution makes sense. I've added 2 tests, modified a printing test that included a field that should be skipped and also modified the metric test since the getChildren method now calls getDirectives for every field.

Haven't had much chance to look at Relay in detail so let me know how I can improve this!

### DIFF
--- a/src/query/RelayQuery.js
+++ b/src/query/RelayQuery.js
@@ -208,7 +208,7 @@ class RelayQueryNode {
           this.__route__,
           this.__variables__
         );
-        if (node && node.shouldBeIncluded()) {
+        if (node && !node.isSkipped()) {
           nextChildren.push(node);
         }
       });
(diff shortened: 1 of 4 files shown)


src/query/RelayQuery.js
```diff
@@ -214,6 +218,36 @@ class RelayQueryNode {
...
+    if (includeNode) {
+      var includeIf = includeNode.arguments.find(
+        arg => arg.name === IF
+      );
+      if (includeIf) {
+        return Boolean(includeIf.value);
+      }
+    }
+
+    return true
```
josephsavona: Because the schema may change, it's definitely good to scan the arguments for each directive. To make this more performant, how about iterating over the directives only once? Something like

```javascript
let isSkipped = false;
let isIncluded = true;
directives.forEach(directive => {
  if (directive.name === SKIP) {
    isSkipped = ...
  } else if (directive.name === INCLUDE) {
    isIncluded = ...
  }
});
return isSkipped || !isIncluded;
```

src/query/RelayQuery.js
```diff
@@ -214,6 +218,36 @@ class RelayQueryNode {
     return children;
   }
 
+  shouldBeIncluded(): ?boolean {
```
josephsavona: :+1: for a new method.

How about naming it `isSkipped(): boolean` to match the `isScalar()`, `isGenerated` pattern? Also note that the return type can be a non-null `boolean`.

src/query/__tests__/RelayQueryRoot-test.js
```diff
@@ -73,6 +73,38 @@ describe('RelayQueryRoot', () => {
     expect(children[2].isGenerated()).toBe(true);
   });
 
+  it('does not return skipped children', () => {
+    var query = getNode(Relay.QL`
+      query {
+        me {
+          id,
+          firstName @skip(if: $cond),
+          lastName @include(if: $cond2),
```
josephsavona: nit: variable names are arbitrary, so you could name this `@skip(if: $true)` to make it even more clear that the variable's value will be `true`

josephsavona:
@xuorig This is awesome!!! I have some minor feedback but overall looks good.

src/query/RelayQuery.js
xuorig: Good idea for the iteration, I will add that in tonight. I dont think `isSkipped || !isIncluded` will work though. Since `skip` should always have priority over `include`, something like `False || !False` would result in the field being skipped even though skip was `False`!

josephsavona:
@facebook-github-bot import

xuorig:
@josephsavona I've addressed your comments. Let me know what you think about my solution for iterating only once over directives. Not the most pretty but it's hard to make it with that Skip > Include priority.

PR #560, opened 2015-11-07
Make GraphQLQueryRunner contextual

devknoll:
@josephsavona Is this more along the lines of what you're looking for?

Part of #558

Issue #558, opened 2015-11-07
[meta] Make all state contextual


josephsavona:
This is a meta-task to track progress toward making all Relay state contextual, such that multiple instances of Relay can be run in the same JS context. The primary motivation is to support server rendering in open source, where sharing JS contexts is the norm.

Tasks:
- [x] `GraphQLQueryRunner`
- [x] `GraphQLPendingQueryTracker` should write query results into the contextual RelayStoreData instance
- [x] `GraphQLDeferredQueryTracker`: should read query root IDs from the contextual RelayStoreData instance
- [x] `RelayMutationTransaction` should write mutation results into the contextual RelayStoreData instance
- [x] `GraphQLStoreRangeUtils` - this holds a global mapping of dataID -> range information, and will have to be made contextual. The primary caller is `readRelayQueryData`.
- [x] `GraphQLStoreChangeEmitter` notifies components (or other observers) of data when it changes.
- [x] Create RelayEnvironment
- [x] Make `RelayStore` a singleton instance of `RelayEnvironment`
- [x] Change `RelayContainer`/`RelayRenderer` to access data via the React context
- [x] Contextualize `RelayRenderer`
- [x] Make `Relay.injectTaskScheduler` a method on `RelayEnvironment`
- [x] Contextualize the network layer and `Relay.injectNetworkLayer`
- [x] Contextualize Relay mutations (cannot create props until the store to which it will be applied is known) (#699)

Note: the following will be completed as part of #559:
- [ ] ~~Document `RelayEnvironment`, make it accessible via `Relay.Environment` or other, and remove any unsupported APIs (or mark them as such).~~
- [ ] ~~Promote `RelayRenderer` to public API~~

josephsavona:
cc @devknoll

devknoll:
> Optionally, Relay.injectTaskScheduler and Relay.injectNetworkLayer could be moved to the Relay Context class to allow different schedulers/network layers per context.

I think this is actually fairly critical for server side rendering.

devknoll:
@josephsavona Do we want `RelayStore` to own the `RelayStoreData` instances? Seems like yes, but wanted to confirm.

devknoll:
@josephsavona Also `GraphQLStoreRangeUtils` seems a little tough. Not sure who should own the instance in this case. Seems like maybe an individual `RelayRecordStore`.

josephsavona:
[...]

PR #561, opened 2015-11-07
Make GraphQLDeferredQueryTracker contextual

devknoll:
Builds on #560. Part of #558

### DIFF
--- a/src/legacy/store/GraphQLDeferredQueryTracker.js
+++ b/src/legacy/store/GraphQLDeferredQueryTracker.js
@@ -17,15 +17,13 @@ var ErrorUtils = require('ErrorUtils');
 var Map = require('Map');
 import type {DataID} from 'RelayInternalTypes';
 import type RelayQuery from 'RelayQuery';
-var RelayStoreData = require('RelayStoreData');
+import type RelayRecordStore from 'RelayRecordStore';
 
 var forEachObject = require('forEachObject');
 var forEachRootCallArg = require('forEachRootCallArg');
 var invariant = require('invariant');
 var isEmpty = require('isEmpty');
 var resolveImmediate = require('resolveImmediate');
-
-var recordStore = RelayStoreData.getDefaultInstance().getRecordStore();
 
 type DeferredQuerySubscriber = {
   dataID: string;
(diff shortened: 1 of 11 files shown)


src/legacy/store/GraphQLDeferredQueryTracker.js
```diff
@@ -17,7 +17,7 @@ var ErrorUtils = require('ErrorUtils');
 var Map = require('Map');
 import type {DataID} from 'RelayInternalTypes';
 import type RelayQuery from 'RelayQuery';
-var RelayStoreData = require('RelayStoreData');
+var RelayRecordStore = require('RelayRecordStore');
```
josephsavona: `import type RelayRecordStore from 'RelayRecordStore'` - only import the type unless you'll use the value at runtime.

src/legacy/store/GraphQLDeferredQueryTracker.js
```diff
@@ -491,7 +500,8 @@ function getQueryParentID(
 }
 
 function getRootCallToIDMap(
```
josephsavona: how about making this a private instance method instead of passing the `recordStore`? e.g. `_getRootCallToIDMap(query)`

src/legacy/store/GraphQLDeferredQueryTracker.js
devknoll: Good catch. Flow (or eslint) should warn about this! :wink:

src/store/RelayStoreData.js
```diff
@@ -15,6 +15,8 @@
 
 var GraphQLStoreChangeEmitter = require('GraphQLStoreChangeEmitter');
 var GraphQLStoreDataHandler = require('GraphQLStoreDataHandler');
+var GraphQLQueryRunner = require('GraphQLQueryRunner');
+var GraphQLDeferredQueryTracker = require('GraphQLDeferredQueryTracker');
```
josephsavona: nit: alpha ordering

src/store/RelayStoreData.js
```diff
@@ -69,6 +71,8 @@ class RelayStoreData {
   _queuedStore: RelayRecordStore;
   _recordStore: RelayRecordStore;
   _queryTracker: RelayQueryTracker;
+  _queryRunner: GraphQLQueryRunner;
+  _deferredQueryTracker: GraphQLDeferredQueryTracker;
```
josephsavona: nit: same, alphabetic ordering

devknoll: Ooh, wasn't aware of this rule. Thanks :smile:

josephsavona:
@facebook-github-bot import

PR #567, opened 2015-11-09
Make RelayPendingQueryTracker contextual

devknoll:
Builds on #566. Part of #558

PR #565, opened 2015-11-09
Make GraphQLStoreChangeEmitter contextual

devknoll:
Builds on #562. Part of #558

PR #568, opened 2015-11-09
Make GraphQLStoreQueryResolver contextual

devknoll:
Builds on #567. Part of #558

### DIFF
--- a/src/legacy/store/GraphQLStoreQueryResolver.js
+++ b/src/legacy/store/GraphQLStoreQueryResolver.js
@@ -22,7 +22,7 @@ import type {DataID} from 'RelayInternalTypes';
 var RelayProfiler = require('RelayProfiler');
 import type RelayQuery from 'RelayQuery';
 import type RelayRecordStore from 'RelayRecordStore';
-var RelayStoreData = require('RelayStoreData');
+import type RelayStoreData from 'RelayStoreData';
 import type {StoreReaderData} from 'RelayTypes';
 
 var filterExclusiveKeys = require('filterExclusiveKeys');
@@ -46,10 +46,10 @@ class GraphQLStoreQueryResolver {
     GraphQLStorePluralQueryResolver |
     GraphQLStoreSingleQueryResolver
   );
-  _store: RelayRecordStore;
+  _storeData: RelayStoreData;
 
   constructor(
-    store: RelayRecordStore,
+    storeData: RelayStoreData,
     fragmentPointer: GraphQLFragmentPointer,
     callback: Function
   ) {
@@ -57,7 +57,7 @@ class GraphQLStoreQueryResolver {
     this._callback = callback;
     this._fragmentPointer = fragmentPointer;
     this._resolver = null;
-    this._store = store;
+    this._storeData = storeData;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


src/legacy/store/GraphQLStoreQueryResolver.js
```diff
@@ -246,8 +243,8 @@ class GraphQLStoreSingleQueryResolver {
       if (subscribedIDs) {
         // always subscribe to the root ID
         subscribedIDs[nextID] = true;
-        this._subscription = GraphQLStoreChangeEmitter.addListenerForIDs(
-          Object.keys(subscribedIDs),
+        this._subscription = this._storeData.getChangeEmitter().addListenerForIDs(
+          Object.keys(subscribedIDs).map(id => this._getCanonicalID(id)),
```
josephsavona: Why this change?

src/legacy/store/GraphQLStoreChangeEmitter.js
```diff
@@ -22,11 +21,10 @@ export type ChangeSubscription = {
   remove: () => void;
 };
 
-var batchUpdate = callback => callback();
-var subscribers = [];
-
-var executingIDs = {};
-var scheduledIDs = null;
+type Subscriber = {
+  callback: Function,
```
josephsavona: nit: it's always better to be as detailed as possible with types, in this case the callback is `() => void`.

src/store/readRelayQueryData.js
```diff
@@ -34,6 +34,8 @@ var callsToGraphQL = require('callsToGraphQL');
 var invariant = require('invariant');
 var validateRelayReadQuery = require('validateRelayReadQuery');
 
+var rangeUtils = RelayStoreData.getDefaultInstance().getRangeData();
```
josephsavona: nit: consistent naming, how about `rangeData` to match the module name?

src/legacy/store/GraphQLStoreQueryResolver.js
devknoll: Right, so `GraphQLChangeEmitter` used to map range -> canonical ID for in `addListenerForIDs`, so moving the logic here just keeps the old behavior intact.

josephsavona: :+1:

src/legacy/store/GraphQLStoreChangeEmitter.js
devknoll: Good catch :+1: `() => void` could probably become its own `SubscriptionCallback` type at this point now too...

josephsavona: yup

src/legacy/store/GraphQLStoreQueryResolver.js
devknoll: That said, I think it might be a good idea to nix as many of these `this._storeData.get*()` calls as possible, and try to expose just the bare minimum API from `RelayStoreData`.

So this would just become `this._storeData.addListenerForIDs`, and `RelayStoreData` would take care of canonicalization and ushering the request to the change emitter.

Thoughts?

josephsavona: Completely agreed, but let's change one thing at a time. First global -> contextual and make sure that's all good, then we can clean up the API.

josephsavona:
@devknoll I'll be importing these internally and recreating your stack of commits to isolate each change, bear with me.

josephsavona:
@facebook-github-bot import

josephsavona:
merging this with #565

PR #569, opened 2015-11-09
Make readRelayQueryData contextual

devknoll:
Builds on #568. Part of #558

PR #570, opened 2015-11-09
WIP: Make RelayStore contextual

devknoll:
Builds on #569. Part of #558

This is actually just a PoC until `RelayContext` is implemented. With this PR, `RelayStoreData. getDefaultInstance` has been completely removed and `RelayStore` turned into a class.

The last steps here are to pass an instance of `RelayStore` via the context and remove the (temporary) `RelayStore.getDefaultInstance` calls.


devknoll:
Whoops, just realized `RelayRecordStore` needs some love too.

devknoll:
Alright, with the latest passing commit, all singleton state (with the exception of tests) should be pushed out to the React components layer. I think at this point we just need to define `RelayContext` and the upgrade path.

josephsavona:
This is awesome!!! Regarding the upgrade path: existing public APIs must continue to work as-is, which means that `RelayStore` must export an object with the same shape as before (`forceFetch`, `primeCache` etc methods). 

At the same time, we also  need to transition RelayRenderer and RelayContainer over to accessing the store instance via `this.props.relay`. A good approach might be something like:
- Create a `RelayContext` class that has instance methods to  match the current RelayStore static methods
- Define RelayStore as an exported instance of RelayContext
- Make RelayRenderer set the `relay` context value to `RelayStore`
- Make RelayContainer access the store via context.

That means everything will technically be contextual on the inside, while still having the same outward API. Then we can deprecate RelayStore, expose RelayContext, and document all this :-)

devknoll:
@josephsavona What about the `injectBatchingStrategy` in `RelayContainer`? Should this just move to the `RelayContainer` constructor?

josephsavona:
how about moving it to RelayRenderer?

josephsavona:
And then make this a public method on the new RelayContext

devknoll:
Something like that?

Next up is cleaning up the `_getStoreData` calls... but, later. I'll let you catch up :wink:

src/container/RelayContainer.js
```diff
@@ -445,7 +438,7 @@ function createContainerComponent(
         }
         if (!deferredSubscriptions.hasOwnProperty(subscriptionKey)) {
           deferredSubscriptions[subscriptionKey] =
-            GraphQLDeferredQueryTracker.addListenerForFragment(
+            deferredQueryTracker.addListenerForFragment(
```
devknoll: @josephsavona Is there some other way we can handle deferred queries? It seems kinda painful to add this to the public API for a single thing to support it... that isn't even OSS supported yet :smile:

devknoll: To elaborate, this method alone requires `RelayContext.prototype.hasPendingQueries`, `RelayContext.prototype.isQueryPending` and `RelayContext.prototype.addListenerForFragment`.

Maybe `hasFragmentData` and the associated state/callbacks could be wrapped in core class and exposed through the public API with a single call...

devknoll:
See `RelayDeferredFragmentState` for an example of what I meant.

Rather than adding a bunch of public API methods, it exposes the state (`hasFragmentData` and `getFragmentError`) for deferred query operations. Then you call `createDeferredFragmentState` if you want one.

The last non-test use of `_getStoreData()` is `GraphQLFragmentPointer.createForRoot` in `RelayRenderer`...

devknoll:
And then there were none!

josephsavona:
So good! I'll have more comments as I work my way through the stack of PRs

devknoll:
:+1: definitely a WIP so don't be afraid to be brutal :wink:

src/store/__mocks__/RelayStore.js
```diff
@@ -79,11 +83,26 @@ RelayStore.forceFetch.mockImplementation((...args) => {
...
+RelayStore.hasOptimisticUpdate.mockImplementation((dataID) => {
+  return storeData.hasOptimisticUpdate(dataID);
+});
+
+RelayStore.resolve.mockImplementation((...args) => {
+  return new GraphQLStoreQueryResolver(storeData, ...args);
+});
+
+RelayStore.buildDeferredFragmentState.mockImplementation((...args) => {
+  return new RelayDeferredFragmentState(
```
devknoll: I don't like how these are just reimplementations... maybe this could become a `requireActual` that then just hooks on the mocking functionality that some of the tests rely on.

src/store/RelayContext.js
```diff
@@ -0,0 +1,279 @@
...
+ *      }
+ *    }
+ *  }
+ *
+ */
+class RelayContext {
+  _storeData: RelayStoreData;
+
+  constructor(storeData?: RelayStoreData) {
+    this._storeData = storeData || new RelayStoreData();
```
devknoll: I don't really like exposing this in the constructor, but it seemed like the lesser evil for removing the `_getStoreData` calls that the tests required. I think ideally the tests wouldn't require so much mocking, but that would require more refactoring...

src/store/RelayDeferredFragmentState.js
```diff
@@ -0,0 +1,130 @@
...
+    }
+  ) {
+    this._deferredErrors = {};
+    this._deferredSubscriptions = {};
+    this._deferredQueryTracker = deferredQueryTracker;
+    this._options = options;
+    this._pendingQueryTracker = pendingQueryTracker;
+  }
+
+  getFragmentError(dataID: DataID, fragmentID: string): ?Error {
```
devknoll: @josephsavona should we create a type alias for `fragmentID`?

src/store/RelayDeferredFragmentState.js
```diff
@@ -0,0 +1,130 @@
...
+    var subscriptionKey = getSubscriptionKey(dataID, fragmentID);
+    return this._deferredErrors[subscriptionKey];
+  }
+
+  hasFragmentData(dataID: DataID, fragmentID: string): boolean {
+    var hasKeys = false;
```
devknoll: I'm not sure how I feel about the Relay pattern of setting objects to nil to make checks faster... but happy to revert that.

(Maybe it would make sense to streamline that pattern with some helpers, at the very least...)

src/store/RelayPendingQueryTracker.js
```diff
@@ -15,11 +15,10 @@
 
 var Deferred = require('Deferred');
 var DliteFetchModeConstants = require('DliteFetchModeConstants');
-var GraphQLDeferredQueryTracker = require('GraphQLDeferredQueryTracker');
 var Promise = require('Promise');
 var PromiseMap = require('PromiseMap');
 import type RelayQuery from 'RelayQuery';
-import type RelayStoreData from 'RelayStoreData';
+var RelayStoreData = require('RelayStoreData');
```
devknoll: Not sure why the type alone stopped being imported here.

src/store/RelayStore.js
```diff
@@ -13,160 +13,6 @@
...
-    mutation: RelayMutation,
-    callbacks?: RelayMutationTransactionCommitCallbacks
-  ): void {
-    var transaction = new RelayMutationTransaction(mutation);
-    transaction.commit(callbacks);
-  }
-};
-
-module.exports = RelayStore;
+module.exports = new RelayContext();
```
devknoll: We should probably hook the methods called on this instance to show a deprecation warning.

src/store/__mocks__/RelayStore.js
josephsavona: :+1:

src/store/RelayDeferredFragmentState.js
josephsavona: good idea!

src/tools/RelayGarbageCollection.js
```diff
@@ -1,69 +0,0 @@
...
- * of patent rights can be found in the PATENTS file in the same directory.
- *
- * @providesModule RelayGarbageCollection
- * @flow
- * @typechecks
- */
-
-'use strict';
-
-var RelayStoreData = require('RelayStoreData');
```
devknoll: Was this a public API? If so, let me know and I can add it back w/ deprecation warnings.

src/store/RelayDeferredFragmentState.js
josephsavona: We always provide default values for all properties in the constructor to ensure that they all get the same hidden class - it makes a big deal for perf

src/store/RelayStore.js
josephsavona: yup

src/tools/RelayGarbageCollection.js
josephsavona: haha. yup, public :-)

src/container/RelayPropTypes.js
```diff
@@ -43,6 +43,8 @@ const RelayPropTypes = {
     queries: PropTypes.object.isRequired,
     uri: PropTypes.object,
   }),
+
+  RelayContext: PropTypes.object, // TODO: instanceof RelayContext
```
devknoll: Not super sure what the right approach here is. `instanceof` doesn't work with the mock objects. That said, if the mock `RelayStore` becomes a real instance + mocked hooks, this might just work.

src/store/RelayDeferredFragmentState.js
devknoll: We might be misunderstanding each other :smile: 

See https://github.com/facebook/relay/blob/master/src/container/RelayContainer.js#L407 for an example of the old implementation here -- specifically using `null` checking aggressively.

devknoll:
I've rebased this on top of the latest changes. I need to look back through to make sure I didn't accidentally bash over any important changes that landed in master (already fixed one instance of this...) but that'll have to wait.

josephsavona:
@devknoll I would recommend rebasing this after #572 closes. The best bet to avoid bashing changes from earlier commits might be to just pull over all the changes starting with "Separate RelayStore and RelayContext" onto a new branch. Obviously no rush :-)

josephsavona:
There's a lot to review here! My main feedback is that the changes should be sequenced in a way that doesn't rely on exposing a "private" API temporarily - we shouldn't need `RelayContext#_getStoreData` ever. There are a few steps to this:

1. Clean up the interdependencies between the core modules - for example RelayStoreData creates a GraphQLQueryRunner, which accepts a RelayStoreData instance. RelayStoreData shouldn't pass a reference itself to any submodule, instead it should pass only the things that module needs (ex: pass just a pending query tracker to query runner)
2. At this point almost nothing should be calling `RelayStoreData.getDefaultInstance()`. Now we can create `RelayContext`: this should have public methods for everything needed by RelayRenderer and RelayContainer. For example, instead of RelayContainer creating a `GraphQLStoreQueryResolver`, we should have a `RelayContext#createResolver(fragment, id)` method. Now RelayStoreData.getDefaultInstance can be replaced, and instead we can have a default RelayContext instance that holds a RelayStoreData object.
3. Finally, convert to passing the instance of RelayContext via react context instead of using the singleton.

devknoll:
:+1:

devknoll:
Hmm... 

So in the case of number 1, `GraphQLQueryRunner` (and several of the other submodules) require another submodule (such as `RelayPendingQueryTracker`) _in addition to_ actual `RelayStoreData` methods (`readFromDiskCache`, `handleQueryPayload`, etc).

devknoll:
Hmm...

So the issue I see with number 1 is that most modules (like `GraphQLQueryRunner`) require references to other objects (like `RelayPendingQueryTracker`) in _addition_ to RelayStoreData itself (for functions like `readFromDiskCache`).

Any thoughts on how to deal with that?

josephsavona:
Gooe point about needing access to those functions, this is definitely pretty involved. If you don't mind, I may go ahead and experiment with this.

devknoll:
Go for it :+1: 

I'm looking at `GraphQLQueryRunner`. If we defined these:

```js
export type RelayQueryDiffer =
  (root: RelayQuery.Root) => Array<RelayQuery.Root>;

export type RelayQueryChecker =
  (root: RelayQuery.Root) => boolean;

export type RelayCacheReader = {
  hasCache: () => boolean;
  readFromCache(queries: RelayQuerySet, callbacks: CacheReadCallbacks): void;
}
```

And then did this:

```js
class GraphQLQueryRunner {
  constructor({
    queryDiffer: RelayQueryDiffer,
    queryChecker: RelayQueryChecker,
    cacheReader: RelayCacheReader,
    pendingQueryTracker: RelayPendingQueryTracker,
  }) {
    // ...
  }
}
```

We gain a lot of types... but avoid leaking `RelayStoreData` and possibly make it a little easier to refactor it down later on...

devknoll:
Here's a more complete example of what I meant... https://github.com/devknoll/relay/commit/e2622321eada07c0c6ca69414a2404778bdfebe4

Issue #542, opened 2015-11-02
Simplify RANGE_ADD mutations


josephsavona:
RANGE_ADD mutations are common cause of confusion for users. In particular, if the application does not query for the connection, the field for the new edge is not included in the mutation fragment. This winds up issuing [this warning](https://github.com/facebook/relay/blob/411780d258fe3d4845ca9ab5d9f292100f2306c4/src/traversal/writeRelayUpdatePayload.js#L326-L333). Users expect the new edge to be fetched, but Relay doesn't have sufficient information to do so. This is because Relay doesn't know what fields should be fetched.

We should make it easier for users to get predictable behavior with range add mutations - either by strict `invariant`s to require the connection to be queried in advance or, possibly, with some other way to define the fields that should be queried on an edge in the case that the connection is not already queried.

josephsavona:
per @steveluscher:

> The task. Given a RANGE_ADD mutation config:
> 1) If Relay never queries for the new edge because they were eliminated during the intersection of the tracked and fat queries, explain that there's a client-side problem and how to fix it (namely to make sure the parent/connection is used – ie. tracked – in your app)
> 2) If the server never responds with the newly created edge despite having been queried for it, explain that the problem is on the server side.

Gregoor:
Shouldn't it be possible as well to have mutations that change connections which might or might not be tracked? Why even throw an error in the case where the connection is not tracked (and thereby no edge is returned)?

dminkovsky:
Re:

> Users expect the new edge to be fetched, but Relay doesn't have sufficient information to do so. This is because Relay doesn't know what fields should be fetched. —@josephsavona 

In troubleshooting this issue, I considered this may be the case, but then thought: "but then why `RANGE_ADD` configuration have the `edgeName` field? Isn't that how Relay determines what to query for?" I guess just knowing the `edgeName` field isn't enough because it doesn't know the structure of the node? Might this structure come from the fat query?

wincent:
I'm going to fold it into a checklist in #538, which is about overhauling the mutations API. (It will be easier to manage with the various interrelated mutations issues in a single place.)

PR #574, opened 2015-11-10
RANGE_ADD - Better warning messages when new edge is missing

xuorig:
Possible improvement for #542

Lot's of issues and misunderstanding with that one, range add mutations can be confusing with the current warning messages.

Better solution as @steveluscher said:
  - When relay does not query the new edge because it was not included in the intersection of tracked and fat queries, show a different warning saying it's a client-side problem ( Connection was never used in the app )
  - If the edge was queried but is not in the payload, explain it's a server side problem.

In `handleRangeAdd`, if the edge is missing from the payload, check if the operation `RelayQueryMutation` contained the newEdge field. If it didn't, it means the new edge was not included because it wasn't in the intersection of the tracked/fat query. If it did, it means the server should've returned that field and didn't.

I've added a function `getNodeByFieldName` to check if the operation contained the field to get the new edge. Not sure if it is the best way to check that, let me know.

The actual warning messages probably need some work, let me know what you think would be the best messages for these cases.

### DIFF
--- a/src/mutation/RelayMutationQuery.js
+++ b/src/mutation/RelayMutationQuery.js
@@ -30,6 +30,7 @@ const nullthrows = require('nullthrows');
 const inferRelayFieldsFromData = require('inferRelayFieldsFromData');
 const intersectRelayQuery = require('intersectRelayQuery');
 const invariant = require('invariant');
+const warning = require('warning');
 
 type BasicMutationFragmentBuilderConfig = {
   fatQuery: RelayQuery.Fragment;
@@ -214,6 +215,17 @@ var RelayMutationQuery = {
           });
         } else {
           // If the connection is not in `rangeBehaviors`, re-fetch it.
+          warning(
+            false,
+            'Relay.Mutation: Since the connection %s of ' +
+            'the field %s with id %s matched none of the rangeBehaviors ' +
+            'specified in your RANGE_ADD config, the entire connection will be ' +
+            'refetched. Configure a range behavior for this connection to fetch ' +
+            'only the new edge and to enable optimistic mutations.',
+            trackedConnection.getStorageKey(),
+            parentName,
+            parentID
+          );
           keysWithoutRangeBehavior[trackedConnection.getShallowHash()] = true;
         }
       });
--- a/src/mutation/__tests__/RelayMutationQuery-test.js
+++ b/src/mutation/__tests__/RelayMutationQuery-test.js
@@ -13,7 +13,9 @@
 
 require('configureForRelayOSS');
 
-jest.dontMock('RelayMutationQuery');
+jest
+  .dontMock('RelayMutationQuery')
+  .mock('warning');
 
 const GraphQLMutatorConstants = require('GraphQLMutatorConstants');
 const Relay = require('Relay');
@@ -456,6 +458,45 @@ describe('RelayMutationQuery', () => {
       `);
       expect(node)
         .toEqualQueryNode(expected);
+    });
+
+    it('warns when rangeBehaviors dont match with tracked connections', () => {
+      tracker.getTrackedChildrenForID.mockReturnValue(getNodeChildren(Relay.QL`
+        fragment on Feedback {
+          comments(orderby:"ranked_threaded",first:"10") {
+            edges {
+              node {
+                body {
+                  text
+                }
+              }
+            }
+          }
+        }
+      `));
+      var node = RelayMutationQuery.buildFragmentForEdgeInsertion({
+        fatQuery,
+        tracker,
+        connectionName: 'comments',
+        parentID: '123',
+        edgeName: 'feedbackCommentEdge',
+        parentName: 'feedback',
+        rangeBehaviors: {
+          '': 'append',
+          'orderby(recent)': 'append',
+        },
+      });
+
+      expect([
+        'Relay.Mutation: Since the connection %s of ' +
+        'the field %s with id %s matched none of the rangeBehaviors ' +
+        'specified in your RANGE_ADD config, the entire connection will be ' +
+        'refetched. Configure a range behavior for this connection to fetch ' +
+        'only the new edge and to enable optimistic mutations.',
+        'comments{orderby:"ranked_threaded"}',
+        'feedback',
+        '123'
+       ]).toBeWarnedNTimes(1);
     });
 
     it('refetches connections in the absence of a range config', () => {
--- a/src/traversal/__tests__/writeRelayUpdatePayload-test.js
+++ b/src/traversal/__tests__/writeRelayUpdatePayload-test.js
@@ -15,8 +15,7 @@ require('configureForRelayOSS');
 
 jest
   .dontMock('GraphQLRange')
-  .dontMock('GraphQLSegment')
-  .mock('warning');
+  .dontMock('GraphQLSegment');
 
 const GraphQLMutatorConstants = require('GraphQLMutatorConstants');
 const Relay = require('Relay');
@@ -735,7 +734,7 @@ describe('writePayload()', () => {
       edgeID = generateClientEdgeID(connectionID, commentID);
     });
 
-    it('warns if the created `edge` field is missing in the payload', () => {
+    it('handles case when created `edge` field is missing in the payload', () => {
       var input = {
         actor_id: 'actor:123',
         [RelayConnectionInterface.CLIENT_MUTATION_ID]: '0',
@@ -786,13 +785,6 @@ describe('writePayload()', () => {
         payload,
         {configs, isOptimisticUpdate: true}
       );
-
-      expect([
-        'writeRelayUpdatePayload(): Expected response payload to include the ' +
-        'newly created edge `%s` and its `node` field. Did you forget to ' +
-        'update the `RANGE_ADD` mutation config?',
-        'feedbackCommentEdge',
-      ]).toBeWarnedNTimes(1);
 
       // feedback is updated, but the edge is not added
       expect(queueStore.getField(connectionID, 'count')).toBe(2);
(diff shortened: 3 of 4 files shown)


src/traversal/writeRelayUpdatePayload.js
```diff
@@ -328,13 +328,23 @@ function handleRangeAdd(
...
   var edgeNode = edge && getObject(edge, NODE);
   if (!edge || !edgeNode) {
-    warning(
-      false,
-      'writeRelayUpdatePayload(): Expected response payload to include the ' +
-      'newly created edge `%s` and its `node` field. Did you forget to ' +
-      'update the `RANGE_ADD` mutation config?',
-      config.edgeName
-    );
+    if (operation.getNodeByFieldName(config.edgeName)) {
```
josephsavona: use `operation.getFieldByStorageKey(config.edgeName)`

src/query/RelayQuery.js
```diff
@@ -247,6 +247,11 @@ class RelayQueryNode {
     return this.getFieldByStorageKey(field.getStorageKey());
   }
 
+  getNodeByFieldName(fieldName: String): ?RelayQueryNode {
+    let children = this.getChildren()
+    return children.find(c => c.__concreteNode__.fieldName === fieldName);
+  }
+
```
josephsavona: unnecessary, see below

src/traversal/__tests__/writeRelayUpdatePayload-test.js
```diff
@@ -735,7 +735,7 @@ describe('writePayload()', () => {
       edgeID = generateClientEdgeID(connectionID, commentID);
     });
 
-    it('warns if the created `edge` field is missing in the payload', () => {
+    it('warns if the created `edge` field is missing in the payload and was never queried by client', () => {
```
josephsavona: nit: 80 char limit, also maybe abbreviate to

    it('warns if a new edge was never queried by client', () => {

josephsavona:
cc @steveluscher @kassens

src/traversal/writeRelayUpdatePayload.js
xuorig: ah of course. Updated the PR!

devknoll:
Thank you!!! :heart:

xuorig:
@josephsavona @kassens any comments on this ? Would be helpful to add that in :dancers:

steveluscher:
Thanks for your patience on this. It's taken us a long time to get to it, I know. Here's what we're thinking.

To recap, these are two distinct cases.

1. If the query includes the field but the response payload doesn't, it indicates a server error (the server should have returned `null` if it can't fetch a field).
2. If the query **doesn't** include the field, it indicates that the _range configs_ do not account for the tracked fields. Since this is a problem with the query rather than the response payload, what do you think about _moving_ this warning out of the writer and into `RelayMutationQuery`, making it clear that the problem is range behaviors in RANGE_ADD not matching up with what the application is actually querying for by way of a helpful and well-written error message?

xuorig:
@steveluscher sorry for the wait, (finals week 😪) I have updated the PR with what you have recommended. I make the check in `buildFragmentForEdgeInsertion`. If the newEdge is not present in `mutatedFields` after it's been populated, we can assume `rangeBehaviors` was not matching the trackedConnection and we can warn.

Should this be an invariant or keep it as a warn ? Let me know if this is what you had in mind!

src/mutation/RelayMutationQuery.js
```diff
@@ -242,6 +243,21 @@ var RelayMutationQuery = {
         }
       }
     }
+
+    // Warn if the newly created edge was not included in mutatedFields
+    const edgeField = mutatedFields.find(field => {
```
josephsavona: How about adding an `else` clause at [line 225](https://github.com/facebook/relay/pull/574/files#diff-e8b6d06b88b10658255cc62756b5c59eR225)? `mutatedEdgeFields` is only blank if no range behaviors matched and the new edge field would be missing.

josephsavona:
Thanks for updating this! We should also remove the warning in `writeRelayUpdatePayload` (and just return if appropriate).

xuorig:
Shouldn't we keep the current warning in the case where the server is not returning the newEdge ?

josephsavona:
We can keep warning if the *server* doesn't return the edge, but we should only warn if we asked for the edge and didn't get it. The warning currently still occurs even if no range behaviors matched.

xuorig:
Ah you're right! Updated the PR to address your comments.

src/mutation/__tests__/RelayMutationQuery-test.js
```diff
@@ -458,6 +460,38 @@ describe('RelayMutationQuery', () => {
...
+        }
+      `));
+      var node = RelayMutationQuery.buildFragmentForEdgeInsertion({
+        fatQuery,
+        tracker,
+        connectionName: 'comments',
+        parentID: '123',
+        edgeName: 'feedbackCommentEdge',
+        parentName: 'feedback',
+        rangeBehaviors,
```
josephsavona: nit: i suspect you borrowed this from other test cases in this file, but it would make the test more self-contained and readable if we were explicit about the range behaviors, e.g `rangeBehaviors: {}`.

src/traversal/__tests__/writeRelayUpdatePayload-test.js
```diff
@@ -735,73 +734,6 @@ describe('writePayload()', () => {
       edgeID = generateClientEdgeID(connectionID, commentID);
     });
 
-    it('warns if the created `edge` field is missing in the payload', () => {
```
josephsavona: maybe keep this test case and just remove the expectation about the warning? would be nice to test that the other parts still work.

src/traversal/writeRelayUpdatePayload.js
```diff
@@ -331,13 +331,6 @@ function handleRangeAdd(
   const edge = getObject(payload, config.edgeName);
   const edgeNode = edge && getObject(edge, NODE);
   if (!edge || !edgeNode) {
-    warning(
-      false,
-      'writeRelayUpdatePayload(): Expected response payload to include the ' +
-      'newly created edge `%s` and its `node` field. Did you forget to ' +
-      'update the `RANGE_ADD` mutation config?',
-      config.edgeName
-    );
```
josephsavona: :+1:

josephsavona:
Sorry - a few more nits, but otherwise this looks great :-)

src/traversal/__tests__/writeRelayUpdatePayload-test.js
xuorig: Good point. I renamed the test and removed the warn assert!

src/mutation/__tests__/RelayMutationQuery-test.js
xuorig: Much clearer that way!

steveluscher:
Thanks for your patience, @xuorig. We just went over this internally, and I'd like to share a summary of our discussion. I wrote:

> Let me try to work through this to make sure I understand it.

> Client devs have the option of configuring a RANGE_ADD config or not. If they do configure one, it's likely because they expect it to be of use.

> Now, if you've specified a RANGE_ADD config, there are tracked connections that match, but none of them match your rangeBehaviors, it's because you have a connection in your app with arguments that you haven't accounted for in your RANGE_ADD config. This is the aim of the warning, to say “hey… we found at least one tracked connection for that parentID/connectionName combination, but we couldn't find an associated rangeBehavior – did you forget to configure one?”

> > @yungsters: …this warning … can be triggered [if] the key does not have a range behavior defined.

> Interesting. To fallback by refetching the whole connection is good (and what the code does now if you don't have a matching range behavior) but one of the nice things about RANGE_ADD configs is to prevent refetching and to enable optimistic mutations. When you miss a rangeBehavior, you lose all of that without warning.

What if we changed the warning to indicate that the entire connection has been refetched and that the developer can't expect optimistic mutations to work, with a little bit about how to write a more efficient and optimism-compatible mutation. Something like:

> “Relay.Mutation: Since the connection `ships(orderby:"latest")` of the field `faction` with id `123` matched none of the rangeBehaviors specified in your `RANGE_ADD` config, the entire connection has been refetched. Configure a range behavior for this connection to fetch only the new edge and to enable optimistic mutations. See http://… for more information.”

The only problem here is that there's no way for the developer to squelch this warning if their desired behavior _was_ to refetch the connection. Thoughts, @yuzhi @xuorig @yungsters?

yungsters:
I discussed with @yuzhi a bit. Currently, setting a range behavior configuration to `null` means the range should not be refetched at all. I think we should change this to `IGNORE` instead of `null`.

We can add a `REFETCH` or `ALL` constant to squelch the warning.

xuorig:
Correct me if I'm wrong but If the desired behavior was to refetch the connection, why would there be a `RANGE_ADD` then ? wouldn't you use something like `REQUIRED_CHILDREN` instead ?

> We can add a REFETCH or ALL constant to squelch the warning.
That would definitly help remove any ambiguity about the behavior.

yungsters:
A `RANGE_ADD` may be an `APPEND` for some but `REFETCH` for others (e.g. if a particular argument causes sorting that can only be determined by the server).

xuorig:
@steveluscher @yungsters I propose that we change the warning message to what @steveluscher proposed, where we clearly state which connection wasn't found in the rangeBehaviors and in another PR we address the `REFETCH` / `ALL`  and `IGNORE` behaviors ?

xuorig:
I have modified the PR with the message @steveluscher proposed. I think this might be a good first step to clear up the small confusion people have with `RANGE_ADD` mutations.

yungsters:
@facebook-github-bot import

PR #575, opened 2015-11-10
Fixed return an incorrect promise in RelayDefaultNetworkLayer.sendQueries.

ErrorPro:
There is return an other promise in sendQueries function. When you do 
```javascript
request.resolve({response: payload.data})
```
you resolve request's promise, but you don't return it further. So we have problem in upper functions which wait for resolving or rejecting promise in sendQueries.

### DIFF
--- a/src/network-layer/default/RelayDefaultNetworkLayer.js
+++ b/src/network-layer/default/RelayDefaultNetworkLayer.js
@@ -89,7 +89,8 @@ class RelayDefaultNetworkLayer {
         }
       }).catch(
         error => request.reject(error)
-      )
+      ),
+      request
     )));
   }
 


src/network-layer/default/RelayDefaultNetworkLayer.js
```diff
@@ -89,7 +89,8 @@ class RelayDefaultNetworkLayer {
         }
       }).catch(
         error => request.reject(error)
-      )
+      ),
+      request
```
kassens: Unless I missed something, the newly added `request` is the second argument to the `requests.map(...)` call on line 70. If that's the case, this should not change the behavior as the arrow function is automatically bound to the parents scope.

ErrorPro: The problem isn't context. There're two promises: 1 - request ( argument ), 2 -this._sendQuery(request). And the function returnes the second promise, but works with the first.

josephsavona: this doesn't seem to wait for `this._sendQuery(request)`, since it maps the `requests` back to themselves. what exact problem is this solving?

ErrorPro: Ok let me explain: Here is my code: I want to intercept query and mutation requests. 
```Javascript
Relay.injectNetworkLayer(
      {
        ...defaultNetworkLayer,
        sendQueries: (queryRequests) => {
          return defaultNetworkLayer.sendQueries(queryRequests).then(
            error => console.error(error, 'ERR'),
          )
        }
      }
    );
```
So i pass array of requests to defaultNetworkLayer.sendQueries and im gonna waiting for a promise.
So the sendQueries takes that array, takes each element in array (instance of RelayQueryRequest) resolves or rejects it and return a different promise.  So you don't resolve that promise which you return, you resolve RelayQueryRequest. I can make example in plnk or smth else.

ErrorPro: http://jsbin.com/favage/26/edit?js,console

ErrorPro: Here is : http://jsbin.com/zelerisose/17/edit?js,console
It is what i have in my project right now. The same result.

josephsavona: The code as implemented is equivalent to `Promise.all(requests.map(req => (foo, req)))` which is equivalent to `Promise.all(requests.map(req => req))` - in other words, the Promise.all does not wait for the results of `_sendQuery` or any of the chained promises. The default network layer intentionally does not provide access to the underlying payloads (note plural: there is no single `response`). 

I would recommend implementing a custom network layer instead.

josephsavona:
I'm going to close this because we don't want to move forward with the code as written, but feel free to continue the discussion if you come up with an alternate approach.

PR #577, opened 2015-11-11
Stop requiring Promise, adjust requires tooling appropriately

zpao:
`fbjs-scripts` 0.5 hasn't been shipped yet but tested locally with `npm link`

cc @DmitrySoshnikov


josephsavona:
@zpao do you mind importing this? lgtm

zpao:
Will do in the AM. Would like to make sure https://github.com/facebook/fbjs/pull/80 is right before shipping 0.5

zpao:
@facebook-github-bot import

zpao:
I got into a state where tests weren't passing but all cleared up now. I split out the changes to the actual `require` statements since @DmitrySoshnikov is doing that separately - this will go in after.

@facebook-github-bot import

Issue #533, opened 2015-11-02
babel-relay-plugin is not working with babel 6.x


mapduty:
```
Module build failed: TypeError: Plugin is not a function
    at getBabelRelayPlugin.js:44:12
```

schweller:
That's most probably because babel 6.x changed smoothly the way that babel's plugin API works.
Changing this lines should get the job done: 
https://github.com/facebook/relay/blob/master/scripts/babel-relay-plugin/src/getBabelRelayPlugin.js#L47-L48

Docs: http://babeljs.io/blog/2015/10/29/6.0.0/

teaegg:
I simply removed the `new Plugin()` and rebuild,  got error

```
Cannot read property 'opts' of undefined
    at PluginPass.Program (node_modules/babel-relay-plugin/lib/getBabelRelayPlugin.js:49:20)
``` 

after converted the router in relay-starter-kit
```javascript
export default class extends Relay.Route {
  static queries = {
    viewer: () => Relay.QL`
      query {
        viewer
      }
    `,
  };
  static routeName = 'AppHomeRoute';
}
```
to: 
```javascript
class AppHomeRouter extends Relay.Route {}
AppHomeRouter.queries = {
  viewer: () => Relay.QL`
    query {
       viewer
    }
  `,
};
AppHomeRouter.routeName = 'AppHomeRoute';
module.exports = AppHomeRouter;
```
error disappeared and transformed successfully. however, it's still not working.

josephsavona:
Yup! The new version *just* came out, and we'll update as soon as feasible.

alvinthen:
Any workaround for now besides reverting back to babel v5.x?

philostler:
+1 just hit this and would love a workaround without have to downgrade

indiejoseph:
+1

mikestaub:
+1

josephsavona:
@sebmck himself is helping out on this - hang in there!

amenuor:
Hello, I hit this issue too but I couldn't wait :)
I have a working solution at this repo: https://github.com/YanchWare/relay 
Maybe it can help for the final solution.

It is not based on the latest version of your code, but on the release 0.3.0 (I think - version from facebook/relay-starter-kit)

leebenson:
This looks pretty easy to fix... I got `scripts/build-lib` working with a combination of the following:

**Bump babel versions**

Something like this in `package.json`:

```
  "devDependencies": {
[...]

PR #579, opened 2015-11-11
Fix babel-relay-plugin for babel 6.x

amenuor:
Fixes issues with transpiling graphql queries code with babel 6.x as described in #533


leebenson:
@amenuor @joshuaevenson - Any news on this? It seems to be failing at:

```
Cannot find module 'babel-types/lib'
```

amenuor:
@leebenson of course the build fails. To have that work correctly the whole project need to be updated to use babel 6. 
I have looked briefly into it and it is quite a major change. 
I can work on it if needed since I started this...

leebenson:
@amenuor I included you on this since you kindly submitted a PR, but this is more a prompt for FB re: next steps for fix + merge. Babel 6.x seems to be gaining traction, and I imagine this is quickly going to become a roadblock to adoption.

amenuor:
Ah ok :) Let me know if I can help.

jwaldrip:
This is definitely becoming a roadblock. :+1: to getting this merged in.

josephsavona:
@amenuor this PR only contains changes to the files in `lib/`, which are generated. Can you commit the changes to the original `src/` files?

amenuor:
@josephsavona as I stated also in the issue #533 this code is mostly for guidance.
I just focused on how to make it work for the projects I am working on which depend on this plugin.
For simplicity I just modified the output of the generated files. However, from that it should be quite straightforward to modify the source files.

I will check later today if you got this to work, if not I will work on the source files.
Let's get it done :)

amenuor:
@josephsavona I have checked-in the major changes to have this plugin working on babel6.

I have had some issues with es6 Spread operator. Somehow it doesn't work on babel6 (it gives Syntax errors) or I couldn't make it work. Based on http://babeljs.io/repl/ (which still run babel 5.x) I have defined an _extends function and use that instead. Not elegant but it seems to work.

I receive several errors when trying to run the tests and check that everything is actually fine.
Flow doesn't like several hundreds statements in /node_modules/*
Jest complains about some syntax errors related to destructuring assignments as far as I can see...

However, the plugin builds fine... we just need to verify that it actually works :)
I do not have so much experience with Flow and Jest, so if you have maybe you can figure this out faster than I can?

leebenson:
@amenuor for rest/spread, make sure you're including `babel-plugin-syntax-object-rest-spread` in package.json and in your plugins config, as well as `babel-plugin-transform-object-rest-spread` to actually do the _extend transform.

amenuor:
@leebenson well spotted! :) Now it works correctly.
@josephsavona I have checked in the changes, and all tests pass on the plugin side. The build is still failing because of two main issues:

1) the main relay package is not supporting babel6 yet. I only worked on the babel-relay-plugin. I could help also on the main package if needed.
2) Flow gives 7000+ errors. I haven't looked much into it yet, but it seems to be related to the new packages in node_modules (as for instance babel-core).

josephsavona:
If you're on master, babel-relay-plugin should work out of the box with babel 6 as of ba004a4. We'll upgrade the examples, tests, etc to use babel 6 at a later date.

josephsavona:
@amenuor thanks for your help with this!

amenuor:
@josephsavona glad I could help :)

Issue #487, opened 2015-10-18
Resolve queries if all errors are on nullable fields


vladar:
Relay currents rejects requests whenever the `errors` response field is non-empty. However, these errors may all be for nullable fields. Relay should not reject requests in this case, and instead should resolve a `RelayQueryRequest` whenever the `data` field is non-null _or_ there are no `errors` (indicating that all required fields could be resolved).

-- original issue --

Currently when server returns errors for nullable fields Relay will execute fail callbacks and not render.

This is different from GraphQL behavior. GraphQL still returns all other data in response, so it should be possible to render anyway.

 Only errors in non-nullable fields should break the application.

josephsavona:
@vladar Thanks for bringing this up. By errors you're referring to application errors while resolving the nullable fields, as opposed to GraphQL validation errors, correct?

vladar:
Yes. When `resolve` method of some nullable field throws, GraphQL will just return `null` for this field, add entry to `errors`, but still evaluate all other fields and return usable result.

Relay doesn't seem to follow this model and breaks if there is any error in response - `renderFailure` is called on such responses.

josephsavona:
@leebyron @dschafer Is there any way for a client (Relay) to detect whether an error occurred on a nullable field instead of a non-null field?

dschafer:
> @leebyron @dschafer Is there any way for a client (Relay) to detect whether an error occurred on a nullable field instead of a non-null field?

Nothing built in; you could check to see if `data` is completely null before erroring; that feels like the right thing to do here. The server can also customize what it returns in the error, so we could add that data there if we wanted.

josephsavona:
@dschafer Thanks for confirming. 

In this case, `RelayDefaultNetworkLayer` should be changed to resolve the request when `data` is non-null or there are no `errors` (if non-null but there errors just log them to console), and only `reject` the request when data is null and there are errors.

@vladar any interest in submitting a PR? ;-)

vladar:
Yeah, will do in next couple days :)

bsr203:
I have posted a related question at SO http://stackoverflow.com/questions/38284556/relay-graphql-nullable-response-or-a-way-to-catch-query-error

any response is appreciated. thanks.

bsr203:
@josephsavona any hint on how would you model a request where error condition not block rendering the component? I hit a roadblock and just a one liner would help me to proceed. thanks in advance

bsr203:
[...]

PR #594, opened 2015-11-13
Do not reject request when there are errors unless data is null

xuorig:
Some improvement for #487 

Errors on fields should only make the request rejected if the data is empty or null. If we receive some data from the GraphQL server and also errors, `console.error` them but resolve the request.

Implemented this in `RelayDefaultNetworkLayer`
  -  if payload has errors
    -  if data is not null/undefined and is not an empty object.
      - log the errors and resolve the request
    - else ( data is null or an empty object )
      - reject the request with errors

Let me know if the logic can be made simpler and if the error message should be different!

### DIFF
--- a/src/network-layer/default/RelayDefaultNetworkLayer.js
+++ b/src/network-layer/default/RelayDefaultNetworkLayer.js
@@ -72,13 +72,22 @@ class RelayDefaultNetworkLayer {
         result => result.json()
       ).then(payload => {
         if (payload.hasOwnProperty('errors')) {
-          var error = new Error(
-            'Server request for query `' + request.getDebugName() + '` ' +
-            'failed for the following reasons:\n\n' +
-            formatRequestErrors(request, payload.errors)
-          );
-          (error: any).source = payload;
-          request.reject(error);
+          if (payload.data && Object.keys(payload.data).length) {
+            console.error('Server request for query `' +
+              request.getDebugName() + '` ' +
+              'resolved with errors:\n\n' +
+              formatRequestErrors(request, payload.errors)
+            );
+            request.resolve({response: payload.data});
+          } else {
+            var error = new Error(
+              'Server request for query `' + request.getDebugName() + '` ' +
+              'failed for the following reasons:\n\n' +
+              formatRequestErrors(request, payload.errors)
+            );
+            (error: any).source = payload;
+            request.reject(error);
+          }
         } else if (!payload.hasOwnProperty('data')) {
           request.reject(new Error(
             'Server response was missing for query `' + request.getDebugName() +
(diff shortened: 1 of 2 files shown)


devknoll:
Not totally sure how I feel about this...

At the very least, resolve should probably _also_ include any errors, and since this changes the semantics, should be documented as a breaking change.

src/network-layer/default/RelayDefaultNetworkLayer.js
```diff
@@ -72,13 +72,22 @@ class RelayDefaultNetworkLayer {
...
-          request.reject(error);
+          if (payload.data && Object.keys(payload.data).length) {
+            console.error('Server request for query `' +
+              request.getDebugName() + '` ' +
+              'resolved with errors:\n\n' +
+              formatRequestErrors(request, payload.errors)
+            );
+            request.resolve({response: payload.data});
+          } else {
+            var error = new Error(
```
josephsavona: nit: `const error = ...`

src/network-layer/default/RelayDefaultNetworkLayer.js
```diff
@@ -72,13 +72,22 @@ class RelayDefaultNetworkLayer {
...
       ).then(payload => {
         if (payload.hasOwnProperty('errors')) {
-          var error = new Error(
-            'Server request for query `' + request.getDebugName() + '` ' +
-            'failed for the following reasons:\n\n' +
-            formatRequestErrors(request, payload.errors)
-          );
-          (error: any).source = payload;
-          request.reject(error);
+          if (payload.data && Object.keys(payload.data).length) {
```
josephsavona: shouldn't this be `if (payload.data != null)`?

xuorig: I originally wrote it this way, but I think payload.data can be `{}`, an empty object. We still need to reject the request in that case I would think

xuorig:
@devknoll I see what you mean, I guess this solution would make errors inaccessible for anybody wanting to use them when there is data. @josephsavona what do you think ?

steveluscher:
@devknoll or @josephsavona, can you give this one last look, then close if you decide we should leave the current behavior as it is?

steveluscher:
Ping @devknoll. Do we have other ideas, or shall we close this up?

devknoll:
@steveluscher I'm no longer maintaining code that relies on this (hooray for objective opinions!) but still stand by my previous thoughts. 

I think that any errors should also be passed to the success callback (could just be sibling to `response`) and this documented as a breaking change.

My reasoning is: A) this matches the GraphQL behavior and B) an invariant of "errors from the server are always passed to application code" just seems nicer for the developer

steveluscher:
Thanks for the contribution @xuorig! I'm going to close this up. Feel free to open it again if any new information comes up that might change things.

valorize:
I think it would be still a nice feature. Maybe it would be better if configurable. I will prepare something if no one minds.

valorize:
I think it would be still a nice feature. Maybe it would be better if configurable like failOnError which default to true.

Issue #550, opened 2015-11-06
Option for skipping mutations in queue


jardakotesovec:
I have several use cases with optimistic updates  where user can make several changes on the same thing quickly which results in several mutations on the same thing. Lets say I have text input for user name.

If I handle that with optimistic updates - it works nicely, only issue is that it sends request for every letter (serially thanks to collision key), which can take a while. It would be handy for these situations be able to set some option on mutation which says that it can skip all queued mutations with that same collision key.

josephsavona:
That's a good idea, it would be great to have support for this. Any interest in submitting a PR?

jardakotesovec:
@josephsavona Yes.. that should not be that difficult as my first time Relay contribution.

But not sure if make more sense to just silently remove pending transactions from the queue or make them fail so the transaction callback gets always called. It would be necessary indicate that the failing reason (probably via status) is that it is skipped.

I am inclining to the first option - just remove pending transactions (and all relevant clean up) as its easier and I don't see much benefit of calling the failure callbacks. But would like your opinion.

Also we need good api name for it, here are some ideas: `shouldClearCollisionQueue`/`shouldSkipCollisionQueue` or shorter options: `shouldClearQueue`/`shouldSkipQueue`

josephsavona:
Two mutations with the same collision key may be initiated from different parts of UI and be unaware of each other. There should be *some* callback that fires in the case of a mutation to allow for any cleanup, though I agree that `onFailure` isn't appropriate for this case. 

@yungsters @wincent Thoughts on switching to either an `onReadyStateChange`-style callback with props such as `{error, aborted, data}`? Calling `onSuccess` when a mutation isn't intuitive.

devknoll:
What if throttling/debounce was made a user responsibility instead? I.e. Relay just provided control to add/remove mutations from the pending store, and a function to finally commit (and send) the mutation?

devknoll:
Implementing throttle/debounce on top of that would be trivial: create a mutation and add it to the store, but don't commit it. If another mutation comes in, the app can remove the last mutation, and add a new one. After the debounce threshold, it can commit the change (with normal collision key behaviors).

jardakotesovec:
I like @devknoll 's idea as it helps cover reasonable use case when user want to control (postpone) when update it sent to server. 

To keep discussion moving forward, here are some api ideas based on that idea:
Relay.Store.queue(mutation).commit(callbacks) could be equivalent to Relay.Store.Update(mutation, callbacks).

[...]

PR #599, opened 2015-11-14
add updateLocal method to create transaction without comitting it

xuorig:
For #550

Adds an `updateLocal` method to `RelayStore` that creates a transaction in the Queue, but doesn't commit it. 

Returns a reference to the transaction so it can be committed or rollbacked afterwards.

There weren't any tests for update so I added a test for `update` and `updateLocal`, just making sure if the correct functions on the Queue and Transaction are called.

### DIFF
--- a/docs/APIReference-Store.md
+++ b/docs/APIReference-Store.md
@@ -20,6 +20,12 @@ The Relay `Store` provides an API for dispatching mutations to the server.
       Initiate processing of a mutation.
     </a>
   </li>
+  <li>
+    <a href="#applyupdate-static-method">
+      <pre>static applyUpdate(mutation, callbacks)</pre>
+      Adds a MutationTransaction to the queue without committing it.
+    </a>
+  </li>
 </ul>
 
 ## Methods
@@ -28,7 +34,7 @@ The Relay `Store` provides an API for dispatching mutations to the server.
 
 ```
 static update(mutation: RelayMutation, callbacks: {
-  onFailure?: (transaction: Transaction) => void;
+  onFailure?: (transaction: RelayMutationTransaction) => void;
   onSuccess?: (response: Object) => void;
 }): void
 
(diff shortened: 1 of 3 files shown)


josephsavona:
This looks good, and thanks for the tests! Having two such similar methods is redundant though. 

@yungsters @wincent thoughts on making it the default for `update` to not commit the transaction, and require users to call `commit()`? We could keep the old behavior for one release with a warning.

wincent:
> thoughts on making it the default for `update` to not commit the transaction

We should optimize for the most common use case, although figuring out _what_ exactly that is is tricky. My intuition is that controlled inputs bound to mutations are actually the edge case, so the existing auto-commit behavior should be _easy_ and the deferred-commit-managed-by-the-user behavior should be _possible_.

So another possible option to consider:

- Make it a single method, with the commit behavior controllable via an options object where `{commit: true}` is the default.

If we go with two methods, I am not sure the meaning of the `updateLocal` method is obvious from its name. Naming is hard, but other ideas that crossed my mind are `enqueueUpdate`/`commitUpdate`.

Finally, as for the idea of making `update` the only method and not committing by default, I could live with that (and it would conveniently liberate us from having to come up with decent names for two methods).

josephsavona:
Good points. So there are three options:
- two methods, which should then both return a MutationTransaction for consistency 
- one method that auto-commits by default, with an option to not commit. It returns the transaction. This means `store.update(mutation, callbacks).commit()` will throw (already committing) but `store.update(mutation, callbacks, {commit: false}).commit()` is ok
- one method, never auto-commits: `store.update(mutation, callbacks).commit()` or leave off the commit() call to defer. 

The last option feels the least awkward (especially with a rename s/update/enqueueUpdate/), and it also makes the option to enqueue a mutation but not commit it more discoverable.

xuorig:
> the existing auto-commit behavior should be easy and the deferred-commit-managed-by-the-user behavior should be possible.

Totally agree with this, I really think the deferred commit way will be used in edge cases and probably should not be the default.

> This means store.update(mutation, callbacks).commit() will throw (already committing) but store.update(mutation, callbacks, {commit: false}).commit() is ok

Which means this would make sense, but it does seem fairly awkward. I would personally probably go for this anyway I think.

Let me know what you guys decide to go for and I'll update the PR

josephsavona:
@xuorig thanks for your patience. We chatted offline and came up with the following API:

1. `Relay.Store.applyUpdate(mutation, callbacks): RelayMutationTransaction` - does not commit the mutation immediately (e.g: rename `updateLocal` to `applyUpdate`)
2. `Relay.Store.commitUpdate(mutation, callbacks): RelayMutationTransaction` - commits the mutation immediately (this replaces the existing `update`)
3. `Relay.Store.update(...): void` - mark this as deprecated and suggest using `commitUpdate`.

Would you mind changing `updateLocal` -> `applyUpdate` and adding docs for this method (item 1)? We can address 2 & 3 separately (or happy to receive contributions if you have time). Thanks!

xuorig:
Alright, I changed the method name to `applyUpdate` + updated the docs with example and some info.

Squashed + reworded commit also.

I will tackle 2 and 3 in another PR

docs/APIReference-Store.md
```diff
@@ -62,3 +68,36 @@ var mutation = new MyMutation({...});
 
 Relay.Store.update(mutation, {onFailure, onSuccess});
 ```
+
+### applyUpdate (static method)
+
+```
+static applyUpdate(mutation: RelayMutation, callbacks: {
+  onFailure?: (transaction: Transaction) => void;
```
josephsavona: nit: `transaction: RelayMutationTransaction` (also should update the above doc for `update` for consistency)

docs/APIReference-Store.md
```diff
@@ -62,3 +68,36 @@ var mutation = new MyMutation({...});
...
+  onSuccess?: (response: Object) => void;
+}): RelayMutationTransaction
+```
+
+The `applyUpdate` adds a mutation just like `update`, but does not commit it. It returns a `MutationTransaction` that can be committed or rollbacked.
```
josephsavona: `It returns a `RelayMutationTransaction``

josephsavona:
@xuorig awesome, thanks! just two minor nits, otherwise this looks good to go.

xuorig:
Addressed your nits!

josephsavona:
@facebook-github-bot import

josephsavona:
:+1:

PR #603, opened 2015-11-16
Make RelayStore Contextual

devknoll:
This is a cleaner shot at #570.

* `RelayGarbageCollection` hasn't been updated yet (needs clarification)
* `RelayRenderer` takes `RelayContext` as a prop (instead of as context from RelayRootContainer).
* Not using Joe's newest `GraphQLStoreQueryResolver` changes.
* Adds `Relay.Store` deprecation calls.
* Fixes `RelayStore` mock.

This ignores the clean up of interdependencies in `RelayStoreData`, since the discussion on that is ongoing.

### DIFF
--- a/src/container/RelayRenderer.js
+++ b/src/container/RelayRenderer.js
@@ -13,18 +13,18 @@
 
 'use strict';
 
-const GraphQLFragmentPointer = require('GraphQLFragmentPointer');
 const React = require('React');
+var ReactDOM = require('ReactDOM');
+import type RelayContext from 'RelayContext';
 import type {RelayQueryConfigSpec} from 'RelayContainer';
 const RelayPropTypes = require('RelayPropTypes');
-const RelayStore = require('RelayStore');
-const RelayStoreData = require('RelayStoreData');
 import type {
   Abortable,
   ComponentReadyState,
   ReadyState,
   RelayContainer
 } from 'RelayTypes';
+const RelayStore = require('RelayStore');
 const StaticContainer = require('StaticContainer.react');
 
 const getRelayQueries = require('getRelayQueries');
@@ -37,6 +37,7 @@ type RelayRendererProps = {
   onReadyStateChange?: ?(readyState: ReadyState) => void;
   queryConfig: RelayQueryConfigSpec;
   render?: ?(renderArgs: RelayRendererRenderArgs) => ?ReactElement;
+  relayContext: RelayContext;
 };
 type RelayRendererRenderArgs = {
   done: boolean;
@@ -123,10 +124,17 @@ class RelayRenderer extends React.Component {
     super(props, context);
     this.mounted = true;
     this.state = this._runQueries(this.props);
+
+    this.props.relayContext.injectBatchingStrategy(
+      ReactDOM.unstable_batchedUpdates
+    );
   }
 
   getChildContext(): Object {
-    return {route: this.props.queryConfig};
+    return {
+      route: this.props.queryConfig,
+      relay: this.props.relayContext,
+    };
   }
 
   /**
@@ -152,7 +160,9 @@ class RelayRenderer extends React.Component {
       if (readyState.ready && !props) {
         props = {
           ...queryConfig.params,
-          ...mapObject(querySet, createFragmentPointerForRoot),
+          ...mapObject(querySet, (query) =>
+            this.props.relayContext.createFragmentPointerForRoot(query)
+          ),
         };
       }
       this.setState({
@@ -171,8 +181,8 @@ class RelayRenderer extends React.Component {
     };
 
     const request = forceFetch ?
-      RelayStore.forceFetch(querySet, onReadyStateChange) :
-      RelayStore.primeCache(querySet, onReadyStateChange);
+      this.props.relayContext.forceFetch(querySet, onReadyStateChange) :
+      this.props.relayContext.primeCache(querySet, onReadyStateChange);
 
     return {
       activeComponent: this.state ? this.state.activeComponent : null,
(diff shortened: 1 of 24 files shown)


src/container/RelayRenderer.js
```diff
@@ -123,10 +124,17 @@ class RelayRenderer extends React.Component {
     super(props, context);
     this.mounted = true;
     this.state = this._runQueries(this.props);
+
+    this.props.relayContext.injectBatchingStrategy(
+      ReactDOM.unstable_batchedUpdates
+    );
```
denvned: `ReactDOM.unstable_batchedUpdates` is not suitable for server-side rendering. And generally, I believe `RelayRenderer` should not be responsible for defining the batching strategy, and should not override it either. It may cause problems such as: https://github.com/denvned/isomorphic-relay-router/issues/5#issuecomment-156920743. Still, `ReactDOM.unstable_batchedUpdates` might be the default injected by the constructor of `RelayContext`.

src/container/RelayRenderer.js
```diff
@@ -283,25 +293,22 @@ class RelayRenderer extends React.Component {
...
-    null;
-}
-
 RelayRenderer.propTypes = {
   Component: RelayPropTypes.Container,
   forceFetch: PropTypes.bool,
   onReadyStateChange: PropTypes.func,
   queryConfig: RelayPropTypes.QueryConfig.isRequired,
   render: PropTypes.func,
+  relayContext: RelayPropTypes.RelayContext.isRequired,
```
denvned: May be just `context` is better?

src/container/RelayRootContainer.js
```diff
@@ -102,6 +105,7 @@ function RelayRootContainer({
   renderFetched,
   renderLoading,
   route,
+  relayContext,
```
denvned: May be just `context` is better?

denvned:
`RelayRenderer` should abort pending request and run new queries when the Relay context is replaced. 

`RelayContainer` should also react to the change of the context.

src/container/RelayRenderer.js
denvned: Or even better, do not inject `ReactDOM.unstable_batchedUpdates` at all. User can inject that himself if he needs it.

devknoll: I think it might be good for Relay to have a sensible default here... Maybe we could introduce a `BatchingStrategyLayer` that by default injects `ReactDOM.unstable_batchedUpdates` on the client side, etc. Then you can override this if you'd like.

(In this particular case, the change was made to keep the same behavior as `master`.)

devknoll:
Good catch :+1: Thanks for the feedback!

src/container/RelayRenderer.js
denvned: I think the following can solve the problem, and be compatible with `master`:
```javascript
const defaultContext = new RelayContext();
defaultContext.injectBatchingStrategy(
  ReactDOM.unstable_batchedUpdates
);

RelayRenderer.defaultProps = {
  context: defaultContext,
};
```

src/container/RelayRenderer.js
```diff
@@ -283,25 +293,22 @@ class RelayRenderer extends React.Component {
...
+};
+
+RelayRenderer.defaultProps = {
+  relayContext: RelayStore,
```
denvned: Why not `new RelayContext()`?

denvned: Since you are deprecating `RelayStore`, may be `RelayContext.getDefaultInstance()` is better?

src/container/RelayRenderer.js
denvned: But the later approach have a problem that only the default instance will have the batching strategy injected. Injecting `ReactDOM.unstable_batchedUpdates` to every new `RelayContext` by default is also bad, because in couples Relay to ReactDOM too much.

I think, that could be solved by subclassing `RelayContext` with `ReactDOMRelayContext`, so only the later will inject `ReactDOM.unstable_batchedUpdates`. And make `ReactDOMRelayContext` the default for the `context` property of `RelayRenderer`:
```javascript
RelayRenderer.defaultProps = {
  context: ReactDOMRelayContext.getDefaultInstance(),
}
```
And make `RelayStore` an alias of `ReactDOMRelayContext.getDefaultInstance()` for backward compatibility.

cc @josephsavona

src/container/RelayRenderer.js
josephsavona: `relayContext` is a required prop, there's no need for a default

src/store/RelayDeferredFragmentTracker.js
```diff
@@ -0,0 +1,130 @@
+/**
+ * Copyright 2013-2015, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @providesModule RelayDeferredFragmentTracker
```
josephsavona: heads up that @steveluscher is working on a diff that will completely remove the need for the deferred tracker

src/store/__mocks__/RelayStore.js
```diff
@@ -9,8 +9,8 @@
```
josephsavona: tests should no longer be relying on RelayStore, so this mock can probably be simplified to exporting the actual module

devknoll: Hmm, lots of tests are relying on the mocks here though. Do you think those should be moved into `RelayContext` then?

src/store/RelayDeferredFragmentTracker.js
devknoll: :heart:

src/container/RelayRenderer.js
devknoll: Default was added here because _tons_ of tests are shallow rendering a `RelayRenderer`. Will update those when I get a chance :+1:

PR #606, opened 2015-11-17
Replace toGraphQL/fromGraphQL with toJSON/fromJSON in RelayQuery

denvned:
`RelayQuery` nodes are [supposed](https://github.com/facebook/relay/blob/v0.5.0/src/query/RelayQuery.js#L92) to be immutable. But `toGraphQL`, which is used to serialize queries, breaks immutability by replacing the concrete node of the `RelayQuery` node with a plain object version of it. So, for example, `RelayQuery.Fragmet.getConcreteFragmentID()` now returns another ID... That sometimes causes problems.

Initially, I was planning to solve that by disabling replacement of the old concrete node with the concrete node produced by `toGraphQL`, since it was performed for caching purposes only. But, ended up with a bigger refactor by completely replacing `toGraphQL` with a new method `toJSON()` on `RelayQuery.Root`, `RelayQuery.Field`, and `RelayQuery.Fragment`. :smile: The implementation was copied from `toGraphQL` without noticeable modifications.

**UPDATE:**
Also replaced `fromGraphQL` with `fromJSON`.


src/query/RelayQueryPath.js
```diff
@@ -179,9 +178,12 @@ class RelayQueryPath {
     while (next) {
       let node = getShallowClone(next._node);
       if (node instanceof RelayQuery.Root) {
-        path.unshift(toGraphQL.Query(node));
+        path.unshift(node.toJSON());
       } else {
-        path.unshift(toGraphQL.Field(node));
+        invariant(node instanceof RelayQuery.Field,
+          'RelayQueryPath.toJSON(): Invalid node.'
+        );
```
josephsavona: this shouldn't be necessary

src/query/RelayQuery.js
```diff
@@ -572,6 +574,42 @@ class RelayQueryRoot extends RelayQueryNode {
     }
     return super.equals(that);
   }
+
+  toJSON(): ConcreteQuery {
```
josephsavona: As of @yuzhi's c9fc00541e282e84b6d5e9f58495de477e0753c6, the concrete node no longer needs to be cached, so this can just be `return QueryBuilder.createQuery({...})`

src/query/RelayQuery.js
```diff
@@ -910,6 +950,19 @@ class RelayQueryFragment extends RelayQueryNode {
     }
     return super.equals(that);
   }
+
+  toJSON(): ConcreteFragment {
```
josephsavona: same, no caching needed

src/query/RelayQuery.js
```diff
@@ -1241,6 +1296,17 @@ class RelayQueryField extends RelayQueryNode {
     return field;
   }
 
+  toJSON(): ConcreteField {
```
josephsavona: and same here

src/query/__tests__/RelayQuery_toJSON-test.js
```diff
@@ -18,9 +18,8 @@ var Relay = require('Relay');
 var filterObject = require('filterObject');
 var fromGraphQL = require('fromGraphQL');
 var splitDeferredRelayQueries = require('splitDeferredRelayQueries');
-var toGraphQL = require('toGraphQL');
 
-describe('toGraphQL', function() {
+describe('RelayQuery.toJSON', function() {
```
josephsavona: :+1:

josephsavona:
A couple nits, but this is great!

denvned:
Fixed.

denvned:
@josephsavona I extended the PR by replacing `fromGraphQL` with `fromJSON` (in a separate commit).

josephsavona:
@denvned Thanks again for working on this. Thinking about the implications of landing this, though, there are really two aspects to this change:
- Change query serialization to stop modifying `__concreteNode__` - this is a simple change to make `getConcreteQueryNode` just return the concrete node and not support overwriting it. The results of `toGraphQL` no longer need to be cached.
- Changing the internal API from `to/fromGraphQL` to `to/fromJSON` and rewriting the modules.

The former has immediate value for server rendering and requires minimal API changes (only affects RelayQuery and toGraphQL) - we're happy to support and land this. The former is aesthetic and requires updating a bunch of APIs internally - we don't have the time to support this second change.

denvned:
> we don't have the time to support this second change for now

Dang! I've just updated the PR with the master. OK, I'll send another PR.

denvned:
BTW, yes, that change is mostly aesthetic, but I thought, FB also desired it, because of this:
> [TODO (#6937314): RelayQueryNode support for toJSON/fromJSON](https://github.com/facebook/relay/blob/v0.5.0/src/query/RelayQuery.js#L103)

josephsavona:
> but I thought, FB also desired it, because of this:

Sorry for the confusion - GitHub issues are current, TODOs in code may be out of date.It's probably better to check before making a large change based on a TODO - sorry for the confusion :-/

denvned:
> It's probably better to check before making a large change based only on a TODO.

Luckily, that wasn't a large change. :smiley:

denvned:
> Change query serialization to stop modifying `__concreteNode__` - this is a simple change to make `getConcreteQueryNode` just return the concrete node and not support overwriting it. The results of `toGraphQL` no longer need to be cached.

Done in #614

Issue #205, opened 2015-09-01
Add capability to send multiple mutations in one request


jardakotesovec:
GraphQL make possible to post multiple mutation tasks in one request - is possible to do that with Relay?

clentfort:
Looking at [`sendMutation`](https://github.com/facebook/relay/blob/master/src/network-layer/default/RelayDefaultNetworkLayer.js#L49-L67) in the default network layer I would say no, not possible. It only ever sends one mutation at a time.
Maybe if you explain why you want to send multiple mutations at once either a suggestion on how to improve your design can be made, or it can serve as an example for when multiple mutations are useful.

jardakotesovec:
Lets say that I have context menu when I can change many things about the image: its metadata, tags and groups.  I like the idea of having several simple mutations with clear purpose rather than having one `updateImage' mutation which does everything based on args.

I understand that I could trigger several mutations via separate requests, but its not optimal - it causes more data to send/receive and possibly more re-renders. 

GraphQL support that..  I would think that there was some flexibility reason for it.

But also I can imagine that things can get complicated to handle that..  and they are pretty complicated already..  so also depends on complexity that it would bring...

yungsters:
Relay does not currently support sending multiple mutations for a single request, but there isn't any reason we can't add support for it.

josephsavona:
This is definitely a nice to have, but we haven't seen strong demand for this feature. I'm going to close it due to inactivity, but feel free to comment here and we can reopen.

jwaldrip:
We would love to see this too.

kareem3d:
so I cant do a mutation like this ?
```
setViewer
createProduct
```
That really hurts :(

wincent:
@kareem3d: What would `setViewer` do? (And specifically, what would it do that couldn't be achieved by adding another argument to the input to `createProduct`?) Independently of that, overriding the viewer is a delicate issue in Relay/GraphQL due to the pervasive use of caching (where switching viewers may involve large parts or all of the cache to be invalidated).

kareem3d:
[...]

PR #607, opened 2015-11-17
add commitUpdate method, deprecate update

xuorig:
Follow up to #599, #550 

Deprecate `RelayStore.update`, rename to `RelayStore.commitUpdate`

`commitUpdate` now returns the `RelayMutationTransaction` just like `applyUpdate`.

Changed to `commitUpdate` in the docs.

I used `warning` to deprecate and also call the correct method, let me know if this is the way we should deprecate the function.

### DIFF
--- a/docs/APIReference-Store.md
+++ b/docs/APIReference-Store.md
@@ -15,8 +15,8 @@ The Relay `Store` provides an API for dispatching mutations to the server.
 
 <ul class="apiIndex">
   <li>
-    <a href="#update-static-method">
-      <pre>static update(mutation, callbacks)</pre>
+    <a href="#commitupdate-static-method">
+      <pre>static commitUpdate(mutation, callbacks)</pre>
       Initiate processing of a mutation.
     </a>
   </li>
@@ -30,13 +30,13 @@ The Relay `Store` provides an API for dispatching mutations to the server.
 
 ## Methods
 
-### update (static method)
+### commitUpdate (static method)
 
 ```
-static update(mutation: RelayMutation, callbacks: {
+static commitUpdate(mutation: RelayMutation, callbacks: {
   onFailure?: (transaction: RelayMutationTransaction) => void;
   onSuccess?: (response: Object) => void;
-}): void
+}): RelayMutationTransaction
 
 // Argument to `onFailure` callback
 type Transaction = {
@@ -44,7 +44,7 @@ type Transaction = {
 }
 ```
 
-The `update` method is analagous to dispatching an action in Flux. Relay processes
+The `commitUpdate` method is analogous to dispatching an action in Flux. Relay processes
 the mutation as follows:
 
 - If the mutation defines an optimistic payload - a set of data to apply locally while waiting for the server response - Relay applies this change and updates any affected React components (note that optimistic updates do not overwrite known server data in the cache).
@@ -66,7 +66,7 @@ var onFailure = (transaction) => {
 };
 var mutation = new MyMutation({...});
 
-Relay.Store.update(mutation, {onFailure, onSuccess});
+Relay.Store.commitUpdate(mutation, {onFailure, onSuccess});
 ```
 
 ### applyUpdate (static method)
--- a/src/store/RelayStore.js
+++ b/src/store/RelayStore.js
@@ -22,6 +22,7 @@ var RelayStoreData = require('RelayStoreData');
 
 var forEachRootCallArg = require('forEachRootCallArg');
 var readRelayQueryData = require('readRelayQueryData');
+var warning = require('warning');
 
 import type {
   Abortable,
@@ -161,6 +162,10 @@ var RelayStore = {
     return new RelayQueryResultObservable(storeData, fragmentPointer);
   },
 
+  /**
+   * Adds an update to the store without committing it. The returned
+   * RelayMutationTransaction can be committed or rolled back at a later time.
+   */
   applyUpdate(
     mutation: RelayMutation,
     callbacks?: RelayMutationTransactionCommitCallbacks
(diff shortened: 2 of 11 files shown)


src/store/RelayStore.js
```diff
@@ -161,6 +162,10 @@ var RelayStore = {
     return new RelayQueryResultObservable(storeData, fragmentPointer);
   },
 
+  /**
+   * Adds an update to the store without committing it. The returned
+   * RelayMutationTransaction can be commited or rollbacked at a later time.
```
josephsavona: yay docblocks! maybe "committed or rolled back"

src/store/RelayStore.js
```diff
@@ -171,11 +176,35 @@ var RelayStore = {
     );
   },
 
+  /**
+   * Adds an update to the store and commits it immediately. Returns
+   * The RelayMutationTransaction.
```
josephsavona: s/The/the/

src/store/RelayStore.js
```diff
@@ -171,11 +176,35 @@ var RelayStore = {
...
+   * Method renamed to commitUpdate
+   */
   update(
     mutation: RelayMutation,
     callbacks?: RelayMutationTransactionCommitCallbacks
   ): void {
-    this.applyUpdate(mutation, callbacks).commit();
+    warning(
+      false,
+      'RelayStore.%s is deprecated. Please use RelayStore.%s instead.',
```
josephsavona: nit: backticks around code (`RelayStore.%s`)

josephsavona: also, it might me helpful to call out the existing of `applyUpdate`, this is a good opportunity to inform users of the new method. maybe

```
'`Relay.Store.update` is deprecated. Please use `Relay.Store.commitUpdate`' +
' or `Relay.Store.applyUpdate` instead.'
```

src/store/RelayStore.js
xuorig: Yep makes more sense :dancers:

xuorig:
Addressed the comments :coffee:

src/store/RelayStore.js
kassens: No need for the `%s` placeholders if you have static strings.

The message string has to be static so we can easily group warnings by it; only dynamic parts have to be added with placeholders.

xuorig: @kassens makes sense, removed the placeholders

docs/APIReference-Store.md
```diff
@@ -15,7 +15,7 @@ The Relay `Store` provides an API for dispatching mutations to the server.
 
 <ul class="apiIndex">
   <li>
-    <a href="#update-static-method">
+    <a href="#commitupdate-static-method">
       <pre>static update(mutation, callbacks)</pre>
```
plievone: Also the link text: update -> commitUpdate

docs/APIReference-Store.md
```diff
@@ -44,7 +44,7 @@ type Transaction = {
 }
 ```
 
-The `update` method is analagous to dispatching an action in Flux. Relay processes
```
plievone: While you are here: I guess 'analogous' is more common? :)

xuorig:
@plievone done

plievone:
Perhaps also the few mentions in examples and docs could be commitUpdated, as they are in the same repo and easy to grep?
https://github.com/facebook/relay/search?utf8=%E2%9C%93&q=%22Store.update%22&type=Code

xuorig:
Good catch, I've updated docs and examples

josephsavona:
@facebook-github-bot import

josephsavona:
@xuorig we'll land this as time permits - shipping it requires a codemod bc we don't want warnings everywhere we call `update`.

xuorig:
:+1:

jardakotesovec:
It seems to me that this implementation does not cover use case, when it just wants to avoid stacking many mutations in collision queue.  Which is the use case for updating text input or anything where user can change values quickly and it wants to remove previous transaction if its just sitting on the queue with newer one.

As far as I can see - to put mutation on the collision queue I have to commit it, but I can't rollback if it just sits on the queue and has not been sent, right?

Maybe if rollback would be also possible when transaction is in `COMMIT_QUEUED`?

jardakotesovec:
@josephsavona @xuorig What do you think? Should I provide another PR that adjust `rollback` function that can rollback also transactions with status `COMMIT_QUEUED `?

josephsavona:
> Should I provide another PR that adjust rollback function that can rollback also transactions with status COMMIT_QUEUED?

@jardakotesovec That would be great - all that should be required is updating the status check in `RelayMutationTransaction` to allow calling `rollback` when the transaction is queued. The actual logic for the rollback in `RelayMutationQueue` should already handle this case.

Globegitter:
Does this cover the use case where I want to commit data to my local cache in different places (don't care about the transactions returned) but then just want to update everything that is stored in the cache, but now out of sync with the backend?

jardakotesovec:
@Globegitter It does not..   as decided in #550,  control of transaction handling is given to the programmer. So with this PR, you have to keep track of the transactions on your own. 

But api could be extended for example to either provide function that commit all uncommited transactions as you suggested, or just provide you list of uncommited transactions so you can commit them.

Globegitter:
@jardakotesovec So right now we would have to manually keep track of each transaction that is being returned by `applyUpdate` (and e.g. store them in an array). Then at some point just go through all the transactions and commit them? Would this then make an individual request for each committed transaction or is there a way to commit all transactions in one HTTP request?

jardakotesovec:
@Globegitter I think you should open new issue/PR to keep track of this use case and discuss potential improvements.

Relay currently sends [one](https://github.com/facebook/relay/issues/205) http request per mutation.

PR #608, opened 2015-11-18
new example use-case: bidirectional connections

rblakeley:
Issue #609
The use-case I was going for with this example is bidirectional data relationships. In an app, at any given point, you may wish to have the UI show a user and her associated groups. In another place you may want to show a group and its associated users. I didn't see this represented in the current examples.
```
user { groups }
group { users }
```
In this example users are soccer players and the groups are their positions. In the UI, each player is listed with switches that can be turned on/off to indicate whether that player can play a given position.

Next to the list of players is a list of positions. Below each position is a list of players who are switched on for that position.

You can also add a new player via the text input below the player list.

### DIFF
--- a/examples/user-group/.gitignore
+++ b/examples/user-group/.gitignore
@@ -1,0 +1,4 @@
+.DS_Store
+node_modules
+npm-debug.log
+data/schema.graphql
--- a/examples/user-group/LICENSE
+++ b/examples/user-group/LICENSE
@@ -1,0 +1,31 @@
+BSD License
+
+For Relay Starter Kit software
+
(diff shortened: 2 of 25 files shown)


steveluscher:
This is great! I don't think that we can accept new examples though, because each one adds work for our very small team – to upgrade it with every API change in Relay.

Would you be interested in breaking this out into a repo of its own?

rblakeley:
@steveluscher 
Thanks, that makes sense. I will break this out as an independent example.

rblakeley:
See https://github.com/rblakeley/relay-example-players-positions for the stand-alone repo.

PR #614, opened 2015-11-19
Do not memoize `toGraphQL`

denvned:
First, memoization of `toGraphQL` is no longer requiered. Second, `RelayQuery.Node` is [supposed](https://github.com/facebook/relay/blob/v0.5.0/src/query/RelayQuery.js#L92) to be immutable, while memoization of the result of `toGraphQL` in the `__concreteNode__` property of `RelayQuery.Node` breaks immutability, as, for example, `RelayQuery.Fragmet.getConcreteFragmentID()` returns another ID after the memoization. That sometimes causes problems.


steveluscher:
Thanks for this! Is there a reason that you chose to keep the getter function, rather than to go straight for `__concreteNode__` in all cases?

josephsavona:
A getter function is more consistent with existing `getRoute` and `getVariables` methods - also mediating access with a method gives us option value at virtually no cost.

steveluscher:
(Y)

PR #617, opened 2015-11-20
Fix variables for implicit fragment

denvned:
The problem was that variables of the implicit query fragment were not updated after query parameters had changed. The query produced by `buildRQL.Query` for an implicit definition was not equal to the query produced for an analogous explicit definition (see the added test).

### DIFF
--- a/src/query/__tests__/buildRQL-test.js
+++ b/src/query/__tests__/buildRQL-test.js
@@ -36,8 +36,11 @@ describe('buildRQL', () => {
     });
     MockComponent = React.createClass({render});
     MockContainer = Relay.createContainer(MockComponent, {
+      initialVariables: {
+        if: null,
+      },
       fragments: {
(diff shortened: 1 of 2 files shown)


src/query/__tests__/buildRQL-test.js
```diff
@@ -36,8 +36,11 @@ describe('buildRQL', () => {
     });
     MockComponent = React.createClass({render});
     MockContainer = Relay.createContainer(MockComponent, {
+      initialVariables: {
+        if: null,
+      },
       fragments: {
-        foo: () => Relay.QL`fragment on Node { name }`,
+        foo: () => Relay.QL`fragment on Node { firstName(if: $if) }`,
```
steveluscher: Rather than to edit `MockContainer` could you push your custom mock down into your test? See `MockContainer2`.

denvned:
@steveluscher, Applied your suggestion. Thanks!

Issue #589, opened 2015-11-12
Make RelayRenderer isomorphic


denvned:
I could make a PR based on this: https://github.com/denvned/isomorphic-relay/blob/v0.2.1/src/IsomorphicRenderer.js

Currently, `_runQueries` is an async operation, so it should not be called during server side rendering. That's why I moved it to `componentDidMount`, which is not called on the server.

`isDataReady` checks if the data is already fetched (current implementation is not so important, and might be replaced). If the data is available, initialize with a preloaded state, otherwise with an empty state.

What do you think?

josephsavona:
Great work on `isomorphic-relay`! It is *very* similar to the API we've been exploring internally, and we'd love to have support for this in Relay proper:

```javascript
// SERVER
Relay.injectNetworkLayer(new RelayServerNetworkLayer(...));
Relay.prepare(route, component, ({props, data, error /*, ...*/ }) => {
  if (error) {
    // stuff...
  }
  React.renderToString(
    <YourPageTemplate 
      data={data}>  // opaque data to be passed as JSON to the client
      <RelayRenderer {...props}>
    </YourPageTemplate>
  );
});

// CLIENT
Relay.injectNetworkLayer(new RelayClientNetworkLayer(data)); // pass in the data from the server
ReactDOM.render(
  <RelayRenderer route={...} component={...} />,
  ...
);
```

It would be relatively simple to change `RelayRenderer` to support synchronously rendering the component when its props contained data created by `Relay.prepare`. Basically: if those props are present, render synchronously, otherwise fetch data.

Any interest in submitting a PR? :-)

denvned:
Sure, I'll be glad to submit a PR.

How about triggering synchronous rendering in `RelayRenderer` by passing it a `ReadyState` received from `Relay.prepare`?

I think, the first render on the client should also be synchronous, otherwise server side rendered DOM will not match virtual DOM after the first render (it will be just `<noscript data-reactid=...></noscript>`), so React will not reuse it, also there will be flickering at page load. That's why, I think that data on the client should also be prepared before the first render.

josephsavona:
Good point, injecting the data from the server into the client's network layer means we can't synchronously render on the client. It might be reasonable to adjust the above such that `Relay.prepare` resolves with `{props, data, error}`, to be used on the client as:

```javascript
// SERVER
// note: use the normal network layer
Relay.prepare(route, component, ({props, data, error}) => {/* see above */});

// CLIENT
// note: use the normal network layer
[...]

PR #625, opened 2015-11-22
Make RelayRenderer run queries after mount

denvned:
...to make sure `RelayRenderer` does not run queries during synchronous server-side rendering.

In the `RelayRenderer` tests, I had to replace `ShallowRenderer.render` with the traditional `ReactDOM.render`, because `ShallowRenderer.render` does not invoke some component lifecycle methods including `componentDidMount` that now run queries. But I did not change the logic of the tests, as **visible behaviour of `RelayRenderer` was fully retained.**

Added a test to check that `RelayRender` does not run queries on the server.

This PR is a forerunner of the PR that will make `RelayRenderer` isomorphic (see #589). That next PR will be pretty small and easily comprehensible, because this PR already did much of the necessary dirty work.


denvned:
I would very much appreciate any feedback on this PR. My further work on [the server-side rendering PR](https://github.com/facebook/relay/issues/589#issuecomment-158766319) depends on it.

src/container/__tests__/RelayRenderer_onReadyStateChange-test.js
```diff
@@ -172,7 +174,7 @@ describe('RelayRenderer.onReadyStateChange', () => {
 
   it('is aborted and not mounted when aborted from unmounting', () => {
     expect(request => {
-      ShallowRenderer.unmount();
+      ReactDOM.render(<div />, container);
```
steveluscher: `ReactDOM.unmountComponentAtNode(container);`

src/container/__tests__/RelayRenderer_abort-test.js
```diff
@@ -57,7 +58,7 @@ describe('RelayRenderer.abort', () => {
       toAbortOnUnmount() {
         const {abort, request} = render();
         this.actual(request);
-        ShallowRenderer.unmount();
+        ReactDOM.render(<div />, container);
```
steveluscher: `ReactDOM.unmountComponentAtNode(container);`

src/container/__tests__/RelayRenderer_render-test.js
```diff
@@ -45,31 +59,31 @@ describe('RelayRenderer.render', () => {
         return this.actual.props.shouldUpdate;
       },
       toBeShallowRenderedChild() {
```
steveluscher: If we're going to stick with a change to the deep renderer, let's rename these matchers.

steveluscher:
Sorry for the delay. Love how you've piped everything through `_buildState()`.

Can you talk for a moment about why you switched from the shallow renderer to the fully-fledged renderer? I can only guess that you wanted to use the same type of renderer between the DOM and the server tests, but I'd like to hear your reasoning before I pull this in.

steveluscher:
Oh. Nevermind. You already explained why. I'll pull this in for a deeper review.

@facebook-github-bot import

src/container/__tests__/RelayRenderer_server-test.js
```diff
@@ -0,0 +1,52 @@
...
+const RelayRenderer = require('RelayRenderer');
+const RelayStore = require('RelayStore');
+
+const getRelayQueries = require('getRelayQueries');
+
+describe('RelayRenderer', function() {
+  let MockComponent;
+  let MockContainer;
+
+  let container;
```
steveluscher: Lint: Never used.

src/container/__tests__/RelayRenderer_server-test.js
```diff
@@ -0,0 +1,52 @@
...
+jest.dontMock('RelayRenderer');
+
+const React = require('React');
+const ReactDOMServer = require('ReactDOMServer');
+const Relay = require('Relay');
+const RelayQueryConfig = require('RelayQueryConfig');
+const RelayRenderer = require('RelayRenderer');
+const RelayStore = require('RelayStore');
+
+const getRelayQueries = require('getRelayQueries');
```
steveluscher: Lint: Never used.

denvned:
@steveluscher Thanks for feedback!
Applied all of your corrections.

steveluscher:
Dumb question: since `GraphQLQueryRunner` already [schedules fetches using the `RelayTaskScheduler`](https://github.com/facebook/relay/blob/master/src/legacy/store/GraphQLQueryRunner.js#L258), could we have prevented fetching in the server rendering case by simply injecting a noop scheduler?

```
Relay.injectTaskScheduler(executeTask => {});
```

josephsavona:
talked offline, this would theoretically work but would leak memory (as the scheduled tasks built up in the queue)

denvned:
Also we will still need to run queries on the server to prepare data before rendering, and injecting a noop scheduler might break it because `RelayTaskScheduler` is a global singleton currently, and there might be multiple requests in parallel.

There might also be an exotic cases like the user needs to simultaneously use both, traditional `ReactDOM.render`, and `ReactDOMServer.renderToString` in the browser, so he cannot inject a noop scheduler because it will break traditional render.

As I see it, injecting a scheduler is generally too hacky solution, and it is hard to tell what additional problems it might have. On the other hand, making `RelayRenderer` run queries after mount is pretty clear and predictable.

src/container/RelayRenderer.js
```diff
@@ -116,19 +115,48 @@ const {PropTypes} = React;
...
+      },
+    };
   }
 
   getChildContext(): Object {
     return {route: this.props.queryConfig};
   }
 
+  componentDidMount(): void {
+    this._runQueries(this.props);
```
yungsters: I assume this does not `this.setState(...)` because it is assumed to be identical to the call in `constructor`. Instead of making `this._runQueries` return the new state, how about completely separating the two calls?

denvned: If I understood you correctly, I just committed your suggestion. Thanks!

yungsters:
@denvned I recently committed f1373602073972ffa06854c676cf61f79c4fa3cc which introduces two new props to `RelayRenderer` — `onPrimeCache` and `onForceFetch` — that allow you to supply the behavior when a set of queries is to be fetched. If instead of reordering when queries are run, you simply supplied dummy functions into these props, would that solve your use case?

denvned:
@yungsters Actually, when upgrading [isomorphic-relay](https://github.com/denvned/isomorphic-relay/) to Relay 0.7, I tried to make use of these new props, but it didn't work out. Not running the queries on the server is only part of the deal, we also need to initialize the state with preloaded data before the first render, i.e. before mount. But if pass prepared ready state to `onReadyStateChange` callback before render, React throws an error because  `onReadyStateChange` calls `setState`, and React does not allow calling `setState` before mount.

So, to make these props useful for isomorphic rendering we probably have to allow calling `onReadyStateChange` callback synchronously, before render.

denvned:
Another problem with `onPrimeCache` and `onForceFetch` and isomorphic rendering is that we need to use prepared data for the first render not only on the server, but also on the client, so we have to supply these props on the client also. But on the client the situation is more complex, because in addition to setting prepared state for initial render, we might also need to run actual queries in the case of `forceFetch`, or if some deferrable data were not prepared. In the end, it turns out that the implementation of the function to be passed as `onPrimeCache` and `onForceFetch` would probably be pretty hacky, complicated, and error prone. But I might be mistaken.

PR #626, opened 2015-11-22
Make GraphQLQueryRunner not call diffRelayQueries synchronously

denvned:
There is no good reason to call `diffRelayQueries` synchronously, it only needlessly block execution of other code (e.g. currently it blocks initial render of `RelayRenderer`). So, it have moved to the rest of the code inside task scheduler callback.

Also cleaned up some duplicate code.

API and behaviour have not changed.

### DIFF
--- a/src/legacy/store/GraphQLQueryRunner.js
+++ b/src/legacy/store/GraphQLQueryRunner.js
@@ -76,39 +76,9 @@ class GraphQLQueryRunner {
   run(
     querySet: RelayQuerySet,
     callback: ReadyStateChangeCallback,
-    fetchMode?: string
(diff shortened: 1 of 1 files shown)


yungsters:
This has been rebased and landed as 8058d21f6b9e33702124658d2c19a715027694b9. I apologize for not pulling it in through the typical import toolchain (because we were doing some internal maintenance on it when I was pulling this in).

sahrens:
Thanks for the fix! This made a material impact in the responsiveness of some of our internal apps that use Relay with React Native - diffRelayQueries can sometimes take quite a while on complex apps running on slow Android phones, so it's really helpful to break it off into another async task.

Keep up the awesome work :)

denvned:
@sahrens Nice to know it was so helpful for you!

Issue #615, opened 2015-11-19
rangeBehaviors should be a function that returns a GraphQLMutatorConstants.RANGE_OPERATION


steveluscher:
As mentioned by @johanobergman in #604 and similar to what @AndrewIngram [proposed](https://github.com/facebook/relay/issues/538#issuecomment-156396132) in #538, let's make `rangeBehaviors` a function that receives the connection arguments and returns one of `GraphQLMutatorConstants.RANGE_OPERATIONS`.

```js
getConfigs: [{
  /* ... */
  rangeBehaviors(
    fieldArgs: {[argName: string]: argValue: string}
  ): ?$Enum<typeof GraphQLMutatorConstants.RANGE_OPERATIONS> {
    return isBefore(fieldArgs.startDate, fieldArgs.endDate) ?
      GraphQLMutatorConstants.RANGE_OPERATIONS.APPEND :
      GraphQLMutatorConstants.RANGE_OPERATIONS.PREPEND;
  }
  /* ... */
}]
```

Add an appropriate deprecation warning to `RelayDeprecated`.

Bonus: write a [jscodeshift](https://medium.com/@cpojer/effective-javascript-codemods-5a6686bb46fb#.pe5meqdiz) codemod that will mod this:

```js
rangeBehaviors: {
  'foo(true)': 'prepend',
  'foo(false)': 'append',
}
```

…to this:

```js
rangeBehaviors({foo}) {
  if (foo === true) {
    return 'prepend';
  } else if (foo === false) {
    return 'append';
  }
}
```

johanobergman:
Awesome, this will help a lot since I won't have to track all cached parameter combinations myself.

You could also pass the entire connection to that function and let the developer manually insert the record wherever he wants it. This would be a great deal more flexible than `append`, `prepend` and `remove` are (even though I don't have a use case for it myself right now).

devknoll:
Just wanted to chuck this in here, since it has been discussed a little before: https://github.com/facebook/relay/issues/293#issuecomment-139417086

AndrewIngram:
As mentioned in the related ticket, I'm still not convinced the mutation is the place to handle this. It feels like it could be defined in the actual query on the connection, or in the schema itself (as briefly discussed in the other ticket). If this doesn't make sense, it'd be good to have some insight as to why.

xuorig:
Questions about this issue:

[...]

Issue #538, opened 2015-11-02
[meta] Simpler Mutations API


josephsavona:
This is a meta-task to track progress toward making Relay mutations simpler to both understand and to define in common cases.

In a client/server system there is an unavoidable complexity in handling writes: 
- Making network requests
- Handling error (and possibly retries)
- Sequencing possibly overlapping writes
- Handling optimistic changes and reverting them on error, or clearing them on success
- Handling server response payloads
- (server) Implementing business logic to process incoming writes

Relay handles the vast majority of this complexity on behalf of developers. The tradeoff is that GraphQL mutations are abstracted from the underlying data store, and therefore the system _cannot automatically know what changed_. This requires the developer to tell the system what changed in the form of mutation configs (`getConfigs`). 

There are several avenues for exploration:
1. Simplify the process of defining mutation configs. @steveluscher proposed creating helper functions that would replace the need for the mutation config objects.
2. Require product developers to manually convert the mutation payload into a set of change operations (e.g. `set('record', 'field', 'value')` or `append('connection', 'edgeId')`). If Relay could understand a small set of these change operations, the community could work to define helpers for converting payloads into this form.
3. Create a standard response format for GraphQL mutations to match the change operations from 2, such that mutations would return a description of what changed. This would not reduce the work of defining mutations, but would allow defining the changes once instead of once per client.

It's important to note that the current mutations API is heavily skewed toward practicality: it has allowed us to iterate quickly and produce resilient applications. We're interested in making this API better and welcome contributions from the community. In particular, the best form of contribution is either links to prototypes or pull requests, which will help us and the community understand the practical tradeoffs of any alternative APIs.

terrisgit:
+1

AndrewIngram:
In terms of getConfigs, the biggest issue for me is that rangeBehaviors don't really work that well. Essentially a connection can be augmented such that it's filterable and sortable, by an arbitrary number of arguments. It's not enough to be able to take a single argument and decide whether to append/prepend/ignore. We need to be able to look at all the arguments (at the same time) and return a sort key that tells Relay where to add it to the list (or whether to remove it).

```
rangeBehaviours: (edge, connectionArgs) => {
  // perform some calculations and return a sort key, or null to remove it from the list
}
```

Of course, it wouldn't be enough to just run this against the new edges, it'd need to be done for the existing edges in the connection too.

[...]

PR #639, opened 2015-12-01
Make rangeBehaviors a function

xuorig:
Fix for #615, proposed in #538

Makes `rangeBehaviors` a function that receives the connection arguments and returns one of `GraphQLMutatorConstants.RANGE_OPERATIONS`.

Will try and write a codemod for this during the week if I have time.

To check if the tracked connections where included in rangeBehaviors, the old behavior was to check if the connection's `rangeBehaviorKey` was in the `rangeBehaviors` map. Now that we don't have access to the `rangeBehaviors` map I've followed @josephsavona 's advice. I am fetching the connectionIds and calling `getRangeFilterCalls` on each one to get all the `rangeBehaviors` using the parentID.

Since these rangeBehaviors are not in the same format I've wrote a new function for connections called `getRangeBehaviors`. It is almost the same as `getRangeBehaviorKey` except it returns a list of the actual calls instead of a string representation. I can then make the correct check using these values.

Other thing I've added is a function `getObjectFromCalls` in `writeRelayUpdatePayload`. It transforms calls like this  `[{name: 'orderby', value: 'recent'}]` to an object like this `{orderby: 'recent'}` that can be passed to the `rangeBehaviors` function.

Rest is pretty straight forward I think, let me know what you think!

### DIFF
--- a/examples/todo/js/mutations/AddTodoMutation.js
+++ b/examples/todo/js/mutations/AddTodoMutation.js
@@ -42,12 +42,15 @@ export default class AddTodoMutation extends Relay.Mutation {
       parentID: this.props.viewer.id,
       connectionName: 'todos',
       edgeName: 'todoEdge',
-      rangeBehaviors: {
-        '': 'append',
-        'status(any)': 'append',
-        'status(active)': 'append',
-        'status(completed)': null,
-      },
+      rangeBehaviors: ({status}) => {
+        if (status === 'any' || status === 'active') {
+          return 'append';
+        } else if (status === 'completed') {
+          return null;
+        } else {
+          return 'append';
+        }
+      }
     }];
   }
   getVariables() {
--- a/docs/Guides-Mutations.md
+++ b/docs/Guides-Mutations.md
@@ -329,9 +329,10 @@ Given a parent, a connection, and the name of the newly created edge in the resp
 
   The field name in the response that represents the newly created edge
 
-- `rangeBehaviors: {[call: string]: GraphQLMutatorConstants.RANGE_OPERATIONS}`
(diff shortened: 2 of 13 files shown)


examples/todo/js/mutations/AddTodoMutation.js
```diff
@@ -42,12 +42,15 @@ export default class AddTodoMutation extends Relay.Mutation {
...
-        '': 'append',
-        'status(any)': 'append',
-        'status(active)': 'append',
-        'status(completed)': null,
-      },
+      rangeBehaviors: ({status}) => {
+        if (status === 'any' || status === 'active') {
+          return 'append';
+        } else if (status === 'completed') {
+          return null;
```
josephsavona: should this be "ignore" or "refetch"?

xuorig: Totally :) this PR is 4 months old but I would gladly rebase and modify it if this is still something you wanted to introduce! What do you think?

josephsavona:
@xuorig This is great! We should definitely move in this direction. However a codemod seems to be non-trivial to write, and either way we'd ideally have at least one release that supports both the old and new behaviors. How about instead of immediately *replacing* ``rangeBehaviors` with a function, we instead add the option of defining a `getRangeBehaviors` function. Only one or the other can be provided, and internally RelayMutationQuery can use `getRangeBehaviors` to construct the `rangeBehaviors` object.

xuorig:
That would make a lot of sense, I'm going to rebase and work on this in the next few days :)

xuorig:
Closing in favor of #1054

PR #645, opened 2015-12-03
Update APIReference-Container.md typo

gustavoguichard:
In the current docs it says we should use `prepareVariables` and `initialVariables` as `static` variables of a class.

That doesn't work and is misleading, so I updated it to display it the way it actually works: As keys of an Object.

i.e.: 
```js
// from
Relay.createContainer(Component, {
  static initialVariables = {...}
  ...
}

// to
Relay.createContainer(Component, {
  initialVariables: {...}
  ...
}
```

Hope it helps.


kassens:
Thanks!

@facebook-github-bot import

PR #646, opened 2015-12-04
queued transaction can be also rollbacked

jardakotesovec:
This is follow up to #607 that adds posibility to rollback transaction with status `COMMIT_QUEUED`.

There is one more missing piece for this use case that I was not sure how to approach. 

To replace transaction that is queued I need to first check if its status is still `COMMIT_QUEUED`. Issue is that transaction is removed from pendingTransactionMap once its successfully committed. So I can easily end up with exception from `getStatus()`, which is bit overreaction :-).

So maybe return some status like `FINISHED` instead of throwing if its not in pending queue? I am of course very opened about naming :-).

### DIFF
--- a/src/mutation/RelayMutationQueue.js
+++ b/src/mutation/RelayMutationQueue.js
@@ -134,6 +134,19 @@ class RelayMutationQueue {
 
   rollback(id: ClientMutationID): void {
     const transaction = this._get(id);
+    const collisionKey = transaction.getCollisionKey();
+    if (collisionKey) {
+      const collisionQueue = this._collisionQueueMap[collisionKey];
+      if (collisionQueue) {
+        const index = collisionQueue.indexOf(transaction);
+        if (index !== -1) {
+          collisionQueue.splice(index, 1);
+        }
+        if (collisionQueue.length === 0) {
+          delete this._collisionQueueMap[collisionKey];
+        }
+      }
+    }
     this._handleRollback(transaction);
   }
 
--- a/src/mutation/__tests__/RelayMutationQueue-test.js
+++ b/src/mutation/__tests__/RelayMutationQueue-test.js
@@ -517,4 +517,67 @@ describe('RelayMutationQueue', () => {
       expect(successCallback1).toBeCalled();
     });
   });
+
+  describe('rollback', () => {
+    var mockMutation1, mutationNode, fatQuery;
+
+    beforeEach(() => {
+      fatQuery = Relay.QL`fragment on Comment @relay(pattern: true) {
+        ... on Comment {
+          doesViewerLike
+        }
+      }`;
+      mutationNode = Relay.QL`mutation{commentCreate(input:$input)}`;
+      RelayMutation.prototype.getFatQuery.mockReturnValue(fatQuery);
+      RelayMutation.prototype.getMutation.mockReturnValue(mutationNode);
+      RelayMutation.prototype.getCollisionKey.mockReturnValue('key');
+      RelayMutation.prototype.getVariables.mockReturnValue({});
+      RelayMutation.prototype.getConfigs.mockReturnValue('configs');
+
+      mockMutation1 = new RelayMutation();
+      mockMutation1.getCollisionKey.mockReturnValue('key');
+
+    });
+
+    it('rollback queued transaction', () => {
+      var transaction1 = mutationQueue.createTransaction(mockMutation1);
+      transaction1.commit();
+
+      expect(transaction1.getStatus()).toBe(
+        RelayMutationTransactionStatus.COMMITTING
+      );
+
+      var transaction2 = mutationQueue.createTransaction(mockMutation1);
+      transaction2.commit();
+
+      expect(transaction2.getStatus()).toBe(
+        RelayMutationTransactionStatus.COMMIT_QUEUED
+      );
+
+      var transaction3 = mutationQueue.createTransaction(mockMutation1);
+      transaction3.commit();
+
+      expect(transaction3.getStatus()).toBe(
+        RelayMutationTransactionStatus.COMMIT_QUEUED
+      );
+
+      expect(RelayNetworkLayer.sendMutation.mock.calls.length).toBe(1);
+
+      transaction2.rollback();
+
+      expect(() => transaction2.getStatus()).toFailInvariant(
+        'RelayMutationQueue: `1` is not a valid pending transaction ID.'
+      );
+
+      var request = RelayNetworkLayer.sendMutation.mock.calls[0][0];
+      request.resolve({response: {'res': 'ponse'}});
+      jest.runAllTimers();
+
+      expect(RelayNetworkLayer.sendMutation.mock.calls.length).toBe(2);
+
+      expect(transaction3.getStatus()).toBe(
+        RelayMutationTransactionStatus.COMMITTING
+      );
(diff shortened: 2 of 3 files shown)


src/mutation/RelayMutationQueue.js
```diff
@@ -133,6 +133,20 @@ class RelayMutationQueue {
 
   rollback(id: ClientMutationID): void {
     const transaction = this._get(id);
+
+    const collisionKey = transaction.getCollisionKey();
+    var collisionQueue =
```
xuorig: nit: use `let` here and for `index` :)

src/mutation/RelayMutationQueue.js
```diff
@@ -133,6 +133,18 @@ class RelayMutationQueue {
...
     const transaction = this._get(id);
+    const collisionKey = transaction.getCollisionKey();
+    const collisionQueue =
+        collisionKey && this._collisionQueueMap[collisionKey];
+    if (collisionQueue) {
+      const index = collisionQueue.indexOf(transaction);
+      if (index !== -1) {
+        collisionQueue.splice(index, 1);
+      }
+      if (collisionKey && collisionQueue.length === 0) {
```
jardakotesovec: Checking for collisionKey here is redundant, but flow was concern that it could be null/undefined, which is not true..

jardakotesovec:
Can I ask for feedback here?

src/mutation/RelayMutationQueue.js
josephsavona: Flow isn't smart enough to realize that `collisionQueue` being null implies that `collisionKey` is also truthy. The length check isn't required, though, so how about just adding this in the outer `if`?

src/mutation/RelayMutationQueue.js
```diff
@@ -133,6 +133,18 @@ class RelayMutationQueue {
 
   rollback(id: ClientMutationID): void {
     const transaction = this._get(id);
+    const collisionKey = transaction.getCollisionKey();
+    const collisionQueue =
+        collisionKey && this._collisionQueueMap[collisionKey];
+    if (collisionQueue) {
```
josephsavona: `if (collisionKey && collisionQueue) {`

src/mutation/__tests__/RelayMutationQueue-test.js
```diff
@@ -509,4 +509,67 @@ describe('RelayMutationQueue', () => {
...
+      request.resolve({response: {'res': 'ponse'}});
+      jest.runAllTimers();
+
+      expect(RelayNetworkLayer.sendMutation.mock.calls.length).toBe(2);
+
+      expect(transaction3.getStatus()).toBe(
+          RelayMutationTransactionStatus.COMMITTING
+      );
+    });
+  });
```
josephsavona: :+1:

src/mutation/RelayMutationQueue.js
jardakotesovec: `collisionQueue.length === 0` ensures that clean up only happens if this was the only transaction in that `collisionQueue`, there can be more.

yungsters:
@jardakotesovec Thanks for your contribution. I definitely want to see this fix land, but I have some feedback.

src/mutation/RelayMutationQueue.js
yungsters: Per @josephsavona, doing the check here makes it so you don't have to do it on line 144. Better yet, you can avoid the redundancy by doing:

```
if (collisionKey) {
  const collisionQueue = this._collisionQueueMap[collisionKey];
  if (collisionQueue) {
    // Flow now knows that both `collisionKey` and `collisionQueue` are truthy.
  }
}
```

src/mutation/RelayMutationQueue.js
```diff
@@ -133,6 +133,18 @@ class RelayMutationQueue {
...
+      const index = collisionQueue.indexOf(transaction);
+      if (index !== -1) {
+        collisionQueue.splice(index, 1);
+      }
+      if (collisionKey && collisionQueue.length === 0) {
+        delete this._collisionQueueMap[collisionKey];
+      }
+    }
```
yungsters: To reduce the number of code paths that rollback follows (which just makes the code slightly easier to reason about), what do you think about moving all of this logic into `_clearPendingTransaction`?

src/mutation/__tests__/RelayMutationQueue-test.js
```diff
@@ -509,4 +509,67 @@ describe('RelayMutationQueue', () => {
...
+      mockMutation1.getCollisionKey.mockReturnValue('key');
+
+    });
+
+    it('rollback queued transaction', () => {
+      var transaction1 = mutationQueue.createTransaction(mockMutation1);
+      transaction1.commit();
+
+      expect(transaction1.getStatus()).toBe(
+          RelayMutationTransactionStatus.COMMITTING
```
yungsters: Nit: De-dent by 2 spaces. Same with the rest of this test case.

src/mutation/RelayMutationQueue.js
jardakotesovec: @yungsters  Thanks a lot for feedback.
`collisionQueue` is managed in `commit`, `_advanceCollisionQueue`, `_failCollisionQueue`.  So If I move it to `_clearPendingTransaction` than both `_failCollisionQueue` and `_clearPendingTransaction` will try to remove it for certain code path (its necessary to slice() queue in `_failCollisionQueue` to iterate it correctly) from the `collisionQueue`.

And its not possible just move this responsibility only to `_clearPendingTransaction` since its necessary to clean `collisionQueue` also for situations when its commited again (not rollbacked).

But if you still prefer it I will move it there and adjust `_failCollisionQueue` to deal with that by using `slice` and iterating on copy:
```
  _failCollisionQueue(transaction: PendingTransaction): void {
    const collisionKey = transaction.getCollisionKey();
    if (collisionKey) {
      const collisionQueue = nullthrows(this._collisionQueueMap[collisionKey]);
      // Remove the transaction that called this function.
      collisionQueue.shift();
      collisionQueue.slice().forEach(
        transaction => this._handleCommitFailure(transaction, null)
      );
      delete this._collisionQueueMap[collisionKey];
    }
  }

```

yungsters: @jardakotesovec Ahh... makes sense. You clearly understand this code better than I do, haha. I'll test this internally and land it. Thanks again!

yungsters:
@facebook-github-bot import

jardakotesovec:
@yungsters Thanks a lot. Do you have some thoughts on getStatus issue I mentioned in my initial comment?  Would be handy if getStatus would return something when its processed instead of exception.

yungsters:
> Do you have some thoughts on getStatus issue I mentioned in my initial comment? Would be handy if getStatus would return something when its processed instead of exception.

Indeed, but the difficultly is in differentiating between completed and invalid IDs. One solution would be to keep around the map of IDs that have been successfully committed, but that will continue to leak memory into the future.

Another solution would be to change `RelayMutationQueue` to expose methods on some instance that callers can obtain a reference for as long as needed.

@josephsavona Do you remember the reason for why we chose to use IDs instead of exposing a public form of the pending transaction (which would have all the methods like `getError`, `getStatus`, `commit`, etc.)?

PR #650, opened 2015-12-04
Lock to a version of GraphQL before the new restrictions on the size of a GraphQLInt

steveluscher:
New limits were imposed on the size of a `GraphQLInt`. While we figure out what to do about this, this pull request will ensure that the examples keep working.

### DIFF
--- a/examples/relay-treasurehunt/package.json
+++ b/examples/relay-treasurehunt/package.json
@@ -13,7 +13,7 @@
     "classnames": "^2.1.3",
     "express": "^4.13.1",
     "express-graphql": "^0.4.0",
-    "graphql": "^0.4.7",
+    "graphql": "0.4.13",
     "graphql-relay": "^0.3.3",
     "react": "^0.14.0",
     "react-dom": "^0.14.0",
--- a/examples/star-wars/package.json
+++ b/examples/star-wars/package.json
@@ -13,7 +13,7 @@
     "classnames": "^2.1.3",
     "express": "^4.13.1",
     "express-graphql": "^0.4.0",
-    "graphql": "^0.4.7",
+    "graphql": "0.4.13",
     "graphql-relay": "^0.3.3",
     "react": "^0.14.0",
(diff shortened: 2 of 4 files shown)


steveluscher:
Change introduced here: https://github.com/graphql/graphql-js/commit/06f97b67491f0215df7536aac50361bd90d5097d#diff-87c5e74dd1f7d923143e0eee611f598eR19

taion:
Are you thinking of modifying how the connection arguments are modelled, or just trying to sort out how to update the examples?

steveluscher:
We're trying to decide whether to change the connection argument types from `GraphQLInt` to a custom `SuperReallyReallyBigInt` type, or whether to just modify the examples to fetch `(first: ${Math.pow(2, 32) -1)})`. The former is what it is, and the latter will break anyone who's been employing the `(first: ${Math.MAX_SAFE_INTEGER})` pattern.

taion:
That makes sense - is there no way to make this be implementation-specific though? Java has gotten away with 32-bit array indices for over a decade now - to the extent that there is a formal `GraphQLInt` type, it's not a bad default for this sort of thing, rather than forcing everyone to use a non-native int type. `int64` is un-natural in JS, but `int53` is un-natural _everywhere_.

I've actually switched all of my code to the above (but as `-1 >>> 1` because I hate readable code), and it wasn't awful.

steveluscher:
It kind of already is implementation specific. `graphql-relay-js` happens to use `GraphQLInt` wholesale for connection arguments ([link](https://github.com/graphql/graphql-relay-js/blob/master/src/connection/connection.js#L25-L49)) but another implementation (Ruby? Python?) could make a different choice. Also, when crafting your schema, you could choose not to use the `graphql-relay-js` helpers, and define connections in your own way.

> because I hate readable code

LOL.

taion:
Oh, right - I see. Relay itself doesn't care, so it's just a `graphql-relay-js` question? That makes sense. Thanks!

PR #661, opened 2015-12-07
Lint

plievone:
Fixes most of the `npm run lint` issues. There are three commits here: the first codemods the dangling comma, the second one fixes a few other eslint issues, and the third one fixes a few unnecessary warnings during tests. Take in full or in part, with or without attribution (as there are quite many lines changed!).

### DIFF
--- a/src/store/RelayPendingQueryTracker.js
+++ b/src/store/RelayPendingQueryTracker.js
@@ -119,7 +119,7 @@ class PendingFetch {
 
   constructor(
     {fetchMode, forceIndex, query}: PendingQueryParameters,
-    {pendingFetchMap, preloadQueryMap, storeData}: {
+    {pendingFetchMap, preloadQueryMap, storeData, }: { // babel-eslint issue
       pendingFetchMap: {[queryID: string]: PendingState};
       preloadQueryMap: PromiseMap<Object, Error>;
       storeData: RelayStoreData;
--- a/examples/relay-treasurehunt/data/schema.js
+++ b/examples/relay-treasurehunt/data/schema.js
@@ -155,5 +155,5 @@ var mutationType = new GraphQLObjectType({
  */
 export var schema = new GraphQLSchema({
   query: queryType,
-  mutation: mutationType
+  mutation: mutationType,
 });
--- a/examples/relay-treasurehunt/server.js
+++ b/examples/relay-treasurehunt/server.js
@@ -38,15 +38,15 @@ var compiler = webpack({
         query: {stage: 0, plugins: ['./build/babelRelayPlugin']},
         test: /\.js$/,
       },
-    ]
+    ],
   },
-  output: {filename: 'app.js', path: '/'}
+  output: {filename: 'app.js', path: '/'},
 });
 var app = new WebpackDevServer(compiler, {
   contentBase: '/public/',
   proxy: {'/graphql': `http://localhost:${GRAPHQL_PORT}`},
   publicPath: '/js/',
-  stats: {colors: true}
+  stats: {colors: true},
 });
 // Serve static resources
 app.use('/', express.static(path.resolve(__dirname, 'public')));
--- a/examples/star-wars/data/database.js
+++ b/examples/star-wars/data/database.js
@@ -63,19 +63,19 @@ var executor = {
 var rebels = {
   id: '1',
   name: 'Alliance to Restore the Republic',
-  ships: ['1', '2', '3', '4', '5']
+  ships: ['1', '2', '3', '4', '5'],
 };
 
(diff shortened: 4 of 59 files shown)


src/store/RelayPendingQueryTracker.js
```diff
@@ -118,12 +118,14 @@ class PendingFetch {
   _errors: Array<?Error>;
 
   constructor(
+    /* eslint-disable comma-dangle */
```
plievone: eslint insists on adding a dangling comma after first `storeData`, so the rule is disabled here.

steveluscher: It complains about this line?

```
     {pendingFetchMap, preloadQueryMap, storeData}: {
```

Do we need to add [a plugin to strip Flow types](https://www.npmjs.com/package/eslint-plugin-flowtype) before we lint? Know why this might be happening, @zpao?

plievone: Or perhaps we'll just add the comma there, it's a single comma mishap in a large codebase. Stripping flowtypes might be suboptimal as currently eslint catches unused type imports.

josephsavona: is there an eslint setting for disabling last comma in single-line objects?

plievone: Yes, and these results should be with 'always-multiline' enabled. So there is something in babel-eslint that gets confused in this rare case.

plievone: ...if somebody wants to dig in this, the multiline case is inferred in https://github.com/eslint/eslint/blob/master/lib/rules/comma-dangle.js and flowtypes are dealt with in https://github.com/babel/babel-eslint/blob/master/index.js and elsewhere.

zpao: Don't strip flow types - babel-eslint parses them and we get the used/unused warnings. I think we just do this or leave it in there as a known failure (which makes lint a bit less useful though).

I would also update all the lint deps here to make sure you have latest eslint, babel-eslint (and then fix the version numbers - the ranges have been annoying in other projects as rules change).

Babel-eslint v5 (in beta) might fix this, but we should also file a test case against them so they can fix this.

steveluscher: To be clear, I'm not saying that what you've done here is wrong, I'm trying to dig into it to determine:

1. If this is solved in a newer version of babel-eslint (as @zpao suggested), or
2. If this is a bug that they might like to know about

Can you check if 1 or 2 are at play?

plievone: Ok! Made a reduced test case and filed an issue eslint/eslint/issues/4634
We could also let the comma be there and let it be found by linter when the bug is fixed.
Btw, used [this eslint plugin diff](https://github.com/eslint/eslint/pull/4419/files#diff-dd33e748952b0c362b1b9fbb73ae1d06R163) for codemodding dangling commas.

hzoo: At first I thought it was an babel-eslint issue but I think the `comma-dangle` rule may have to be rewritten (maybe through https://github.com/babel/eslint-plugin-babel). The `isMultiline` doesn't account for flow types.

screen shot 2015-12-08 at 12 08 21 pm

So it looks like a multline which causes the error (without the flow type it's fine since it starts/ends on the same line without it)

hzoo: So really the issue is that the `ObjectPattern` loc is also accounting for the typeAnnotation so the column start and column end are 2 and 3 so it thinks it is actually multiline. Not sure what we can do other than making a forked rule in https://github.com/babel/eslint-plugin-babel. Or we would have to change the range/loc of nodes in either babel/babel-eslint.

kassens:
Hi @plievone, sorry for letting this sit for a while. We've had some bigger updates to the codebase. I see you're change consists of a codemod diff and some manual changes. Would you mind redoing the automated part on master? This should reduce the work required to rebase the manual part.

kassens:
This is great, thanks!

@facebook-github-bot import

plievone:
There's also a bonus commit silencing most of the unnecessary warnings during tests, you can cherry pick what you need.

plievone:
> I would also update all the lint deps here to make sure you have latest eslint, babel-eslint (and then fix the version numbers - the ranges have been annoying in other projects as rules change).

@kassens One could also pin some devDependencies here to a point version as zpao suggested.

yungsters:
This is amazing.

kassens:
Landing all the changes from this PR internally now. It was easier to include the test fixes now.

PR #662, opened 2015-12-07
Upgrade flow and jest, and remove a few obsolete flow issues

plievone:
Tried out what happens if one upgrades flow and removes some type suppressions that may be obsolete now. I don't know if removing them is intended, but you can cherrypick the changes you care about.


scripts/babel-relay-plugin/src/RelayQLAST.js
```diff
@@ -383,7 +383,7 @@ class RelayQLArgument {
       case 'EnumValue':
         return value.value;
       case 'ListValue':
-        return value.values.map(
+        return (value: types.GraphQLList).values.map( // Flow
```
plievone: Don't know which is your preferred way to handle this kind of flow issues.

wincent:
This is fantastic, @plievone. I've wanted to get rid of those workarounds for the longest time!

Sorry that we let this one get stale; since you pushed this we already updated (to an even later version of Flow) and fixed some of the issues independently. I've gone through and picked out the changes that are still relevant, so I am going to mark this one as closed. Thanks once again for your contribution!

plievone:
Thanks!

PR #678, opened 2015-12-11
[Relay] don't use the reserved word "await" as a method name

### LINKED ISSUE #665: Error Running Relay in MS Edge
I'm getting an error while trying to run an app which has Relay in the Edge browser. The browser keeps throwing this error:

```
The use of a keyword for an identifier is invalid
```
I have tracked it down to the file: [RelayTaskScheduler.js](https://github.com/facebook/relay/blob/288d6a16b45ba098e742c664bfb2f9e225d0f39f/src/tools/RelayTaskScheduler.js#L94) in the **compiled** code (which is generated by Babel):

```
var RelayTaskScheduler {
  ...
  await: function await() {...}
}
```

I've verified that this is the problem by replacing the await function name with something non-reserved. `await` is a future reserved keyword: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar?redirectlocale=en-US&redirectslug=JavaScript%2FReference%2FReserved_Words#Future_reserved_keywords

kassens:
Apparently `await` is missing from some lists of reserved words. Typically Babel or minifiers should quote `await` like they quote `function` in `var x = {function: 1, await: 2}`.

Instead of waiting for changes in Babel or webpack to trickle down, let's rename this for now.

Fixes #665.

### DIFF
--- a/src/legacy/store/GraphQLQueryRunner.js
+++ b/src/legacy/store/GraphQLQueryRunner.js
@@ -255,7 +255,7 @@ function runQueries(
     );
   }
 
-  RelayTaskScheduler.await(() => {
+  RelayTaskScheduler.enqueue(() => {
     var forceIndex = fetchMode === DliteFetchModeConstants.FETCH_MODE_REFETCH ?
       generateForceIndex() : null;
 
--- a/src/store/RelayGarbageCollection.js
+++ b/src/store/RelayGarbageCollection.js
@@ -95,10 +95,10 @@ function scheduler(run: () => boolean): void {
     }
     // This is effectively a (possibly async) `while` loop
     if (hasNext) {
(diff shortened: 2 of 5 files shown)


kassens:
@facebook-github-bot import

kumarharsh:
Thanks for the PR. +1

gschammah:
is this fix going to be published to npm soon? thanks!

kassens:
@gschammah Yes, we're planning on publishing 0.7.0 on Wednesday.

hannesj:
@kassens any update on this?

rosskevin:
This is currently an issue with babel-runtime falling behind on versions, everything else is `6.25.0` while `babel-runtime` is `6.23.0`.

https://github.com/babel/babel/issues/5873

PR #680, opened 2015-12-14
fix non-stardard arrow function call syntax on updateSchema

sibeliusseraphini:
based on this [discussion](https://phabricator.babeljs.io/T2027), this:

```
() => {}()
```

is a a non-standard arrow function call syntax allowed on babel 5, but it doesn't work on babel 6


kassens:
Thanks!

@facebook-github-bot import

PR #683, opened 2015-12-15
Add Relay context to Relay containers

denvned:
This PR is meant to finish the work started by @devknoll on contextualizing the Relay store. The previous his attempt on that was #603, but he didn't have enough time to finish that soon, so he gave me a permission to continue his work.

Escape from the global Relay store is very important for server side rendering (https://github.com/denvned/isomorphic-relay/issues/6). It also solves the problem of resetting Relay store (#233).

I tried to make this PR as simple as possible to make the initial review easier. That's why there are few things not included yet:
- Unit tests for change of `relayContext` property.
- Updated docs.
- Updated examples.
- Fixed references to changed API in the code comments.

I removed the global Relay store singleton altogether to make sure it is not used anywhere. But it might  make sense to add placeholders for replaced API with deprecation messages, for backward compatibility (like [here](https://github.com/devknoll/relay/blob/0d545bb0b833d1631f55e4bc2fb8919ec898c005/src/store/RelayStore.js)).


josephsavona:
@denvned My apologies for the delay in reviewing this. I'll write up more detailed thoughts in the morning, but I just wanted to thank you for your contribution :-)

src/mutation/RelayMutation.js
```diff
@@ -230,10 +229,16 @@ class RelayMutation<Tp: Object> {
     return null;
   }
 
-  _resolveProps(props: Tp): void {
-    const fragments = this.constructor.fragments;
-    const initialVariables = this.constructor.initialVariables || {};
+  /**
+   * @internal
+   */
+  static resolveProps(
```
josephsavona: What necessitates this change? Mutations will never need a `RelayContext` instance - the changes here seem focused on passing in the instance of `RelayStoreData` instead of using the shared one. 

Let's split the changes to `RelayMutation` and `RelayMutationQueue` into a separate PR.

src/container/RelayRootContainer.js
```diff
@@ -27,6 +28,7 @@ type RootContainerProps = {
   Component: RelayContainer;
   forceFetch?: ?boolean;
   onReadyStateChange?: ?(readyState: ReadyState) => void;
+  relayContext: RelayContext;
```
josephsavona: We definitely want `RelayRenderer` to accept a `RelayContext` as a prop. However, we also need to make the transition easy for developers. Let's do this:
- Keep the addition of `relayContext` to RelayRenderer.
- Add `RelayContext.getDefaultInstance()` (static) and have `RelayRootContainer` pass that default instance as the context to RelayRenderer.

src/store/RelayContext.js
```diff
@@ -6,7 +6,7 @@
  * LICENSE file in the root directory of this source tree. An additional grant
```
josephsavona: Let's keep `RelayStore.js` as a module; it should export a default instance of `RelayContext`. We don't need a test for this, only a mock

src/store/RelayContext.js
```diff
@@ -76,7 +75,25 @@ var queuedStore = storeData.getQueuedStore();
...
  *
  */
-var RelayStore = {
+class RelayContext {
+  _storeData: RelayStoreData;
+
+  constructor() {
+    this._storeData = new RelayStoreData();
+
+    this.injectBatchingStrategy(ReactDOM.unstable_batchedUpdates);
```
josephsavona: let's move this out of the module definition and set it in `RelayPublic.js`

src/mutation/RelayMutationQueue.js
```diff
@@ -350,14 +355,15 @@ class PendingTransaction {
...
+      this._collisionKey =
+        this.mutation.getCollisionKey(this.getProps()) || null;
     }
     return this._collisionKey;
   }
 
   getConfigs(): Array<RelayMutationConfig> {
     if (!this._configs) {
-      this._configs = this.mutation.getConfigs();
+      this._configs = this.mutation.getConfigs(this.getProps());
```
josephsavona: Why the change to pass props? It doesn't seem like `getConfigs` method signature has changed. Again, let's split the `RelayMutation*` changes into a (single) separate PR.

src/store/RelayGarbageCollection.js
```diff
@@ -1,104 +0,0 @@
-/**
```
josephsavona: We're using this module, and outright deleting it will make importing this PR more difficult. Let's keep the module in place, but change it to call the corresponding functions on `RelayStore` (which should be an instance of `RelayContext`).

```
const RelayGarbageCollection = {
  initialize(stepLength: number) {
    RelayStore.initialize(stepLength);
  }
  ...
}
```

josephsavona:
At a high level, this is definitely the direction we want to go and is a good step towards implementing #558. However, it will be difficult to import and land this PR in the current form due to the sheer volume of code changes, the impact on public APIs, and the risk inherent in significantly changing core functionality. The only practical way to land this will be to split it up: first prepare the big change without affecting public APIs, then add one final PR to make the cutover. This will a) make it much easier to review, b) make the dependencies between changes explicit and c) isolate blame in the case something does go wrong. Kent Beck summed up this approach more succinctly than I can: "for each desired change, make the change easy (warning: this may be hard), then make the easy change" - https://twitter.com/KentBeck/status/250733358307500032.

Here's how that might look:
- Split off any changes that can be made safely today. An example is the changes to `RelayMutation` and `RelayMutationQueue` - these aren't dependent on the new `RelayContext` work and are a perfect example of something that can be safely changed now.
- Prepare the big change. We could create `RelayContext` first and make `RelayStore` export a default instance of it - without having any other callers of `RelayContext`. This will allow us to test and verify that module on its own first. The same PR could update `RelayGarbageCollection` to delegate to the context, since they're related. Note that this PR should/will completely preserve the existing public API.
- Again, change `RelayContainer` to access Relay data via `context.relay`, and change `RelayRenderer` to pass `RelayStore` (the default context instance) as `context.relay`. This will mean that we're now *using* contextual data, without impacting public API.
- Finally, change RelayRender to accept the context instance as a prop, and change RelayRootContainer to pass the default instance instead. At this point, we can a) document RelayRenderer, b) mark RelayRootContainer as deprecated and c) document and make RelayContext part of RelayPublic.

denvned:
@josephsavona Thanks for the great review! I have started working on splitting this. Hope to submit the "prepare the big change" PR tomorrow.

src/mutation/RelayMutationQueue.js
denvned: In the description to #699 I tried to explain why I am passing props to `RelayMutation` methods. But if you know a better solution I will be happy to change the PR.

josephsavona: Gotcha, the description there makes sense

denvned:
@josephsavona Should I submit the PR making `RelayContainer` access Relay data via `context.relay` now, or is it better to wait until the `RelayContext` PR (#698) is merged?

josephsavona:
Probably better to wait until the other PR is merged so you can base your work on that.

fabiosantoscode:
Excellent work!

though, I'm surprised this does not touch the query caches (https://github.com/facebook/relay/search?utf8=%E2%9C%93&q=querycache)

Their `.size` seems to grow unbounded with each new query. Under a minigun-powered load test, I've performed 600 requests to the same page, and have observed that getRelayQueries.js's `queryCache` grows to 1171 in size, and that buildRQL.js's `queryCache` grows to 2342 in size.

Is this out of scope for this pull request?

josephsavona:
> Is this out of scope for this pull request?

@fabiosantoscode Yup, those query caches are unrelated to this PR. If you're seeing a behavior that seems incorrect or that might be a bug, definitely file a separate PR so we can confirm. Thanks!

fabiosantoscode:
I'll file an issue, but I only to confirmed this using the isomorphic-relay package (although the code that was leaking was relay, it had some things instrumented from the outside).

josephsavona:
The core of this was split off and landed in #761, so let's close this and continue discussion via follow-ups in #699 and #704.

PR #684, opened 2015-12-15
DefaultNetworkLayer returns more valuable promise

jardakotesovec:
`sendMutation` / `sendQueries` returns promise, but always fulfilled with undefined.  This PR suggest to return `request` promise which is what most people would expect I think. 

For me its valuable that I can do some extra logging/error detecting stuff by extending DefaultNetworkLayer class.

Let me know if that change makes sense and if you want tests for it.

### DIFF
--- a/src/network-layer/default/RelayDefaultNetworkLayer.js
+++ b/src/network-layer/default/RelayDefaultNetworkLayer.js
@@ -60,9 +60,11 @@ class RelayDefaultNetworkLayer {
       } else {
         request.resolve({response: payload.data});
       }
-    }).catch(
-      error => request.reject(error)
-    );
+      return request;
+    }).catch(error => {
+      request.reject(error);
+      return request;
+    });
   }
 
   sendQueries(requests: Array<RelayQueryRequest>): Promise {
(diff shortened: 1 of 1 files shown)


src/network-layer/default/RelayDefaultNetworkLayer.js
```diff
@@ -60,9 +60,11 @@ class RelayDefaultNetworkLayer {
...
         request.resolve({response: payload.data});
       }
-    }).catch(
-      error => request.reject(error)
-    );
+      return request;
+    }).catch(error => {
+      request.reject(error);
+      return request;
+    });
```
josephsavona: Hmm. It's already possible for users to wait for the results of `request` directly:

```javascript
class MutationLoggingNetworkLayer extends DefaultNetworkLayer {
  sendMutation(mutation) {
    // log the response or error (note that `mutation` is a promise)
    mutation.then(
      error => console.error(error),
    );
    // Send the mutation using the default network implementation
    return super.sendMutation(mutation);
  }
```

Have you tried this approach?

jardakotesovec: Right.. this would work..  so just [example](https://facebook.github.io/relay/docs/api-reference-relay.html#injectnetworklayer-static-method-example) misled me to using promise returned from `DefaultNetworkLayer.sendMutation`.

josephsavona: What do you think about closing this PR and moving forward with the one to update the example?

PR #685, opened 2015-12-15
Override sendMutation via superclass

jardakotesovec:
Creating super class seems to me as cleaner approach, so options can be easily passed to the newly created network layer when creating instance.

To make this example work (including old version), its necessary to return better promise (#684).

### DIFF
--- a/docs/APIReference-Relay.md
+++ b/docs/APIReference-Relay.md
@@ -150,23 +150,19 @@ As an example, we can log each mutation that is sent to the server as follows:
 ```
 var DefaultNetworkLayer = Relay.DefaultNetworkLayer;
 
-var MutationLoggingNetworkLayer = {
-  // Use the defaults for other options.
-  ...DefaultNetworkLayer,
-
+class MutationLoggingNetworkLayer extends DefaultNetworkLayer {
   sendMutation(mutation) {
-    // Log the query text exactly as it will be sent to the server.
-    console.log(mutation.getQueryString());
-    // Send the mutation using the default network implementation, but
-    // log the response or error.
-    return DefaultNetworkLayer.sendMutation(mutation).then(
+    // log the response or error (note that `mutation` is a promise)
+    mutation.then(
       response => console.log(response),
       error => console.error(error),
     );
+    // Send the mutation using the default network implementation
+    return super.sendMutation(mutation);
   }
 };
 
-Relay.injectNetworkLayer(MutationLoggingNetworkLayer);
+Relay.injectNetworkLayer(new MutationLoggingNetworkLayer());
 ```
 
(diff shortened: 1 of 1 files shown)


josephsavona:
Either composition or inheritance would work here, so this seems fine. See my [comment](https://github.com/facebook/relay/pull/684/files#r50766453) though about a simpler approach for waiting for the results of the mutation though.

jardakotesovec:
Thanks for feedback. If I use the construction from [docs](https://facebook.github.io/relay/docs/api-reference-relay.html#injectnetworklayer-static-method-example)
```
var MutationLoggingNetworkLayer = {
  // Use the defaults for other options.
  ...DefaultNetworkLayer,

  sendMutation(mutation) {
    // Log the query text exactly as it will be sent to the server.
    console.log(mutation.getQueryString());
    // Send the mutation using the default network implementation, but
    // log the response or error.
    return DefaultNetworkLayer.sendMutation(mutation).then(
      error => console.error(error),
    );
  }
};
```

It does not copy prototype or really anything from `DefaultNetworkLayer` and I end up with simple object having `sendMutation` prop.

I am using babel 6 with latest `babel-plugin-syntax-object-rest-spread` and `babel-plugin-transform-object-rest-spread`. 

So not sure how the composition is expected to work here.

docs/APIReference-Relay.md
```diff
@@ -150,23 +150,20 @@ As an example, we can log each mutation that is sent to the server as follows:
...
-  ...DefaultNetworkLayer,
-
+class MutationLoggingNetworkLayer extends DefaultNetworkLayer {
   sendMutation(mutation) {
     // Log the query text exactly as it will be sent to the server.
     console.log(mutation.getQueryString());
     // Send the mutation using the default network implementation, but
     // log the response or error.
-    return DefaultNetworkLayer.sendMutation(mutation).then(
+    return super.sendMutation(mutation).then(
```
josephsavona: How about updating this with approach from https://github.com/facebook/relay/pull/684/files#r50766453 ?

josephsavona:
Yup, a class-based approach is cool too, let's keep that and just apply the changes in my comment above.

For reference, what I meant by composition was something like:

```javascript
const network = new DefaultNetworkLayer(...);
const MyNetworkLayer = {
  ...network,
  sendMutation(mutation) {
    mutation.then(...);
    return network.sendMutation(mutation);
  },
}
Relay.injectNetworkLayer(MyNetworkLayer);
```

jardakotesovec:
@josephsavona Thanks for explanation! PR is updated with your code.

josephsavona:
@facebook-github-bot import

yungsters:
@josephsavona Hmm... looks like the bot failed.

PR #689, opened 2015-12-17
[relay] fix dependencies of examples

kassens:
`npm install` of the todo example failed for 2 reasons:

- The version of babel-relay-plugin was bumped to 6.0.1 without an update of the
  `peerDependency` or `relay` itself. I published 6.0.1 and bumped the peer
  dependency.
- `history@0.14` and `react-router` had incompatible dependencies. We should
  probably fix all dependency versions and bump them with a test instead of
  having them update whenever new versions are published.

### DIFF
--- a/examples/todo/package.json
+++ b/examples/todo/package.json
@@ -15,7 +15,7 @@
     "express-graphql": "^0.4.0",
     "graphql": "0.4.13",
     "graphql-relay": "^0.3.3",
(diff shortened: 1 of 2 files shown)


kassens:
@facebook-github-bot import

josephsavona:
@kassens should we have bumped `react-relay` to 0.6.1 for this?

PR #698, opened 2015-12-22
Add RelayContext

denvned:


josephsavona:
Awesome, quick turnaround! I'll review in the morning but at a quick glance this looks great.

src/store/RelayContext.js
```diff
@@ -0,0 +1,269 @@
...
+  StoreReaderData,
+  StoreReaderOptions,
+} from 'RelayTypes';
+
+import type {
+  DataID,
+  RelayQuerySet,
+} from 'RelayInternalTypes';
+
+let defaultInstance;
```
josephsavona: let's use `RelayStore` as the default instance, so we don't need a default instance or `getDefaultInstance` here

src/store/RelayContext.js
```diff
@@ -0,0 +1,269 @@
...
+ */
+class RelayContext {
+  _storeData: RelayStoreData;
+
+  static getDefaultInstance(): RelayContext {
+    if (!defaultInstance) {
+      defaultInstance = new RelayContext(RelayStoreData.getDefaultInstance());
+    }
+    return defaultInstance;
+  }
```
josephsavona: see above

src/store/RelayContext.js
```diff
@@ -0,0 +1,269 @@
...
+  constructor(storeData: RelayStoreData = new RelayStoreData()) {
+    this._storeData = storeData;
+  }
+
+  /**
+   * @internal
+   */
+  getStoreData(): RelayStoreData {
+    return this._storeData;
+  }
```
josephsavona: Remove this until we need it for something?

src/store/RelayContext.js
```diff
@@ -0,0 +1,269 @@
...
+
+  /**
+   * Initializes garbage collection: must be called before any records are
+   * fetched. When records are collected after calls to `scheduleCollection` or
+   * `scheduleCollectionFromNode`, records are collected in steps, with a
+   * maximum of `stepLength` records traversed in a step. Steps are scheduled
+   * via `RelayTaskScheduler`.
+   */
+  initializeGarbageCollection(stepLength: number): void {
```
josephsavona: Hmm. Let's keep the `scheduler` implementation in `RelayGarbageCollection` and make this a simple proxy method through to the initialize method on `RelayStoreData`. The signature here can then be `initializeGarbageCollection(scheduler)`.

src/store/RelayContext.js
```diff
@@ -0,0 +1,269 @@
...
+          hasNext = run();
+          iterations++;
+        }
+        // This is effectively a (possibly async) `while` loop
+        if (hasNext) {
+          RelayTaskScheduler.enqueue(runIteration);
+        }
+      };
+      RelayTaskScheduler.enqueue(runIteration);
+    }
```
josephsavona: Thanks for considering this, but let's move the scheduler back to `RelayGarbageCollection`

src/store/RelayContext.js
```diff
@@ -0,0 +1,269 @@
...
+   * NOTE: If the given record is still referenced, no records are collected.
+   */
+  scheduleGarbageCollectionFromNode(dataID: DataID): void {
+    const garbageCollector = this._storeData.getGarbageCollector();
+
+    if (garbageCollector) {
+      garbageCollector.collectFromNode(dataID);
+    }
+  }
+}
```
josephsavona: :+1:

src/store/RelayContext.js
```diff
@@ -0,0 +1,269 @@
...
+  _storeData: RelayStoreData;
+
+  static getDefaultInstance(): RelayContext {
+    if (!defaultInstance) {
+      defaultInstance = new RelayContext(RelayStoreData.getDefaultInstance());
+    }
+    return defaultInstance;
+  }
+
+  constructor(storeData: RelayStoreData = new RelayStoreData()) {
```
josephsavona: This type of optional with a default API seem easy to use but are too unpredictable - it's easy to forget to pass an instance somewhere and wonder why tests are failing. Let's require the instance to passed in for now (which you're already doing in tests).

src/store/RelayGarbageCollection.js
```diff
@@ -37,28 +31,14 @@ var RelayGarbageCollection = {
...
-      stepLength > 0,
-      'RelayGarbageCollection: step length must be greater than zero, got ' +
-      '`%s`.',
-      stepLength
-    );
-    _stepLength = stepLength;
-    RelayStoreData
-      .getDefaultInstance()
-      .initializeGarbageCollector(scheduler);
+    RelayContext.getDefaultInstance().initializeGarbageCollection(stepLength);
```
josephsavona: `RelayStore.initializeGarbageCollection(...)`

josephsavona:
Sweet, this looks great. Main theme of my feedback is that we should use `RelayStore` as the "default" RelayContext instance to aid in the transition: all existing Relay codebases are already targeting that module so they'll continue to work.

denvned:
@josephsavona Committed your suggestions. Did I understand everything right?

josephsavona:
nice job.

josephsavona:
@facebook-github-bot import

PR #699, opened 2015-12-22
Contextualize Relay mutations

denvned:

Currently, Relay resolve mutations props using the global `RelayStore`, and store them in the `props` instance field of `RelayMutation`. That makes it impossible to contextualize Relay mutations without changing the public API.

<s>After analyzing different alternatives, I have found that the probably cleanest solution is not to store resolved mutation props in an instance field, but to pass them as an argument to each of the mutation methods. That way mutations can be reused in different `RelayStoreData` contexts, and also they can be reused in the same context but at different times, because props are re-resolved at each mutation execution.</s>


src/mutation/RelayMutation.js
```diff
@@ -230,10 +229,16 @@ class RelayMutation<Tp: Object> {
...
   }
 
-  _resolveProps(props: Tp): void {
-    const fragments = this.constructor.fragments;
-    const initialVariables = this.constructor.initialVariables || {};
+  /**
+   * @internal
+   */
+  static resolveProps(
+    storeData: RelayStoreData, mutation: RelayMutation<Tp>
```
josephsavona: one argument per line

josephsavona:
Thanks for splitting this up. It makes it clear, however, that this is a breaking change for all existing mutations. The `RelayMutation` API indicates that mutations can access `this.props` - see the example at http://facebook.github.io/relay/docs/api-reference-relay-mutation.html#fragments-static-property-example - and this PR removes that property. 

We're going to need to think this through and find a way to not change the public API of mutations, or come up with a transition plan. 

cc @yungsters

yungsters:
> Currently, Relay resolve mutations props using the global `RelayStore`, and store them in the `props` instance field of `RelayMutation`. That makes it impossible to contextualize Relay mutations without changing the public API.

Can you explain why this is the case? Why can't we simply change *when* we resolve the props (i.e. when enqueued) but continue to assign it onto the instance as `this.props`?

denvned:
@yungsters Yes, I thought about that, but came to a conclusion that it might *silently* break existing mutations that might expect that props are resolved in the constructor and never changed after that. So, probably, explicitly changing the API here is the lesser of two evils, because it will break existing mutations *loudly*. :smile: By passing props as an argument to the methods we make it more clear that the props are not permanent, and that the mutation instance must not keep any state derived from the props.

But maybe I am exaggerating the problem too much, and we actually can do it. You decide.

denvned:
> it might silently break existing mutations that might expect that props are resolved in the constructor and never changed after that.

On second though, we probably can mitigate the problem greatly by prohibiting reuse of the same mutations in different contexts (e.g. throw an error).

josephsavona:
> we probably can mitigate the problem greatly by prohibiting reuse of the same mutation instances in different Relay contexts (e.g. throw an error)

Yes, let's do this. A given instance of a mutation should not be applied to multiple Relay contexts - that isn't possible today, so adding the limitation will have zero impact on existing products. This is a much cleaner solution - thanks again for working on this!

denvned:
@josephsavona Done. This is much cleaner indeed.

src/mutation/RelayMutation.js
```diff
@@ -35,6 +35,8 @@ export type RelayMutationFragments<Tk> = {
   [key: Tk]: RelayQLFragmentBuilder;
 };
 
+type RelayContext = typeof RelayStore;
```
denvned: I will remove it when `RelayContext` is merged.

josephsavona:
Sweet, thanks again. Will review and import today.

src/mutation/RelayMutation.js
josephsavona: nice workaround

src/store/__tests__/RelayStore-test.js
```diff
@@ -168,6 +168,15 @@ describe('RelayStore', () => {
...
     describe('applyUpdate', () => {
+      it('sets context in mutation before creating transaction', () => {
+        mockMutation.bindContext.mockImplementation(() => {
+          expect(createTransactionMock).not.toBeCalled();
+        });
+        RelayStore.applyUpdate(mockMutation);
+        expect(mockMutation.bindContext).toBeCalled();
+        expect(mockMutation.bindContext.mock.calls[0][0]).toBe(RelayStore);
+      });
+
```
josephsavona: add a similar test for `update`?

josephsavona:
@facebook-github-bot import

src/store/__tests__/RelayStore-test.js
denvned: done

josephsavona: Thanks!

src/mutation/__tests__/RelayMutation-test.js
```diff
@@ -47,37 +49,76 @@ describe('RelayMutation', function() {
...
-    ));
-    expect(Relay.Store.read.mock.calls[0]).toEqual(
-      [mockBarRequiredFragment, 'bar']
+    jasmine.addMatchers(RelayTestUtils.matchers);
+  });
+
+  it('throws if used in different Relay contexts', () => {
+    mockMutation.bindContext(relayContext);
+    expect(() => {
+      mockMutation.bindContext({});
```
josephsavona: `mockMutation.bindContext(new RelayContext());`

src/mutation/RelayMutation.js
```diff
@@ -50,12 +50,29 @@ class RelayMutation<Tp: Object> {
...
+    this._unresolvedProps = props;
+  }
+
+  /**
+   * @internal
+   */
+  bindContext(context: RelayContext): void {
+    if (!this._context) {
+      this._context = context;
+      this._resolveProps();
```
josephsavona: Existing products may have logic in mutation constructors that depends on `this.props` being set. With this new change, there isn't a clear place to put that logic. I wonder if we should add something like an optional "did receive props" hook that would provide a clear way to upgrade mutations. We could probably even write a codemod that would move existing constructor logic after the `super` call into the callback method.

cc @wincent

denvned: I think it is a very good idea. I have added `didResolveProps` (is the name ok?) hook in a separate commit.

josephsavona:
@denvned nice. i'll import this week.

josephsavona:
@facebook-github-bot import

src/mutation/RelayMutation.js
denvned: > We could probably even write a codemod that would move existing constructor logic after the `super` call into the callback method.

@josephsavona, I've never written any codemod before, but I could try if it would help to merge this PR sooner. Should I do it?

josephsavona: If you have time that would be great, but we can move forward with the PR without it (most mutations don't rely on the removed behavior).

denvned: Here you go: https://github.com/denvned/codemod/blob/master/transforms/RelayMutation-didResolveProps.js

I tried to make it as safe as possible, consequently it makes transformation only if `super(props)` is the first statement in the constructor (I hope it is the most common case, please tell if it is not). There are also other safety checks. And it reports the reason for every skipped case.

The first codemod I ever wrote, writing it was exciting :neckbeard:

denvned: Forgot about one more limitation, the codemod detects a mutation only if it subclass `RelayMutation` directly.

src/mutation/RelayMutation.js
```diff
@@ -317,7 +342,7 @@ class RelayMutation<Tp: Object> {
         );
         var dataID = RelayFragmentPointer.getDataID(propValue, fragment);
         if (dataID) {
-          resolvedProps[fragmentName] = RelayStore.read(fragment, dataID);
+          resolvedProps[fragmentName] = this._context.read(fragment, dataID);
```
denvned: @josephsavona, Should I replace `RelayContext` type here with the new `RelayContextInterface`, and add `read` and `readAll` to its definition?

Issue #700, opened 2015-12-22
Docs need to mention `null` usage for initialVariables


RavenHursT:
The docs for setting `initialVariables` for the Container API do not mention how to initialize unknown values for variables (i.e. The initial `cursor` value for pagination).

The correct way to initialize variables w/ an unknown value is to use `null`.  It took over half a day of searching and asking before this answer was given to me:

http://stackoverflow.com/questions/34406667/relay-pagination-how-to-initialize-after-value/34407181#34407181

This should _definitely_ be in the documentation for this feature.

devknoll:
@RavenHursT This is technically true, but I'm curious: why are you managing the `after` cursor yourself? Relay will automatically add the argument and its value when you change the `first` variable.

RavenHursT:
@devknoll : Server-side rendering.. if I go to `www.mysite.com/widgets?pageNumber=5` then we need a way to initialize what will be rendered when using https://github.com/denvned/isomorphic-relay

PR #701, opened 2015-12-22
Documentation for null initialization of initialVariables

RavenHursT:
Added information in Container API docs for null usage in initialVariables. Fixes https://github.com/facebook/relay/issues/700


yungsters:
Thanks for documenting these. It's not great that these are necessary, but it's better to have the documented than not.

yungsters:
@facebook-github-bot shipit

RavenHursT:
Thanks @yungsters

PR #703, opened 2015-12-24
Remove fragmentPointer field from GraphQLStoreQueryResolver

andimarek:
The fragmentPointer field was not used except in the resolve function. But resolve gets a fragmentPointer anyway, so just use that. 

This makes it also more consistent and less confusing imho.

### DIFF
--- a/src/container/RelayContainer.js
+++ b/src/container/RelayContainer.js
@@ -471,7 +471,6 @@ function createContainerComponent(
         } else if (!queryResolver) {
           queryResolver = new GraphQLStoreQueryResolver(
             storeData,
-            fragmentPointer,
             this._handleFragmentDataUpdate.bind(this)
           );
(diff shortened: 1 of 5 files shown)


josephsavona:
:+1:

josephsavona:
@andimarek Thanks for submitting this. I recently updated `GraphQLStoreQueryResolver` to add better error messages, which uses the fragment passed to the constructor. We might reconsider this at a future date, but for now let's close this. Thanks again for contributing, please continue!

andimarek:
@josephsavona I will ... one day a PR will go through :-)

josephsavona:
@andimarek Proposing ideas that *don't* pan out can be just as helpful as proposing ones that do. I've closed tons of my own PRs and diffs - we really appreciate your contributions!

PR #704, opened 2015-12-25
Contextualize RelayNetworkLayer

denvned:
This PR is another step toward making all Relay state contextual (#558).

Contextual `RelayNetworkLayer` is important for server-side rendering. It will allow to inject a contextual network layer initialized with client request specific params (e.g. authentication cookies) on the server. Currently it is impossible because the injected network layer is global, and shared between client requests.


src/network/__mocks__/RelayNetworkLayer.js
```diff
@@ -9,4 +9,46 @@
...
+    requests.push({
+      resolve(response, error) {
+        resolve({error: error || null, response});
+      },
+      reject,
+    });
+  }
+}
+
+RelayNetworkLayerMock.prototype = RelayNetworkLayer.prototype;
```
josephsavona: Hmm, would be nice if there was a clearer way to override just the one function as necessary. @yungsters any ideas?

src/RelayPublic.js
```diff
@@ -47,7 +46,8 @@ var RelayPublic = {
   createContainer: RelayContainer.create,
   createQuery: createRelayQuery,
   getQueries: getRelayQueries,
-  injectNetworkLayer: RelayNetworkLayer.injectNetworkLayer,
+  injectNetworkLayer: RelayInternals.NetworkLayer.injectNetworkLayer
```
josephsavona: `RelayInternals` is meant for exporting some functions for dev tools - to reference a module, import it normally. In this case something alone the lines of `RelayStoreData.getDefaultInstance().getNetworkLayer().bind(...)`

src/network/__mocks__/RelayNetworkLayer.js
denvned: Looks like only `RelayPendingQueryTracker-test` needs the mock for `fetchRelayQuery`, and only `RelayMutationQueue-test` needs the mock for `sendMutation`. Should I mock these methods right in these tests?

salbito:
Any updates on this front?

josephsavona:
@denvned @salbito I'll be working on the react-relay / Relay core split over the next couple of months, and i'll be working to get the remaining "Contextualize ..." PRs landed soon.

denvned:
@josephsavona Really nice to hear! The global network layer is the last thing that blocks many projects from using Relay isomorphically.

I've just resolved the conflicts with the master, and added an ability to inject a network layer to `RelayContext`, with `RelayDefaultNetworkLayer` injected by default. Hope this will help.

josephsavona:
@denvned awesome, quick turnaround as always! I'll import the latest version.

src/__forks__/Relay.js
```diff
@@ -16,10 +16,6 @@
 var RelayDefaultNetworkLayer = require('RelayDefaultNetworkLayer');
 var RelayPublic = require('RelayPublic');
 
-// By default, assume that GraphQL is served at `/graphql` on the same domain.
-// $FlowFixMe(>=0.16.0)
-RelayPublic.injectNetworkLayer(new RelayDefaultNetworkLayer('/graphql'));
```
denvned: I have removed the injection here because the default network layer is now injected in the constructor of `RelayContext`.

josephsavona:
@facebook-github-bot import

Issue #677, opened 2015-12-11
Validate connection calls after excluding null values at runtime


BerndWessels:
Hello
Using the following query:

      fragment on Person {
        firstName
        email
        posts(first: $first, after: $after, last: $last, before: $before) {
          edges {
            cursor
            node {
              id
              title
            }
          }
          totalCount
          pageInfo {
            hasNextPage
            hasPreviousPage
            startCursor
            endCursor
          }
        }
      }

causes the following error:

    Uncaught Error: GraphQL validation/transform error ``Connection arguments `posts(before: <cursor>, first: <count>)` are not supported. Use `(first: <count>)`, `(after: <cursor>, first: <count>)`, or `(before: <cursor>, last: <count>)`.``

So obviously some combinations of parameters are not valid - and that is fair enough.

But I was expecting that the validation only happens after it excluded variables with value `null`. Unfortunately that doesn't seem to be the case, otherwise we could use the previous query with these variables:

    export default Relay.createContainer(Home, {
      initialVariables: {
        first: 2,
        after: "YXJyYXljb25uZWN0aW9uOjQ=",
        last: null,
        before: null,
      },
      fragments: {
        viewer: () => Relay.QL`
          fragment on Person {

And it should succeed since only the `first` variable is non-null and all the other variables should be ignored / not included.

So for now we have to use terrible workarounds like this:

    export default Relay.createContainer(Home, {
      initialVariables: {
        forward: true,
        first: 2,
        after: "YXJyYXljb25uZWN0aW9uOjQ=",
        last: null,
        before: null,
      },
      fragments: {
        viewer: () => Relay.QL`
          fragment on Person {
            firstName
            email
            posts(first: $first, after: $after) @include(if: $forward) {
              edges {
                cursor
                node {
                  id
                  title
                }
              }
              totalCount
              pageInfo {
                hasNextPage
                hasPreviousPage
                startCursor
                endCursor
              }
            },
            posts(last: $last, before: $before) @skip(if: $forward) {
              edges {
                cursor
                node {
                  id
                  title
                }
              }
              totalCount
              pageInfo {
                hasNextPage
                hasPreviousPage
                startCursor
                endCursor
              }
            },
            ${UpdatePersonMutation.getFragment('person')}
          }
        `
      }
    });

So basically my feature request is that variables with `null` value will be ignored and excluded before query validation and submitting.

josephsavona:
[...]

PR #708, opened 2015-12-29
Allow variable connection arguments

NevilleS:
Relay Connections use a handful of arguments for pagination (first, last, before, after). These connection arguments are statically validated by the babel-relay-plugin, and then checked again at runtime by GraphQLRange. However, if the developer wants to use variables for the connection arguments (for example to support switching between `first` & `last` at runtime), the validation performed by babel-relay-plugin gets in the way and prevents this, which forces the developer to use some workarounds to "trick" the validation (see issue #677 for some examples).

This commit changes the babel-relay-plugin validation to allow otherwise invalid combinations (like `first`+`before`) as long as all the arguments are variables. When the connection arguments are variables, babel-relay-plugin effectively "trusts" that the developer will provide the correct arguments at runtime, and it becomes the job of GraphQLRange to validate the calls given the actual values. GraphQLRange was already doing this validation, so I simply added another unit test to verify that it does, in fact, reject the only unsupported runtime combination (`first`+`last`).

NOTE: babel-relay-plugin and GraphQLRange have different "expectations" for what combinations of arguments to a connection are, in fact, valid. For example, babel-relay-plugin claims that `before`+`first` is invalid, even though GraphQLRange happily accepts that combination. I chose not to change this, trusting that there was probably a good reason for the discrepancy and that babel-relay-plugin was just being more defensive. However, I did notice that babel-relay-plugin will (amusingly) allow the one combination of arguments that GraphQLRange cannot support: `first`+`last`, so I added a new validation to babel-relay-plugin to prevent that... unless they are both variables, of course!

This fixes issue #677.


yungsters:
Sure, thanks!

NevilleS:
Alright @yungsters I rebased to latest master, and fixed up some things in my changes. It looks like there was a decent amount of refactoring in the relay plugin code, but I think I'm following the new behaviours correctly.

For the "happy path" tests, I still feel a little weird just basically copying & pasting the generated code into the test fixture expected output, but I assume that's what you do too :smile:

Note that when I initially wrote this PR I tested that the change worked in a "real" app too, but in this case I'm just trusting the unit tests...

josephsavona:
My workflow for the plugin is to make changes, use `npm run update-fixtures` to regenerate the output portions of the fixture files, and then use git diff to verify that the change in output is what I expected. 

Looks like the Babel config in transformGraphQL.js needs `compact: true` for this to work as expected.

NevilleS:
`npm run update-fixtures`, aha, noted. I copy & pasted the failure message and then prettified it :smile:

scripts/babel-relay-plugin/src/RelayQLPrinter.js
```diff
@@ -671,20 +671,29 @@ module.exports = function(t: any, options: PrinterOptions): Function {
   }
 
   function validateConnectionField(field: RelayQLField): void {
+    let [first, last, before, after] = [field.findArgument('first'), field.findArgument('last'),
+      field.findArgument('before'), field.findArgument('after')]
```
yungsters: Nit: Multiline (with trailing comma).

```
let [...] = [
  field.findArgument('first'),
  // ...
  field.findArgument('after'),
];
```

yungsters:
Thanks for rebasing, @NevilleS. This is great.

Currently, `npm run update-fixtures` is a bit borked due to dependency hell, but #944 will fix it.

Just one nit before I import this.

scripts/babel-relay-plugin/src/RelayQLPrinter.js
NevilleS: Haha, sure. It's your house!

NevilleS:
Alright, fixed the style issue. Let me know if there's anything else amiss, I don't mind. I think I stuck to 80 character line lengths in this case...

yungsters:
@facebook-github-bot import

NevilleS:
That `@facebook-github-bot import` command is a missed opportunity to use `ship it` instead... Or `make it so`...

yungsters:
@NevilleS Haha, I generally only use **shipit** if the change is documentation-only. For everything else, I run through tests internally to triple-check that nothing breaks.

NevilleS:
Wait, `shipit` is a command for your bot? 

Nevermind, well played Facebook :clap:

yungsters:
@NevilleS Haha, yup. **import** will create an internal Phabricator revision that we can then land internally (which will get automatically sync'd out to GitHub and close this PR). It has something to do with unidirectional data flow or something like that. ಠ◡ಠ

On the other hand, **shipit** will automatically land the revision internally.

josephsavona:
@NevilleS thanks for this!

NevilleS:
No problem! What else do you have for me @josephsavona? Or are we going to figure out what to do with #782?

Issue #559, opened 2015-11-07
[meta] Relay Core API


josephsavona:
# Relay Core API

The current (public) Relay API is primarily focused on high-level integration with React via RelayContainer and RelayRenderer. Internally at Facebook, we make use of Relay internals in a variety of ways, notably to implement several variations on server "rendering" and for GraphQL subscriptions. We've also experimented with things like running parts of Relay in a web worker.

These types of features are currently somewhat difficult to implement in open source due to the restricted set of functionality exposed in the public API. This issue is to track work toward a new, more modular public API. We plan to strike a balance between overly monolithic and overly decomposed, and split Relay into two main parts:
- Relay Core: imperative API for fetching, observing, and updating data.
- React/Relay: React integration in the form of `RelayContainer` and `RelayRenderer`, implemented purely in terms of the public Relay Core API.
## Goals

Our goal with this approach is to enable more experimentation in user space, while retaining the ability to efficiently fetch data for entire view hierarchies in a single round trip.

Examples of things that should be possible with this new API:
- Server rendering (or pre-fetching data on the server and rendering on the client).
- Implementing real-time updates (subscriptions) in user-space.
- Offline persistence.
- Implementing your own alternatives to RelayContainer/RelayRenderer.
- Using Relay Core with non-React view libraries.

As mentioned, we're already doing some of these things internally: what's involved here is exposing more of the internal APIs in a slightly more user-friendly way, documenting them, and providing examples.
## Strawman API

Below is a possible list of core primitives along with examples of how they could be composed:
- readFragmentData - get the data for a fragment give some node id
- observeFragmentData - like read, but get notifications as data changes
- readQueryData - get the data for a query
- observeQueryData - like read, but get notifications of changes
- buildQueryFromFragment - turn a fragment into a query (useful for refetching select parts of data)
- canResolveQueryLocally - can a query be resolved from cached data?
- fetchQueryFromCache - load data from persisted storage. returns a payload that can be applied with `writeQueryPayload`
- fetchQueryFromNetwork(forceFetch) - fetch data from the network, optionally riffing against the cache. returns a payload that can be applied with `writeQueryPayload`
- sendOperation(operation) - send a mutation/subscription to the server, asynchronously get the response
- createOperation - given a mutation/subscription description and props, construct a mutation transaction
- createQuery - construct a runtime query descriptor from a Relay.QL expression
- createFragment - construct a runtime fragment descriptor from a Relay.QL expression
[...]

PR #721, opened 2016-01-05
Using Relay without React

andimarek:
This PR has to the goal to make Relay usable without React.

__Overview:__
The core functions of Relay don't depend really on React, but currently it's only really useable with `RelayContainer`, which is a higher order React component.

The idea behind this PR is to introduce two new classes: `GenericRelayRootContainer` and `GenericRelayContainer`. Both are very similar to `RelayRootContainer/RelayRenderer` and `RelayContainer`. The main difference is that they don't wrap any React Component and don't deal with any rendering. They just maintain the data specified by fragments. 
Instead of rendering a Component they call a registered callback function to inform about new/changed data.
This way it is possible to use co-located data declaration with any rendering technology.

To make it more clear I added a modified star-wars example to use Angular instead React: [examples/star-wars-angular](https://github.com/andimarek/relay/tree/generic-relay-container/examples/star-wars-angular)

This is already working but it's not perfect. I'm sharing this here to get some feedback, if this is even possible to get merged and what to improve/change.
 
__Implementation Details:__
Like I said I added to new classes: [GenericRelayRootContainer](https://github.com/andimarek/relay/blob/generic-relay-container/src/container/GenericRelayRootContainer.js) and [GenericRelayContainer](https://github.com/andimarek/relay/blob/generic-relay-container/src/container/GenericRelayContainer.js).

The `GenericRelayRootContainer` combines some aspects of `RelayRootContainer` and `RelayRenderer`: It basically just calls `getRelayQueries` for the first data fetching.

`GenericRelayContainer` is a reduced version of `RelayContainer`: It's not a React Component, but a simple class.The important parts for the data declaration are the same as in `RelayContainer`: There is still a `RelayContainerSpec` and Routes are also the same. But for example because there is no React Context anymore, the route has to provided along with all the other props.

I added a few new types, but reused all of the existing types where possible/made sense.

There are also two new test files [GenericRelayContainer-test](https://github.com/andimarek/relay/blob/generic-relay-container/src/container/__tests__/GenericRelayContainer-test.js) and [GenericRelayRootContainer-test](https://github.com/andimarek/relay/blob/generic-relay-container/src/container/__tests__/GenericRelayRootContainer-test.js)

The tests are kept minimal because I think some aspects of `GenericRelayContainer` and `RelayContainer` should be pulled out into own modules to make it more DRY. I wanted to wait for some feedback before duplicating too much test cases.


josephsavona:
Thanks for your contribution! It's exciting to see interest in using Relay with other view libraries. As noted in the [roadmap](https://github.com/facebook/relay/wiki/Roadmap#in-progress), we're actively on working on making Relay more modular - including separating Relay into a core, imperative data-fetching API and a declarative React integration layer. The core could then be used to implement integrations with other view libraries. 

The community has been actively contributing toward a prerequisite of this - #558 - to make all Relay state contextual instead of global, and this is almost finished. We (Relay core) are working on the core/react split - see #559.

We definitely would like to support the community in integrating Relay with other libraries. For this PR, it might be best to create a new package/repository - e.g. `relay-angular` - since you and the community will have more expertise about Angular best-practices. Note that `GenericRelay(Root)Container` is a good temporary solution for resolving Relay data until #558 and #559 are complete, at which point there will be a documented, public API for resolving data outside of Relay containers.

andimarek:
Thanks for your feedback and the references to the issues. 

Especially your comment about `DataContainer`(https://github.com/facebook/relay/issues/559#issuecomment-155703059):
This is I think the same idea as the `GenericRelayContainer`.

I will comment on this specific issues, as it seems the better place.

andimarek:
About the `relay-angular` suggestion: I just wanted to make clear, that the goal of this PR was not to make Relay work with Angular specifically. It was just the example I created with Angular, because I know Angular better than e.g. ember.

andimarek:
I'm closing this PR, because the topic will be handled in #559.

Issue #693, opened 2015-12-21
Using NODE_DELETE lists become [null] array with 1 item


oguzbilgic:
Using `NODE_DELETE`, related lists become `[null]` array with 1 item in them.

Given this mutation payload:

```js
fragment on DeleteCarPayload {
  id
  viewer { cars }
}
```

with this `getConfigs`:

```js
return [{
  type: 'NODE_DELETE',
  parentName: 'viewer',
  parentID: this.props.viewer.id,
  connectionName: 'cars',
  deletedIDFieldName: 'id'
}];
```

There is also another field that has cars in it: `shop.cars`. Car should also be removed from `shop.cars` list. However It's only converted to `null`, It's not removed, unlike `viewer.cars`

```js
// mutation applied
console.log(viewer.cars.length, viewer.cars)
// 0, []
console.log(shop.cars.length, shop.cars)
// 1, [null]
```

I believe this is a bug. I can't think of a reason why this should be the behavior.

josephsavona:
Thanks for reporting this. Relay removes deleted nodes from connections, and we should probably remove them from lists for consistency. In this case that would mean returning an empty list instead of a list of `[null]`. 

@yuzhi can you think of any reason we shouldn't/can't do this?

yuzhi:
@oguzbilgic Yeah `NODE_DELETE` mainly support nodes and nodes in connections since we do keep some extra metadata about each connection. If you want to update `viewer {cars}` and you know that can change as a part of this mutation, you can add another `FIELDS_CHANGE` config.
`
      {
        type: RelayMutationType.FIELDS_CHANGE,
        fieldIDs: { viewer: this.props.viewer.id},
      },
`
@josephsavona  We can probably do some filtering, although I don't know if there are legit cases for having nulls in plural fields.

oguzbilgic:
@yuzhi Thank you. Thats exactly how I got around. However It's really pita at the moment. Basically `NODE_DELETE` doesn't do what its suppose to, and we need to find every list that may include this , get their ids in `getFragment`, and maybe add to `payload` ...

josephsavona:
It definitely seems that filtering nulls would be the correct behavior here. @oguzbilgic any interest in submitting a PR to fix? 

[...]

PR #722, opened 2016-01-05
Filter records in plural linked fields which are deleted

andimarek:
This is a fix for issue #693.

The method `getLinkedRecordIDs` in `RelayRecordStore` now filters ids of records,
which are marked as "NONEXISTENT".

A new test is added and a single assertion in `readRelayDiskCache-test` is removed,
which failed after the change and was not really relevant for the test.

### DIFF
--- a/src/store/RelayRecordStore.js
+++ b/src/store/RelayRecordStore.js
@@ -480,7 +480,7 @@ class RelayRecordStore {
         dataID
       );
       return item.__dataID__;
-    });
+    }).filter(id => this.getRecordState(id) !== 'NONEXISTENT');
   }
 
   /**
--- a/src/store/RelayRecordStore.js
+++ b/src/store/RelayRecordStore.js
@@ -473,7 +473,7 @@ class RelayRecordStore {
     const linkedIDs = [];
     field.forEach((item, ii) => {
       invariant(
-        typeof item === 'object' && item.__dataID__,
+        typeof item === 'object' && item != null && item.__dataID__,
         'RelayRecordStore.getLinkedRecordIDs(): Expected element at index %s ' +
         'in field `%s` for record `%s` to be a linked record.',
         ii,
--- a/src/store/RelayRecordStore.js
+++ b/src/store/RelayRecordStore.js
@@ -459,7 +459,7 @@ class RelayRecordStore {
     dataID: DataID,
     storageKey: string
   ): ?Array<DataID> {
-    var field = this._getField(dataID, storageKey);
+    const field = this._getField(dataID, storageKey);
     if (field == null) {
       return field;
     }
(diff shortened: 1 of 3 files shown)


src/store/RelayRecordStore.js
```diff
@@ -516,7 +516,7 @@ class RelayRecordStore {
         dataID
       );
       return item.__dataID__;
-    });
+    }).filter(id => this.getRecordState(id) !== 'NONEXISTENT');
```
josephsavona: Sorry for the delay in reviewing this. The logic here is exactly what we want, but this will end up allocating an extra array. Maybe allocate a new array and use `forEach` instead of map + filter?

andimarek:
Hi,

I changed the implementation to use just one array.

src/store/RelayRecordStore.js
```diff
@@ -470,17 +470,22 @@ class RelayRecordStore {
       storageKey,
       dataID
     );
-    return field.map((item, ii) => {
+    const linkedIDs = [];
+    field.forEach((item, ii) => {
+      const id = item.__dataID__;
```
josephsavona: it's possible that the value could be an array of scalars, i'm surprised Flow accepted this. Can you revert this part to only access `__dataID__` after confirming `item` is a non-null object?

src/store/RelayRecordStore.js
```diff
@@ -470,17 +470,22 @@ class RelayRecordStore {
...
         ii,
         storageKey,
         dataID
       );
-      return item.__dataID__;
+      if (this.getRecordState(id) !== 'NONEXISTENT') {
+        linkedIDs.push(id);
+      }
     });
+    return linkedIDs;
```
josephsavona: otherwise :+1:

andimarek:
@josephsavona I changed it to make sure, that the item is only accessed after the invariant check. This what you mean, right?

andimarek:
Thanks for taking time to review and thanks for the suggestions to improve

josephsavona:
yup, this is great. technically have to add `item !== null` since typeof null is object (thanks javascript!), but if you don't mind i can just add that when i import it.

josephsavona:
@facebook-github-bot import

andimarek:
Yes, you are right: This check was missing missing before my change. I added it together with a test for it and improved a second test a bit to test also for the error message, not only if it's thrown any error.

andimarek:
@josephsavona I'm not sure why the test is failing now.

andimarek:
It's probably because of the github outage.

andimarek:
@josephsavona Anything else here I can do/should do? Thanks!

wincent:
@andimarek Your PR has been imported into our internal Phabricator instance, but I see @josephsavona marked it as "Changes planned". Can you chime in on what you're wanting Joe?

josephsavona:
@wincent @yuzhi Thoughts on this approach? Note that the equivalent function for reading range data returns both a filtered and unfiltered list of edges, but `getLinkedRecordID` returns the id regardless of the linked record's status. I'm torn about whether the filtering really belongs in the store or if it should happen at a higher level (such as the reader). Thoughts?

josephsavona:
@andimarek thanks for your patience on this PR. Summarizing an offline discussion with @yuzhi: It turns out that this change would have unintended consequences and we can't proceed with it as-is. The cache reader needs to know what IDs are linked, even if they don't exist in the store at the moment (this is why `getRangeMetadata` returns both a filtered and unfiltered list of edges). Filtering the IDs at the store level also changes which IDs are used to set up component subscriptions, and could cause components not to update when they should. 

A more appropriate place to add this logic would be in `readRelayQueryData` when reading the results for plural linked fields.

yuzhi:
Just adding to what @josephsavona already mentioned. The similar problem can also occur when we GC the linked nodes but not the parent. Filtering out at the RelayRecordStore level can cause us to think there are no linked nodes when in reality, these nodes just got temporarily cleared by GC. We won't be able to tell the difference between the node is null vs node is node missing there due to GC or DiskCache since many of our internal functions depend on this function.

andimarek:
@josephsavona  @yuzhi  Thanks for the background infos. I'm closing this PR, because as you explained it's the wrong approach.

PR #726, opened 2016-01-07
Build the Relay Playground in DEV mode.

steveluscher:


steveluscher:
@facebook-github-bot import

PR #730, opened 2016-01-08
Don't try to setState after a fragment data update if component not mounted

skevy:
In certain cases, if a RelayContainer is mounted and unmounted quickly, a warning about setting state on an unmounted component is thrown.

This happens due to the fact that in the callback for fragment data updates from the GraphQLStoreQueryResolver, we `setState` without checking whether or not the component is mounted.

This PR resolves that issue.


steveluscher:
Thanks!

@facebook-github-bot import

skevy:
:rocket:

Issue #256, opened 2015-09-08
babel-relay-plugin => babel-plugin-relay


thejameskyle:
The babel plugin in `scripts/babel-relay-plugin` should be named according to the `babel-plugin-*` naming scheme. However, the name is currently taken by an empty project.

I've contacted the package owner on npm and CC'd npm support according to https://docs.npmjs.com/misc/disputes. Hopefully the owner releases the name without any issue.

thejameskyle:
The module name has been released by the owner https://www.npmjs.com/package/babel-plugin-relay

thejameskyle:
I've reseved the name for now and added all the owners of the `babel-relay-plugin` module on npm to it. Feel free to remove me and replace it, if you have any issues just ping me.

devknoll:
p.s. the intention was to publish #206 separately under this package name, but since there's interest in having this functionality in the current `babel-relay-plugin` / I haven't had time to actually work on the project yet, I went ahead and released the name.

josephsavona:
@devknoll thanks!

josephsavona:
also thanks @thejameskyle!

devknoll:
@josephsavona this happening? :wink:

josephsavona:
@devknoll Yeah, we should eventually move to the recommended naming scheme. This might take a few rounds of PRs:
- change the package name in the code (and then publish it)
- update everything to depend on it
- also update the docs

Community help appreciated! ;-)

devknoll:
Thanks for the list and the tag, that's what I wanted to see :+1:

thejameskyle:
ping

BerndWessels:
@devknoll steveluscher @josephsavona Will this replace the `babel-relay-plugin` on NPM ?
If so it should still be in sync with the version number of `react-relay` to avoid confusion ?!
But it looks like there is a `1.0.0` published to NPM - but `react-relay` is only `0.6.1`.
Can you please fix this or am I mistaken here?

frodosamoa:
:+1:

schickling:
Until this is happening we've built [babel-plugin-react-relay](https://github.com/graphcool/babel-plugin-react-relay) supporting different source types such as JSON files, JS schema definitions and even URL schema endpoints.

This is meant to provide a more convenient and powerful solution without the need for scripts like `build/babelRelayPlugin.js`.

josephsavona:
[...]

PR #732, opened 2016-01-10
Rename babel plugin

devknoll:
Renames the babel plugin internally, fixing #256. A follow up PR will change the package.json to use the new plugin name.


BerndWessels:
@devknoll @steveluscher @josephsavona Will this replace the `babel-relay-plugin` on NPM ?
If so it should still be in sync with the version number of `react-relay` to avoid confusion ?!
But it looks like there is a `1.0.0` published to NPM - but `react-relay` is only `0.6.1`.
Can you please fix this or am I mistaken here?

devknoll:
@BerndWessels Yes, the intention would be to publish a `babel-plugin-relay@0.6.3` at the same time as deprecated `babel-relay-plugin@0.6.3`.

BerndWessels:
@devknoll OK, but why is there a `babel-plugin-relay` with version `1.0.0` already published to NPM? That seems wrong and might lead to confusion?

BerndWessels:
@devknoll Oh, I see - just for holding on to the name on NPM - so you will delete the `1.0.0` later I guess.

josephsavona:
@devknoll import this!

devknoll:
@facebook-github-bot import

yungsters:
So besides rebasing, what else do we need to do to get this in?

josephsavona:
A quick update here: we understand that this can be helpful in OSS. At the same time, we use this plugin heavily internally, and this PR requires quite a bit of work to successively land and reconfigure things in a few environments. It's a lot of work and doesn't seem to be super high-value relative to other feature requests and core improvements. 

The easiest thing might be to temporarily provide two packages, where `babel-relay-plugin` just re-exports `babel-plugin-relay` (or vice-versa). That way we can make the change, and then as a follow-up go back and update our internal configuration w/o breakage.

Anyone interested in working on this?

josephsavona:
I'm going to close this - see my previous comment for more context. We're happy to do the rename, but it has to be gradual in order to not break things.

Issue #541, opened 2015-11-02
[meta] Support real-time updates via event subscriptions


josephsavona:
Realtime data in GraphQL is something that we and the community are actively exploring. There are many ways to achieve "realtime" or near-realtime updates: polling, "live" queries, or event-based approaches (more on these tradeoffs [on the GraphQL blog](http://graphql.org/blog/subscriptions-in-graphql-and-relay/)). Furthermore, there are a variety of transport mechanisms to choose from depending on the platform: web sockets, MQTT, etc.

Rather than support any one approach directly, we would prefer to allow developers to implement any of these approaches. Therefore, we don't plan to create a `RelayMutation`-style API for subscriptions. Instead we're working create a "write" API that will make it easy for developers to tell Relay about new data (along the lines of `store.write(query, data)`). See #559 for more information.

For now, we recommend checking out @edvinerikson's [`relay-subscriptions`](https://github.com/edvinerikson/relay-subscriptions) module.

skevy:
Just fwiw, I've actually started implementing this. I'm only so far into it - but I just wanted to put here that I'm actively working on it.

josephsavona:
@skevy I'm not sure why, but I can't assign this to your directly - but thanks for the heads up, looking forward to this!

skevy:
@josephsavona it's cuz I'm not a collaborator. Silly Github.

Is anyone at FB working on this for OSS release? Or only for internal use

josephsavona:
Aha! No one is actively working on the OSS Subscriptions (the API described above) - so you won't conflict :-)

dallonf:
The "subscribe/dispose" API seems awfully imperative and fiddly to control... what about something like this where RelayContainers define subscriptions declaratively:

```js
module.exports = Relay.createContainer(Story, {
  fragments: {
    story: () => Relay.QL`
      fragment on Story {
        text,
        likeCount
      }
    `,
  },
  subscriptions: (props) => [
    new StoryLikeSubscription({storyId: props.storyId})
  ]
});
```

josephsavona:
@dallonf That's a great idea. Ultimately there has to be an imperative API *somewhere*: in Relay, that's `Relay.Store`. We do need imperative APIs for opening subscriptions and disposing them, but the API you described would be a great way to provide access to those methods from containers.

Note that the query tree is static and `props` are unavailable: instead the API should use `variables`:

```javascript
  subscriptions: variables => [...]
```

skevy:
[...]

PR #736, opened 2016-01-12
Change clientMutationID invariant to a conditional check to work with subscriptions

eyston:
I'm working on #541 and `handleRangeAdd` in `writeRelayUpdatePayload` has an invariant on `clientMutationID` being in the payload.  This won't be the case for a subscription payload configured as a RANGE_ADD.

From what I can tell the `clientMutationID` is used in order to link up a temporary node ID added by an optimistic mutation with the real one from the mutation payload.  This won't be necessary for subscriptions.

I replaced the invariant with a conditional check on the `clientMutationID`.  I also duplicated the test the for non-optimistic RANGE_ADD's without a `clientMutationID` which I'm not sure is necessary or not.

### DIFF
--- a/src/traversal/writeRelayUpdatePayload.js
+++ b/src/traversal/writeRelayUpdatePayload.js
@@ -321,13 +321,6 @@ function handleRangeAdd(
   config: OperationConfig,
   isOptimisticUpdate: boolean
 ): void {
-  const clientMutationID = getString(payload, CLIENT_MUTATION_ID);
-  invariant(
-    clientMutationID,
-    'writeRelayUpdatePayload(): Expected operation `%s` to have a `%s`.',
-    operation.getName(),
-    CLIENT_MUTATION_ID
-  );
   const store = writer.getRecordStore();
 
   // Extracts the new edge from the payload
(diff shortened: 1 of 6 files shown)


eyston:
I updated the tests to use a `subscription` query to make it clear what functionality is being tested.  This required adding a subscription root with a field.  I'm not sure the proper naming conventions that would like to be maintained for the subscription schema but can change them to match.  Also, `Relay.QL` requires subscriptions to have a single `input` with at least `clientSubscriptionID` so that was done.  I'd be interested in the requirement of `clientSubscriptionID` if time permits :).

src/traversal/writeRelayUpdatePayload.js
```diff
@@ -381,25 +374,29 @@ function handleRangeAdd(
...
-    // server ID.
-    const clientNodeID =
-      RelayMutationTracker.getClientIDForMutation(clientMutationID);
-    if (clientNodeID) {
-      RelayMutationTracker.updateClientServerIDMap(
-        clientNodeID,
-        nodeID
+  const clientMutationID = getString(payload, CLIENT_MUTATION_ID);
+  // subscriptions won't have a clientMutationID so never optimistically add
+  if (clientMutationID) {
```
josephsavona: How about changing the outer `if` condition to `if (operation instanceof RelayQuery.Mutation)`, and move the invariant on `clientMutationId` inside the if block? This will get the intended effect of not running the invariant for subscriptions, while keeping the validation for mutations.

josephsavona:
@eyston Sorry for the delay in providing feedback: this is great. Other than the one piece of feedback above, this is ready to import.

eyston:
Updated with feedback -- which was way smarter check condition.  Also rebased off master to fix CI.  I can squish this or something too, dunno conventions!

Sorry for delay myself.  Busy week!  Good times.

josephsavona:
@eyston thanks for this, looks great and we'll import next week. There's no need to squash commits for a single PR, our tool imports the combined patch. Thanks for offering though ;-)

Globegitter:
@josephsavona Any update on getting this merged?

taion:
Following up on this – did this code ever get pulled in? Right now it's not possible to apply a RANGE_ADD subscription response without setting a dummy `clientMutationId`.

edvinerikson:
I think we can solve this by moving: [writeRelayUpdatePayload.js#L320-L326](https://github.com/facebook/relay/blob/master/src/traversal/writeRelayUpdatePayload.js#L320-L326]) to [writeRelayUpdatePayload.js#L376-L383](https://github.com/facebook/relay/blob/master/src/traversal/writeRelayUpdatePayload.js#L376-L383)

like it is done here: [RelayStoreData.js#L376-L383](https://github.com/facebook/relay/blob/master/src/store/RelayStoreData.js#L376-L383)

taion:
I feel like this PR does the trick pretty well, though.

taion:
Ping! We're working around this in relay-subscriptions right now by adding a dummy `clientMutationId` that won't conflict, but it'd be good if we could remove that hack.

PR #746, opened 2016-01-15
Make Treasure Hunt tutorial color blind friendly :)

chrisbolin:
Trivial change: making the "Treasure" box color blue instead of green. I honestly couldn't tell the difference between the red and green at first :tongue:


steveluscher:
Awesome.

@facebook-github-bot import

PR #748, opened 2016-01-16
Updates for future updates to jest and node-haste.

cpojer:


scripts/jest/preprocessor.js
```diff
@@ -45,7 +45,10 @@ var babelOptions = assign(
     blacklist: babelDefaultOptions.blacklist.concat('validation.react'),
     _moduleMap: fixModules(assign({}, babelDefaultOptions._moduleMap, require('fbjs/module-map'), {
       'React': 'react',
+      'reactComponentExpect': 'react/lib/reactComponentExpect',
+      'ReactDOM': 'react-dom',
       'ReactUpdates': 'react/lib/ReactUpdates',
+      'ReactTestUtils': 'react/lib/ReactTestUtils.js',
```
cpojer: the only reason why these three modules work right now is because node-haste1 looks into node_modules folders which node-hast2 doesn't do anymore (unless explicitly required).

package.json
```diff
@@ -73,7 +73,7 @@
     "persistModuleRegistryBetweenSpecs": true,
     "modulePathIgnorePatterns": [
       "<rootDir>/lib/",
-      "<rootDir>/node_modules/(?!(fbjs/lib/|react/lib/))"
+      "<rootDir>/node_modules/(?!(fbjs/lib/|react/lib/|fbjs-scripts/jest))"
```
cpojer: this is ignored but for some reason haste1 allows you to require the setup env script anyway.

cpojer:
@facebook-github-bot import

cpojer:
@facebook-github-bot shipit

PR #749, opened 2016-01-16
Update license year range to 2016

pra85:

PR #757, opened 2016-01-20
Future-proof copyright year ranges

wincent:
We use the "-present" range in other open source projects, so let's do the same here, which means we won't need to update these every year.


wincent:
Closing in favor of #749.

PR #758, opened 2016-01-20
Fix import react and relay in the tutorial example

tupy:
Add missing imports to fix the the tutorial following the example in the relay repository: https://github.com/facebook/relay/blob/master/examples/relay-treasurehunt/js/components/App.js


yungsters:
This was accomplished by 72580206f1e8eb790f78a6b4ffa3c2803d9654b6.

PR #761, opened 2016-01-22
Contextualize RelayContainer

denvned:

### DIFF
--- a/src/container/RelayContainer.js
+++ b/src/container/RelayContainer.js
@@ -21,6 +21,7 @@ const React = require('React');
 const ReactDOM = require('ReactDOM');
 const RelayContainerComparators = require('RelayContainerComparators');
 const RelayContainerProxy = require('RelayContainerProxy');
+const RelayContext = require('RelayContext');
 const RelayFragmentReference = require('RelayFragmentReference');
 import type {RelayQuerySet} from 'RelayInternalTypes';
 const RelayMetaRoute = require('RelayMetaRoute');
@@ -29,14 +30,11 @@ const RelayPropTypes = require('RelayPropTypes');
 const RelayProfiler = require('RelayProfiler');
 const RelayQuery = require('RelayQuery');
 const RelayRecord = require('RelayRecord');
-const RelayStore = require('RelayStore');
-const RelayStoreData = require('RelayStoreData');
 import type {
   Abortable,
   ComponentReadyStateChangeCallback,
   RelayContainer,
   RelayProp,
-  Subscription,
   Variables,
 } from 'RelayTypes';
 import type URI from 'URI';
@@ -74,14 +72,9 @@ export type RootQueries = {
 };
 
 var containerContextTypes = {
+  relay: React.PropTypes.instanceOf(RelayContext).isRequired,
   route: RelayPropTypes.QueryConfig.isRequired,
 };
-
-var storeData = RelayStoreData.getDefaultInstance();
-
-storeData.getChangeEmitter().injectBatchingStrategy(
-  ReactDOM.unstable_batchedUpdates
-);
 
 /**
  * @public
@@ -110,7 +103,6 @@ function createContainerComponent(
 
   class RelayContainer extends React.Component {
     mounted: boolean;
-    _deferredSubscriptions: ?{[subscriptionKey: string]: Subscription};
     _didShowFakeDataWarning: boolean;
     _fragmentPointers: {[key: string]: ?GraphQLFragmentPointer};
     _hasStaleQueryData: boolean;
@@ -128,7 +120,15 @@ function createContainerComponent(
     constructor(props, context) {
       super(props, context);
 
-      var {route} = context;
+      var {relay, route} = context;
+      invariant(
+        relay instanceof RelayContext,
+        'RelayContainer: `%s` was rendered without a valid `relayContext`. ' +
+        'Make sure the `relayContext` is valid, and make sure that it is ' +
+        'correctly set on the parent component\'s context ' +
+        '(e.g. using <RelayRootContainer>).',
+        containerName
+      );
       invariant(
         route && typeof route.name === 'string',
         'RelayContainer: `%s` was rendered without a valid route. Make sure ' +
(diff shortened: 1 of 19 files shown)


src/container/RelayContainer.js
```diff
@@ -74,15 +72,10 @@ export type RootQueries = {
 };
 
 var containerContextTypes = {
+  relay: React.PropTypes.instanceOf(RelayContext).isRequired,
```
josephsavona: How about moving this to `RelayPropTypes`?

src/container/RelayContainer.js
```diff
@@ -128,7 +120,15 @@ function createContainerComponent(
...
-      var {route} = context;
+      var {relay, route} = context;
+      invariant(
+        relay instanceof RelayContext,
+        'RelayContainer: `%s` was rendered without a valid `relayContext`. ' +
+        'Make sure the `relayContext` is valid, and make sure that it is ' +
+        'correctly set on the parent component\'s context ' +
+        '(e.g. using <RelayRootContainer>).',
+        containerName
+      );
```
josephsavona: Nice. The key `relayContext` in the message doesn't match the actual key name, though. Maybe something like 

    'RelayContainer: `%s` was rendered without setting the `relay` property of the React context....'

src/container/RelayContainer.js
```diff
@@ -110,7 +103,6 @@ function createContainerComponent(
 
   class RelayContainer extends React.Component {
     mounted: boolean;
-    _deferredSubscriptions: ?{[subscriptionKey: string]: Subscription};
```
josephsavona: nice! thanks for cleaning this up

src/container/RelayRenderer.js
```diff
@@ -126,7 +134,7 @@ class RelayRenderer extends React.Component {
   }
 
   getChildContext(): Object {
-    return {route: this.props.queryConfig};
+    return {relay: RelayStore, route: this.props.queryConfig};
```
josephsavona: nit: multi-line

src/container/__tests__/RelayContainer-test.js
```diff
@@ -310,18 +325,42 @@ describe('RelayContainer', function() {
...
+    RelayTestRenderer.render(
+      () => <MockContainer foo={mockFooPointer} />,
+      mockRelayContextB,
+      mockRoute
+    );
+
+    const mockResolvers = GraphQLStoreQueryResolver.mock.instances;
+    expect(mockResolvers.length).toBe(2);
+    expect(mockResolvers[1].mock.store).toBe(mockRelayContextB.getStoreData());
+    expect(mockResolvers[0].reset).toBeCalled();
```
josephsavona: These are great tests. My only feedback is that nothing is tested between the first and second render call - ideally we'd verify the state after the first render and then test how it changed after the second render.

src/container/__tests__/RelayContainer-test.js
```diff
@@ -373,6 +416,50 @@ describe('RelayContainer', function() {
...
+      () => <MockContainer foo={mockFooPointer} />,
+      mockRelayContext,
+      mockRouteB
+    );
+
+    const mockResolvers = GraphQLStoreQueryResolver.mock.instances;
+    expect(mockResolvers.length).toBe(1);
+    expect(mockResolvers[0].resolve.mock.calls.length).toBe(2);
+  });
+
```
josephsavona: see note above

src/container/__tests__/RelayContainer_hasFragmentData-test.js
```diff
@@ -34,13 +34,14 @@ describe('RelayContainer', () => {
           foo: () => Relay.QL`fragment on Node{id}`,
         },
       });
+      const relayContext = new RelayContext();
       const RelayTestRenderer = RelayTestUtils.createRenderer();
       mockContainerInstance = RelayTestRenderer.render(genMockPointer => {
         return <MockContainer foo={genMockPointer('42')} />;
-      });
+      }, relayContext);
```
josephsavona: supernit: we generally prefer each argument on its own line unless it's a single argument - it makes it easy to miss the last argument otherwise.

src/store/RelayStoreData.js
```diff
@@ -86,10 +84,7 @@ class RelayStoreData {
    * Get the data set backing actual Relay operations. Used in GraphQLStore.
    */
   static getDefaultInstance(): RelayStoreData {
-    if (!_instance) {
-      _instance = new RelayStoreData();
-    }
-    return _instance;
+    return RelayStore.getStoreData();
```
josephsavona: there's a circular dependency here:

- `RelayStore` requires `RelayContext` to create the default instance
- `RelayContext` requires `RelayStoreData` to create the underlying data object
- `RelayStoreData` requires `RelayStore` in order to return the default instance

This can break depending on the build system used. What's the motivation for changing the implementation here? Seems like the previous approach would still work.

josephsavona:
@denvned Thanks for this! I have lots of minor feedback but this is great. I'm going to import now to start testing it out, but feel free to make updates (if you can!) and i'll re-import.

josephsavona:
@facebook-github-bot import

src/store/RelayStoreData.js
denvned: The motivation was to get rid of `storeData` argument in the constructor of `RelayContext`. That makes us sure that no two instances of `RelayContext` share the same instance of `RelayStoreData`, so it is enough to write `relayContextA !== relayContextB` to check that `relayContextA.getStoreData() !== relayContextB.getStoreData()`.

I think, we could break the circular dependency by removing `getDefaultInstance` from `RelayStoreData`, and replacing `RelayStoreData.getDefaultInstance()` with `RelayStore.getStoreData()` everywhere (it's only 3 modules and 2 tests in OSS). What do you think?

denvned:
@josephsavona Thanks for review! I updated PR with all your corrections except the one about the circular dependency (see my [comment](https://github.com/facebook/relay/pull/761#discussion_r50822499)).

src/store/RelayStoreData.js
josephsavona: Thanks for clarifying - that makes total sense, and :+1: to replacing `RelayStoreData.getDefaultInstance()` with `RelayStore.getStoreData()`.

denvned: Replaced `RelayStoreData.getDefaultInstance()` with `RelayStore.getStoreData()`.

josephsavona:
@facebook-github-bot import

Issue #706, opened 2015-12-27
Support node 5.x


kasselTrankos:
Hi,
Im trying to use relay with node 5.x, and give me allways 
Current node version is not supported for development, expected "5.3.0" to satisfy "4.x"

Could i use it?
tks

josephsavona:
Thanks for filing this. I'm not aware of any reason that we *can't* support node 5.x, but others will know more.

cc @zpao

kasselTrankos:
Ok, meanwhile i'll try to use in one isomorphic project

kasselTrankos:
thanks

zpao:
I don't recall if there was a reason not to support node 5 for development, but npm 2 was still a requirement due to Flow configs IIRC.

kasselTrankos:
hi @zpao, i put the error code.
```javascript
AssertionError: Current node version is not supported for development, expected "5.3.0" to satisfy "4.x".
    at Object.<anonymous> (/home/vera/js/relay/base/node_modules/fbjs-scripts/node/check-dev-engines.js:38:3)
```
 :(, so if npm requirement Flow change will be usefull,

edvinerikson:
This seems to have been resolved in https://github.com/facebook/relay/commit/60c8a76a47385d8f843e35de72d16f758689abc0

josephsavona:
As @edvinerikson noted this should now be resolved.

zpao:
Fwiw, bumping the npm version there is probably not right. Flow likely
won't work (because of nested node_modules in flowconfig)

josephsavona:
Reopening per @zpao

edvinerikson:
@zpao can you elaborate a little bit more? I ran `flow check src/` and got `Found 0 errors` back.
When I added a type error I got:
```
~/code/relay(master ✗) flow check src/
src/RelayPublic.js:39
 39:   return 1 * 1;
              ^^^^^ number. This type is incompatible with
 38: function bar(): React$Element {
                     ^^^^^^^^^^^^^ React$Element

src/RelayPublic.js:42
 42: foo('string');
     ^^^^^^^^^^^^^ function call
 42: foo('string');
         ^^^^^^^^ string. This type is incompatible with
 34: function foo(a: number): React$Element {
                     ^^^^^^ number

Found 2 errors
```
and the code:
```jsx
function foo(a: number): React$Element {
  return <span />;
}

function bar(): React$Element {
  return 1 * 1;
}

foo('string');
bar();

```

[...]

PR #772, opened 2016-01-27
Allow Node 5 and npm 3

wincent:
I think we can probably loosen our restriction on Node and npm versions
at this point.

I just tried this with Node 5.5.0 and npm 3.3.12, and compared with Node
4.2.6 and npm 2.15.12. There are no new lint or Flow issues under the
new versions, and install and build works. Also confirmed that examples
still run (actually found a real issue in one but will fix that
separately).

Closes: https://github.com/facebook/relay/issues/706

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,8 +62,8 @@
     "webpack-stream": "^2.1.0"
   },
   "devEngines": {
-    "node": "4.x",
-    "npm": "2.x"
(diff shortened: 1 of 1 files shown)


wincent:
No longer necessary, as already happened in 60c8a76a47385d8f843e3.

PR #773, opened 2016-01-27
Update example schemas

wincent:
While testing each of the examples under Node 5 and npm 3 I ran `npm run
update-schema`. Turns out these haven't been refreshed in a while, so
the output format doesn't match that produced by the current version of
our "graphql" dependency. May as well update it.


wincent:
@facebook-github-bot import

PR #774, opened 2016-01-27
Track fragments on null linked fields

### LINKED ISSUE #745: Mutation query missing fields that should have been tracked by previous queries
This is a hard one to describe succinctly, so excuse the length. I think this is a bug, but can't rule out the possibility that we're just doing something wrong here...

We noticed some peculiar behaviour where our mutation queries get formed incorrectly in some circumstances. It seems that if multiple components register fragments for a linked field that is initially returned as `null`, the Relay store can get into a bad state wherein future mutation queries will only refetch the first fragment on that field, even though the fat query should include both. Basically, the store doesn't seem to be tracking fragments correctly in this case. Let me try to explain that in some more detail.

### Reproduction
I haven't built a repro for this in something like Relay Playground yet. I can probably figure out a way to do that, but I'm not particularly familiar with writing schemas with `graphql-js` yet, so instead I've just typed up the simplest example I could come up with... sorry.

(EDIT: I added a unit test that reproduces the issue here: https://github.com/NevilleS/relay/commit/a39ab59e62c0d466e7138cc90ea022e659f2408c)

### Example
Let's say we have this great app where we track `Users` and `Birthdays`.  A `User` type has a single `birthday` field, and a `Birthday` type has `year, `month`, and `day` fields, which are strings. Both `User` and `Birthday` are nodes, so they have ids. Our app has two components: `UserComponent` and `UserDetailsComponent`. We have a single `SetBirthdayMutation` that is used to create or update a `User`'s `birthday`.

Generally speaking, the user flow is something like this:
1) Navigate to some kind of user profile, where we use a `UserComponent` to render basic information about the user, including their birth year if the birthday is set.
2) Click on a "details" link which uses a `UserDetailsComponent` to render all the birthday information: year, month, and day - if it's set. There's also a form here that lets you edit the birthday data.
3) Complete a form that submits a `SetBirthdayMutation` that creates or updates the `birthday` field.

Step 1) `UserComponent` just shows the user's birthday year:
```
fragment on User {
  birthday  {
    year
  }
}
```

Step 2) `UserDetailsComponent` shows the month and day
```
fragment on User {
  birthday  {
    month
    day
  }
}
```

Step 3) `SetBirthdayMutation` takes a `user_id`, `year, `month`, and `day`, and creates/updates a `Birthday` for the given `User`. The fat query looks like this:
```
fragment on SetBirthdayMutationPayload {
  user {
    birthday
  }
}
```
Let's assume we use `FIELDS_CHANGE` and provide the `user_id` as the parentID, etc. etc.

### The Problem
Here's the thing - the `birthday` field for a user isn't defined until that mutation is run at least once, which creates the field. That means if we run those steps in order, the server returns `birthday: null` for the first step, and Relay doesn't even include the `birthday { day }` fragment in the second query, since it already knows that the `birthday` doesn't exist, I guess (that's a reasonable optimization). That looks something like this:
```
// Query 1)
query {
  node(id: "some-user") {
    id
    birthday {
      id
      year
    }
  }
}

// Response 1)
data = {
  node: {
    id: "some-user",
    birthday: null
  }
}

// Query 2)
query {
  node(id: "some-user") {
    id
    // other fields, but does not request birthday { year, month, day }
  }
}

// Response 2)
data = {
  node: {
    id: "some-user",
    // other fields, no birthday
  }
}
```

Alright, here's the curious part. When we run the mutation following steps 1 & 2, it looks something like this (omitting some obvious things like clientMutationId for brevity):
```
mutation set_birthday_mutation(input: { user_id: "some-user", month: "8", day: "13" }) {
  user {
    birthday {
      year    // ONLY THE YEAR IS REQUESTED, WHY?
    }
  }
}
```

For some reason, **the mutation doesn't try to fetch birthday { month, day }*** if you run that sequence. However, if you go straight to the `UserDetailsComponent` without doing the `UserComponent` first (e.g. only steps 2 and 3), the sequence looks more like this
```
// Query 1) is not sent

// Query 2) is the new first query
query {
  node(id: "some-user") {
    id
    birthday {
      year
      month
      day
    }
  }
}

// Response 2)
data = {
  node: {
    id: "some-user",
    birthday: null
  }
}

// Mutation
mutation set_birthday_mutation(input: { user_id: "some-user", month: "8", day: "13" }) {
  user {
    birthday {
      year
      month
      day    
    }
  }
}
```
In this sequence, Relay correctly tracks the second `Birthday` fragment, so we get back the new birthday data we were expecting.

### Theory
It seems that the problem is somehow related to there being two fragments defined on `Birthday`, and only the first one is being tracked. I'm guessing that the fact that `birthday: null` is returned is important, and probably ends up meaning that the fragment isn't added to the tracker. Since the record store isn't updated by either query (no data), and the tracker isn't updated, the intersection performed by `RelayMutationQuery` results in a query that only contains the fragment of the first query.

Thanks for your help!

NevilleS:
This fixes #745.

Essentially, what's happening is that the query tracker isn't being updated by `diffRelayQuery` when the query is on a node that is currently `null` in the local store. This causes Relay to lose track of what queries the components are expecting and therefore future mutations form incorrect queries.

My current fix just changes a bit of logic in `diffRelayQuery` to always track null linked fields. I'm not sure it's correct though - I'd like to write some code that ensures the `RelayQueryTracker` correctly merges the two tracked queries that `diffRelayQuery` registers via `trackNodeForID`, but maybe that's already covered by existing tests.

### DIFF
--- a/src/traversal/__tests__/diffRelayQuery-test.js
+++ b/src/traversal/__tests__/diffRelayQuery-test.js
@@ -24,7 +24,7 @@ const diffRelayQuery = require('diffRelayQuery');
 describe('diffRelayQuery', () => {
   var RelayRecordStore;
 
-  var {defer, getNode, getVerbatimNode} = RelayTestUtils;
+  var {defer, getNode, getVerbatimNode, writePayload} = RelayTestUtils;
 
   var rootCallMap;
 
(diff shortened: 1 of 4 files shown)


josephsavona:
The general approach here feels correct, but there are other places within `diffRelayQuery` where a null link can be reached, and those should be handled similarly (plural linked fields and null connections). It might be helpful to add tests for these cases to find a common place to add the new logic.

> I'd like to write some code that ensures the RelayQueryTracker correctly merges the two tracked queries that diffRelayQuery registers

Tracked queries aren't merged, so no need to test this.

NevilleS:
I expected as much. I'll take a look later this week and make some changes.

Does a similar issue exist with the RelayQueryWriter, though? It also calls
into the query tracking logic. When I was poking around it seemed like that
wasn't the case (since that code only deals with what comes back from the
server, not what was asked for in the first place), but I could be wrong.

josephsavona:
The writer adds tracked queries for newly created nodes, which by definition are non-null, so i don't believe this issue could occur there.

NevilleS:
OK, thanks for confirming that particular suspicion :smile: 

I think this issue could theoretically affect any field type that can have a sub-selection (and therefore need fragment tracking). I suppose that only rules out scalar fields, so I'll spend some time writing tests that exercise:

- [X] tracking fragments on null linked fields
- [ ] tracking fragments on null connections
- [ ] tracking fragments on null plural fields
- [ ] tracking fragments on null edges (?)

Does that sound about right?

josephsavona:
That sounds right. Note that we don't track queries for records with a "client" id - (ie any record where the server didn't return an `id`) - and edges fall into this category. So we don't need to worry about tracking queries on null edges.

NevilleS:
OK, I added some tests this morning for plural fields & connections, and found that the same issue seemed to be present there (unsurprisingly). I've rebased this PR to the latest master and re-pushed. Let me know what to clean up, change, etc.

josephsavona:
This is a phenomenal PR - awesome job on tests. Thank you!

josephsavona:
@facebook-github-bot import

NevilleS:
No problem. Shall I rebase to the latest master?

josephsavona:
@NevilleS its ok, our tool can rebase on import.

NevilleS:
:open_mouth: We live in the future!

PR #776, opened 2016-01-28
Fix fixture that was missing `canHaveSubselections` metadata.

steveluscher:


steveluscher:
@facebook-github-bot shipit

PR #779, opened 2016-01-30
network layer support for subscriptions

eyston:
This adds support to the network layer for subscriptions.  And by support I mean properly throwing errors -- so not much really!

This builds off of https://github.com/facebook/relay/pull/736 purely for having schema support for subscriptions during tests.  It doesn't depend on the code change to `writeRelayUpdatePayload`.  So really the only diff should be the last commit: https://github.com/facebook/relay/commit/690da4d9e14bfb9a18ed5ebbe153aaccb786708d

I'm going to try adding some sweet, sweet github code comments for areas I have questions with!

### DIFF
--- a/scripts/jest/testschema.graphql
+++ b/scripts/jest/testschema.graphql
@@ -24,6 +24,10 @@ type Mutation {
   viewerNotificationsUpdateAllSeenState(input: UpdateAllSeenStateInput): ViewerNotificationsUpdateAllSeenStateResponsePayload
 }
 
+type Subscription {
+  commentCreate(input: CommentCreateSubscriptionInput): CommentCreateSubscriptionResponsePayload
+}
+
 input ActorSubscribeInput {
   clientMutationId: String
   subscribeeId: ID
@@ -36,6 +40,11 @@ input ApplicationRequestDeleteAllInput {
 
 input CommentCreateInput {
   clientMutationId: String
+  feedbackId: ID
+}
+
+input CommentCreateSubscriptionInput {
+  clientSubscriptionId: String,
   feedbackId: ID
 }
 
@@ -117,6 +126,13 @@ type Comment implements Node {
 
 type CommentCreateResponsePayload {
   clientMutationId: String
+  comment: Comment
+  feedback: Feedback
+  feedbackCommentEdge: CommentsEdge
+}
+
+type CommentCreateSubscriptionResponsePayload {
+  clientSubscriptionId: String
   comment: Comment
   feedback: Feedback
   feedbackCommentEdge: CommentsEdge
--- a/scripts/jest/testschema.graphql
+++ b/scripts/jest/testschema.graphql
@@ -25,7 +25,7 @@ type Mutation {
 }
 
 type Subscription {
-  commentCreate(input: CommentCreateSubscriptionInput): CommentCreateSubscriptionResponsePayload
+  commentCreateSubscribe(input: CommentCreateSubscribeInput): CommentCreateSubscribeResponsePayload
 }
 
 input ActorSubscribeInput {
@@ -43,7 +43,7 @@ input CommentCreateInput {
   feedbackId: ID
 }
 
-input CommentCreateSubscriptionInput {
+input CommentCreateSubscribeInput {
   clientSubscriptionId: String,
   feedbackId: ID
 }
@@ -131,7 +131,7 @@ type CommentCreateResponsePayload {
   feedbackCommentEdge: CommentsEdge
 }
 
-type CommentCreateSubscriptionResponsePayload {
+type CommentCreateSubscribeResponsePayload {
   clientSubscriptionId: String
   comment: Comment
   feedback: Feedback
--- a/scripts/jest/testschema.json
+++ b/scripts/jest/testschema.json
@@ -8122,7 +8122,7 @@
           "description": null,
           "fields": [
             {
-              "name": "commentCreate",
+              "name": "commentCreateSubscribe",
               "description": null,
               "args": [
                 {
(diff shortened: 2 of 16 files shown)


josephsavona:
I have lots of feedback above, but this is a great start. Thanks again for working on this, we're all really excited to see subscription support in OSS :-)

KyleAMathews:
Count me in the "we" :-)

eyston:
I had thought about the subject thing a bit, but ended up with this because the use case I ended up having was a single subscriber per request.  That is how the observer got in the constructor.  I'm happy to move it to an explicit `subscribe` method but have a few questions:

- Does FB end up with multiple subscribers for a single subscription request?  In the issue you had mentioned possibly de-duping subscriptions and I wasn't sure how that would end up happening.

- What are the semantics around `dispose` on the subject?  With a single subscriber I was essentially copying the `observable.create` semantics from RxJS.  The subscription was a cold observable and each 'subscribe' allocated / disposed resources independently based on the subscription -- when a user called `dispose` on the subscription it `disposed` the resources around the subscription.  With a subject it now becomes 'hot' and something has to call `dispose` on the subject (disposing a subscription just unsubscribes, right?).  Do we want the semantics that when observer count goes from X to 0 we dispose the subject resources (those defined by the network layer)?

Here is the subscriber observer implementation I had: https://github.com/eyston/relay/blob/subscriptions/src/subscription/RelaySubscriptionObserver.js

eyston:
Okay, I'm dumb and the link you sent defines the semantics:

- X -> 0 disposes
- 0 -> 1 a second time throws an invariant (e.g. you are trying to subscribe to something that has been disposed)

eyston:
Updated from comments other than the subject stuff.  I'm still slightly confused on what we want.

The `SubscriptionRequest` is a subject.  Should the `sendSubscription` method of an injected network layer view the `request` as:

- an observer
- an observable
- a subject (both observer / observable)

If the answer is subject, that is weird to me... the network layer is the one calling 'onNext', why would it subscribe to itself?

Also confusing to me is that the `sendSubscription` methods job is to initialize a source of events -- right?  e.g. on `sendSubscription` I might allocate resources such as open a socket / long poll an end point / etc.

We don't want to do that work until someone `subscribes` to the subject.  e.g.: https://github.com/facebook/relay/blob/master/src/store/RelayQueryResultObservable.js#L85-L87

What the analogous `observe` / `unobserve` code for `SubscriptionRequest` be:

```js
class RelaySubscriptionRequest {
  constructor(networkLayer) {
    this.networkLayer = networkLayer;
  }

  // only called on observer count 0 -> 1
  observe() {
    this.networkLayerSubscription = this.networkLayer.sendSubscription(this);
  }

  // only called once, when observer count drops to 0
  unobserve() {
    this.networkLayerSubscription.dispose();
  }
}
```

Or is `observe` assumed to have been done externally?

```js
class RelaySubscriptionRequest {
  constructor(networkLayerSubscription) {
    this.networkLayerSubscription = networkLayerSubscription;
  }

  unobserve() {
    this.networkLayerSubscription.dispose();
  }
}
```

Or do i have no idea what I'm talking about :)

eyston:
Updated / merged to be current with master.  I took a guess on the observer stuff and made a billion tests in an attempt to make clear the implementation although it is probably wrong.

I'm stumped on how this model works though.  The implementation I had been using was that the subscription was a cold observable.  When a user starts a subscription to the store it creates the observable and subscribes and hands this subscription back to the user.  When the user calls dispose it cleans up / disposes the subscription (network layer dispose is called).

If instead the request is a subject then it is unclear to me when we would call the observable (network layer) dispose.  We now can have multiple subscribers to the subject but some subscribers really shouldn't hold the subscription open -- they should dispose when the 'user' calls dispose on their subscription handle.  For example if the network layer itself calls subscribe on the request then it would hold that subscription open.

Would you model the initial creation of the subscription as two subscribes: one by the store which writes data, one to the user supplied callbacks?  If the user calls `dispose` on their subscription it would close one / both / all of the subscriptions?

I feel like I lack the big picture.  The picture I had in my mind is not the same one here :)

Globegitter:
@eyston What is the reason for closing this? Your two PRs seems like some really great features that would benefit quite a few people :)

josephsavona:
@Globegitter check out [this thread](https://github.com/facebook/relay/issues/541#issuecomment-213093469) for more context.

Globegitter:
@josephsavona I have seen that, which I understood as, that you (i.e. relay core team are not actively working towards it). But that also means community contributions won't be accepted?

The reason I am pinging on this PR is that we will start adding websocket to one of our apps and I am right now evaluating the best option. If subscriptions are close, then it would of course make most sense to use that. And given these PRs plus https://github.com/eyston/relay-chat-subscriptions it seems the pieces for relay native subscriptions are already here. But if there is no way any of that will make it into core any time soon I will of course use `handleQueryPayload`, `forceFetch`, etc.

eyston:
I wasn't moving it forward and didn't want to discourage other takes on subscriptions.

I haven't looked at this in a bit, but if `handleUpdatePayload` works with subscription queries (the PR for this was small and I'm not sure if still required) then you could do it outside relay core.  The main benefit of `handleUpdatePayload` vs `handleQueryPayload` (and I might have the names slightly off) is you can use mutation configs on update which lets you append / prepend / remove edges.

josephsavona:
> which I understood as, that you (i.e. relay core team are not actively working towards it). But that also means community contributions won't be accepted?

@Globegitter In general we're absolutely open to community contributions. In this particular case, @eyston did a ton of amazing work building subscription support, but there's a lot of complexity here and we simply don't have the bandwidth to review, import, and maintain all of this code. Upon further consideration, we realized that there a number of features such as this - server rendering being another one - that ideally could be built in user space instead of within the framework itself. Subscriptions can be implemented via `handleQueryPayload` (with a RelayQuery.Subscription for mutation-like behavior) or `handleQueryPayload` (with a RelayQuery.Root), so we'd like to see what people come up with in user-space before building something into the framework.

Globegitter:
@eyston thanks for the clarification, that makes sense.

@josephsavona thanks for giving some background and a bot more information on this. That will help quite a bit with our decision on how to implement subscription/subscription-like features in our apps.

PR #781, opened 2016-01-31
Small fix for Flow v0.21.0

gabelevi:
Flow doesn't infer generics yet, so we need to annotate these
functions in order for Flow to understand what's going on.

Without this change, Flow didn't understand the type of `FIELDS`

Test Plan: Ran Flow v0.21.0 and v0.20.1

### DIFF
--- a/scripts/babel-relay-plugin/src/RelayQLPrinter.js
+++ b/scripts/babel-relay-plugin/src/RelayQLPrinter.js
@@ -43,7 +43,7 @@ type PrinterOptions = {
 
 module.exports = function(t: any, options: PrinterOptions): Function {
   const formatFields = options.snakeCase ?
-    fields => {
+    function<T: Object>(fields: T): {[keys: $Keys<T>]: string} {
       const formatted = {};
       Object.keys(fields).forEach(name => {
         formatted[name] =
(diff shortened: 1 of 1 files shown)


gabelevi:
The version of babel that build-lib uses didn't seem to like the generics syntax for arrow functions, so I just switched to using regular anonymous functions. I think it's Babel 5, so it wasn't just a matter of installing the right plugin.

From the travis build for the previous arrow function version:

```
> scripts/build-lib
/home/travis/build/facebook/relay/scripts/babel-relay-plugin/node_modules/babel-core/lib/transformation/file/index.js:671
      throw err;
      ^
SyntaxError: /home/travis/build/facebook/relay/scripts/babel-relay-plugin/src/RelayQLPrinter.js: Unexpected type cast (46:30)
  44 | module.exports = function(t: any, options: PrinterOptions): Function {
  45 |   const formatFields = options.snakeCase ?
> 46 |     <T: Object>(fields: T): {[keys: $Keys<T>]: string} => {
     |                               ^
  47 |       const formatted = {};
  48 |       Object.keys(fields).forEach(name => {
  49 |         formatted[name] =
```

gabelevi:
@facebook-github-bot shipit

PR #798, opened 2016-02-06
Edit README.md to discribe requirement of babel-cli

tkshi:
This example requirement two global module.`babel-core`and `babel-cli`.Although,There is not describe in `README.md`.
So I fixed it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,12 @@ Relay is a JavaScript framework for building data-driven React applications.
 
 The repository comes with an implementation of [TodoMVC](http://todomvc.com/). To try it out:
 
+You must have installed *babel-core* and *babel-cli* with global.If you still,install them follow command.
+```
+npm i -g babel-core
+npm i -g babel-cli
+```
+
 ```
 git clone https://github.com/facebook/relay.git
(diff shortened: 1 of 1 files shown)


yungsters:
We actually need more than just these steps. Some of our build steps require some upgrades and/or a `.babelrc` file in order to be compatible with Babel 6 out of the box.

The safest command to use for now is `npm install -g babel@5.8.35`. But an even better improvement is upgrading everything to support Babel 6 without any global installs.

edvinerikson:
Seems like https://github.com/facebook/relay/pull/813 fixed both the upgrade to Babel 6 and the global dependency requirement.

edvinerikson:
However the examples install docs doesn't work out of the box (https://github.com/facebook/relay/issues/906). I could install all dependencies without any errors if I added `--ignore-scripts` to the install command. Maybe we could change this PR to document that instead?

josephsavona:
Per @edvinerikson's comment it looks like this isn't sufficient for running the examples. Let's continue discussion in #906 and move forward with a new PR when a full solution is identified.

Issue #790, opened 2016-02-04
Remove deprecated `Component` prop to `RelayRenderer`


wincent:
f5dc55da46517d9f3 renames it to `Container` but continues to support `Component` with a deprecation warning. This issue is a reminder to remove the deprecated code once existing internal call sites have been updated. There should be no callers in the open source community yet because `RelayRenderer` is not yet documented.

Eventually `Relay.RootContainer` will go away, leaving only `RelayRenderer`, but that is a separate step and one that will take longer because there _are_ external callers that are using it.

Related: #785

- [x] Sync f5dc55da46517d9f3 to central internal repo
- [x] Update all existing call sites
- [ ] Remove deprecation code path

PR #799, opened 2016-02-06
Remove deprecated `Component` prop from `RelayRenderer`

wincent:
This is the final task required for:

https://github.com/facebook/relay/issues/790


wincent:
@facebook-github-bot import

Issue #784, opened 2016-02-02
Relay can create invalid queries when fields with arguments overlap


leebyron:
Please see the issue originally filed by @fson at https://github.com/graphql/graphql-js/issues/263 where Relay could generate an invalid query:

```
query ($orderBy: _GroupOrdering, $orderBy2: _GroupOrdering) {
  viewer {
    allGroups(orderBy: $orderBy) { count }
    allGroups(orderBy: $orderBy2) { count }
  }
}
```

ryancole:
I've actually been encountering this issue over the past few days. In my particular case, a relay mutation's payload query seems to be where I'm seeing an invalid query format, resulting in the `"Fields \"_scrimInvitesBKmb7\" conflict because they have differing arguments."` error.

In my particular case, it only happens some times and not others. It also only seems to happen when I'm using `this.props.relay.forceFetch`, as well. if I follow a longer user navigation flow, and sign in, navigate to my react view that has the mutation and then execute the mutation, I will see the error. If I perform a hard refresh (clearing the relay store) and run the mutation, then it works every time without the duplicate field in the query.

I'm not sure if `forceFetch` or having data in the relay store already determines the query that gets built by relay, but I wanted to mention these two components of the issue, from my project. For now, just disabling `forceFetch` seems to fix it.

josephsavona:
@ryancole thanks for the extra info, this is super helpful for debugging!

josephsavona:
There are two basic approaches here:
- Flatten the query prior to printing it.
- Print duplicated fields, but in a way that passes validation.

Flattening the query requires a full extra traversal (O(n) operation), so this isn't tenable from a performance perspective. The simpler solution would be to avoid creating new variables when the values are actually the same. This could be achieved by keeping a map in `printRelayQuery` of serialized value -> variable name, and only generating new variables if the current value is not already a key in the map. This logic would go in [`createVariable`](https://github.com/facebook/relay/blob/master/src/traversal/printRelayOSSQuery.js#L298-L310)

@ryancole any interest in submitting a PR for this? ;-)

ryancole:
I'll take a look right now. Ideally I could first come up with some test case that reproduces the issue, so I may focus on that first before focusing on the proposed fix.

josephsavona:
@ryancole Great, thanks! For a test case, take a look at the [test schema](https://github.com/facebook/relay/blob/master/scripts/jest/testschema.graphql) for fields to use in constructing a repro case - something like

```
  query {
[...]

PR #801, opened 2016-02-07
Account for duplicate variables in printRelayQuery

ryancole:
This pull request relates to #784. I've adjusted the `createVariable` function, within `src\traversal\printRelayOSSQuery.js`, so that it first checks the existing variables before creating a new variable. This is to assist Relay when printing queries that contain overlapping arguments. (If the variable type and value is the same, do we need to continue creating variables with the same type and value? This PR adds the logic that answers that with a *no*.)

Without knowing what existing code is at my disposal within all of the FB libraries, I went ahead and just use a simple for loop over the existing keys, since the return state needs the key anyway.

I also created a test case that, prior to adjusting `createVariable`, failed. After adjusting `createVariable` the test passes, and no other existing tests seem to fail. Yay?

### DIFF
--- a/src/traversal/__tests__/printRelayOSSQuery-test.js
+++ b/src/traversal/__tests__/printRelayOSSQuery-test.js
@@ -195,21 +195,20 @@ describe('printRelayOSSQuery', () => {
       const query = getNode(Relay.QL`
         query FooQuery {
           defaultSettings {
-            web: notifications(environment: WEB),
+            web: notifications(environment: WEB)
             foo: notifications(environment: $env)
           }
         }
       `, {
         env: enumValue,
       });
-      const alias1 = query.getChildren()[0].getSerializationKey();
-      const alias2 = query.getChildren()[1].getSerializationKey();
+      const alias = generateRQLFieldAlias('notifications.environment(WEB)');
       const {text, variables} = printRelayOSSQuery(query);
       expect(text).toEqualPrintedQuery(`
         query FooQuery($environment_0:Environment) {
           defaultSettings {
-            ${alias1}: notifications(environment:$environment_0),
-            ${alias2}: notifications(environment:$environment_0)
+            ${alias}: notifications(environment:$environment_0),
+            ${alias}: notifications(environment:$environment_0)
           }
         }
       `);
--- a/src/traversal/__tests__/printRelayOSSQuery-test.js
+++ b/src/traversal/__tests__/printRelayOSSQuery-test.js
@@ -190,6 +190,34 @@ describe('printRelayOSSQuery', () => {
       });
     });
 
+    it('dedupes arguments', () => {
+      const enumValue = 'WEB';
+      const query = getNode(Relay.QL`
+        query FooQuery {
+          defaultSettings {
+            web: notifications(environment: WEB),
+            foo: notifications(environment: $env)
+          }
+        }
+      `, {
+        env: enumValue,
+      });
+      const alias1 = query.getChildren()[0].getSerializationKey();
+      const alias2 = query.getChildren()[1].getSerializationKey();
+      const {text, variables} = printRelayOSSQuery(query);
+      expect(text).toEqualPrintedQuery(`
+        query FooQuery($environment_0:Environment) {
+          defaultSettings {
+            ${alias1}: notifications(environment:$environment_0),
+            ${alias2}: notifications(environment:$environment_0)
+          }
+        }
+      `);
+      expect(variables).toEqual({
+        environment_0: enumValue,
+      });
+    });
+
     it('throws for ref queries', () => {
       const query = RelayQuery.Root.build(
         'RefQueryName',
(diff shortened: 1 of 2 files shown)


src/traversal/__tests__/printRelayOSSQuery-test.js
```diff
@@ -190,6 +190,34 @@ describe('printRelayOSSQuery', () => {
...
+          defaultSettings {
+            web: notifications(environment: WEB),
+            foo: notifications(environment: $env)
+          }
+        }
+      `, {
+        env: enumValue,
+      });
+      const alias1 = query.getChildren()[0].getSerializationKey();
+      const alias2 = query.getChildren()[1].getSerializationKey();
```
josephsavona: The strictest option here is to verify the exact alias used - let's follow the example from line 375 and use `const alias = generateRQLFieldAlias('notifications.environment(WEB)')`, and then `${alias}: ...` for both fields below to make it clear that the alias should be identical.

src/traversal/__tests__/printRelayOSSQuery-test.js
```diff
@@ -190,6 +190,34 @@ describe('printRelayOSSQuery', () => {
...
+      const {text, variables} = printRelayOSSQuery(query);
+      expect(text).toEqualPrintedQuery(`
+        query FooQuery($environment_0:Environment) {
+          defaultSettings {
+            ${alias1}: notifications(environment:$environment_0),
```
josephsavona: nit: commas not required in GraphQL

src/traversal/printRelayOSSQuery.js
```diff
@@ -301,6 +301,14 @@ function createVariable(
...
   printerState: PrinterState
 ): string {
+  const existingIds = Object.keys(printerState.variableMap);
+  for (let i = 0; i < existingIds.length; i++) {
+    const existingId = existingIds[i];
+    const existingVariable = printerState.variableMap[existingId];
+    if (existingVariable.type === type && Object.is(existingVariable.value, value)) {
+      return existingId;
+    }
+  }
```
josephsavona: This will make each call to `printRelayQuery` O(n^2) with the number of variables in the query. At the moment we inline most argument values instead of creating variables, but if we ever changed that the use of a `for` loop here could be prohibitively expensive. 

Instead, how about changing the printer state's `variableMap` to a `Map` (to allow objects as keys since `value` is mixed), and changing it from `{[variableID: string]: {type,value}}` to `Map<value, {type, variableID}>`. Then the logic here can be roughly:

```javascript
const {variableMap} = printerState;
if (variableMap.has(value)) {
  return variableMap.get(value).variableID;
} else {
  const variableID = name + '_' + base62(printerState.variableCount++);
  variableMap.set(value, {type, variableID});
  return variableID;
}
```

The final printed variables can be created by iterating over `variableMap` to produce an object of variableID -> value.

josephsavona:
@ryancole Thanks for working on this! See feedback for how to make this even better :-)

src/traversal/__tests__/printRelayOSSQuery-test.js
ryancole: Ah nice, alright. I had skimmed looking for how to generate those ids, and used the one that jumped out at me.

ryancole:
@josephsavona Alright, I think I've addressed your feedback. The code now uses `Map { value => { type, variableID }}`. Changing it from the previous `Object { variableID: { type, value }}` required a few changes elsewhere, and also removed the usage of a couple imports.

src/traversal/printRelayOSSQuery.js
```diff
@@ -28,11 +26,11 @@ type PrinterState = {
   fragmentNameByText: {[fragmentText: string]: string};
   fragmentTexts: Array<string>;
   variableCount: number;
-  variableMap: {[variableID: string]: Variable};
+  variableMap: Map<mixed, Variable>;
```
ryancole: I went ahead and use the existing Flow type, `Variable`, here. I don't particularly feel like the new contents of this `Variable` object (`type` and `variableID`) are properly described by the name Variable, but I'm not sure if changing this to something like `VariableDetails` or something is necessary. I just left it as-is. I've also never used Flow, so if this is flat out incorrect, I can change it. :)

src/traversal/printRelayOSSQuery.js
```diff
@@ -64,12 +62,13 @@ function printRelayOSSQuery(node: RelayQuery.Node): PrintedQuery {
     queryText,
     'printRelayOSSQuery(): Unsupported node type.'
   );
+  const variables = {};
+  for (const [key, value] of variableMap) {
```
ryancole: The previously used `mapObject` function didn't work on a `Map`, so I'm just recreating the end result with this `for` logic. If you think this would be cleaner being in its own function, I can pull it out into its own function.

ryancole:
Finally, let me know if you'd like me to squash these commits. I don't have a working Flow binary, on Windows, that was compatible with the required version, so one of the commits ended up being to use TravisCI's Flow. :)

josephsavona:
nice, thanks for the update!

josephsavona:
@facebook-github-bot import

PR #808, opened 2016-02-09
Drop an unnecessary line from .flowconfig

wincent:
Noticed this diagnostic message when running `npm run typecheck`:

    node_modules/fbjs/node_modules/promise No such file or directory

The directory didn't exist because I had previously used NPM 3, which
flattens the node_modules directory, and stores the "promise" module at
"node_modules/promise".

So, the line had no effect -- we still typecheck with zero errors
without this line, and if I introduce an error I see it reported
correctly -- which means that Flow has probably gotten smart enough
about promises to not need this special case any more.

### DIFF
--- a/src/.flowconfig
+++ b/src/.flowconfig
@@ -6,7 +6,6 @@
 
 [include]
 ../node_modules/fbjs/lib
-../node_modules/fbjs/node_modules/promise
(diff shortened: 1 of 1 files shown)


wincent:
@facebook-github-bot import

PR #827, opened 2016-02-12
add reploy.io to USERS

jsierles:


steveluscher:
Thanks!

@facebook-github-bot shipit

Issue #830, opened 2016-02-12
Cannot successfully run examples


ishanthukral:
Trying to run the example projects but running into the issues during `npm install`. Node version `5.5.0`, NPM version `3.3.12` running on El Capitan.

```
cd relay/examples/todo && npm install

> babel-relay-plugin@0.7.0-rc.1 prepublish /Users/<path...>/relay/scripts/babel-relay-plugin
> npm run build


> babel-relay-plugin@0.7.0-rc.1 build /Users/<path...>/relay/scripts/babel-relay-plugin
> scripts/build-lib

module.js:341
    throw err;
    ^

Error: Cannot find module 'babel'
    at Function.Module._resolveFilename (module.js:339:15)
    at Function.Module._load (module.js:290:25)
    at Module.require (module.js:367:17)
    at require (internal/module.js:16:19)
    at Object.<anonymous> (/Users/<path...>/relay/scripts/babel-relay-plugin/scripts/build-lib:3:15)
    at Module._compile (module.js:413:34)
    at Object.Module._extensions..js (module.js:422:10)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Function.Module.runMain (module.js:447:10)

npm ERR! Darwin 15.3.0
npm ERR! argv "/Users/<username...>``/.nvm/versions/node/v5.5.0/bin/node" "/Users/<username...>/.nvm/versions/node/v5.5.0/bin/npm" "run" "build"
npm ERR! node v5.5.0
npm ERR! npm  v3.3.12
npm ERR! code ELIFECYCLE
npm ERR! babel-relay-plugin@0.7.0-rc.1 build: `scripts/build-lib`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the babel-relay-plugin@0.7.0-rc.1 build script 'scripts/build-lib'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the babel-relay-plugin package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     scripts/build-lib
npm ERR! You can get their info via:
npm ERR!     npm owner ls babel-relay-plugin
npm ERR! There is likely additional logging output above.


> react-relay@0.6.1 prepublish /Users/<path...>/relay
> node node_modules/fbjs-scripts/node/check-dev-engines.js package.json && npm run build

module.js:341
    throw err;
    ^

[...]

PR #833, opened 2016-02-12
Loosen babel-core dependency in the examples to prevent version contention

steveluscher:
Resolves, among other things, #830.


steveluscher:
@facebook-github-bot shipit

PR #834, opened 2016-02-13
Run linter against babel-relay-plugin

enaqx:


kassens:
@facebook-github-bot import

PR #835, opened 2016-02-13
Add babel package to babel-relay-plugin

enaqx:
It will make `update-schema` and `update-fixtures` work. Currently they produce error on lacking `babel-node`. Also there is version contention in babel-relay-plugin and relay for these packages: 5.8.35 against 5.8.25 respectively.

https://github.com/facebook/relay/blob/master/scripts/babel-relay-plugin/package.json#L23
https://github.com/facebook/relay/blob/master/package.json#L41

### DIFF
--- a/scripts/babel-relay-plugin/package.json
+++ b/scripts/babel-relay-plugin/package.json
@@ -20,6 +20,7 @@
     "lib/"
   ],
   "devDependencies": {
+    "babel": "^5.8.35",
     "babel-core": "^5.8.35",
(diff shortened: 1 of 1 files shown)


edvinerikson:
Thanks for the PR! It seems that babel have been updated to 6.x which causes this to be invalid. However if you change the `babel` dependency to `babel-cli` I think we can merge this. :)

steveluscher:
We have this now (`babel-cli` in the package.json). Thanks for your help!

Issue #832, opened 2016-02-12
Make tests pass using Node 5 / NPM 3


steveluscher:
Right now, if you `npm install` using Node >=5 and NPM >=3 you'll run into a bunch of problems.

Most of these are due to the fact that npm installs dependencies in a flat directory structure, which makes the paths in `unmockedModulePathPatterns` (see `package.json`) no longer match.

Another problem is that the paths in our `.flowconfig` files may no longer match.

There are undoubtedly other problems.

The task is to make `npm test` pass all-green with both NPM 2 & 3, or make it pass exclusively with NPM 3 and update the `engines` config in `package.json` if dual-compatibility is too much of a pain.

raineroviir:
I'll give this a try

raineroviir:
`npm test` will give the following similar error for 96 tests:
`Error: Failed to get mock metadata: /Users/raineroviir/relay/node_modules/core-js/modules/$.global.js`

Okay so I narrowed down what is broken, the `getType` function in `relay/node_modules/jest-cli/src/lib/moduleMocker.js` has a failsafe return null when it cannot define the type.  In our case when our tests start to mock $.global.js, the type is of type 'global' which isn't an option right now so it returns null

steveluscher:
Related: https://github.com/facebook/jest/issues/554

raineroviir:
Okay so based on that issue and also related: https://github.com/facebook/react/issues/5183, the issue is with jest and to solve it you unmock all the deps.  Although if I set the package.json to look like:
```
"unmockedModulePathPatterns": [
      "<rootDir>/node_modules/"
]
```
We get
18 tests failed, 1149 tests passed (1167 total in 96 test suites, run time 61.989s)

cpojer:
I'll publish a new version of `jest-cli@next` today that should resolve the npm3 mocking issues.

kassens:
We've been running node 5 for a while on travis by now.

PR #840, opened 2016-02-15
Make tests pass using Node 5 / NPM 3 #832

raineroviir:


raineroviir:
ugh I'm having some trouble merging all these commits .. I might give up and submit a new pr

kassens:
No worries, when we merge this, it will all be combined into one commit.

raineroviir:
Ok cool thanks!

package.json
```diff
@@ -85,13 +85,7 @@
...
     ],
     "unmockedModulePathPatterns": [
-      "<rootDir>/node_modules/fbjs/node_modules/core-js/",
-      "<rootDir>/node_modules/fbjs-scripts/",
-      "<rootDir>/node_modules/fbjs/node_modules/promise/",
-      "<rootDir>/node_modules/fbjs/lib/(?!(ErrorUtils.js$|fetch.js$|fetchWithRetries.js$))",
-      "<rootDir>/node_modules/react/",
-      "<rootDir>/node_modules/react-dom/",
-      "<rootDir>/node_modules/react-static-container/"
+      "<rootDir>/node_modules/(?!(fbjs/lib/ErrorUtils.js$|fbjs/lib/fetch.js$|fbjs/lib/fetchWithRetries.js$))"
```
kassens: If you keep the deleted lines here and just have 2 patterns and add `4` and `5` to the travis config, do the tests pass in both environments? It'd be nice to work in both if it's just adding the new patterns instead of replacing them.

Is there something that makes it difficult to keep the npm 2 patterns in this file?

raineroviir: The only difficulty is I don't know how to.  Let me see what I can do

kassens:
This is great, thanks for taking the extra step for cross compatibility! Just one more thing

package.json
```diff
@@ -30,7 +30,8 @@
   "dependencies": {
     "babel-runtime": "5.8.24",
     "fbjs": "^0.7.0",
-    "react-static-container": "^1.0.0-alpha.1"
+    "react-static-container": "^1.0.0-alpha.1",
+    "babel-relay-plugin": "0.7.1"
```
kassens: I think the idea is that people would need to add `babel-relay-plugin` to their project with npm 3, so it should remain (just) a peer dependency. I'd love someone with more npm knowledge here.. 

cc @steveluscher

raineroviir: Npm3 will throw an error if it isn't listed under 'dependencies', that's why I put it there

kassens: Just guessing here, but should it maybe be in devDependencies in that case? It's needed to run the unit tests, but if you're using it in your own project, you should add it to your dependencies (to use it in your transform step).

raineroviir: I'm not sure, let me try it out

cpojer:
fyi jest 0.9.0 (without the fb-suffix) will have fixed npm3 support in Jest. I'll try to get it out next week :)

raineroviir:
OK great! I moved the babel-relay-plugin into devDeps and you were right @kassens ! it passed all tests locally so not sure why one of the tests failed here on travis.

package.json
```diff
@@ -92,7 +93,8 @@
       "<rootDir>/node_modules/fbjs/lib/(?!(ErrorUtils.js$|fetch.js$|fetchWithRetries.js$))",
       "<rootDir>/node_modules/react/",
       "<rootDir>/node_modules/react-dom/",
-      "<rootDir>/node_modules/react-static-container/"
+      "<rootDir>/node_modules/react-static-container/",
+      "<rootDir>/node_modules/(?!(fbjs/lib/ErrorUtils.js$|fbjs/lib/fetch.js$|fbjs/lib/fetchWithRetries.js$))"
```
cpojer: you should be able to revert this change once I update to jest 0.9.0-fb3.

package.json
```diff
@@ -40,6 +40,7 @@
     "babel": "^5.8.25",
     "babel-core": "^5.8.25",
     "babel-eslint": "^4.1.1",
+    "babel-relay-plugin": "0.7.1",
```
cpojer: also this one. I'll update the preprocessor in my diff so this shouldn't be needed.

package.json
raineroviir: Ok sure! I'll be awaiting jests 0.9.0-fb3 release

package.json
raineroviir: I found that travis-ci will not green light the PR in npm3 since it no longer looks at peer dependencies, so I threw it in devDeps

cpojer: I just updated the path here: https://github.com/facebook/relay/blob/master/scripts/jest/preprocessor.js#L67 to the one in `scripts` and that worked for me.

raineroviir: Im not able to repro passing tests without having babel-relay-plugin in devdeps or deps.  What do you mean `scripts`?

cpojer: I updates the path in https://github.com/facebook/relay/blob/master/scripts/jest/preprocessor.js#L66 to https://github.com/facebook/relay/blob/master/scripts/babel-relay-plugin/package.json and that made all the tests pass :)

kassens:
Looks like the flow fix didn't quite work. Sorry this is such a journey and thank you for keeping the updates coming!

raineroviir:
Yeah! Its a big project, not easy! We can either ignore the 'base62' node_module or include it under Flow, in the flowconfig which one makes more sense?

raineroviir:
Getting this error on `npm install`
Error: ModuleNotFoundError: Module not found: Error: Cannot resolve 'file' or 'directory' ./Set in /Users/raineroviir/relay/lib.

Seems like ES6 Set isn't getting loaded?

josephsavona:
@raineroviir that's from a commit of mine yesterday, i'm trying to figure out why Set isn't available and will report back here.

josephsavona:
see #881 for the fix to the missing `Set` module

raineroviir:
Thanks Joseph!

kassens:
Awesome, great to see it finally all working :+1: 

@facebook-github-bot import

cpojer:
Yes! This is great.

raineroviir:
Glad to help =)

PR #844, opened 2016-02-16
Make arguments accecpt number|boolean|string|object

iamchenxin:
Please review me.
Template commit for review.
This fix just covert the input `identifyingArgValue`(string|number|object) in  `getDataID` and `putDataID` to a string.
Because GraphQL is already a type system ,so the number is covert to plain string,do not use maker to  distinguish from string.
The object is covert to something like JSON,but with sorted key-value.

At this time , i m a bit confused by Object args, maybe should fix more file to accept this.
at now unit test stop at `RelayQLAST.js:435`.
 [The failed unit test is object arg](https://github.com/iamchenxin/relay/blob/fixGraphQLFragment/src/query/__tests__/RelayFragmentPointer-test.js#L82)
```
query{
        usertemp(tuple:{name:"tom",age:18}){ id }
}
```
worked in raw GraphQL server,but failed in relay. So this should be an additional type check error in RelayQLAST?  
Continue tomorrow.


src/query/RelayFragmentPointer.js
```diff
@@ -66,11 +66,13 @@ class RelayFragmentPointer {
         });
       });
     }
+    let argType = typeof identifyingArgValue;
     invariant(
-      typeof identifyingArgValue === 'string' || identifyingArgValue == null,
+      typeof argType === 'string' || identifyingArgValue == null
```
steveluscher: Since your aim is to support every type of argument, you should simply be able to remove this invariant.

src/query/__tests__/RelayFragmentPointer-test.js
```diff
@@ -15,6 +15,7 @@ require('configureForRelayOSS');
 
 jest.dontMock('RelayFragmentPointer');
 
+const stringifyArg = require('RelayRecordUtil').stringifyArg;
```
steveluscher: Can you use `require('stableStringify')` instead?

src/store/RelayRecordStore.js
```diff
@@ -124,11 +126,12 @@ class RelayRecordStore {
         'cannot be null or undefined.',
         storageKey
       );
-      return identifyingArgValue;
+      return stringifyArg(identifyingArgValue);
```
steveluscher: This section of code is a special case for the `Node` interface: a field that you can use to fetch a record by its global ID. We want to leave this special case intact; you should never be able to use `node(id: …)` with anything other than an `id` argument of type `ID!`. See https://facebook.github.io/relay/graphql/objectidentification.htm for more information.

src/store/RelayRecordStore.js
```diff
@@ -124,11 +126,12 @@ class RelayRecordStore {
...
     }
     if (identifyingArgValue == null) {
       identifyingArgValue = EMPTY;
     }
+    identifyingArgValue=stringifyArg(identifyingArgValue);
```
steveluscher: Since this is basically a hash of an identifying argument and not the value itself, consider:

```js
const identifyingArgHash = stableStringify(identifyingArgValue);
```

src/query/__tests__/RelayFragmentPointer-test.js
```diff
@@ -63,6 +64,40 @@ describe('RelayFragmentPointer', () => {
       );
     });
 
+    it('creates a wrapped fragment pointer with number arg', () => {
```
steveluscher: I suspect that the tests you want to write will live in `RelayQueryRoot-test.js`. Start from `it('returns the identifying argument type', …)` and work outward to assert that non-string identifying arguments are permitted on root fields.

scripts/jest/testschema.graphql
```diff
@@ -8,6 +8,7 @@ type Root {
   nodes(ids: [ID!]): [Node]
   username(name: String!): Actor
   usernames(names: [String!]!): [Actor]
+  usertemp(tuple: tupleUser!): Actor
```
steveluscher: Basically, we need to add support for three new types of identifying argument: **arrays**, **numbers**, and **objects**. How about this for the sample schema:

```graphql
checkinSearchQuery(query: CheckinSearchInput!): CheckinSearchResult
fastestRoute(waypoints: [Waypoint!]!): Route
task(number: Int!): Task
```

iamchenxin:
@steveluscher Thank you,i will fix them all this evening.

iamchenxin:
@steveluscher Need one more day to complete this. 
Find that accept objects need to modify `babel-relay-plugin`.  
Seems at now Relay makes the argument as an array. I think maybe the object-arg would be made as a value for key 'callValue' . 

```
calls: [
   {
	kind: 'Call',
	metadata: {},
	name: 'first',
	value: {
		kind: 'CallValue',
		callValue: 5
	 }
    }, 
],
```

iamchenxin:
@steveluscher Need help , When i use [t.valueToNode](https://github.com/facebook/relay/blob/master/scripts/babel-relay-plugin/src/babelAdapter.js#L35) from `babelAdapter` , Pass a number type to it still return something like `{"type":"Literal","value": ` .  
How could i make it return a numberType. What implement  the `valueToNode` use.

Its my first time to look into a babel plugin,so there are so much things i do not known.
I search the modules find a function `Literal(node)` in babel's `inferers.js`
```
function Literal(node) {
  var value = node.value;
  if (typeof value === "string") return t.stringTypeAnnotation();
  if (typeof value === "number") return t.numberTypeAnnotation();
  if (typeof value === "boolean") return t.booleanTypeAnnotation();
  if (value === null) return t.voidTypeAnnotation();
  if (node.regex) return t.genericTypeAnnotation(t.identifier("RegExp"));
}
```
And i search the [babel-handbook](https://github.com/thejameskyle/babel-handbook/blob/master/translations/en/plugin-handbook.md) ,find some function like `t.numberLiteral`.

Go to sleep,see you tomorrow,thanks.

src/store/RelayRecordUtil.js
```diff
@@ -0,0 +1,79 @@
...
+          'function stringilyArg() only allow string|number|object' +
+          'but shoule not be here' +
+          'the Arg is `%s`,type is `%s`',
+          arg, argType
+        );
+        return '';
+    }
+
+  }
+};
```
josephsavona: how is this different than the existing `stableStringify` function?

src/query/__tests__/RelayFragmentPointer-test.js
josephsavona: +1 - no need for a new module

src/store/RelayRecordUtil.js
iamchenxin: ! Oh! i havent notice this function `stableStringify` .
Its my fault, i thought you let me rename my function to  `stableStringify`  , how stupid am i.

iamchenxin: Need to sleep, will check that function tomorrow.

iamchenxin:
@steveluscher Almost done. and i need some help. did not find the function to let babel generate Array `[]` . So i template write `[]` as key-value [RelayQLPrinter.js#L556](https://github.com/iamchenxin/relay/blob/fixGraphQLFragment/scripts/babel-relay-plugin/src/RelayQLPrinter.js#L556). Need to modify it tomorrow evening.
And i also a bit confused where to test the `getDataID` for new types.

scripts/babel-relay-plugin/src/RelayQLPrinter.js
```diff
@@ -534,12 +534,35 @@ module.exports = function(t: any, options: PrinterOptions): Function {
...
+        });
+      }else {
+        return codify({
+          kind: t.valueToNode('CallValue'),
+          callValue: t.valueToNode(value)
+        });
+      }
     }
 
+    // todo array!
```
steveluscher: Does one of `objectify` or `codify` already do most of this?

josephsavona: Yeah, `objectify` already implements most of `printObject`. Let's just change the core line:

    properties.push(property(key, t.valueToNode(value)));

To check the type of `value` and recurse if it is a nested object/array. Note that any helper functions can be moved out of the class definition to the bottom of the file.

scripts/babel-relay-plugin/src/RelayQLPrinter.js
```diff
@@ -534,12 +534,35 @@ module.exports = function(t: any, options: PrinterOptions): Function {
           value.map(element => this.printArgumentValue(element))
         );
       }
-      return codify({
-        kind: t.valueToNode('CallValue'),
-        callValue: t.valueToNode(value),
-      });
+      if(true||!value.hasOwnProperty("0")){
```
josephsavona: what is `hasOwnProperty("0")` checking?

scripts/babel-relay-plugin/src/RelayQLAST.js
```diff
@@ -381,6 +381,32 @@ class RelayQLArgument {
...
+            return _recurValue(v);
+          });
+        case 'ObjectValue':
+          var ob={};
+          astValue.fields.map(function (field) {
+            ob[field.name.value] = _recurValue(field.value);
+          });
+          return ob;
+      }
     }
```
josephsavona: Let's move this to the bottom of the file to make it clear that it's a helper and doesn't rely on any scope here. also, how about naming it something like `getInputObjectValue(inputObject)` to make it more clear what it's doing?

src/store/RelayRecordStore.js
```diff
@@ -112,10 +113,11 @@ class RelayRecordStore {
   /**
    * Get the data ID associated with a storage key (and optionally an
    * identifying argument value) for a root query.
+   * toDO ?any should be something like number|string|object
    */
   getDataID(
     storageKey: string,
-    identifyingArgValue: ?string
+    identifyingArgValue: ?any
```
josephsavona: Let's use `identifyingArgValue: mixed` here (no need for the `?` optional annotation since mixed includes undefined/null). With `any` Flow will stop doing inference on these values which will reduce the chance of catching bugs.

src/store/RelayRecordStore.js
```diff
@@ -129,6 +131,7 @@ class RelayRecordStore {
     if (identifyingArgValue == null) {
       identifyingArgValue = EMPTY;
     }
+    identifyingArgValue=stableStringify(identifyingArgValue);
```
josephsavona: nit: please try to stay consistent with our code formatting. we prefer whitespace around assignments: `var = fn(value)`

src/store/RelayRecordWriter.js
```diff
@@ -34,6 +34,7 @@ import type {
 import type {RecordState} from 'RelayRecordState';
 const RelayRecordStatusMap = require('RelayRecordStatusMap');
 import type {CacheWriter} from 'RelayTypes';
+const stableStringify= require('stableStringify');
```
josephsavona: nit: we follow a convention that function modules are grouped separately than class/object imports - this should go after the block around lines 39-40

src/store/RelayRecordWriter.js
```diff
@@ -94,7 +95,7 @@ class RelayRecordWriter {
    */
   getDataID(
     storageKey: string,
-    identifyingArgValue: ?string
+    identifyingArgValue: ?any
```
josephsavona: replace `?any` with `mixed`

src/store/RelayRecordWriter.js
```diff
@@ -120,7 +122,7 @@ class RelayRecordWriter {
    */
   putDataID(
     storageKey: string,
-    identifyingArgValue: ?string,
+    identifyingArgValue: ?any,
```
josephsavona: replace `?any` with `mixed`

src/store/RelayRecordWriter.js
```diff
@@ -135,6 +137,7 @@ class RelayRecordWriter {
     if (identifyingArgValue == null) {
       identifyingArgValue = EMPTY;
     }
+    identifyingArgValue=stableStringify(identifyingArgValue);
```
josephsavona: nit: missing whitespace

src/store/stableStringify.js
```diff
@@ -0,0 +1,73 @@
...
+        false,
+        'function stringilyArg() only allow string|number|object' +
+        'but shoule not be here' +
+        'the Arg is `%s`,type is `%s`',
+        arg, argType
+      );
+  }
+}
+
+module.exports = stableStringify;
```
josephsavona: `stableStringify` already exists in the `fbjs` module - there should be no need to create a new one. was there some behavior that it is missing that is necessary here?

src/store/RelayRecordWriter.js
```diff
@@ -108,6 +109,7 @@ class RelayRecordWriter {
...
```
josephsavona: nit: whitespace

src/query/__tests__/RelayQueryRoot-test.js
```diff
@@ -342,6 +342,80 @@ describe('RelayQueryRoot', () => {
...
+        }
+      }
+    `);
+    wayQuery.getConcreteQueryNode().metadata= {
+      identifyingArgName: 'waypoints',
+      identifyingArgType: '[Waypoint!]!' };
+    const nodeIdentifyingArg = wayQuery.getIdentifyingArg();
+
+    expect(nodeIdentifyingArg).toBeDefined();
+    expect(nodeIdentifyingArg.type).toBe('[Waypoint!]!');
```
josephsavona: nit: redundant, it's checked on the next line.

src/query/__tests__/RelayQueryRoot-test.js
```diff
@@ -342,6 +342,80 @@ describe('RelayQueryRoot', () => {
...
+        }
+      }
+    `);
+    nQuery.getConcreteQueryNode().metadata= {
+      identifyingArgName: 'query',
+      identifyingArgType: 'CheckinSearchInput!' };
+    const nodeIdentifyingArg = nQuery.getIdentifyingArg();
+
+    expect(nodeIdentifyingArg).toBeDefined();
+    expect(nodeIdentifyingArg.type).toBe('CheckinSearchInput!');
```
josephsavona: nit: redundant, checked on next line

src/query/__tests__/RelayQueryRoot-test.js
```diff
@@ -342,6 +342,80 @@ describe('RelayQueryRoot', () => {
...
+        }
+      }
+    `);
+    nQuery.getConcreteQueryNode().metadata= {
+      identifyingArgName: 'number',
+      identifyingArgType: 'scalar' };
+    const nodeIdentifyingArg = nQuery.getIdentifyingArg();
+
+    expect(nodeIdentifyingArg).toBeDefined();
+    expect(nodeIdentifyingArg.type).toBe('scalar');
```
josephsavona: checked on next line

src/query/RelayFragmentPointer.js
```diff
@@ -66,11 +66,12 @@ class RelayFragmentPointer {
...
     }
+    let argType = typeof identifyingArgValue;
     invariant(
-      typeof identifyingArgValue === 'string' || identifyingArgValue == null,
+      typeof argType !== 'function' && argType !== 'symbol',
       'RelayFragmentPointer: Value for the argument to `%s` on query `%s` ' +
-      'should be a string, but it was set to `%s`. Check that the value is a ' +
-      'string.',
+      'should be boolean|string|number|object, but it was set to `%s`. ' +
+      'And this should not happen, must be some bugs in code of Relay',
```
josephsavona: I think (hope) that it's more likely that someone will pass a function as an argument (perhaps doing `value: someFn` and forgetting to call the function) than that we'll write a bug here. Let's drop the last line ;-)

iamchenxin:
@josephsavona @steveluscher 
Maybe i should learn jest first, i do unit test whole night by hand, have no idea why `stableStringifyTmp.js` be mocked by some test files.(but when i touched that file,it suddenly can require the real  `stableStringifyTmp.js`).   (temporary rename my function to `stableStringifyTmp` ,later,i should check if it is work right by using `stableStringify`).

Now the only failed test is this:
```
 FAIL  src/traversal/__tests__/printRelayOSSQuery-test.js (4.189s)
● printRelayOSSQuery › roots › it prints object call values
  - query PrintRelayOSSQuery($query_0:CheckinSearchInput!){checkinSearchQuery(query:$query_0){query}}
    to equal
    query PrintRelayOSSQuery($query_0:CheckinSearchInput){checkinSearchQuery(query:$query_0){query}}
```
CheckinSearchInput! vs CheckinSearchInput ( should check this,tomorrow)

And i have not set the webstorm right for eslint, so i will modify the code style later by hand,when i finish them. (i need to search how to get webstorm load the eslint plugins ?).

scripts/babel-relay-plugin/src/RelayQLAST.js
```diff
@@ -752,6 +754,31 @@ function stripMarkerTypes(schemaModifiedType: GraphQLSchemaType): {
...
+    case 'FloatValue':
+      return parseFloat(inputObject.value);
+    case 'StringValue':
+    case 'BooleanValue':
+    case 'EnumValue':
+      return inputObject.value;
+    case 'ListValue':
+      return inputObject.values.map(v=>
+        getInputObjectValue(v)
+      );
```
josephsavona: could simplify with `return inputObject.values.map(getInputObjectValue)`

scripts/babel-relay-plugin/src/RelayQLAST.js
```diff
@@ -752,6 +754,31 @@ function stripMarkerTypes(schemaModifiedType: GraphQLSchemaType): {
...
+    case 'ObjectValue':
+      var ob={};
```
josephsavona: nit: `const`

src/store/RelayRecordWriter.js
```diff
@@ -37,7 +37,7 @@ import type {CacheWriter} from 'RelayTypes';
 
 const invariant = require('invariant');
 const rangeOperationToMetadataKey = require('rangeOperationToMetadataKey');
-
+const stableStringifyTmp = require('stableStringifyTmp');
 const {CURSOR, NODE} = RelayConnectionInterface;
```
josephsavona: nit: newline

src/query/__tests__/RelayQueryRoot-test.js
```diff
@@ -342,6 +342,77 @@ describe('RelayQueryRoot', () => {
...
+    var nQuery = getNode(Relay.QL`
+      query {
+        task(number:5){
+          dumb,
+        }
+      }
+    `);
+    nQuery.getConcreteQueryNode().metadata= {
+      identifyingArgName: 'number',
+      identifyingArgType: 'scalar' };
```
josephsavona: What happens if you don't assign `metadata` here? Will the test fail? If so, this means that this PR won't work in practice - we'll need to ensure that the metadata is set by the plugin.

iamchenxin: @josephsavona nope, that assign `metadata` is something i copyed from a previous exist test. I am not sure what is mean for ...  just because i do not completely know the whole work process of  Relay ,so i copyed some unit test code from some similar exist test.

josephsavona: That's cool, just checking. If the test passes without explicitly setting the metadata (`.metadata = ...`), then let's skip setting it. That way we're testing whether a query as written by a user will produce the correct runtime behavior.

iamchenxin:
@josephsavona Now , i mainly confused by some mechanism of Relay's unit test. I can not get the node-debug for bugger worked for unit test .and also I need to learn lots of things (jest , flow, babel-plugin ~~).
And need some help! the file `stableStringifyTmp` i wrote, cannt be require-actual ,in some unit test file. ( seems in that file `stableStringifyTmp` will be auto mocked, but if i modified  any character in that file,and save the file, suddenly the  real  `stableStringifyTmp` will be used rightly. How could i solve this problem？what things does i miss)

src/query/__tests__/RelayQueryRoot-test.js
iamchenxin: @josephsavona Get it, thanks for your help ,i think i should finish this PR tonight, this PR stimulate me to planning some times to learn the right work process for Javascript. I think if i learned (jest and babel-plugins ) before, this pull request should be much simpler.  ~_~ at night, i should make ESlint work right in webstrom first, then i can automatic do the right style,when inputting characater.

iamchenxin:
@josephsavona Finished . 
Merge my function to `stableStringify`. and extent original `stableStringify` for possible expantions.
But stay the returned string from `stableStringify`  to be a none standard JSON. Cause some unit test files assume the returned string has a raw ` key ` style. See [stableStringify.js#L107](https://github.com/iamchenxin/relay/blob/fixGraphQLFragment/src/query/stableStringify.js#L107)

scripts/jest/testschema.graphql
```diff
@@ -625,3 +627,21 @@ type Settings {
   notificationSounds: Boolean
   notifications(environment: Environment): Boolean
 }
+
+input tupleUser{
+  name: String!
+  age: Int!
+}
```
josephsavona: Unused?

src/query/__tests__/RelayQueryRoot-test.js
```diff
@@ -336,12 +336,77 @@ describe('RelayQueryRoot', () => {
...
+        }
+      }
+    `);
+    const nodeIdentifyingArg = nQuery.getIdentifyingArg();
+    expect(nodeIdentifyingArg).toBeDefined();
+    expect(nodeIdentifyingArg).toEqual({
+      name: 'number',
+      value: 5,
+    });
+  });
```
josephsavona: The toEqual checks means we don't need the toBeDefined check

scripts/jest/testschema.graphql
iamchenxin: oh seems forgot to delete it,yes unused. resolving conflicts now.

src/query/__tests__/RelayQueryRoot-test.js
iamchenxin: okay

src/store/RelayRecordWriter.js
```diff
@@ -94,7 +94,7 @@ class RelayRecordWriter {
    */
   getDataID(
     storageKey: string,
-    identifyingArgValue: ?string
+    identifyingArgValue: any
```
josephsavona: Let's change any to mixed to get type checking here.

josephsavona:
Can you clarify what needed to change in stableStringify?

iamchenxin:
@josephsavona for `stableStringify` the different is in unit test.
There is two change:
1.  string type old `'"test"'` => `'test'` (This is change is required by query arguments)
2.  array `[0:"a",1:"b"]` =>  `["a","b"]`  (This change is not required,but since the old style is only used by one unit test file, and do not affect old src code, so i change this, to make it simpler ,shorter and Standard, and,the stringify string can be parsed back to object by JSON.parse.  ).

src/store/RelayRecordWriter.js
iamchenxin: mixed can not be passed . flow warn me `mixed` can not return to `DataID`.
but i think i should be resolved by `return stableStringify(identifyingArgValue) `.

```
getDataID( storageKey: string, identifyingArgValue: any): ?DataID {
    if (RelayNodeInterface.isNodeRootCall(storageKey)) {
      return identifyingArgValue;
    }
    ......
}
```
but i think i should be resolved by check `string` in function?
```
getDataID( storageKey: string, identifyingArgValue: mixed): ?DataID {
    if (RelayNodeInterface.isNodeRootCall(storageKey)) {
      invariant(identifyingArgValue!=null && typeof identifyingArgValue =='string',"...");
      return stableStringify(identifyingArgValue);
    }
    ......
}
```
But im not sure if its suitable for doing this.

iamchenxin: Or maybe just use a `return Object.prototype.toString.call(identifyingArgValue)`;
i should search if flow has some type cast function.

iamchenxin:
@josephsavona i havent test this in a real relay project yet. I should test it tomorrow to check if it is all right.

src/store/RelayRecordWriter.js
josephsavona: I see what's happening. The type for `identifyingArgValue` has to be `mixed` here because we will now support arbitrary values, but line 106 returns the value directly which conflicts with the string return type...

src/store/RelayRecordWriter.js
```diff
@@ -94,7 +94,7 @@ class RelayRecordWriter {
...
   ): ?DataID {
     if (RelayNodeInterface.isNodeRootCall(storageKey)) {
       invariant(
```
josephsavona: If the root call is `node`/`nodes`, then the storage key *must* be a string or int per the Relay GraphQL specification. We can add an invariant here for this and coerce the return value to a string.

src/store/RelayRecordStore.js
```diff
@@ -112,10 +113,11 @@ class RelayRecordStore {
   /**
    * Get the data ID associated with a storage key (and optionally an
    * identifying argument value) for a root query.
+   * toDO ?any should be something like number|string|object
    */
   getDataID(
     storageKey: string,
-    identifyingArgValue: ?string
+    identifyingArgValue: any
```
josephsavona: Again, please do not use `any` as a type. As I've mentioned on this PR before, this bypasses Flow type checking and reduces our chances of catching edge case bugs just like the one in the previous comment.

src/store/RelayRecordStore.js
```diff
@@ -30,6 +30,7 @@ import type {
   RecordMap,
 } from 'RelayRecord';
 import type {RecordState} from 'RelayRecordState';
+const stableStringify= require('stableStringify');
```
josephsavona: nit: move to line 37

src/store/RelayRecordStore.js
iamchenxin: ok, i modify it to mixed, i check that type inside function.   steveluscher told me to keep `identifyingArgValue` as a plain return value for `root node`.

iamchenxin:
@josephsavona Since Relay  will allow use complex object as Arguments for query later.
Maybe there will be some strange values passed in  by user.
This is one reason i seperated the `type check` and `stringify algorithm` from `stableStringify`. So will be easier to add some relay store type which different from jstype. 
```
function stableStringify(input: mixed): string {
  switch (getStableType(input)) {
    case 'string':
    case 'number':
    case 'boolean':
      return input.toString();
   case 'bigdata':  // like this, for some big data use md5-like instead of json
     return cachedMD5(input);
    default :
      return orderedJSON(input);
  }
}
```
Now i have a thought about `holes-array` [stableStringify-test.js#L55](https://github.com/iamchenxin/relay/blob/fixGraphQLFragment/src/query/__tests__/stableStringify-test.js#L55)
Consider this:
```
      expect(stableStringify(
        [null, undefined, null, 'foo']
      )).toBe('[null,,null,"foo"]');
      
      expect(stableStringify(
        [null, , null, 'foo']
      )).toBe('{0:null,2:null,3:"foo"}');
```
Should we treat `[null, undefined, null, 'foo']` and ` [null, , null, 'foo']` to a same data or not ?

iamchenxin:
@josephsavona @steveluscher I test this fix in real relay project. And find something strange,this is the last point do not support object. [forEachRootCallArg](https://github.com/facebook/relay/commit/425f309e82976b974d291e63b4676ff9d45d4390#diff-41479ecf3670e4e60511fb37625a0e1aR54) 
In this function , i modified the code from old `fn('' + identifyingArgValue)` to fn( identifyingArgValue), but the `identifyingArgValue`object was still automaticly converted to `[object object]` , im not sure why this will happen, its just a callback,

josephsavona:
@iamchenxin thanks for working on this. it looks like it's pretty close. Rather than go back and forth here on stylistic concerns and minor functionality details, i'm going to go ahead and import this and make the last few edits necessary to ship it.

@facebook-github-bot import

iamchenxin:
@josephsavona Its okay! I find it is some bugger from chrome's debug mode ? maybe from some cached code? when i flash the `script` many times , the ` fn( identifyingArgValue)` suddenly work well. and do not call toString() automaticly.

iamchenxin:
@josephsavona 
Maybe still have some hidden problem .. `forEachRootCallArg` with `mixed` type the flow result is (i will check them tomorrow as soon as i have my free time)
```
src/interface/RelayOSSNodeInterface.js:96
 96:         results.push({
             ^ call of method `push`
 99:           rootCallInfo: {storageKey, identifyingArgValue},
                                          ^^^^^^^^^^^^^^^^^^^ mixed. This type is incompatible with
 32:   identifyingArgValue: ?string;
                             ^^^^^^ string

src/store/RelayDiskCacheReader.js:166
166:         forEachRootCallArg(query, identifyingArgValue => {
             ^ function call
171:           this.visitRoot(storageKey, identifyingArgValue, query);
                                          ^^^^^^^^^^^^^^^^^^^ mixed. This type is incompatible with
207:     identifyingArgValue: string,
                              ^^^^^^ string

```

josephsavona:
@iamchenxin That's cool, i can look into it.

iamchenxin:
@josephsavona Seems , it is finally finished , i test this in some real project. But still need to write more new unit tests for this.I should add the unit test later ,after i learn jest .(at now i have some difficulties to write complex unit test).

iamchenxin:
@josephsavona Rebased to master,But when i test with the new master.
even in the master branch , i get this error too.
```
src/query/buildRQL.js:25
 25: const Set = require('Set');
Set. Required module not found
Found 1 error
```

kassens:
@iamchenxin That should be fixed with #881. Try to rebase on master again.

iamchenxin:
@kassens thank you , rebased

iamchenxin:
@josephsavona should this need reimport? it rebased after import.

josephsavona:
Yup, i'll rebase internally.

josephsavona:
@facebook-github-bot import

josephsavona:
@iamchenxin Thanks so much for working on this. When I imported and reviewed the code, I realized that all of the call sites of `forEachRootCallArg` were actually doing the same logic to convert the value to a key. Normally I'd make edits on the original PR, but in this case the changes ended up being quite extensive. I've created two new PRs that build on your implementation in #894 and #895, so I hope you don't mind if I close this PR and continue discussion there. Thanks again for doing so much on this!

iamchenxin:
@josephsavona Its okay. Hope i did not add some hidden bugs  in ,  and will keep a watchful eye on them.

PR #846, opened 2016-02-17
Update Relay to jest 0.9.

cpojer:
The babel-relay-plugin is failing right now but it seems unrelated to the Jest update.


kassens:
@facebook-github-bot import

PR #857, opened 2016-02-20
Update Guides-Routes.md

nodkz:
Avoid unnecessary misunderstanding:  word `container` for target render DOM element is not the best name. Cause term `container` in Relay used for another thing.


josephsavona:
@facebook-github-bot shipit

PR #867, opened 2016-02-22
Use 'babel-polyfill' instead of 'babel/polyfill' in examples

chentsulin:
use `babel-polyfill` which is listed in `package.json`.


kassens:
Thanks, looks like this is also the recommended way on https://babeljs.io/docs/usage/polyfill/

@facebook-github-bot shipit

PR #872, opened 2016-02-23
fix corrupted import for babel polyfill

gouroujo:


kassens:
Thanks for sending the pull request. I'm closing this as I'm already merging #867 which has the same fix.

PR #879, opened 2016-02-24
Fix RelayContext mock for strict class transforms

kassens:
With strict class transforms the methods are not enumarable, so I'm changing
this to an explicit list.


kassens:
@facebook-github-bot import

PR #882, opened 2016-02-25
Fix most warnings in jest tests

kassens:
For the most part this adds missing `__typename` fields in the payloads which
triggered a warning and spammed full test runs.


kassens:
@facebook-github-bot import

PR #889, opened 2016-03-01
Make babelAdapter compatible with Babel 6.6

### LINKED ISSUE #887: Babel Relay Plugin breaks with babel-core v6.6.0
Currently `babel-relay-plugin` checks for the existence of the `Plugin` field on the babel object passed to the plugin, and as of version 6.6.0, it is now a function which throws the error `Error: The (relay-query) Babel 5 plugin is being run with Babel 6.`. The Babel Relay plugin detects this function and runs in Babel 5 mode, which means this gets called and the whole thing bombs out. Or, at least that's my understanding after a quick glance at the code.

kassens:
Babel 6.6 added `Plugin` back so it's truthy now, but it just throws a warning.
Changing to a simple regex to check the version.

Check in Babel was added with babel/babel@6425c475374972d3f11b9340a08a37d171714b25

Fixes #887.


alexhawkins:
Please merge this pull request, currently having to manually make this change. The above fix worked.

chirag04:
works for me

schickling:
Maybe it would make sense to create a separate repository for the `babel-relay-plugin` in order to make releases more flexible?

mrjackdavis:
Yes, this is frustrating. Not sure if facebook would separate this one though. They like their mono-repositories

josephsavona:
> Maybe it would make sense to create a separate repository for the babel-relay-plugin in order to make releases more flexible?

Intuitively this makes sense, but in practice having the plugin in the same repository doesn't impact our ability to release, while it does make iteration on Relay core simpler (most updates to the plugin correspond with changes in Relay core).

mrjackdavis:
When are we likely to see a release?

PR #573, opened 2015-11-09
Implement "introduce ship" in relay compontent

bfwg:
Hi, 
I implemented the mutation on app side, now the example not only can add ship from `POST` requests but also from the app its self.

PR #892, opened 2016-03-01
Implement `AddShipMutation` for Star Wars Example

yungsters:
I'm simply picking up where #573 left off.


yungsters:
@facebook-github-bot shipit

Issue #674, opened 2015-12-10
Allow disabling Relay container shouldComponentUpdate?


taion:
I have a few use cases where I have components that receive data via React `context`. The current implementation of `shouldComponentUpdate` on Relay containers blocks updates to those components via `context`.

Would it make sense to allow disabling the optimized `shouldComponentUpdate` as a configuration option for `Relay.createContainer`? Right now I have to do something a bit ugly to get the behavior I want.

devknoll:
This seems pretty reasonable. You should open a PR :smile:

taion:
I wanted someone to tell me if it seemed reasonable :p

Given https://github.com/facebook/react/issues/2517#issuecomment-169838716, it looks like this will at most just be a temporary workaround – but it might be useful for some months.

peterhorne:
I'd like to pick this up if no one else is currently working on it?

I've had a quick stab at it here: https://github.com/peterhorne/relay/commit/3e437dfb7e5ae2029686c56247c041b8ccfa0945. If that looks alright please let me know and I shall add tests/docs and submit a PR.

taion:
I never started working on this.

Your code is roughly what I had in mind. For consistency with e.g. React-Redux containers, it might make sense to make the setting be something like `pure: false`, though.

peterhorne:
I've changed the setting name to `pure` as suggested and have submitted a PR https://github.com/facebook/relay/pull/897

taion:
This is inconvenient timing, but @gaearon has convinced me that `shouldComponentUpdate` blocking context updates is something that should be addressed by the library using context, not the library providing the container with `shouldComponentUpdate`: https://github.com/reactjs/react-router/issues/470.

Specifically in the context of React Router, we're working on making our context use bypass intervening SCU checks by default, and extracting the utilities into a separate library for use by other context-using libraries: https://github.com/reactjs/react-router/pull/3430/files#diff-933e86e8d3d465b7c3d503e664607d59.

It's possible, then, that this is a bit of an anti-feature now, since I'm not aware of any other good reason to disable the Relay container pure render optimizations other than to allow context updates to propagate.

josephsavona:
@taion thanks for the follow-up. We'd be happy to remove the sCU option, we'll probably wait until `react-router` and other popular libraries using context have updated with the approach you described.

taion:
[...]

PR #897, opened 2016-03-03
Allow disabling Relay container shouldComponentUpdate

peterhorne:
This PR introduces an optional flag `pure` to `RelayContainer` which overrides `shouldComponentUpdate` to return true.

Closes https://github.com/facebook/relay/issues/674.

### DIFF
--- a/src/container/RelayContainer.js
+++ b/src/container/RelayContainer.js
@@ -54,6 +54,7 @@ type FragmentPointer = {
   dataIDs: DataID | Array<DataID>
 };
 export type RelayContainerSpec = {
+  pure?: boolean;
   initialVariables?: Variables;
   prepareVariables?: (
     prevVariables: Variables,
@@ -103,6 +104,7 @@ function createContainerComponent(
   const fragmentNames = Object.keys(fragments);
   const initialVariables = spec.initialVariables || {};
   const prepareVariables = spec.prepareVariables;
+  const pure = spec.pure !== false;
 
   class RelayContainer extends React.Component {
     mounted: boolean;
@@ -691,6 +693,10 @@ function createContainerComponent(
       nextState: any,
       nextContext: any
     ): boolean {
+      if (!pure) {
+        return true;
+      }
+
       // Flag indicating that query data changed since previous render.
       if (this._hasStaleQueryData) {
         this._hasStaleQueryData = false;
--- a/docs/APIReference-Container.md
+++ b/docs/APIReference-Container.md
@@ -37,6 +37,12 @@ Relay containers are created using `Relay.createContainer`.
       A method to modify the variables based on the runtime environment or previous variable values.
     </a>
   </li>
+  <li>
+    <a href="#pure">
+      <pre>pure</pre>
+      A boolean for disabling shouldComponentUpdate.
+    </a>
+  </li>
 </ul>
 
 *Properties and Methods*
@@ -169,6 +175,23 @@ module.exports = Relay.createContainer(ProfilePicture, {
       size: prevVariables.size * window.devicePixelRatio,
     };
   },
+  // ...
+});
+```
+
+### pure
+
+```
+pure: ?boolean
+```
+
+Relay containers may block updates to components that receive data via context. Setting pure to `false` will override the container's `shouldComponentUpdate` to always return true.
+
+#### Example
+
+```{2}
+module.exports = Relay.createContainer(ProfilePicture, {
+  pure: false,
   // ...
 });
 ```
--- a/src/container/__tests__/RelayContainer-test.js
+++ b/src/container/__tests__/RelayContainer-test.js
@@ -864,4 +864,45 @@ describe('RelayContainer', function() {
     );
     expect(render.mock.calls.length).toBe(3);
   });
+
+  it('applies `pure` properly', () => {
+    var mockDataSet = {
+      '42': {__dataID__: '42', name: 'Tim'},
+    };
+    var render = jest.genMockFunction().mockImplementation(() => <div />);
+    var shouldComponentUpdate = jest.genMockFunction();
+    shouldComponentUpdate.mockReturnValue(true);
+
+    var MockComponent = React.createClass({render, shouldComponentUpdate});
+
+    var MockImpureContainer = Relay.createContainer(MockComponent, {
+      pure: false,
+      fragments: {
+        foo: jest.genMockFunction().mockImplementation(
+          () => Relay.QL`fragment on Node{id,name}`
+        ),
+      },
+    });
+
+    GraphQLStoreQueryResolver.mockResolveImplementation(0, (_, dataID) => {
+      return mockDataSet[dataID];
+    });
+    mockFooFragment =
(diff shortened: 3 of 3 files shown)


src/container/RelayContainer.js
```diff
@@ -106,6 +107,7 @@ function createContainerComponent(
   var fragmentNames = Object.keys(fragments);
   var initialVariables = spec.initialVariables || {};
   var prepareVariables = spec.prepareVariables;
+  var pure = spec.pure !== false;
```
josephsavona: Let's be more specific:

    var pure = spec.pure === true;

peterhorne: This is written as a double negative to handle the case when `spec.pure` is `undefined` and it should default to `true`.

josephsavona: yup, realized that after - makes sense

wincent:
Instead of making this configurable, could we instead make our `this.context` vs `nextContext` check a little more sophisticated?

At the moment, we short-circuit with `true` as soon as we see Relay-specific context has changed. But if you look at what we do for non-query props (ie. non-Relay props) we also return `true` if any of them don't match. I wonder if we could/should do the same with context.

peterhorne:
That would be ideal but I don't think it's possible because you can't introspect all possible context items. Perhaps I've overlooked some mechanism for doing so?

wincent:
> Perhaps I've overlooked some mechanism for doing so?

I guess by default you're only going to get the `contextTypes` that you registered for, and the only way around it would be to reach into internal React book-keeping data structures, which probably isn't going to happen...

peterhorne:
Is there anything outstanding preventing this from being merged?

wincent:
> Is there anything outstanding preventing this from being merged?

I'd like to leave this one up for a bit to give people a chance to comment on it before we proceed. Three reasons why I'm inclined to be pretty conservative about this:

- It increases the (currently tiny) API surface area of Relay, something that we want to keep small (partly because it's much harder to revise or remove an API once it's out in the wild).
- Naming is hard: others might have opinions on `pure` as a name (my concerns: it is a very generic dictionary word, and it has substantial conceptual overlap with stateless functional components in React, yet saying `pure: false` doesn't necessarily mean that a component is impure in the the technical sense).
- That unresolved React thread (https://github.com/facebook/react/issues/2517) makes me wary that things may change in React context or alternative APIs/semantics may be introduced into React, and these in turn could make this new API at best temporary and at worst a liability for us.

So, let's gather a bit more input first.

peterhorne:
Okay, thanks for the update :smiley:

The option name was originally `disableShouldComponentUpdate: true` and I changed it to `pure: false` following a suggestion on the original issue. On reflection I think the original name is better due to its explicitness.

KyleAMathews:
I don't see the point of this tbh. If you really want this, just wrap your Relay component with https://github.com/reactjs/react-static-container

peterhorne:
From react-static-container readme:

> This component should be used when you know that a subtree of components will never need to be updated.

However, this PR introduces the ability to override Relay and *enable* updating of the subtree once again.

josephsavona:
I agree that this is a valid use-case, but it also seems to be uncommon enough that it would be preferable to avoid increasing the API surface area, as @wincent mentioned. 

Rather than add more API/options, I think we'll make more progress as a community if we focus on low-level primitives that can be composed together. What if it was so easy to build your own Relay Container that you could just do that and write your own `shouldComponentUpdate`?

KyleAMathews:
That's exactly what react-static-container is designed for. It doesn't introduce any new ReactElements so you can add or remove it without triggering any DOM changes e.g.

```js
if (SHOULD_DISABLE_UPDATES) {
  // Wrap Relay component.
  return <StaticContainer>{relayComponent}</StaticContainer>
} else {
  return relayComponent
}
```

josephsavona:
I should clarify that i'm referring to #559.

josephsavona:
@KyleAMathews the problem is that Relay will currently act as a static container and block updates, when the OP would prefer that it doesn't (bc data has changed on context).

KyleAMathews:
@josephsavona ah :) didn't read the OP very closely I guess.

taion:
Hi, sorry. This is partially my fault, but I forgot to subscribe to this PR.

I suggested `pure: false` because it mirrors the syntax for React Redux. It might not be the best name in a vacuum, but for better or for worse @gaearon has already staked out ground there, and I think it'd be less confusing overall to use a consistent name.

This is useful in the case where I'm using forms that communicate data via `context`, and I need to cause components to re-render based on contextual updates from the forms that would otherwise be blocked by the Relay containers.

The current userspace implementation is a bit gross:

```js
import Relay from 'react-relay';

function shouldComponentUpdate() {
  return true;
}

export default function createFormInputContainer(...args) {
  const Container = Relay.createContainer(...args);

  function FormInputContainer(props, context) {
    const container = new Container(props, context);

    // Always update; this will be nested in a contained form anyway, and we
    // need to not block updates from the form on context.
    container.shouldComponentUpdate = shouldComponentUpdate;

    return container;
  }

  Object.keys(Container).forEach(key => {
    FormInputContainer[key] = Container[key];
  });

  return FormInputContainer;
}
```

This dance is necessarily of the lazy class initialization in `Relay.createContainer`.

josephsavona:
I'm not crazy about increasing the surface area of RelayContainer, but the workaround (as documented by @taion) is not at all obvious. Let's move forward with this.

josephsavona:
@facebook-github-bot import

taion:
We actually just hit the opposite problem, BTW, in that there are a few places where we wish we could configure the Relay container would do shallow equality checks on non-scalar props the way the Redux container does. Probably too much to ask that this part of the code have yet another branch, though. 😉

chirag04:
It would be nice to have redux style shallow equality checks for non-scalars props. In react-native all my `RelayContainer` get a `navigator` object, Relay's SCU is not very helpful there.

josephsavona:
@chirag04 Relay's `shouldComponentUpdate` is intended as a safe default. If you have complex objects (such that Relay's SCU always returns true), we recommend implementing `shouldComponentUpdate` in your component.

taion:
Is there no downside to putting the SCU check both in the Relay container and in the contained component, though?

This came up in the context of the example F8 app, which uses a number of no-op Redux containers to make components be pure, like https://github.com/fbsamples/f8app/blob/b5df451259897d1838933f01ad4596784325c2ad/js/tabs/schedule/SharingSettingsModal.js#L91.

josephsavona:
> Is there no downside to putting the SCU check both in the Relay container and in the contained component, though?

The render logic of `RelayContainer` is very minimal - it basically just renders the component. Are you seeing a case where 2 SCUs is that much slower than one (compared to actually rendering the subtree)?

taion:
Nope – I'm good for now, thanks!

taion:
Out of curiosity, per my line note at https://github.com/facebook/relay/commit/76477bba6c3e4d61ef8cefdc8bfc4cb1e528503d#commitcomment-17315176, what's the benefit of not allowing an actual `shouldComponentUpdate` implementation here? It looks like the API otherwise allows doing so.

PR #898, opened 2016-03-03
Reset store functionality

### LINKED ISSUE #233: Implement RelayStore.reset()

Globegitter:
Fixes #233.

Tested it in one of our apps and seems to do the trick.

Since this is my first PR I am not so sure about all the conventions but tried to follow it as much as possible for unit tests, etc.


src/store/__tests__/RelayContext-test.js
```diff
@@ -70,6 +77,16 @@ describe('RelayContext', () => {
     });
   });
 
+  describe('reset', () => {
+    fit('invokes its invariant', () => {
```
josephsavona: s/fit/it/

josephsavona:
Thanks for working on this. Eventually we won't need a reset method because you will just create a new RelayContext instead. I'm not sure if we want to add an API that we'll delete so soon, but in this case there is enough demand that it might be best to proceed. Let's get some feedback here before proceeding. 

Cc @wincent

schickling:
Would love to use this! :+1:

wincent:
This is great!

_But_ as @josephsavona points out, won't be needed once everything is isolated inside `RelayContext`. The concern that we'll soon delete the `reset()` API is valid, but I actually think there's a deeper reason that even having the API in the first place is a bit wrongful. Being able to reset to a clean state via mutating is a fundamentally brittle design (even if it's right now, what if somebody ends up adding additional global state outside the `RelayStoreData`?) compared to literally replacing the entire context. So, I think we should hold off on this for now and if you need something like this you can run with a patch on top of base Relay. Thoughts?

josephsavona:
> So, I think we should hold off on this for now and if you need something like this you can run with a patch on top of base Relay.

I'm inclined to agree. We're making good progress on #558 and expect to have `RelayContext` be public within the next few releases.

Globegitter:
Felt like dipping a little bit into relay and this one seemed like a good, simple start. But that is totally fair and good to see things are progressing.

josephsavona:
@Globegitter Thanks for opening this PR - it helps us see what features people are most interested in and moves the discussion forward (even if the PR itself doesn't land). Definitely check out the [help-wanted tag](https://github.com/facebook/relay/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22) for other places to contribute!

Closing per discussion above - there are pending diffs for all the remaining items on #558.

tlvenn:
>Eventually we won't need a reset method because you will just create a new RelayContext instead.

Now that #558 has landed, can we expose RelayEnvironment to the public API (`Relay.Environment`) so we can instantiate it ?

Issue #166, opened 2015-08-25
How to get full mutation response?


vslinko:
I have `CreateSessionMutation` with `sid` field defined in `getMutation` and `getFatQuery`.

```js
export default class CreateSessionMutation extends Relay.Mutation {
  getMutation() {
    return Relay.QL`mutation {createSession { session { sid } }}`
  }
  getVariables() {
    return {
      email: this.props.email,
      password: this.props.password
    }
  }
  getFatQuery() {
    return Relay.QL`
      fragment on CreateSessionPayload {
        session {
          sid
        }
      }
    `
  }
  getConfigs() {
    return []
  }
  getCollisionKey() {
    return `createSession`
  }
}
```

I want to call that mutation inside flux, get id of created session, store that id inside localStorage.
`Relay.update(mutation, {onSuccess})` returns only `{createSession:{clientMutationId:"0"}}` without `sid`.
How to get created `sid` from flux?

josephsavona:
Thanks for filing an issue! There's a couple steps to get this mutation working:
- `getConfigs` should be configured to handle a `FIELD_CHANGE`  - see the example at https://facebook.github.io/relay/docs/guides-mutations.html#a-complete-example
- To get the results of a field in the fat query, at least one component should be querying for that data. In this case, the component that triggers the mutation should query for `... { session { sid } }`. When the mutation is complete, the value of that data will be the new `sid`. 

Here's a rough example:

```javascript
class LoginComponent extends React.Component {
  ...
  createSession() {
    Relay.update(new CreateSessionMutation(...), {
      onSuccess: () => {
        // after the mutation completes the `sid` will be populated
        console.log(this.props.user.session.sid);
      }
    });
  }
}
Relay.createContainer(LoginComponent, {
  fragments: {
    user: () => Relay.QL`
      # query for session{sid} which is initially undefined
      fragment on YourTypeWithSession {
        session {
          sid
        }
      }
    `
  }
});
```

yuzhi:
If what you are adding is an completely new node in the system, where you have not had a chance to fetch any fields to yet, `FIELDS_CHANGE` might not see any fields that it needs to update. If that's the case, you can use `REQUIRED_CHILDREN` with a query to fetch any field you want on that might not have been fetched yet. Since no view needs this information, you can use the `onSuccess` callback with the mutation to do wtv you want with the payload.

vslinko:
@yuzhi Thank you! It works! But why `REQUIRED_CHILDREN` isn't documented?

[...]

PR #900, opened 2016-03-04
Document `REQUIRED_CHILDREN`

wincent:
I made a somewhat artificial example up for this. If anybody wants to suggest a more realistic use case derived from actual practice, please chime in!

Note that I also mentioned `@relay(pattern: true)` to give the technique a little more exposure. Maybe dragging connections into this is complicating the issue unnecessarily...

Also note that I explicitly avoided the "sign-in mutation" example from #166 because until we have `RelayContext` fully rolled out, I don't think we want to encourage the use of session handling too overtly.

Closes: https://github.com/facebook/relay/issues/166

### DIFF
--- a/docs/Guides-Mutations.md
+++ b/docs/Guides-Mutations.md
@@ -431,6 +431,56 @@ class RemoveTagMutation extends Relay.Mutation {
 }
 ```
 
+### `REQUIRED_CHILDREN`
+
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
In case it helps, my use case was a mutation that created a new node. I wanted to get a reference to the new node's `id` so that I could immediately `router.push` somewhere based on it (instead of just letting the UI re-render).

wincent:
@facebook-github-bot import

PR #901, opened 2016-03-04
Make plugin build checker independent of Babel version

kassens:
Currently, the build checker test is tied to the exact output of the transform.
This means if people run different minor versions of babel, the test fails.

This change makes the test more resilient. Everytime the build is run, the
`SRC_HASH` is updated. The test hashes the source files again and makes sure the
hash matches. This makes sure we ran the build after modifying some sources.


kassens:
@facebook-github-bot import

PR #909, opened 2016-03-04
Test the version of the specific Babel that was passed into the babel…

steveluscher:
Previously, we were testing that `require('babel-core').version` started with `6.`. This could lead to unexpected results, since the return value of that require might or might not be the version of Babel you're using.


steveluscher:
@facebook-github-bot import

PR #913, opened 2016-03-04
build-lib script is not generated

kassens:
Breaks up the `@generated` string so the file is not incorrectly classified as generated.


kassens:
@facebook-github-bot import

PR #914, opened 2016-03-05
Update plugin to Babel 6

kassens:
Update the dependencies to the latest Babel version using Facebook's set of plugins from fbjs-scripts since Babel no longer has default plugins.


kassens:
@facebook-github-bot import

Issue #896, opened 2016-03-03
Support interpolation of route-conditional fragments that return an array of fragment references


steveluscher:
Presently there is a limitation in Relay that requires that any interpolation you perform in a `Relay.QL` block must evaluate to:

* a fragment reference

  `${Foo.getFragment('viewer')}`

* an array of fragment references

  `${COMPONENTS.map(c => c.getFragment('viewer'))}`

* a route-conditional function that returns *exactly one* fragment reference

  `${route => COMPONENTS[route].getFragment('viewer')}`

The task at hand is to enable the interpolation of **route-conditional functions that return an array of fragment references**.

```js
${route => COMPONENTS.map(c => 
  c.getFragment('treasure').if(variables[`${route}Unlocked`])
)}
```

steveluscher:
Origin issue: http://stackoverflow.com/a/35471007/802047

NevilleS:
This is the same issue too: #775

shaimo:
I have a similar use case. Can someone please elaborate on that 'route-conditional function'? Where is that 'route' parameter coming from? Thanks

steveluscher:
If you interpolate a function into your query:

```js
viewer: () => RelayQL`
  fragment on Viewer {
    ${route => ...}
  }
```

That function will be receive the `routeName` of the `Relay.Route` you used with `Relay.RootContainer` as its first argument, and will expect that you return a single fragment reference (eg. from `Container.getFragment('viewer')`).

shaimo:
@steveluscher Thanks. I guess it won't work for me. I'm using react-router so the given route seems like something that was generated by it. I actually need to choose the component based on a parameter (called 'hash' in my case), but I don't see any way to do it...

steveluscher:
Closed by #915.

taion:
Apologies if I'm missing something here, but with the

```js
${route => COMPONENTS.map(c => 
  c.getFragment('treasure').if(variables[`${route}Unlocked`])
)}
```

syntax – is it the case that you _must_ define this function inline to the fragment if you want to have access to `variables`? In other words, I couldn't have a separate:

```js
const conditionalFragment = route => (
  COMPONENTS[route.name].getFragment('treasure', variables)
);
```

josephsavona:
@taion You can define the function wherever you want, as long as you make sure that `variables` are in scope.

taion:
[...]

PR #915, opened 2016-03-06
Support route conditional fragments returning an array

xuorig:
Gave #896 a try.

modifies `createNode` and `getChildren` to possibly return an array of nodes when a route conditional fragment reference returns an array.

Let me know what you think, It's seems weird to have `createNode` possibly return an Array but I'm not sure how it could be done without it!

### DIFF
--- a/src/query/RelayQuery.js
+++ b/src/query/RelayQuery.js
@@ -214,14 +214,25 @@ class RelayQueryNode {
           if (concreteChild == null) {
             return;
           }
-          const node = createNode(
+
+          let nodes = createNode(
             concreteChild,
             this.__route__,
             this.__variables__
           );
-          if (node && node.isIncluded()) {
-            nextChildren.push(node);
+
+          // Route conditional functions can return an array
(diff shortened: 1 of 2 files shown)


src/query/RelayQuery.js
```diff
@@ -211,14 +211,20 @@ class RelayQueryNode {
...
           );
-          if (node && node.isIncluded()) {
-            nextChildren.push(node);
-          }
+
+          nodes.forEach(node => {
+            if (node && node.isIncluded()) {
+              nextChildren.push(node);
+            }
+          });
```
josephsavona: I suspect that `[].concat(...)` will be slower than a manual check, especially since most cases will return a single value not an array. Let's explicitly check for array/scalar similar to what you did below.

xuorig:
Switched to manual check!

wincent:
Sorry for leaving this one unattended for so long, @xuorig. Going to pull it in and see if it applies, tests pass etc.

PR #918, opened 2016-03-06
Update Jest to 0.9.0.

cpojer:
This puts Jest back from the pre-release track on to `0.9.x`. Apparently npm doesn't automatically go from `0.9.0-fb3` to above or equal `0.9.0`.


kassens:
@facebook-github-bot shipit

PR #921, opened 2016-03-08
Make RelayRenderer isomorphic

denvned:
The first commit of this PR does refactoring that simplifies state handling, but does not change behaviour. And the second, very simple one, actually makes `RelayRenderer` isomorphic.

Now it is possible to pass an `initialReadyState` to `RelayRenderer`, allowing server-side rendering and initial rendering in the browser to use pre-primed data.

See #589

### DIFF
--- a/src/container/RelayRenderer.js
+++ b/src/container/RelayRenderer.js
@@ -313,21 +313,23 @@ class RelayRenderer extends React.Component {
    */
   _calculateRenderArgs(): RelayRendererRenderArgs {
     const {readyState} = this.state;
-    return readyState ?
-      {
+    if (readyState) {
+      return {
         done: readyState.done,
         error: readyState.error,
         props: readyState.ready ? this._resolveContainerProps() : null,
         retry: this._retry.bind(this),
         stale: readyState.stale,
-      } :
-      {
+      };
+    } else {
+      return {
         done: false,
         error: null,
         props: null,
         retry: this._retry.bind(this),
         stale: false,
       };
+    }
   }
 
   render(): ?React.Element {
--- a/src/container/RelayRenderer.js
+++ b/src/container/RelayRenderer.js
@@ -37,6 +37,7 @@ const mapObject = require('mapObject');
 type RelayRendererProps = {
   Container: RelayContainer;
   forceFetch?: ?boolean;
+  initialReadyState?: ?ReadyState;
   onForceFetch?: ?(
     querySet: RelayQuerySet,
     callback: (readyState: ReadyState) => void
@@ -143,7 +144,9 @@ class RelayRenderer extends React.Component {
     this.mounted = true;
     this.pendingRequest = null;
     this.querySet = null;
-    this.state = {active: false, readyState: null};
+    this.state = props.initialReadyState ?
+      {active: true, readyState: {...props.initialReadyState, mounted: true}} :
+      {active: false, readyState: null};
   }
 
   getChildContext(): Object {
@@ -154,7 +157,10 @@ class RelayRenderer extends React.Component {
   }
 
   componentDidMount(): void {
-    this._runQueries(this.props);
+    const {readyState} = this.state;
+    if (!readyState || !readyState.done || this.props.forceFetch) {
+      this._runQueries(this.props);
+    }
   }
 
   /**
@@ -287,7 +293,10 @@ class RelayRenderer extends React.Component {
    */
   _resolveContainerProps(): Object {
     if (!this.containerProps) {
-      const {environment, queryConfig} = this.props;
+      const {Container, environment, queryConfig} = this.props;
+      if (!this.querySet) {
+        this.querySet = getRelayQueries(Container, queryConfig);
+      }
       this.containerProps = {
         ...queryConfig.params,
         ...mapObject(
@@ -360,6 +369,7 @@ function createFragmentPointerForRoot(
 RelayRenderer.propTypes = {
   Container: RelayPropTypes.Container,
   forceFetch: PropTypes.bool,
+  initialReadyState: RelayPropTypes.ReadyState,
   onReadyStateChange: PropTypes.func,
   queryConfig: RelayPropTypes.QueryConfig.isRequired,
   environment: RelayPropTypes.Environment,
--- a/src/container/RelayRenderer.js
+++ b/src/container/RelayRenderer.js
@@ -23,6 +23,7 @@ const RelayPropTypes = require('RelayPropTypes');
 import type {
   Abortable,
   ComponentReadyState,
+  ComponentReadyStateChangeCallback,
   ReadyState,
   RelayContainer,
 } from 'RelayTypes';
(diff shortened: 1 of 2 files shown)


src/container/RelayRenderer.js
```diff
@@ -342,12 +287,50 @@ class RelayRenderer extends React.Component {
...
+        retry: this._retry.bind(this),
+        stale: readyState.stale,
+      } :
+      {
+        done: false,
+        error: null,
+        props: null,
+        retry: this._retry.bind(this),
+        stale: false,
+      };
```
wincent: I'd probably just use an `if`/`else` here rather than a many-line ternary.

denvned: But isn't it good practice to follow functional programming style when possible? Besides there is even fancier ternary in `_runQueries`. :smile:

josephsavona: +1 for if/else

denvned: Not a problem, replaced with if/else. But it looks bit bulkier and harder to read, IMHO.

josephsavona:
@denvned Can you provide an example of how this would be used to implement server rendering? Would the new `initialReadyState` prop be used on the server, client, or both? It isn't obvious how this prop would be sufficient to render on the client given that data was fetched on the server. Data would have to be provided to the store from the server regardless. 

As an alternative to this approach, how about creating an "isomorphic" RelayEnvironment implementation to be used as

```javascript
// server
var serverEnvironment = new ServerRelayEnvironment(new RelayEnvironment());
React.renderToString(<RelayRender environment={serverEnvironment} ... />
var data = serverEnvironment.getDataForClient(); // <-- return to client

// client
dataFromServer; // from last step of server
var clientEnvironment = new ClientRelayEnvironment(
  new RelayEnvironment(), 
  dataFromServer
);
React.render(<RelayRenderer environment={clientEnvironment} ... />
```

denvned:
@josephsavona 
> Can you provide an example of how this would be used to implement server rendering?

Sure, it mostly follows your recommendations in https://github.com/facebook/relay/issues/589#issuecomment-156169239, and https://github.com/facebook/relay/issues/589#issuecomment-156195977:
```js
// server
const environment = new RelayEnvironment();
environment.prepareData(Container, queryConfig).then(
  ({props, data}) => {
    const reactMarkup = ReactDOMServer.renderToString(<RelayRenderer {...props} />);
    renderHtmlPage(reactMarkup, JSON.stringify(data));
  }
);

// client
const environment = new RelayEnvironment();
const data = JSON.parse(getData());
environment.injectData(Container, queryConfig, data).then(
  (props) => {
    ReactDOM.render(<RelayRenderer {...props} />, root);
  }
);
```
In both cases `props` are `{Container, environment, initialReadyState, queryConfig}`.

> Would the new `initialReadyState` prop be used on the server, client, or both?

We need it on both, because the data should be ready on the client during the initial render, otherwise React wouldn't reuse server-side rendered markup.

> override the implementation of key methods - forceFetch, primeCache - to skip work when data was already fetched

Actually, I have already tried that approach but unsuccessfully: https://github.com/facebook/relay/pull/625#issuecomment-185375491. Also I think passing `initialReadyState` is more clean solution.

josephsavona:
I see, thanks for clarifying. You're referring to your [comment](https://github.com/facebook/relay/pull/625#issuecomment-185375491):

> ...React throws an error because onReadyStateChange calls setState, and React does not allow calling setState before mount.

This makes sense - `RelayRenderer` currently assumes rendering will be asynchronous, but if we're rendering on the client with data from the server it needs to be synchronous.

My concern here is that the shape and value of `initialReadyState` is an implementation detail that users shouldn't have to know about. However, exposing it as a prop makes it part of the public API and requires us to continue supporting it which we'd prefer to avoid.

I'd like to see a proposal that uses some form of composition in order to avoid the need to change the `RelayRenderer` API. For example, the API I proposed above could be achieved by creating an `<IsomorphicRelayRenderer>` that uses RelayRenderer internally *except* for the first render on the client (add whatever prop you need to indicate that the first render should resolve synchronously from pre-populated data). In general, we'd like to move towards providing lower level primitives (#559) and encourage composition over a monolithic core.

To summarize, rather than proceed with this PR I propose that we:
- Implement an IsomorphicRelayRenderer outside of `react-relay` so that users can do server rendering today
- Provide lower-level primitives to make `isomorphic-relay` easier to implement via all public methods.

@denvned what do you think?

yungsters:
I have also been working on a lower-level component that RelayRenderer would compose. I came to the same conclusion as @denvned that feeding in a `readyState` object was sufficient. Where my line of thinking differed was that I was thinking of building a lower level `RelayReadyStateRenderer` that always synchronously rendered the supplied `readyState`.

@denvned Do you want to take a stab at this? Or want me to go ahead and build it (which is what I was planning on doing today)?

denvned:
> My concern here is that the shape and value of `initialReadyState` is an implementation detail that users shouldn't have to know about. However, exposing it as a prop makes it part of the public API and requires us to continue supporting it which we'd prefer to avoid.

I thought that it could be an undocumented, internals only prop that is passed exclusively as a part of opaque `props` returned by `prepareData` and `injectData` (as in my example above).

> creating an <IsomorphicRelayRenderer> that uses RelayRenderer internally *except* for the first render on the client

Nice idea! I think I even see all the details of a possible implementation (probably the trickiest part is to not allow `RelayRenderer` to destroy DOM tree on the second render on the client). But we need to make a public API in Relay that will allow to get `props` (containing fragment pointers) by a given `Container` and `queryConfig` that we then would pass to `Container` during the first render, or alternatively we probably can use `RelayReadyStateRenderer` proposed above by @yungsters to solve this.

> Do you want to take a stab at this? Or want me to go ahead and build it (which is what I was planning on doing today)?

@yungsters, I can not wait to see it, but would not be able to build it myself today. So please do it! :+1: 

> To summarize, rather than proceed with this PR

@josephsavona, I think the first commit of this PR is still useful, it cleans up `RelayRenderer` a lot without changing API or behaviour. Maybe you can cherry-pick it?

josephsavona:
> I thought that it could be an undocumented, internals only prop

I understand this line of thinking, and it's really tempting. Unfortunately if someone *can* pass it as a prop and depend on its behavior, the *will*, and this limits our flexibility. I really like @yungsters idea of a lower-level `RelayReadyStateRenderer` that you can use if you know what you're doing, and keeping a higher-level component that most people will use.

> the trickiest part is to not allow RelayRenderer to destroy DOM tree on the second render on the client

I would imagine that this new component would render the root component directly at first, until such point as the props changed such that data needed to be fetched from the server. At that point you'd render a `RelayRenderer` and let it do the data-fetching (or possibly implement that yourself)

> ...Maybe you can cherry-pick it?

Agree that the first commit is worth discussing on its own; no rush, maybe split that off into a separate PR as time permits?

denvned:
> Agree that the first commit is worth discussing on its own; no rush, maybe split that off into a separate PR as time permits?

@josephsavona I've made a separate PR, #942.

src/container/RelayRenderer.js
everdimension: No need for the `else` when `return`ing conditionally.

yungsters:
@denvned Does 8074aa0cdc7fb3fb85be9634d1a34d6650dfa307 suit your needs?

denvned:
@yungsters Looks like it should. Thanks!

PR #922, opened 2016-03-08
Treat all fat queries as patterns and preserve their empty nodes

steveluscher:
So, here's the thing. Currently, the plugin allows you to write an invalid query: a subselection-supporting field that has no subselections:

    fragment on Viewer { todos }

That should fail, unless you've specifically marked the fragment as being a pattern:

    fragment on Viewer @relay(pattern: true) {
      todos
    }

Pattern fragments are those on which certain validations are disabled.

For now, we're going to treat all fat queries as pattern fragments to unbreak products. Next, we're going to start to warn in the Babel Relay plugin when you have a query like the first one above. Lastly, we'll revert this PR and make the Babel Relay plugin throw on a query like the first one above.

### DIFF
--- a/src/mutation/RelayMutationQueue.js
+++ b/src/mutation/RelayMutationQueue.js
@@ -386,7 +386,19 @@ class PendingTransaction {
       this._fatQuery = nullthrows(flattenRelayQuery(
         fragment,
         {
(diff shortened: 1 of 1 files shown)


steveluscher:
@facebook-github-bot import

josephsavona:
:ship: it

PR #930, opened 2016-03-09
Mark all of the fat queries in the examples directory as containing pattern fragments

steveluscher:
See #922 for a detailed discussion.


steveluscher:
@facebook-github-bot import

PR #934, opened 2016-03-09
Reorder installs based on which one's more likely to fail

steveluscher:
`npm install -g react-native-cli ` is less likely to fail than `npm install`, so let's encourage people to run that one first when installing the React Native / Relay example app.


steveluscher:
@facebook-github-bot shipit

PR #938, opened 2016-03-10
Fix flow error in getRelayQueries

kassens:
The return value of `Map.prototype.get` is nullable, but the standard flow type
currently doesn't match that yet. Internally, we have the return value as
nullable which is why the `@FlowFixMe` annotation was added.


kassens:
@facebook-github-bot import

PR #942, opened 2016-03-10
Refactor RelayRenderer

denvned:
This simplifies state handling in `RelayRenderer`, but does not change external behaviour and API. It follows principles described in section "[Thinking in React: Identify the minimal (but complete) representation of UI state](https://facebook.github.io/react/docs/thinking-in-react.html#step-3-identify-the-minimal-but-complete-representation-of-ui-state)". I identified that only two things actually determine internal state of `RelayRenderer`:
- the most recent ready state (the `readyState` state param),
- and whether we received a ready state at least once (the new `active` state param).

The `active` param means that we made a render while having a ready state at least once, and next time we will rerender only if we also have a ready state (we might have reseted the ready state to `null` after sending new queries).

Previously other state params were either redundant or derivable from `readyState` and `active`.

After this refactoring it will probably be easy for `RelayRenderer` to make use of `RelayRedayStateRenderer` [proposed](https://github.com/facebook/relay/pull/921#issuecomment-194392149) by @yungsters.

But originally it was submitted as a part of #921 to simplify implementation of isomorphic rendering for `RelayRenderer`.

### DIFF
--- a/src/container/RelayRenderer.js
+++ b/src/container/RelayRenderer.js
@@ -33,6 +33,9 @@ const getRelayQueries = require('getRelayQueries');
 const invariant = require('invariant');
 const mapObject = require('mapObject');
 
+type RelayContainerProps = {
+  [propName: string]: mixed;
+};
 type RelayRendererProps = {
   Container: RelayContainer;
   forceFetch?: ?boolean;
@@ -54,7 +57,7 @@ export type RelayRendererRenderCallback =
 type RelayRendererRenderArgs = {
   done: boolean;
   error: ?Error;
-  props: ?{[propName: string]: mixed};
+  props: ?RelayContainerProps;
   retry: ?Function;
   stale: boolean;
 };
@@ -80,7 +83,7 @@ const {PropTypes} = React;
  *
  * The `render` callback is called with an object with the following properties:
  *
- *   props: ?{[propName: string]: mixed}
+ *   props: ?RelayContainerProps
  *     If present, sufficient data is ready to render the container. This object
  *     must be spread into the container using the spread attribute operator. If
  *     absent, there is insufficient data to render the container.
@@ -125,7 +128,7 @@ const {PropTypes} = React;
  *
  */
 class RelayRenderer extends React.Component {
-  containerProps: ?{[propName: string]: mixed};
+  containerProps: ?RelayContainerProps;
   gcHold: ?GarbageCollectionHold;
   mounted: boolean;
   pendingRequest: ?Abortable;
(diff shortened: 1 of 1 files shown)


denvned:
@yungsters Of couse you are right! And actually I also knew it when did refactoring. I don't know why I wrote "with truthy `ready` flag" in the description. :smile: Corrected the description. The actual implementation is correct, I believe.

src/container/RelayRenderer.js
```diff
@@ -253,29 +212,7 @@ class RelayRenderer extends React.Component {
...
-    this.setState(
-      this._buildState(
-        this.state.activeContainer,
-        this.state.activeEnvironment,
-        this.state.activeQueryConfig,
-        null,
-        null
-      )
-    );
+    return !!this.state.readyState || !this.state.active;
```
yungsters: Why isn't this just `return !this.state.active`?

src/container/RelayRenderer.js
```diff
@@ -342,12 +281,53 @@ class RelayRenderer extends React.Component {
     this.mounted = false;
   }
 
+  /**
+   * @private
+   */
+  _resolveContainerProps(): Object {
```
yungsters: Not much better, but slightly more documenting.

```
_resolveContainerProps(): {[propName: string]: mixed} {
```

src/container/RelayRenderer.js
```diff
@@ -342,12 +281,53 @@ class RelayRenderer extends React.Component {
...
+    }
+  }
+
   render(): ?React$Element {
     let children;
     let shouldUpdate = this._shouldUpdate();
     if (shouldUpdate) {
       const {Container, render} = this.props;
-      const {renderArgs} = this.state;
+      const renderArgs = this._calculateRenderArgs();
```
yungsters: Nice. I like how you changed the `renderArgs` (and `containerProps`) to be lazily computed.

src/container/RelayRenderer.js
denvned: After receiving a `readyState` the first time, we set `active` to `true` and never change back. So, with just `return !this.state.active;` we would never update after that.

yungsters: Oooh, got it. To cement my understanding...

- `active` will only be `false` in between the initial render and the first "tick" (when `onReadyStateChange` first gets fired).
- `readyState` will only be non-null if the `onReadyStateChange` callback has ever fired for the current set of `(container, environment, route)`.

Did I get that right?

denvned: Yeah, that's correct.

src/container/RelayRenderer.js
denvned: Agree, but I added `type RelayContainerProps = {[propName: string]: mixed}` for it, because it needs to be changed in multiple places in this file.

yungsters:
@facebook-github-bot import

PR #944, opened 2016-03-11
Upgrade Plugin to use `babel-preset-fbjs`

yungsters:
Completely upgrades the `babel-relay-plugin` package to use Babel 6.

Depends on facebook/fbjs#135.

Also, I got rid of all extraneous package dependencies and upgraded the rest.

And for the real reason I even did all of this... `npm run update-fixtures` now preserves pretty printing.

### DIFF
--- a/scripts/babel-relay-plugin/lib/HASH
+++ b/scripts/babel-relay-plugin/lib/HASH
@@ -1,1 +1,1 @@
-fNfgCJRDc73nHs+YJdbdS1IqV+I=
+p3hoT8+mnHP5S+T0/caqPCNnwCo=
--- a/scripts/babel-relay-plugin/lib/tools/transformGraphQL.js
+++ b/scripts/babel-relay-plugin/lib/tools/transformGraphQL.js
@@ -13,8 +13,8 @@
 'use strict';
 
 var babel = require('babel-core');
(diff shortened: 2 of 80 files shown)


josephsavona:
@yungsters is this blocked on something? I need it to solve https://github.com/facebook/relay/pull/943#issuecomment-195069985 - just setting `compact: false` doesn't seem to be enough.

yungsters:
This is blocked on facebook/fbjs#127.

kassens:
Now blocked on facebook/fbjs#135 instead.

josephsavona:
@steveluscher do we need this anymore?

steveluscher:
Generally I think we should be building all of our Babel transforms atop the fbjs preset, but last time I tried this it proved to be non-trivial. I think someone just needs to sit down with it, start with the fbjs preset, and add in transforms until it works.

wincent:
Going to close this (spring cleaning) seeing as this is woefully stale now and won't apply, and the original motivating cause is now resolved:

> And for the real reason I even did all of this... `npm run update-fixtures` now preserves pretty printing.

Confirmed by changing a fixture, running `npm run update-fixtures` and confirming that the desired pretty printing was applied.

We can look at using the `fbjs` preset as a base separately.

PR #945, opened 2016-03-11
Add REFETCH and IGNORE rangeBehaviors

xuorig:
As discussed in #574, #542

Introduce 2 new range behaviors:
  - `REFETCH`: Will refetch the entire connection (to squelch the warning when no `rangeBehavior` matches the tracked connection)
  - `IGNORE`: Replaces using null, means the range should not be refetched at all.

I've deprecated `null`, maybe we don't need to, what do you think ?

@yungsters @steveluscher 

let me what you think! 🍻

### DIFF
--- a/src/legacy/mutation/GraphQLMutatorConstants.js
+++ b/src/legacy/mutation/GraphQLMutatorConstants.js
@@ -14,7 +14,9 @@
 
 const GraphQLMutatorConstants = {
   APPEND: 'append',
+  IGNORE: 'ignore',
   PREPEND: 'prepend',
+  REFETCH: 'refetch',
   REMOVE: 'remove',
 
   NODE_DELETE_HANDLER: 'node_delete',
@@ -49,7 +51,11 @@ GraphQLMutatorConstants.UPDATE_TYPES
 GraphQLMutatorConstants.RANGE_OPERATIONS
   [GraphQLMutatorConstants.APPEND] = true;
 GraphQLMutatorConstants.RANGE_OPERATIONS
+  [GraphQLMutatorConstants.IGNORE] = true;
+GraphQLMutatorConstants.RANGE_OPERATIONS
   [GraphQLMutatorConstants.PREPEND] = true;
+GraphQLMutatorConstants.RANGE_OPERATIONS
+  [GraphQLMutatorConstants.REFETCH] = true;
 GraphQLMutatorConstants.RANGE_OPERATIONS
   [GraphQLMutatorConstants.REMOVE] = true;
 
--- a/src/mutation/RelayMutationQuery.js
+++ b/src/mutation/RelayMutationQuery.js
@@ -23,6 +23,7 @@ const RelayQuery = require('RelayQuery');
 import type RelayQueryTracker from 'RelayQueryTracker';
 const RelayRecord = require('RelayRecord');
 import type {Variables} from 'RelayTypes';
+const {REFETCH} = require('GraphQLMutatorConstants');
 
 const flattenRelayQuery = require('flattenRelayQuery');
 const forEachObject = require('forEachObject');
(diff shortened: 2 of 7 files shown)


src/traversal/writeRelayUpdatePayload.js
```diff
@@ -413,7 +413,14 @@ function addRangeNode(
     null;
 
   // no range behavior specified for this combination of filter calls
-  if (!rangeBehavior) {
+  if (!rangeBehavior || rangeBehavior === IGNORE) {
+    if (!rangeBehavior) {
+      warning(
+        false,
```
wincent: What do you think of?

```
warning(
  !rangeBehaviour,
  ...
);
```

instead of the `if`?

src/mutation/__tests__/RelayMutationQuery-test.js
```diff
@@ -305,6 +305,96 @@ describe('RelayMutationQuery', () => {
...
+              }
+            }
+          }
+        }
+      `);
+      expect(node)
+        .toEqualQueryNode(expected);
+    });
+
+    it('Range is not refetched at all when rangeBehavior is set to IGNORE', () => {
```
wincent: (nit) Lower case "r" in "Range"

src/mutation/RelayMutationQuery.js
```diff
@@ -210,26 +212,32 @@ const RelayMutationQuery = {
...
-            'and to enable optimistic mutations.',
-            trackedConnection.getStorageKey(),
-            parentName,
-            parentID
-          );
+          // If the connection is not in `rangeBehaviors` or we have explicitly
+          // set the behavior to `refetch`, re-fetch it.
+          if (rangeBehaviorValue !== REFETCH) {
+            warning(
+              false,
```
wincent:
```
warning(
   rangeBehavior === REFETCH,
  ...
);
```

?

It's possible we're doing the `warning(false, ...)` thing all over the place to avoid the overhead of a called to a function, but this is not a hot code path and outside of `__DEV__` the function is just `emptyFunction`, so close to free...

xuorig: yup that makes a lot of sense! Will change it.

src/legacy/mutation/GraphQLMutatorConstants.js
```diff
@@ -16,6 +16,8 @@ const GraphQLMutatorConstants = {
   APPEND: 'append',
   PREPEND: 'prepend',
   REMOVE: 'remove',
+  IGNORE: 'ignore',
+  REFETCH: 'refetch',
```
wincent: (nit) Should probably alpha order these.

xuorig:
Just realised failing tests, fixing this.

src/mutation/RelayMutationQuery.js
```diff
@@ -23,6 +23,8 @@ const RelayQuery = require('RelayQuery');
 import type RelayQueryTracker from 'RelayQueryTracker';
 const RelayRecord = require('RelayRecord');
 import type {Variables} from 'RelayTypes';
+const GraphQLMutatorConstants = require('GraphQLMutatorConstants');
+const REFETCH = GraphQLMutatorConstants.REFETCH;
```
wincent: Could do:

    const {REFETCH} = GraphQLMutatorConstants;

of even collapse the two lines:

    const {REFETCH} = require('GraphQLMutatorConstants');

src/mutation/__tests__/RelayMutationQuery-test.js
```diff
@@ -305,6 +305,96 @@ describe('RelayMutationQuery', () => {
       };
     });
 
+    it('refetches the whole range when the matching rangeBehavior is set to REFETCH', () => {
```
wincent: Our internal linter is probably going to complain about this long line. Can you rephrase it as something like:

    it('refetches the whole range when the rangeBehavior is REFETCH', () => {

wincent:
Looks pretty good @xuorig. Thanks for this!

I'll let it sit here for couple of days to give others a chance to comment before importing.

xuorig:
thanks for the quick review @wincent! I addressed your comments.

xuorig:
@wincent anything I can do to see this merged :) ?

wincent:
I'll pull it into our internal CR system and get the ball rolling.

wincent:
@facebook-github-bot import

xuorig:
I don't know if that worked @wincent :panda_face:

wincent:
That bot... I've used it to import at least a couple things today. This time I'm just going to import it manually.

wincent:
Turns out it was an issue with Unicode in your commit (your name). I believe the import will work now, so giving it another try.

wincent:
@facebook-github-bot import

xuorig:
Blame my parents for naming me this way :sob:

wincent:
After a long saga, the import worked, code review happened, I fixed some lint warnings and nits for you and it landed. Thanks once again for contributing!

xuorig:
:tada:

PR #946, opened 2016-03-12
fix gh-pages page to display nicely on narrow mobile screens

everdimension:
Currently the documentation website adapts to mobile screens, but some layout looks broken:

* Navigation items "fall out" from the nav wrapper;
* Marketing columns' text is center-aligned, but columns themselves are not centered;
* Buttons-unit buttons have no space between them when stacked on top of each other.

This commit fixes the issues.

Not sure if modifying the `./css/relay.css` file directly is the correct way, but if it's not, please point me in the right direction.

### DIFF
--- a/css/relay.css
+++ b/css/relay.css
@@ -989,10 +989,14 @@ p + .apiIndex {
     float: none;
   }
   .marketing-col {
-    margin-left: 0;
     float: none;
     margin-bottom: 30px;
     text-align: center;
+  }
+  .marketing-col,
+  .marketing-col:first-child {
+    margin-left: auto;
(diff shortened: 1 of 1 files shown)


wincent:
Thanks for this @everdimension. Can you resubmit please against [this file](https://github.com/facebook/relay/blob/master/website/src/relay/css/relay.css) in the master branch? The website files get generated from that.

Also, any chance of a before/after screenshot, to make this easier to review?

everdimension:
@wincent 
Ok, thanks for advice! I've resubmitted the PR here: https://github.com/facebook/relay/pull/955

PR #950, opened 2016-03-14
Plugin: Add `compactLines` to Fixture Transformer

yungsters:
Fixes printing of the fixtures so that "pretty printing" is preserved.

This will tie us over until #944 is ready to be merged.


yungsters:
@facebook-github-bot import

PR #952, opened 2016-03-14
Clean up some context -> environment cruft

wincent:
For consistency, and to eliminate potential confusion, let's make sure we always refer to the `RelayEnvironment` as an environment and not a context.


wincent:
Travis is too slow. Importing anyway.

wincent:
@facebook-github-bot import

PR #953, opened 2016-03-15
Relay: Reduce Scope of Tests to Relay Plugin

yungsters:
There is some broken integration between the Babel 6 plugins in `fbjs-scripts` and the transform tests in the `babel-relay-plugin`. For now, I am going to reduce the scope of the `babel-relay-plugin` tests until we're fully upgraded to Babel 6. This revision should fix the breaking tests in master.


yungsters:
@facebook-github-bot import

PR #956, opened 2016-03-16
Simplify build step for babel-relay-plugin.

cpojer:
cc @yungsters


cpojer:
@facebook-github-bot import

PR #957, opened 2016-03-16
Fix two old Flow typing issues

kassens:
Flow now know to keep refinements inside callbacks if there are no assignments.


kassens:
@facebook-github-bot import

PR #958, opened 2016-03-17
Fix warnings in tests due to missing __typename

kassens:


kassens:
@facebook-github-bot import

PR #960, opened 2016-03-17
Fix warnings in star wars example

kassens:


kassens:
@facebook-github-bot shipit

PR #961, opened 2016-03-17
[website] Fix let not working outside 'use strict';

vjeux:
The awesome codemod to rename var into let broke the website as it wasn't using 'use strict';. Adding it to the files fixes it :)

### DIFF
--- a/website/core/DocsSidebar.js
+++ b/website/core/DocsSidebar.js
@@ -7,7 +7,9 @@
  * of patent rights can be found in the PATENTS file in the same directory.
  *
  * @providesModule DocsSidebar
-*/
+ */
+
+'use strict';
 
 const Metadata = require('Metadata');
 
--- a/website/core/Prism.js
+++ b/website/core/Prism.js
@@ -7,7 +7,9 @@
  * of patent rights can be found in the PATENTS file in the same directory.
  *
  * @providesModule Prism
-*/
+ */
+
+'use strict';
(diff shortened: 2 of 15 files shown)


website/core/Prism.js
```diff
@@ -498,9 +502,7 @@ if (Prism.languages.markup) {
 			alias: 'language-javascript'
 		}
 	});
-}
-;
-(function(Prism) {
```
vjeux: this self calling function was pretty useless

website/core/Prism.js
```diff
@@ -22,7 +24,7 @@ const unindent = require('unindent');
 // Private helper vars
 const lang = /\blang(?:uage)?-(?!\*)(\w+)\b/i;
 
-const _ = Prism = {
```
vjeux: turns out that it made Prism global which 'use strict'; doesn't like so much :)

website/core/DocsSidebar.js
```diff
@@ -7,7 +7,9 @@
  * of patent rights can be found in the PATENTS file in the same directory.
  *
  * @providesModule DocsSidebar
-*/
+ */
```
vjeux: the indent on the last line of the comment was off

kassens:
lgtm, if you want to import or shipit.

vjeux:
@facebook-github-bot shipit

PR #972, opened 2016-03-20
Support null-prototype mutation payload objects

taion:
This came to my attention via https://github.com/relay-tools/relay-local-schema/pull/15. Recent versions of graphql-js use `null`-prototype objects for maps, which leads to the `payload.hasOwnProperty` call below failing when directly feeding the output of in-process graphql-js into Relay (e.g. with relay-local-schema).

I felt that it was trivial enough to just make this little patch here.


wincent:
Thanks, @taion. Looks reasonable. Will import.

wincent:
@facebook-github-bot import

PR #976, opened 2016-03-22
Fix the generated GraphQL string in example

koistya:


wincent:
Thanks for this, @koistya. It's still not entirely consistent though:

- Component is called `ProfilePicture`.
- Fragment name is `ProfilePhoto` (before your change) and `ProfilePicture` (after your change).
- GraphQL field is called `profilePhoto`.

Want to update that field name to `profilePicture` as well, so that it's the same throughout?

josephsavona:
I'm going to close this due to inactivity, but please feel free to open a new PR or reopen with an update (depending on which is easier/possible with GH).

PR #977, opened 2016-03-22
Fix another let not working outside 'use strict';

chentsulin:
Fix another let not working outside strict mode:

```js
let is_executing = false;
```


kassens:
Thanks!

@facebook-github-bot import

PR #980, opened 2016-03-22
Strip trailing commas from fixtures

kassens:
Commas are considered whitespace in GraphQL and we can just skip them. They're
nice inline, but at end of lines they're useless.

Done using [remove_commas.py](https://gist.github.com/kassens/6449143a9d1adb31b1a9):

```
find scripts/babel-relay-plugin/src/__fixtures__ -type f | xargs -n1 python remove_commas.py
```


kassens:
@facebook-github-bot import

Issue #927, opened 2016-03-08
Enforce that mutation config's `connectionName` references an actual connection field


wincent:
As per @steveluscher's [astute diagnosis of a bug](https://github.com/facebook/relay/issues/768#issuecomment-193887941), we probably want to enforce that fields specified as `connectionName` actually correspond to real connections (ie. with `isConnection` metadata set to `true`).

Issue #768, opened 2016-01-27
NODE_DELETE getOptimisticResponse causes error


BerndWessels:
Hi
I have a `NODE_DELETE` mutation that should remove a `post` from the `viewer.posts` connection.

It works fine without `getOptimisticResponse` !

BUT with `getOptimisticResponse` it kind of breaks with a short flashing red screen and the following error:

Warning: Failed propType: Required prop `post` was not specified in `Post`. Check the render method of `Relay(Post)`.

So it seems after the `optimistic response` and before the `server response` the `deleted edge` is still being rendered but the `deleted post` is already gone which causes the `post component` render to fail.

It looks like that the `optimistic response` is not handled in the same way as the `server response` even though they are identical.

Is this a bug in relay?

[Here is the complete project](https://github.com/BerndWessels/react-webpack) to reproduce the problem.

This is the mutation:

    class DeletePostMutation extends Relay.Mutation {
      getMutation() {
        return Relay.QL`mutation {deletePost}`;
      }
      getVariables() {
        return {
          id: this.props.post.id
        };
      }
      getFatQuery() {
        return Relay.QL`
          fragment on DeletePostPayload {
            viewer {
              id
              posts
            }
            deletedId
          }`;
      }
      getConfigs() {
        return [{
          type: 'NODE_DELETE',
          parentName: 'viewer',
          parentID: this.props.viewer.id,
          connectionName: 'posts',
          deletedIDFieldName: 'deletedId'
        }];
      }
      static fragments = {
        viewer: () => Relay.QL`
          fragment on Person {
            id
          }`,
        post: () => Relay.QL`
          fragment on Post {
            id
          }`
      };
      getOptimisticResponse() {
        const { viewer, post } = this.props;
        const viewerPayload = { id: viewer.id };
        return {
          viewer: viewerPayload,
          deletedId: post.id
        };
      }
    }

This is the post component:

    class Post extends React.Component {
      // Expected properties.
      static propTypes = {
        post: React.PropTypes.object.isRequired
      };
    
      // Initialize the component.
      constructor(props) {
        super(props);
      }
    
      _handleDeletePost = () => {
        const { viewer, post } = this.props;
        // We commit the update directly to the database.
        Relay.Store.commitUpdate(new DeletePostMutation({viewer, post}), {
          onFailure: (err) => {
            // TODO: Deal with it!
            console.log(err);
          },
          onSuccess: (result) => {
            // TODO: Maybe nothing todo here?
          }
        });
      };
[...]

PR #989, opened 2016-03-24
Enforce that connectionName refers to a connection

wincent:
Should help catch issues like this one:

https://github.com/facebook/relay/issues/768

Closes:

https://github.com/facebook/relay/issues/927


wincent:
@facebook-github-bot import

PR #990, opened 2016-03-24
DRY things up in RelayMutationRequest

wincent:
Avoid verbatim repetition of the query-printing code.


wincent:
@facebook-github-bot import

PR #991, opened 2016-03-25
Use right type of nodes in @relay directive definition

kassens:
This was just working by accident, but using the wrong type of node.


kassens:
@facebook-github-bot import

PR #997, opened 2016-03-28
Restore ability to pass filter to `npm test`

wincent:
Prior to 8c386dc7130f9fb7e you could pass a "filter" pattern to `npm test` to limit the scope of the test run (eg. `npm test RelayMutationQuery` to run just the `RelayMutationQuery` tests).

While you can still pass a filter to `npm jest`, we shouldn't break things for people who might be used to running `npm test`. This commit restores that ability.

Manually tested `npm test` with and without a filter.


wincent:
@facebook-github-bot import

Issue #782, opened 2016-02-01
Fragment issue when committing Mutation


danielgriese:
Hi there,

I have just implemented a new feature in our product using Relay and it worked like a charm so far. Thank you for your awesome work and for open sourcing Relay!

Unfortunately I have encountered a tricky issue, that I currently do not know how to circumvent and if it is a bug in relay or if my schema is just programmed in a stupid way. Both options are valid at this point in time... :D
### Situation

I have page showing a "Claim" that refers to a "Medium". This claim has an applied "Policy" that can be changed by a mutation. All these types implement a common Resource interface.
The shortened Schema looks like this.

```
interface Resource {
  id: ID!
  name: String
}

type Claim implements Node, Resource {
  id: ID!
  name: String
  medium: Medium
  policy: Policy
}

type Medium implements Node, Resource {
  id: ID!
  name: String
  remoteUrl: String
}

type Policy implements Node, Resource {
  id: ID!
  name: String
}
```

The mutation is described as this:

```
type Mutation {
  updateClaimPolicy(input: UpdatePolicyInput!): UpdatePolicyPayload
}

input UpdatePolicyInput {
  id: ID!
  policy: String!
  clientMutationId: String!
}

type UpdatePolicyPayload {
  claim: Claim
  clientMutationId: String!
}
```

The client side mutation code:

```
export default class UpdateClaimPolicyMutation extends Relay.Mutation {

  // we depend on the claim id, so make sure we get it
  static fragments = {
    claim: () => Relay.QL`
      fragment on Claim {
        id
      }
    `,
  };

  getMutation() {
    return Relay.QL`mutation {updateClaimPolicy}`;
  }

  getVariables() {
    return { id: this.props.claim.id, policy: this.props.policy._id };
  }

  /**
  * All that could change...
  * For now this is only the claims policy and the updater
  */
  getFatQuery() {
    return Relay.QL`
      fragment on UpdatePolicyPayload {
        claim {
          policy
          updatedOn
          updatedBy
        }
      }
    `;
  }

  // tell relay how to react to the payload...
  // in this case, update the claim
  getConfigs() {
    return [{
      type: 'FIELDS_CHANGE',
      fieldIDs: {
        claim: this.props.claim.id,
      },
    }];
  }

}
```

The displaying container has the following (shortened) fragment:

```
claim: () => Relay.QL`
      fragment on Claim {
        _id
        medium {
          name
        }
        policy {
          name
        }
    }
    `,
```

[...]

PR #1000, opened 2016-03-28
Add an "indirection layer" to avoid type conflicts

NevilleS:
...this is a first stab at a fix for #782. It should fix the issue for at least some subset of FIELDS_CHANGE mutations, but might create new issues in other cases, doesn't correct the query building code for RANGE_ADD or RANGE_DELETE, and still needs more work.

But it does create a test case that at least demonstrates one example of the issue!

### DIFF
--- a/src/mutation/RelayMutationQuery.js
+++ b/src/mutation/RelayMutationQuery.js
@@ -67,8 +67,8 @@ type OptimisticUpdateQueryBuilderConfig =
     response: Object;
   };
 
-const {CLIENT_MUTATION_ID} = RelayConnectionInterface;
-const {ANY_TYPE, ID, TYPENAME} = RelayNodeInterface;
+var {CLIENT_MUTATION_ID} = RelayConnectionInterface;
+var {ANY_TYPE, ID, TYPENAME, NODE_TYPE} = RelayNodeInterface;
 
 /**
  * @internal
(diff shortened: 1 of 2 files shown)


NevilleS:
See the discussion in #782 for some context as to why I'd open a PR that definitely isn't ready :+1: 

(also, PR #1000 :100: :100: :100: :100: :100: :100: :100: :100: :100: :100:)

josephsavona:
cc @wincent, @yuzhi re problems with automatic query construction. This is more evidence to suggest for requiring that mutation query construction be more static/manual/explicit.

NevilleS:
Did you guys ever get a chance to think about this? We've got several places in our codebase with `// TODO: https://github.com/facebook/relay/issues/782` and hoping there'd be a better solution than the indirection layers... 😢

sibelius:
👍

wincent:
> Did you guys ever get a chance to think about this? We've got several places in our codebase with `// TODO: https://github.com/facebook/relay/issues/782` and hoping there'd be a better solution than the indirection layers...

I think there are a couple of main alternatives here:

1. Switch to Relay 2 when it comes out ("soon", watch our meeting sync notes to get the latest status on our progress towards being able to push a preview to a branch on GitHub); here all queries, including mutations, are entirely static, and there is no dynamic runtime query generation (no tracked queries, no fat queries, no intersection), so this issue just goes away.
2. Use the existing low-level mutation API, `RelayGraphQLMutation` (shipped, but not documented, as it was really just a step towards what we wanted to build in Relay 2): this also an entirely static mutation API, where you supply your own query and there are no fat/tracked queries or intersection involved. This is still somewhat of a rough API, and because it sits on top of all the "legacy" underpinnings in terms of mutation queues, transactions etc it does have some limitations that make it not as flexible or powerful as what we'll have in Relay 2.

In the meantime, @NevilleS, what do you want to do with this PR? I see two options: either abandon it or update it. What do you think?

NevilleS:
I think the right thing to do is abandon this PR and we'll wait for Relay 2. When we originally hit this issue the low-level mutation API wasn't available, so if this becomes a larger problem for us that workaround will also work - in the meantime, our "manually-inserted-indirection-layer" hack lets us use the declarative mutation API so we'll continue to use that.

Looking forward to fixing some of this stuff with Relay 2, though!

PR #1004, opened 2016-03-30
Be more explicit about trackers

wincent:
In these files we have change trackers, query trackers, and now fragment trackers. This can be confusing.

This commit makes sure that whenever we have more than one tracker in a local area, we use explicit names:

- `_tracker` becomes `_queryTracker`
- `tracker` becomes `queryTracker`

And so on. Even if we simplify things in the future to involve fewer trackers I still think this is a worthwhile change (another one is probably tackling the code smell of `writePayload` and the `RelayQueryWriter` constructor taking seven and six parameters respectively, but that is beyond the scope of this commit).


wincent:
@facebook-github-bot import

PR #1008, opened 2016-03-30
Fix RelayContainer-test

kassens:
The argument should always be a `DataID` and never an array.


kassens:
@facebook-github-bot import

Issue #968, opened 2016-03-18
Error: TypeError: Object  has no method 'find'


varunarora:
I have been constantly getting the error `Error: TypeError: Object  has no method 'find'` and can't seem to understand this `Relay Transform Error` for days now. Tried all combinations. My GraphQL is setup and working (tested on Graphiql), but this is really confusing me.

```
Error: TypeError: Object  has no method 'find'
    at new RelayQLFragment (/home/varun/Code/opencurriculum/oc_platform/node_modules/babel-relay-plugin/lib/RelayQLAST.js:143:49)
    at RelayQLTransformer.processDocumentText (/home/varun/Code/opencurriculum/oc_platform/node_modules/babel-relay-plugin/lib/RelayQLTransformer.js:156:16)
    at RelayQLTransformer.transform (/home/varun/Code/opencurriculum/oc_platform/node_modules/babel-relay-plugin/lib/RelayQLTransformer.js:67:29)
    at PluginPass.TaggedTemplateExpression (/home/varun/Code/opencurriculum/oc_platform/node_modules/babel-relay-plugin/lib/getBabelRelayPlugin.js:124:36)
    at newFn (/home/varun/Code/opencurriculum/oc_platform/node_modules/babel-core/node_modules/babel-traverse/lib/visitors.js:293:19)
    at NodePath._call (/home/varun/Code/opencurriculum/oc_platform/node_modules/babel-core/node_modules/babel-traverse/lib/path/context.js:75:18)
    at NodePath.call (/home/varun/Code/opencurriculum/oc_platform/node_modules/babel-core/node_modules/babel-traverse/lib/path/context.js:47:17)
    at NodePath.visit (/home/varun/Code/opencurriculum/oc_platform/node_modules/babel-core/node_modules/babel-traverse/lib/path/context.js:105:12)
    at TraversalContext.visitQueue (/home/varun/Code/opencurriculum/oc_platform/node_modules/babel-core/node_modules/babel-traverse/lib/context.js:156:16)
    at TraversalContext.visitSingle (/home/varun/Code/opencurriculum/oc_platform/node_modules/babel-core/node_modules/babel-traverse/lib/context.js:113:19)
    at TraversalContext.visit (/home/varun/Code/opencurriculum/oc_platform/node_modules/babel-core/node_modules/babel-traverse/lib/context.js:200:19)
[...]

PR #1013, opened 2016-03-31
Use find helper

edvinerikson:
Fixes https://github.com/facebook/relay/issues/968 (maybe..)
I couldn't verify that this fixes it. When I tried it locally I could run the plugin in 0.10 without the patch.

#### Edit
I managed to verify that it is working now.
I used the schema/babel config in the todo example and ran `node node_modules/.bin/babel test.js`
**test.js**
```js
import Relay from 'react-relay';
const q = Relay.QL`
fragment on User @relay(pattern: true) {
  id
}
`;
```
cc @kassens @josephsavona

### DIFF
--- a/scripts/babel-relay-plugin/lib/HASH
+++ b/scripts/babel-relay-plugin/lib/HASH
@@ -1,1 +1,1 @@
-x2plDoafKBiENN03Y8mggKLhqmo=
+67cv/FJ+K+p0zcmHMXl78y+vPoY=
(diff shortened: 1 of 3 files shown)


josephsavona:
Thanks!

josephsavona:
@facebook-github-bot import

edvinerikson:
I managed to verify that it is working now.
I used the schema/babel config in the todo example.
`node node_modules/.bin/babel test.js`
**test.js**
```js
import Relay from 'react-relay';
const q = Relay.QL`
fragment on User @relay(pattern: true) {
  id
}
`;
```

Issue #690, opened 2015-12-17
Detect when a polyfill is missing


kassens:
Should we add some simple basic detection for missing polyfills so we can provide a helpful error message instead of some "method x is undefined" errors deep in the stack? (Example: #681)

Not full feature detection, but maybe something basic as:
```javascript
warning(
  Promise && Array.prototype.find,
  'Relay relies on polyfills for older browsers. Babel '
  'provides a good one: https://babeljs.io/docs/usage/polyfill/'
)
```

josephsavona:
This is a great idea. Ship it! ;-)

KyleAMathews:
:+1: this would of saved me some debugging time this morning :)

edvinerikson:
If this is something that we still want to do I'll be happy to hack on it.
```
if (process.env.NODE_ENV === 'development') {
  warning(
    -es6-features-here,
    'message here'
  );
}
```
Would something like that work and where should the check live? RelayPublic?

josephsavona:
@edvinerikson the best place is probably [Relay.js](https://github.com/facebook/relay/blob/master/src/__forks__/Relay.js). The check should probably use `if (__DEV__) { ... }`, which i think we transform appropriately (probably easiest to just try that and see what it builds to).

PR #1014, opened 2016-04-01
added polyfill message

edvinerikson:
Fixes https://github.com/facebook/relay/issues/690
I couldn't really find more es6 stuff other than what it checks (except for `Map` which got transpiled to `_Map` for some reason).
Not sure about tests.. Do we need a test for this and where should it go if so?
cc @josephsavona

### DIFF
--- a/src/__forks__/Relay.js
+++ b/src/__forks__/Relay.js
@@ -16,6 +16,15 @@
 const RelayDefaultNetworkLayer = require('RelayDefaultNetworkLayer');
 const RelayPublic = require('RelayPublic');
 const RelayStore = require('RelayStore');
+const warning = require('warning');
+
+if (__DEV__) {
+  warning(
+    typeof Promise === 'function' && Array.prototype.find,
+    'Relay relies on polyfills for es6 features in older browsers. ' +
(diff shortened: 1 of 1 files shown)


edvinerikson:
Flow install seems to have some issues.

src/__forks__/Relay.js
```diff
@@ -16,6 +16,15 @@
 const RelayDefaultNetworkLayer = require('RelayDefaultNetworkLayer');
 const RelayPublic = require('RelayPublic');
 const RelayStore = require('RelayStore');
+const warning = require('warning');
+
+if (__DEV__) {
+  warning(
+    Promise && Array.prototype.find,
```
kassens: maybe change this to `typeof Promise === 'function'` to not throw if it's not defined.

kassens:
Thanks for doing this, just a small update please and we should be good to merge this.

src/__forks__/Relay.js
edvinerikson: I thought of doing `global.Promise` first but wasn't sure if it gets transformed correctly in browsers.
I'll go with the first. :+1:

edvinerikson:
Awesome, I've updated the PR :)

kassens:
Thanks for the update!

PR #1017, opened 2016-04-04
Relax connectionName validation slightly

wincent:
In f2aa5b0e78c2f5 (D3095672) we tightened up mutation config validation a little too much, breaking product code that didn't explicitly include the connection field in the fat query definition.

In this commit we make it a little less aggressive:

- If the connection field is explicit in the fat query, we validate it.
- If it is not explicit but we can get it by intersecting with a tracked query, we validate that.
- Otherwise we let things slide and give up.

Note that in doing this I addressed the TODO comment and moved the code that deals with fields on the fat parent outside the conditional, so that it runs even when we don't have any tracked queries. That TODO predates even the open source release of Relay, so if somebody has context on if this is a terrible idea, please let me know.

### DIFF
--- a/src/mutation/RelayMutationQuery.js
+++ b/src/mutation/RelayMutationQuery.js
@@ -141,11 +141,12 @@ const RelayMutationQuery = {
       tracker,
     }: EdgeDeletionMutationFragmentBuilderConfig
   ): ?RelayQuery.Node {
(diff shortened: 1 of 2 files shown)


wincent:
@facebook-github-bot import

PR #1024, opened 2016-04-06
Update jest in plugin to 0.10

kassens:


kassens:
@facebook-github-bot import

PR #1037, opened 2016-04-08
Remove dependency on graphql from main package.json

kassens:
I couldn't find a place where we depend on `graphql` directly from the main repo. The only requires I saw were inside the plugin.


kassens:
@facebook-github-bot import

kassens:
According to travis, we need this. ¯\_(ツ)_/¯

PR #1041, opened 2016-04-11
Update README.md

ringzhz:
Changing ports `adb reverse` command to match up with the port that the app actually uses: 8080 instead of 8081.


josephsavona:
@facebook-github-bot shipit

PR #1043, opened 2016-04-12
Update examples to use new IGNORE range behavior

steveluscher:


steveluscher:
@facebook-github-bot shipit

PR #1050, opened 2016-04-14
Prevents RelayNetworkDebug from breaking fetch

finneganh:
RelayNetworkDebug overwrites the global 'fetch' function with one that
does not return a Promise. Additionally, it calls that 'fetch' without a
correct 'this'.

Fixes the overwrite to correctly invoke a global 'fetch' and also
preserve its return value.

I think this should be covered under an existing corporate CLA with Twitter.

### DIFF
--- a/src/tools/RelayNetworkDebug.js
+++ b/src/tools/RelayNetworkDebug.js
@@ -55,10 +55,10 @@ class RelayNetworkDebugger {
     this._fetch = global.fetch;
     global.fetch = (url, options, ...args) => {
       const name = url.split('/')[2];
-      this.logRequest(createDebuggableFromFetch(
+      return this.logRequest(createDebuggableFromFetch(
         name,
         {url, options, args},
-        this._fetch(url, options, ...args)
+        this._fetch.call(global, url, options, ...args)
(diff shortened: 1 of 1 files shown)


yuzhi:
@finneganh Thank you for your pull request. After using this a bit internally, we found that this way of blatantly overwriting the inject network layer is somewhat fragile.  Instead of making it wrap around an `injectedNetworkLayer`, I am actually in the process of rewriting RelayNetworkDebug to be more of a subscriber to `RelayNetworkLayer`.

finneganh:
Sounds good! Thanks for the update.

wincent:
I think we can close this out now that @yuzhi's new take on `RelayNetworkDebug` has landed in 85158ce9f88cf3c93015106e1d1df64b64b49d85. Let me know if I'm mistaken @finneganh, and thanks for contributing!

finneganh:
Cool, looking forward to that getting in a release.

wincent:
Yep, probably worth cutting a release soon.

kassens:
I just published 0.8.1 with the fix.

PR #1055, opened 2016-04-16
Update Flow to v0.23.0

gabelevi:
The only new errors upon upgrading to v0.23.0 come from the use of non-standard Promise APIs, like `done()`. Flow removed them from the builtin library definition because they are non-standard. The fix is to provide a custom definition of what `Promise` looks like.


steveluscher:
@facebook-github-bot shipit

PR #1060, opened 2016-04-18
Pass RelayRootContainer environment prop to RelayRenderer

dguillamot:
Pass RelayRootContainer environment prop to RelayRenderer

I'm not 100% sure on the type checking `(type RootContainerProps... environment: RelayStore)` and the propTypes `(environment: PropTypes.Environment) `

But I needed to add line 107 and line 115 `(environment={environment || RelayStore}` in order to pass the environment prop down to RelayRenderer. All of my screens (using relay with React Native) use Relay.RootContainer and pass in the environment prop.

I know environment is still very new to the Public API  (only 1 week old) so if this method is incorrect please let me know. 

Thanks,

### DIFF
--- a/src/container/RelayRootContainer.js
+++ b/src/container/RelayRootContainer.js
@@ -35,6 +35,7 @@ type RootContainerProps = {
   ) => React$Element;
   renderLoading?: ?() => React$Element;
   route: RelayQueryConfigInterface;
+  environment: RelayStore;
 };
 
 const {PropTypes} = React;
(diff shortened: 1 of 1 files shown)


josephsavona:
@dguillamot The current approach - no `environment` prop on `RelayRootContainer` - is intentional. A lot of existing products use that component *and* use `RelayStore` for data access. It's easy to imagine a case in which a product forgets to set the `environment` variable, only to have some of their data go to the wrong store, so we'd prefer to have the prop be explicitly required (as in RelayRenderer) or fixed (RelayRootContainer). 

If you'd like to explicitly set an environment, we'd recommend using `RelayRenderer` directly.

dguillamot:
Thanks, will take a look at RelayRenderer instead of RootContainer.

PR #1062, opened 2016-04-19
Use Promise Flow lib from fbjs

zpao:
Reverts part of #1055 to keep the Promise definition in fbjs (where IMO it belongs). I shipped fbjs@0.8.1 with that earlier today.


kassens:
@facebook-github-bot import

PR #1067, opened 2016-04-20
Documented all `babel-relay-plugin` options.

Globegitter:
Some of the options that are supported by `getBabelRelayPlugin` are not documented at all. This just makes sure that all options are listed and have a minimal description of what they do.

I am not using the `validator` myself so did not figure out how exactly that would be used. If it is easy to give a minimal example I would love to add this, so we have some more complete documentation.

### DIFF
--- a/docs/Guides-BabelPlugin.md
+++ b/docs/Guides-BabelPlugin.md
@@ -124,7 +124,8 @@ babel.transform(source, {
       debug: false, 
       // Supresses all warnings that would be printed
       suppressWarnings: false, 
-      // Can add custom schema validation rules? Is there a simple example for this?
+      // Can add custom validation rules. 
+      // If none specified some of the default rules are skipped (e.g. for fat queries)
       validator: (GraphQL) => {
         return (schema, ast) => {
           return []
--- a/docs/Guides-BabelPlugin.md
+++ b/docs/Guides-BabelPlugin.md
@@ -114,12 +114,31 @@ If you're using a different GraphQL server implementation, we recommend adapting
 
 By default, `babel-relay-plugin` catches GraphQL validation errors and logs them without exiting. The compiled code will also throw the same errors at runtime, making it obvious that something went wrong whether you're looking at your terminal or browser console.
 
-When compiling code for production deployment, the plugin can be configured to immediately throw upon encountering a validation problem:
+When compiling code for production deployment, the plugin can be configured to immediately throw upon encountering a validation problem. The plugin can be further customized for different environments with the following options:
 
 ```javascript
 babel.transform(source, {
   plugins: [
-    [getBabelRelayPlugin(schemaData), {enforceSchema: true}],
+    [getBabelRelayPlugin(schemaData), {
+    // will throw an error when it validates the queries at build time
+    enforceSchema: true, 
+    // Only if enforceSchema: false and debug: true it will still log validation errors at build time
+    debug: false, 
+    //Customizes the name of the input variable when queries/muations are being sent off to the server.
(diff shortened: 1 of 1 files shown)


docs/Guides-BabelPlugin.md
```diff
@@ -114,12 +114,32 @@ If you're using a different GraphQL server implementation, we recommend adapting
...
+When compiling code for production deployment, the plugin can be configured to immediately throw upon encountering a validation problem. The plugin can be further customized for different environments with the following options:
 
 ```javascript
 babel.transform(source, {
   plugins: [
-    [getBabelRelayPlugin(schemaData), {enforceSchema: true}],
+    [getBabelRelayPlugin(schemaData, {
+      // Only if enforceSchema: false and debug: true it will still log validation errors at build time
+      debug: false, 
+      //Customizes the name of the input variable when queries/muations are being sent off to the server.
```
wincent: Missing a space here after the `//`.

docs/Guides-BabelPlugin.md
```diff
@@ -114,12 +114,32 @@ If you're using a different GraphQL server implementation, we recommend adapting
...
+      inputArgumentName: 'custom_input_name',
+      // Transforms any field names to snake_case when the queries/mutations are sent to the server
+      snakeCase: true,
+      //Supresses all warnings that would be printed
```
wincent: Also missing a space after the `//` here.

wincent:
Thanks for this, @Globegitter. I'm not actually sure that we do want to document all of these options, as that would mean supporting them going forward. `inputArgumentName` and `snakeCase`, for example, might only exist because we needed to distinguish internally between open source GraphQL conventions and our preexisting internal ones.

@josephsavona, @yuzhi, @steveluscher. Any thoughts on which, if any of these options you'd prefer to remain undocumented?

docs/Guides-BabelPlugin.md
```diff
@@ -114,12 +114,32 @@ If you're using a different GraphQL server implementation, we recommend adapting
...
+      suppressWarnings: false, 
+      // enables the usage of variables in your relay queries via ES6 template strings
+      substituteVariables: true,
```
josephsavona: Let's not document this one

josephsavona:
I agree with @wincent - `inputArgumentName`, `snakeCase`, and *especially* `substituteVariables` exist only to support switching between FB and OSS configuration and we'd prefer that people avoid relying on them. Especially `substituteVariables`, which we're working to deprecate.

Globegitter:
Great to get some feedback on that. Any reason for deprecating `substituteVariables`?  Is it going to be on by default?

josephsavona:
`substituteVariables` allows developers to inject a runtime value into a query. Relay tries to only evaluate each query/fragment once and memoize it, so dynamically injected values don't necessarily work the way you'd expect (i.e. updated the injected JS value doesn't change the query). Going forward we'd like to find a way for developers to define these variables statically within GraphQL.

Globegitter:
Ah ok yeah that makes sense and sounds like a good path forward (we are actually using `substituteVariables` currently). Also is there a simple example for the `validator` option? Will update the changes asap.

Globegitter:
I removed the documentation of `substituteVariables`

josephsavona:
For `validator` it's useful if you want to use custom validation rules. Otherwise Relay uses most but not all of the default rules (have to skip some when validating fat queries).

Globegitter:
Updated the comment on the validator.

steveluscher:
@facebook-github-bot import

PR #1073, opened 2016-04-22
use const, let always in examples

chentsulin:


wincent:
Thanks!

wincent:
@facebook-github-bot import

PR #1075, opened 2016-04-22
Update to Jest 11.

cpojer:
Should be backwards compatible. Will land Jest 11 on fbsource next week.


cpojer:
@facebook-github-bot shipit

PR #1076, opened 2016-04-22
Added basic docs for Relay.Renderer

Globegitter:
It is public API but has not documentation whatsoever, which can be confusing as this could indicate that the RelayRenderer API is not (yet) meant to be used. This should make it easier to discover and more clear that the RelayRenderer API can be used.

Specifically the usage of `RelayRenderer` would solve some of our use-cases. Also having some documentation for this would mean that https://github.com/relay-tools/react-router-relay/pull/123 could be merged in.

Happy to add any changes etc. I just thought it would be better to have some documentation, rather than none.

### DIFF
--- a/docs/APIReference-Renderer.md
+++ b/docs/APIReference-Renderer.md
@@ -1,0 +1,187 @@
+---
+id: api-reference-relay-renderer
+title: Relay.Renderer
+layout: docs
+category: API Reference
+permalink: docs/api-reference-relay-renderer.html
+next: api-reference-relay-ql
+---
+
+**Relay.Renderer** is a replacement for `Relay.RootContainer` that composes a `Relay.ReadyStateRenderer` and performs data fetching for a given `queryConfig`. 
+
+## Overview
+
+*Props*
+
+<ul class="apiIndex">
+  <li>
+    <a href="#container">
+      <pre>Container</pre>
+      Relay container that defines fragments and the view to render.
+    </a>
+  </li>
+  <li>
+    <a href="#forcefetch">
+      <pre>forceFetch</pre>
+      Whether to send a server request regardless of data available on the client.
+    </a>
+  </li>
+  <li>
+    <a href="#onforcefetch">
+      <pre>onForceFetch</pre>
+      Called whenever forceFetch is being called
+    </a>
+  </li>
+  <li>
+    <a href="#onprimecache">
+      <pre>onPrimeCache</pre>
+      Called whenever primeCache is being called
+    </a>
+  </li>
+  <li>
+    <a href="#queryconfig">
+      <pre>queryConfig</pre>
+       QueryConfig or Route that defines the query roots.
+    </a>
+  </li>
+  <li>
+    <a href="#environment">
+      <pre>environment</pre>
+      The API for Relay core that provides methods for fetching/mutating data, an in-memory cache, etc.
+    </a>
+  </li>
+    <li>
+    <a href="#render">
+      <pre>render</pre>
+      Called to render when data requirements are being fulfilled.
+    </a>
+  </li>
+  <li>
+    <a href="#onreadystatechange">
+      <pre>onReadyStateChange</pre>
+    </a>
+  </li>
+</ul>
+
+## Props
+
+### Container
+
+```
+Container: RelayContainer
+```
+
+Must be a valid `RelayContainer`. Relay will attempt to fulfill its data requirements before rendering it.
+
+
+### forceFetch
+
+```
+forceFetch: boolean
+```
+
+If supplied and set to true, a request for data will always be made to the server regardless of whether data on the client is available to immediately fulfill the data requirements.
+
+
+### onForceFetch
+
+```
+onForceFetch(
+  querySet: RelayQuerySet,
+  callback: (readyState: ReadyState) => void
+): Abortable
+```
+
+This callback prop is called whenever a forceFetch occurs.
+
+
+### onPrimeCache
+
+```
+onPrimeCache(
+  querySet: RelayQuerySet,
+  callback: (readyState: ReadyState) => void
+): Abortable
+```
+
+This callback prop is called whenever a primeCache call occurs.
+
+### QueryConfig
+
+```
+queryConfig: RelayRoute
+```
+
+Either an instance of `Relay.Route` or an object with the `name`, `queries`, and optionally the `params` properties.
+
+
+### Environment
+
+```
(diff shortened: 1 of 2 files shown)


docs/APIReference-Renderer.md
```diff
@@ -0,0 +1,187 @@
...
+  <li>
+    <a href="#forcefetch">
+      <pre>forceFetch</pre>
+      Whether to send a server request regardless of data available on the client.
+    </a>
+  </li>
+  <li>
+    <a href="#onforcefetch">
+      <pre>onForceFetch</pre>
+      Called whenever forceFetch is being called
```
wincent: Needs trailing full-stop.

docs/APIReference-Renderer.md
```diff
@@ -0,0 +1,187 @@
...
+    </a>
+  </li>
+  <li>
+    <a href="#onprimecache">
+      <pre>onPrimeCache</pre>
+      Called whenever primeCache is being called
```
wincent: Trailing full stop.

docs/APIReference-Renderer.md
```diff
@@ -0,0 +1,187 @@
...
+Must be a valid `RelayContainer`. Relay will attempt to fulfill its data requirements before rendering it.
+
+
+### forceFetch
+
+```
+forceFetch: boolean
+```
+
+If supplied and set to true, a request for data will always be made to the server regardless of whether data on the client is available to immediately fulfill the data requirements.
```
wincent: Might read a little better if you simplify:

> whether data on the client is available to immediately fulfill the data requirements.

to:

> whether data on the client is available already.

docs/APIReference-Renderer.md
```diff
@@ -0,0 +1,187 @@
...
+### onForceFetch
+
+```
+onForceFetch(
+  querySet: RelayQuerySet,
+  callback: (readyState: ReadyState) => void
+): Abortable
+```
+
+This callback prop is called whenever a forceFetch occurs.
```
wincent: Can you backtick-ize `forceFetch` here?

docs/APIReference-Renderer.md
```diff
@@ -0,0 +1,187 @@
...
+### onPrimeCache
+
+```
+onPrimeCache(
+  querySet: RelayQuerySet,
+  callback: (readyState: ReadyState) => void
+): Abortable
+```
+
+This callback prop is called whenever a primeCache call occurs.
```
wincent: Backticks around `primeCache`.

docs/APIReference-Renderer.md
```diff
@@ -0,0 +1,187 @@
...
+Either an instance of `Relay.Route` or an object with the `name`, `queries`, and optionally the `params` properties.
+
+
+### Environment
+
+```
+environment: RelayEnvironment
+```
+
+Either `Relay.Store`, an instance of `Relay.Evironment` or an object with the `forceFetch`, `getFragmentResolver`, `getStoreData`, `primeCache`, and `read` function defined.
```
wincent: s/Evironment/Environment/

Rather than explicitly listing the methods (that may change over time), perhaps just say:

> An object that conforms to the `Relay.Environment` interface, such as `Relay.Store`.

docs/APIReference-Renderer.md
```diff
@@ -0,0 +1,187 @@
...
+render(
+  props: ?{[propName: string]: mixed},
+  done: boolean,
+  error: ?Error,
+  retry: ?Function,
+  stale: boolean
+): ?RelayRenderCallback
+```
+
+Whenever anything is rendered `render` is called to render the view. If the callback is not supplied, the default behavior is to render the container if data is available, the existing view if one exists, or nothing.
```
wincent: > Whenever anything is rendered `render` is called to render the view.

This sentence seems redundant to me. The word "render" is repeated three times, and I could reasonably expect this to be the behavior just based on the name, `render`.

docs/APIReference-Renderer.md
```diff
@@ -0,0 +1,187 @@
...
+    aborted: boolean;
+    done: boolean;
+    error: ?Error;
+    ready: boolean;
+    stale: boolean;
+  }
+): void
+```
+
+This callback prop is called as the various events of data resolution occurs.
```
wincent: Grammar: occurs -> occur

wincent:
Thanks for this. The omission of from the docs here was very much intentional, because we didn't want to churn the public API by throwing this out there before we were sure it was the right design.

It's been stable for a while now, though, and we've already pointed people to this API's existence in various places in order to help them solve problems, so I think the time is probably right to document this.

I left some minor nits inline. Let me know what you think.

As a follow-up to this, we should think about actually marking `RelayRootContainer` as deprecated, but I think that is beyond the scope we need to address here.

docs/APIReference-Renderer.md
```diff
@@ -0,0 +1,187 @@
...
+          }
+        } else {
+          if (renderLoading) {
+            return renderLoading();
+          }
+        }
+        return undefined;
+      }}
+/>
+```
```
wincent: Not sure how much value this example adds. Thinking forward to an imminent time when `RelayRootContainer` is deprecated and most people using Relay won'd have ever heard of it, the references to and comparisons with `RelayRootContainer` won't be super useful (in fact, they might be distracting and therefore detrimental).

Globegitter:
@wincet Good point about the `RelayRootContainer`. Removed its reference and addressed the comments.

wincent:
@facebook-github-bot import

wincent:
Thanks for this @Globegitter. This is landing internally and will get synced out soon. I made a couple of tweaks which you'll see reflected when it lands on master:

- Fixed whitespace errors (trailing whitespace).
- Missing backticks.
- Route -> Relay.Route.
- Killed off some excess blank lines.
- Wrapped long line in comment sample.
- Reworded Relay.Environment mention based on feedback.
- Removed reference to `onForceFetch` and `onPrimeCache`, which only exist to support an experimental internal feature (persisted queries), and we're not ready to publicly support yet.

denvned:
> Removed reference to `onForceFetch` and `onPrimeCache`, which only exist to support an experimental internal feature (persisted queries), and we're not ready to publicly support yet.

@wincent, I guess, since introduction of the `environment` prop to `RelayRenderer` there should be no need for `onForceFetch` and `onPrimeCache` extra props at all anymore, as now we can pass an `environment` with custom `forceFetch` and `primeCache`.

PR #1077, opened 2016-04-22
Fix test failures in RelayMutationQueue

wincent:
Sample failure run: https://travis-ci.org/facebook/relay/jobs/124994485

These look like harmless failures caused by 5943007d03b (D3210894). We're no longer setting the `tracker` property in our call to `buildQueryForOptimisticUpdate`, so the test shouldn't be including it in its assertions.


wincent:
@facebook-github-bot import

Issue #825, opened 2016-02-10
Idea: Remove the mutation GraphQL specification requirements?


ianstormtaylor:
One of the things I found annoying while trying to implement mutations is how much opinion Relay inflicts on my GraphQL schema—having to create `Input` and `Payload` definitions for each operation is really verbose, very Relay-specific, and not necessarily that “clean” of an API.

For example, ignoring Relay, it might be nice for a mutation in the GraphQL schema to look like:

```
type Mutation {
  createTeam(userId: ID!, team: JSON!): Team
}
```

That would be a fairly simple `createTeam` mutation, that is easy to understand. You pass the `userId` of the user you’re creating the team for, and a `team` JSON object containing team attributes (name, billing email, etc.). It maps fairly simply to a `POST /users/:userId/teams` call of traditional REST APIs (not that that’s a requirement, but REST does have a lot of good patterns).

Unfortunately, implementing that same mutation for Relay requires the GraphQL schema to be very Relay-specifc instead:

```
type Mutation {
  createTeam(input: CreateTeamInput!): CreateTeamPayload
}

input CreateTeamInput {
  userId: ID!
  team: JSON!
  clientMutationId: String!
}

type CreateTeamPayload {
  team: Team
  clientMutationId: String!
}
```

A lot more verbose, but also less easily used by other, non-Relay, GraphQL implementations now. Or at least without them having to adopt Relay’s naming opinions.

It seems like one reason the `Input` is used is so that creating the `$input` variable is as easy as possible for the client code. But it doesn’t seem like this should come at the expense of changing the schema itself. It seems like it would still be easy to implement arbitrary variable names, and have them be stringified properly by the query configuration logic in Relay. Such that passing something like:

``` js
{
  userId: 'user_1',
  team: {
    name: 'Team',
    email: 'team@example.com'
  }
}
```

Could be stringified into a mutation of:

```
mutation CreateTeam($userId: String, $team: JSON) {
  createTeam(userId: $userId, team: $team) ...
}
```

Another reason for the `Input` and `Payload` is to attach a `clientMutationId`, which the documentation suggests is  `used to reconcile requests and responses`. But this seems like something that is NetworkLayer specific, and could be better implemented as a separate parameter to any HTTP calls made to the server, and even something that might need to be use-case specific. 

[...]

PR #1083, opened 2016-04-25
Stop relying on clientRequestId in graphql response.

dylanahsmith:
For issue #825

I would like to fully remove a requirement to have clientMutationId field on mutation input or payload, but currently graphql-relay-js defines these fields as non-null.  This pull request changes the spec to say that the server should make these clientMutationId fields nullable as I have done in https://github.com/graphql/graphql-relay-js/pull/79.

In order to keep compatibility for servers with non-null clientMutationId fields, this pull request continues to send the clientMutationId in the mutation input to avoid validation errors.

However, there is no need to get the clientMutationId value from the server, since the client already has it associated with the pending mutation, so it no longer even requests the clientMutationId for the response.  Instead, the handleUpdatePayload and writeRelayUpdatePayload internal functions have been updated to take the clientMutationId in an additional argument.
## Follow-up

We should stop sending automatically sending clientMutationId in the future, which will be a breaking change.  However, the only change that would be needed to handle this breaking change would be to make the clientMutationId field nullable on the server or to manually specify a clientMutationId field on the client.  At that point we can remove the need to have a clientMutationId field from the relay spec.

### DIFF
--- a/src/mutation/__tests__/RelayMutationQuery-test.js
+++ b/src/mutation/__tests__/RelayMutationQuery-test.js
@@ -42,6 +42,11 @@ describe('RelayMutationQuery', () => {
       node => !node.getSchemaName || node.getSchemaName() !== 'source';
     return filterRelayQuery(RelayTestUtils.getNode(...args), filterCallback);
   }
+  function getMutationWithoutClientId(mutation) {
+    return mutation.clone(mutation.getChildren().filter(child =>
+      !child.getSchemaName || child.getSchemaName() !== RelayConnectionInterface.CLIENT_MUTATION_ID
+    ));
+  }
 
   let tracker;
 
@@ -1108,24 +1113,26 @@ describe('RelayMutationQuery', () => {
         mutation,
       });
 
-      const expectedMutationQuery = getNodeWithoutSource(Relay.QL`
-        mutation {
-          commentDelete(input:$input) {
-            ${Relay.QL`
-              fragment on CommentDeleteResponsePayload {
-                feedback {
-                  id
+      const expectedMutationQuery = getMutationWithoutClientId(
+        getNodeWithoutSource(Relay.QL`
+          mutation {
+            commentDelete(input:$input) {
+              ${Relay.QL`
+                fragment on CommentDeleteResponsePayload {
+                  feedback {
+                    id
+                  }
                 }
-              }
-            `},
-            ${Relay.QL`
-              fragment on CommentDeleteResponsePayload {
-                deletedCommentId
-              }
-            `},
-          }
-        }
-      `, variables);
+              `},
+              ${Relay.QL`
+                fragment on CommentDeleteResponsePayload {
+                  deletedCommentId
+                }
+              `},
+            }
+          `, variables
+        )
+      );
 
       expect(query)
         .toEqualQueryNode(expectedMutationQuery);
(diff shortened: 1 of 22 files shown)


wincent:
Thanks for the PR, @dylanahsmith.

The use of client mutation IDs comes from their extraction from Facebook's internal implementation, where they have existed for years (perhaps since the original implementation of mutations, although my memory doesn't extend back that far). Perhaps @dschafer can speak to why/how they arose in the design.

The deal-breaker here, however, is that Relay internals themselves assume a client mutation ID, so this would be a breaking change (as you note), and I am not sure we want to do that. Disclaimer: I haven't fully read #825 yet so I am not sure whether there is a rationale in there strong enough to overturn whatever reasons Dan (or somebody else) can share with us for why things currently are the way they are.

Adding the discussion label for now so we can get some more opinions on this.

dylanahsmith:
I have tried to keep this PR free fron breaking changes. My note about breaking changes was for a proposed follow-up.

dschafer:
> Perhaps @dschafer can speak to why/how they arose in the design.

So that clients could have a unique identifier to associate with mutations, and to link a mutation to its payload. It's also really nice for subscriptions, since you can determine if a given subscription payload is associated with a mutation that you yourself created.

I think I'd be fine with making this nullable, and clientMutationId will be null if you pass up null in the input. I'd still want to enforce that we always set it, but that can be a convention and not part of the spec.

davidcelis:
It's worth mentioning that in #258, changes were made to the Relay Specification that appear to allow for `clientMutationId` to be declared as either non-nullable _or_ nullable...

src/mutation/__tests__/RelayMutationQuery-test.js
```diff
@@ -43,6 +43,11 @@ describe('RelayMutationQuery', () => {
       node => !node.getSchemaName || node.getSchemaName() !== 'source';
     return filterRelayQuery(RelayTestUtils.getNode(...args), filterCallback);
   }
+  function getMutationWithoutClientId(mutation) {
```
steveluscher: If you taught `RelayQLPrinter` not to print `clientMutationId` for mutation queries, could you get rid of all of this?

steveluscher:
Thoughts on this, @josephsavona, @jkassens, @yuzhiz?

src/mutation/__tests__/RelayMutationQuery-test.js
josephsavona: Yes, let's stop adding this field in the plugin rather than filter it in tests. We do not strictly need to fetch the `clientMutationId` field.

josephsavona:
The overall direction looks good here: we can stop querying for this field. Let's make the change noted above in the plugin to stop adding `clientMutationId`/`clientSubscriptionId` fields on mutations/subscriptions. The main requirement is that we keep https://github.com/facebook/relay/pull/1083/files#diff-7fd56ba716b5f798f10ea57b840cc989R485 since our internal schema requires this field.

src/mutation/__tests__/RelayMutationQuery-test.js
dylanahsmith: Maybe it makes more sense to fetch the clientMutationId if the clientMutationId input field is non-null so that relay knows whether it needs to provide the clientMutationId in the input.  That would avoid any need for breaking changes.

josephsavona: That would work if we needed that field, but we don't so let's just remove it.

dylanahsmith:
Updated babel-relay-plugin to avoid adding clientMutationId to the query.

The clientMutationId is still automatically sent as input, so this will keep backwards compatibility.

Note that clientSubscriptionId has always been outside the scope of this PR.  It isn't a part of the relay spec and subscriptions aren't even a part of the GraphQL spec.  Feel free to open a PR if you feel similar changes should be done for subscriptions, but this PR is big enough already.

taion:
@dschafer Can you elaborate on how that "own mutation" association works? Generated `clientMutationId`s aren't unique across clients, so it doesn't seem like I can use the ID by itself to figure out whether a mutation was my own.

josephsavona:
Im going to close this. In a future version of Relay we hope to remove this restriction. Thanks for starting the discussion!

PR #1084, opened 2016-04-25
Update doc metadata

wincent:
I just regenerated the website in order to update the videos page. This meant that our metadata got updated to reflect a new page that we added.


wincent:
@facebook-github-bot import

PR #1091, opened 2016-04-27
Also test node v6 on travis

chentsulin:


wincent:
Thanks for this, @chentsulin. Looks like we might have some work to do to make things pass before we can merge this. Want to take a look at those failures?

chentsulin:
ok, i will take a look and try to figure out what happened.

wincent:
@chentsulin: check out @kassens diff links above. That *might* fix it, and is *at least* related.

kassens:
Yes, the `jest` update was required to make most of the tests pass, but there still seems to be some legit errors.

kassens:
Thanks for sending this out, I took the liberty to include this small change with #1126 which fixes the broken unit tests.

PR #1097, opened 2016-04-28
Added addNetworkSubscriber feature to changelog

Globegitter:
Assuming that this is public, since it is not marked as internal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.8.1 (April 27, 2016)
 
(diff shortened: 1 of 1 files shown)


wincent:
Thanks, @Globegitter!

This was added to support `RelayNetworkDebug` in v0.8.1. `RelayNetworkDebug` was itself added in v0.8.0. As such, it's recently been in a state of flux, which is why we didn't rush to document it. I think for now we should let this bake at least a couple more releases to be sure it won't need to change before mentioning it in any public docs.

Globegitter:
That is fair. As a note on that the revised `RelayNetworkDebug` is working like a charm so far in our app (the previous implementation did not).

Issue #1101, opened 2016-04-29
Relay not passing down fragment


Globegitter:
We are having a tree of components which has two separate fragments being passed down:

```js
// SomeComponent.js
fragments: {
      customer: () => Relay.QL`
        fragment on Customer {
          id,
          fullName,
          ...
        }
      `
    viewer: ({customerId}) => Relay.QL`
       fragment on Query {
         ${OtherComponent.getFragment('viewer', {customerId})}
       }
    `
}
```

```js
// OtherComponent.js
initialVariables: {
  customerId: null
}
fragments: {
  viewer: ({customerId}) => Relay.QL`
       fragment on Query {
         statement(customerId: $customerId) {
           edges {
             node {
               some
               fields
             }
           }
         }
       }
    `
}
```

The `customer` fragment is used to just displays a single customer that is fetched via the `node(id: $id)` query. The other fragment returns a list of `statements` that can (optionally) be selected just for a single customer. So from the `Relay.Route` we are passing all the way down the Relay ID for the customer (and on the backend we are converting it to our local id),

So on OtherComponent.js we want to display a single customer's statements which at a glance seems like it should just work.

Strangly though as it stands, when we access `this.props.viewer`  in `OtherComponent.js` we have an object, but apart from some private properties it is empty.

If now we change the `customerId` from being passed down to hardcoded (so: `... statement(customerId: valid_global_id) {...`) it all of the sudden works. we have all everything on `this.props.viewer` that we requested.

What is going on here? Is relay trying to do some matching and because it is seeing a valid id for a customer here and we are asking for something else than a customer it just freaks out?

This issue is also very hard to debug as there is no error, we actually get the expected response back it just seems to go 'missing' on the way.

josephsavona:
@Globegitter Can you paste the route queries as well? Anything else interesting between the route and the top fragment given above?

Globegitter:
We are actually using `react-router-relay` which injects the params from the route (https://github.com/relay-tools/react-router-relay#path-parameters) and we have that defined as:

```
const customerNodeQueries = {
  customer: () => Relay.QL`query { node(id: $customerId) }`,
  viewier: (Component, {customerId}) => Relay.QL`query { viewer(){ ${Component.getFragment('viewer', {customerId})} } }`
};
```

Nothing else interesting that I can think off. Though I am also just seeing there is a `prepareParams` in here: https://github.com/relay-tools/react-router-relay#additional-parameters Maybe that would have been a better way of passing down the variable?

[...]

PR #1104, opened 2016-04-30
Added missing docs for prepareParams to Relay.Route.

Globegitter:
Given the issue https://github.com/facebook/relay/issues/1101 and that there are no docs for `prepareParams` yet I thought I could do both at once.

I haven't actually tested this example out as such, but assuming that `prepareParams` in `react-router-relay` (https://github.com/relay-tools/react-router-relay/tree/6c36030e8bdfba7ca2e344a6436135610b6e2256#additional-parameters) do behave exactly the same as stated.

### DIFF
--- a/docs/APIReference-Route.md
+++ b/docs/APIReference-Route.md
@@ -125,7 +125,7 @@ const Container = Relay.createContainer(({viewer}) => (
   }
 });
 ```
-In this example query parameters are being passed down to the `Child` component: One comes from the `ProfileRoute` via `prepareParams` and given `ProfileRoute` and `Container` are both being passed into `Relay.Renderer`, it automatically passed it down. The `Container` then passes two different `limit` variables down to the `Child` one from the `Route` and one from within.
+In this example `prepareParams` overrides any `limit` variables the route gets initialised with. Given `ProfileRoute` and `Container` are both being passed into `Relay.Renderer`, that limit variable will automatically be passed down to `Container`.
 
 ### queries (static property)
 
--- a/docs/APIReference-Route.md
+++ b/docs/APIReference-Route.md
@@ -92,38 +92,19 @@ class ProfileRoute extends Relay.Route {
     viewer: () => Relay.QL`query { viewer }`
   };
   static prepareParams = (prevParams) => {
-    // Make sure any params that ProfileRoute gets initialised with will still be passed down.
     return {
+      // Ensure any params that ProfileRoute gets initialised with will still be passed down.
       ...prevParams,
+      //  Override the `limit` variable in the top level container.
       limit: 10
     }
   }
   // ...
 }
 
-const Child = Relay.createContainer((props) => ..., {
-  initialVariables: {
-    limit: null
-  },
-  fragments: {
-    viewer: ({limit}) => Relay.QL`
-      fragment on User {
-        friends(first: $limit) {
-          edges {
-            node {
-              name
-            }
-          }
-        }
-      }
-    `
-  }
-});
-
-const Container = Relay.createContainer((props) => (
+const Container = Relay.createContainer(({viewer}) => (
   <View>
-    <Child viewer={viewer} limit={this.props.relay.variables.limit} />
-    <Child viewer={viewer} limit={20} />
+    {viewer.edges.map((node) => <div>node.name</div>)}
   </View>
   ), {
   initialVariables: {
@@ -131,10 +112,14 @@ const Container = Relay.createContainer((props) => (
   },
   fragments: {
     viewer: ({limit}) => Relay.QL`
-      fragment on User {
-        name
-        ${Child.getFragment('viewer', {limit})}
-        ${Child.getFragment('viewer', {limit: 20})}
+      fragment on Viewer {
+        friends(first: $limit) {
+          edges {
+            node {
+              name
+            }
+          }
+        }
       }
     `
   }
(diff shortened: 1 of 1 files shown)


Globegitter:
I know this example is a bit overloaded, since it actually covers bits of the `Route` as well as the `Relay.Container` but it does kind of fit and would have at helped us when we were trying to solve that specific use-case.

docs/APIReference-Route.md
```diff
@@ -70,6 +76,72 @@ class ProfileRoute extends Relay.Route {
...
+  }
+  // ...
+}
+
+const Child = Relay.createContainer((props) => ..., {
+  initialVariables: {
+    limit: null
+  },
+  fragments: {
+    viewer: ({limit}) => Relay.QL`
```
josephsavona:
```
viewer: () => Relay.QL`...`
```

The `limit` argument is unnecessary here.

docs/APIReference-Route.md
```diff
@@ -70,6 +76,72 @@ class ProfileRoute extends Relay.Route {
...
+
+#### Example
+
+```
+class ProfileRoute extends Relay.Route {
+  static queries = {
+    viewer: () => Relay.QL`query { viewer }`
+  };
+  static prepareParams = (prevParams) => {
+    // Make sure any params that ProfileRoute gets initialised with will still be passed down.
```
josephsavona: nit: move this comment before or to the end of the `...prevParams` line? it's ambiguous what it's referring to here.

docs/APIReference-Route.md
```diff
@@ -70,6 +76,72 @@ class ProfileRoute extends Relay.Route {
...
+    return {
+      ...prevParams,
+      limit: 10
```
josephsavona: to help clarify the other comment, it might be good to comment this line too, maybe something like "and override the `limit` variable in the top level container.

docs/APIReference-Route.md
```diff
@@ -70,6 +76,72 @@ class ProfileRoute extends Relay.Route {
...
+        name
+        ${Child.getFragment('viewer', {limit})}
+        ${Child.getFragment('viewer', {limit: 20})}
+      }
+    `
+  }
+});
+```
+In this example query parameters are being passed down to the `Child` component: One comes from the `ProfileRoute` via `prepareParams` and given `ProfileRoute` and `Container` are both being passed into `Relay.Renderer`, it automatically passed it down. The `Container` then passes two different `limit` variables down to the `Child` one from the `Route` and one from within.
+
```
josephsavona: It's worth documenting this, but it doesn't really have to do with routes or prepareParams. What about adding a section to the RelayContainer guide (linked to from the Route guide) about variable passing?

Globegitter: Yeah that is a good idea, will update it that way asap.

Globegitter:
@josephsavona Have addressed most of your comments. Seeing as you have now more thoroughly documented the variable passing down (https://github.com/facebook/relay/pull/1142) I simplified the example.

I did however think that some of the other sections could do with some further explanation. For me it is often more easy to understand things based on an (explained) example.

Let me know if anything else needs fixing up.

docs/APIReference-Route.md
```diff
@@ -70,6 +76,57 @@ class ProfileRoute extends Relay.Route {
...
+      //  Override the `limit` variable in the top level container.
+      limit: 10
+    }
+  }
+  // ...
+}
+
+const Container = Relay.createContainer(({viewer}) => (
+  <View>
+    {viewer.edges.map((node) => <div>node.name</div>)}
```
akre54: Shouldn't this be `viewer.edges.map((edge) => <div>{edge.node.name}</div>)`?

Globegitter: Thanks for the catch. The one disadvantage of being fullstack and not having been able to write any meaningful JS in about a month.

Globegitter:
@josephsavona Anything still blocking this PR being merged?

docs/APIReference-Route.md
josephsavona: ping

josephsavona:
@Globegitter see my comment about the extraneous `limit` argument

josephsavona:
otherwise looks good

Globegitter:
Somehow missed. Just updated.

docs/APIReference-Route.md
```diff
@@ -70,6 +76,57 @@ class ProfileRoute extends Relay.Route {
...
+            }
+          }
+        }
+      }
+    `
+  }
+});
+```
+In this example `prepareParams` overrides any `limit` variables the route gets initialised with. Given `ProfileRoute` and `Container` are both being passed into `Relay.Renderer`, that limit variable will automatically be passed down to `Container`.
+
```
josephsavona: @Globegitter on second reading, this example conflates two things: `prepareParams` and passing route vars to the root container fragment.

Can you separate these? Note that `prepareParams` can be used for things like setting default values or converting from strings (e.g. in a URL) to integers. I'd recommend keeping this PR focused on `prepareParams` only, and maybe add a follow-up for the route -> root container variable example.

wincent:
Going to pick this up and polish off any pending details necessary to get this through internal code-review. Thanks, @Globegitter.

wincent:
@facebook-github-bot import

facebook-github-bot-3:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/503983499788689/int_phab) to review.

wincent:
@Globegitter: I landed the modified version of this, incorporating @josephsavona's suggestions:

> Can you separate these? Note that `prepareParams` can be used for things like setting default values or converting from strings (e.g. in a URL) to integers. I'd recommend keeping this PR focused on `prepareParams` only, and maybe add a follow-up for the route -> root container variable example.

Globegitter:
@wincent Sorry for the delay and thanks for making the last changes.

PR #1105, opened 2016-05-01
Fix formatErrors when offset is negative

xuorig:
I was looking at this stackoverflow question: https://stackoverflow.com/questions/36948277/relay-js-error-repeat-argument-must-be-greater-than-or-equal-to-0-and-not-be

It seems like if your server returns errors with column `0`, offset would be `-1`, and `repeat` would 💥  because `.repeat` takes a count argument which is
> An integer between 0 and +∞.

This fix is just `Math.max(0, offset)`.

I was looking for the spec for errors, I'm not totally sure if `0` is an accepted value for column in the first place so I might be missing context.

### DIFF
--- a/src/network-layer/default/__tests__/RelayDefaultNetworkLayer-test.js
+++ b/src/network-layer/default/__tests__/RelayDefaultNetworkLayer-test.js
@@ -174,6 +174,34 @@ describe('RelayDefaultNetworkLayer', () => {
       expect(error.source).toEqual(response);
     });
 
+    it('handles errors with column 0', () => {
+      const response = {
+        errors: [{
+          message: 'Something went wrong.',
+          locations: [{
+            column: 0,
+            line: 1,
+          }],
+        }],
+      };
+
+      fetch.mock.deferreds[0].resolve(genResponse(response));
+      jest.runAllTimers();
+
+      expect(rejectCallback.mock.calls.length).toBe(1);
+      const error = rejectCallback.mock.calls[0][0];
+      expect(error instanceof Error).toBe(true);
+      expect(error.message).toEqual([
+        'Server request for mutation \`FeedbackLikeMutation\` failed for the ' +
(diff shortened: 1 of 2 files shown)


src/network-layer/default/__tests__/RelayDefaultNetworkLayer-test.js
```diff
@@ -174,6 +174,38 @@ describe('RelayDefaultNetworkLayer', () => {
       expect(error.source).toEqual(response);
     });
 
+    it('handles errors with 0 column', () => {
```
josephsavona: nit maybe "...with column 0"?

src/network-layer/default/__tests__/RelayDefaultNetworkLayer-test.js
```diff
@@ -174,6 +174,38 @@ describe('RelayDefaultNetworkLayer', () => {
...
+          locations: [{
+            column: 0,
+            line: 1,
+          }],
+        }],
+      };
+
+      expect(fetch).not.toBeCalled();
+      networkLayer.sendMutation(request);
+      expect(fetch).toBeCalled();
```
josephsavona: these are tested elsewhere; probably okay to remove and just test the error message

xuorig:
Addressed your comments @josephsavona

kassens:
Thanks!

@facebook-github-bot import

facebook-github-bot-7:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/548900535290502/int_phab) to review.

sjchmiela:
+1 When is this going to be released?

wincent:
We're likely to cut a release this week, unless there's a show-stopper.

sjchmiela:
Great, thanks!

PR #1109, opened 2016-05-03
Build with babel 6

steveluscher:


steveluscher:
This builds well, and the website examples work, but the Jests don't pass. It's having trouble with `instanceof` checks. Seeing a lot of this:

```
Expected a `RelayQueryRoot`, got `RelayQueryRoot`.
```

https://gist.github.com/steveluscher/dd534fcd2790b2efa1ed40e06b92bc15

steveluscher:
Does removing `babel-runtime` from the build affect your builds at all, @skevy, @taion? I don't know how you build your apps.

taion:
I use `babel-polyfill` for all my application builds, so it wouldn't make a difference to me. Why drop `babel-runtime` though?

steveluscher:
> Why drop babel-runtime though?

I actually tried to drop `babel-polyfill` from the website examples, and discovered that `babel-runtime` alone was not enough to run them. It seems that `core-js/modules/es6.promise` doesn't implement `Promise.done()` – something that we use in Relay quite a bit. This got me thinking that we've been transforming the source with `babel-runtime` in service of nobody, this whole time.

Thoughts?

taion:
Isn't that orthogonal? I thought `Promise.done` came from the special Promise impl in fbjs.

There do seem to be calls to e.g. `Object.assign` in the code that `babel-runtime` would do something useful for, and I do see usages to things not polyfilled by `babel-runtime` like `Array.prototype.find` getting dropped: https://github.com/facebook/relay/issues/968

steveluscher:
>  I thought Promise.done came from the special Promise impl in fbjs.

Indeed. I just couldn't get the built product to stop referring to the `core-js` version that doesn't implement `Promise.done()`.

I think that `babel-polyfill` covers `Object.assign` and `Array.prototype.find`.

Also, this nightmare when you plug in the runtime transformer: https://gist.github.com/steveluscher/b6934e8cc2de8af6019ded4473f5d5ff

steveluscher:
OK, I put `babel-runtime` back, but I'm skeptical that it's doing what we need it to do. This whole system could use a look-over.

taion:
Oh gosh, yeah, that does sound messy. Maybe "bring your own ES6 environment" isn't such a bad approach, then.

`babel-polyfill` does cover everything that `babel-runtime` covers, but it's only helpful if people remember to use it 😛

steveluscher:
It is kind of messy.

I'm really tempted to strip out all the polyfills, and write up instructions on how to polyfill where necessary, but tonight is not the night.

taion:
Quite familiar with that pain: https://github.com/reactjs/react-router/pull/3331/files

It's sort of funny that it's easier to use new syntax, which transpiles down cleanly, than to use new standard library functionality, which instead requires a polyfill (or runtime, but then `babel-runtime` replaces even things like `Object.keys` with ugly core-js calls).

steveluscher:
Yeah. We still have use cases for `Object.assign` in Relay that object rest spread syntax doesn't cover (like mass assigning some properties without making a copy).

Vanuan:
Would there be a release soon?

Vanuan:
Is there a workaround?

steveluscher:
If we can get #1118 in this week, we'll do a release. cc/ @leebyron

PR #1125, opened 2016-05-06
Use async function for async test

kassens:
This is a followup on 581d87a to try out if async functions are supported on our
different targets to make some code more readable.


kassens:
I had this idea after seeing 581d87a. Open discussion if we want to use async function.

wincent:
Another option, which I am not sure Jest supports, is to use Jasmine's [standard async test mechanism](http://jasmine.github.io/2.0/introduction.html#section-Asynchronous_Support) instead of non-standard `pit`. They all end up producing pretty similar-looking code, so I don't feel too strongly about it.

kassens:
Doesn't seem worth the change.

PR #1126, opened 2016-05-06
Fix unit test for node 6 and add node 6 to travis

kassens:
The format of the JSON token error changed slightly with node 6:
```
node < 6: Unexpected token @
node   6: Unexpected token @ in JSON at position 0
```


kassens:
@facebook-github-bot import

PR #1139, opened 2016-05-13
Upgrade fbjs preset

zpao:
This removes a bunch of overlapping dependencies now that we do
everything you need in the fbjs preset.

This is pending the actual release of v2 of the preset, which hasn't happened yet - I wanted to try this and see what the outcome would be. Luckily nothing really needed to be done :)

Here's the diff of `lib/`: https://gist.github.com/zpao/79879855ac21d12094087a754daa7fa4. Overwhelmingly the difference is that have the function name transform on now. Otherwise there are a couple ordering differences with the helpers and then L1620 (not sure what's happening there)

It looks like you weren't actually making use of anything that was in stage-1, and es2015 wasn't even used, so got rid of those.


wincent:
Yay! Thanks.

zpao:
Bot comment incoming saying I updated this. I shipped `babel-preset-fbjs@2.0.0-alpha.1` last night and updated package.json to use that. It's just a devDep that we control so not a big deal to use that but I'll work to ship final. I wanted to get you and Draft using it and then ship final after a week in case anything shakes out.

kassens:
Seems good to ship!

steveluscher:
@facebook-github-bot shipit

steveluscher:
I'll remove the -alpha in a follow up.

facebook-github-bot-2:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1599338557047738/int_phab) to review.

steveluscher:
Hey @zpao, I wasn't able to get the `fbjs` preset to work with the example apps (that use object spread, and `import` syntax). Maybe that's just the way it is, but even when I stacked up `['es2015', 'stage-0', 'fbjs']` Babel choked on JSX.

PR #1155, opened 2016-05-20
Improve type definition for Higher Order Component

nmn:
WIP: I'll probably have to make more changes to make this work correctly. Creating the PR so I can get early feedback.

The community at large has not been able to property define types for higher-order components for a long time

PropTypes on React Components are extremely important. They are perhaps, one of the most important types in a React App.
Higher Order Components are extremely common in React Apps.
Therefore, the fact that Higher Order Components eat the type definition of React Components is unacceptable.

After days of research and trial and error, I've found the type definitions that work.

This PR, adds the 'correct' (\* see footnote) type definitions for the HOC, so that when the time comes to publish the
flow type definition file(s) to NPM, Relay.CreateContainer does not eat type definitions.
- technically speaking, Relay.CreateContainer takes all the props that match fragment names and hydrates the values using ids.
  So the types aren't completely correct, but it's close enough. Perhaps the return value should be something like:
  `React$Component<DefaultProps, Props & {[key: $Keys<FragmentObj>]:{__DATA_ID__: string}}, State>`

### DIFF
--- a/src/container/RelayContainer.js
+++ b/src/container/RelayContainer.js
@@ -84,10 +84,10 @@ const containerContextTypes = {
  *  - Propagate the `route` via context (available on `this.props.relay`).
  *
  */
-function createContainerComponent(
-  Component: ReactClass<any>,
+function createContainerComponent<DefaultProps, Props, State, C: React$Component<DefaultProps, Props, State>>(
+  Component: Class<C>,
   spec: RelayContainerSpec
-): RelayContainerClass {
+): React$Component<DefaultProps, Props, State> {
   const ComponentClass = getReactComponent(Component);
   const componentName = getComponentName(Component);
   const containerName = getContainerName(Component);


steveluscher:
Phoning a friend. @samwgoldman, your thoughts on this?

samwgoldman:
1. It's not true that the `State` of the returned component class is the same as the state of the input component class. This is fine, though, since state is a purely internal notion and callers don't care about the state. Personally I'd remove the `State` type parameter and replace it's instances with `any`.

1. The return type `React$Component<DefaultProps, Props, State>` describes instances of such a class, but this function returns a class. This is what `ReactClass<Config>` is for, where `Config = $Diff<Props, DefaultProps>`.

1. There are actually a few places inside the function that might be updated to use more specific types. In particular `_getQueryData` takes the props (actually the config, because the inner component will supply defaults), but is currently annotated as `Object`.

Does this work?

```
function createContainerComponent<Config>(
  Component: ReactClass<Config>,
  spec: RelayContainerSpec
): ReactClass<Config> {
  /* ... */
}
```

nmn:
@samwgoldman Thanks for the quick reply:

1. I agree. I can probably replace both the places is State is used with `any`.
2. Oops. That's a typo. I've been doing that right in so many other places.
3. I agree. Guess should have put *WIP* in bold. I had been frustrated with not having types for Higher Order Components as that is pretty important in a React App. After trying many different things I got something working in my project locally. So I just wanted to quickly create a PR to remind myself to make PRs to share it as I'm guessing others have the same frustration.

(Flow is pretty awesome to be able to handle it BTW!)

As for the type definition. I've tried the `ReactClass<Config>` form before for higher-order functions, but it's never worked. In the case of Relay, it may actually work as from a user point of view the Class appears unchanged. But in most cases, the higher order component needs to merge some props in.

And since I have your attention for a second, here's some other type definition work I've been doing for Higher Order Components:

- https://github.com/facebook/flow/pull/1821
- https://github.com/reactjs/react-redux/pull/389

Would love to have a conversation about a few missing features over in Discord/flow PR thread.

steveluscher:
Ping! Anything left to do here folks, or should we close this out?

samwgoldman:
I wouldn't take this as-is, but might be worth seeing if the type I recommended would work.

josephsavona:
Marking as needs revision, per @samwgoldman's suggestions.

nmn:
@samwgoldman the ReactClass types don't actually work most of the time.  I'll try to see if it works now with flow 0.30

chollier:
Just came across this PR after spending the day trying to figure out how to type my exported RelayContainers.

I concur with @nmn, `ReactClass<Config>` indeed doesn't work but :
```patch
-function createContainerComponent(
-  Component: ReactClass<any>,
+function createContainerComponent<DefaultProps, Props, C: React$Component<DefaultProps, Props, any>>(
+  Component: Class<C>,
    spec: RelayContainerSpec
-): RelayContainerClass {
+): Class<C> {
```

would indeed be super useful.

@nmn note that the returned type needs to be a `Class<C>` or `Class<React$Component<DefaultProps, Props, any>>`

josephsavona:
Closing due to inactivity. Thanks for the PR, we'll keep this in mind going forward!

PR #1161, opened 2016-05-22
extends: .eslintrc to .eslintrc.js

iamchenxin:
use fbjs's new .eslintrc.js


kassens:
@facebook-github-bot import

kassens:
Thanks, there might be some lint regressions that we should fix as a followup.

PR #1163, opened 2016-05-23
RelayDefaultNetworkLayer, throw Error also on non-200 response.

Globegitter:
Right now the mutation transacion gets its error attribute set differently on server-side non-200 responses  (e.g. 500) compared to 200 responses that include the `errors` attribute in the json. On the latter it is an `Error` object as described in the docs but on the first we still get the normal fetch `Response`. (So https://facebook.github.io/relay/docs/api-reference-relay-store.html#commitupdate-static-method is not completely right)

That also means handling these two different cases in the mutation `onFailure` call requires two very different pieces of logic. 

This PR now makes it so that transaction.getError() always returns an `Error`, and if it was a non-200 response it also includes that status code.


.eslintrc
```diff
@@ -2,7 +2,7 @@
 parser: babel-eslint
 
 extends:
-  - ./node_modules/fbjs-scripts/eslint/.eslintrc
+  - ./node_modules/fbjs-scripts/eslint/.eslintrc.js
```
wincent: This is going to conflict with https://github.com/facebook/relay/pull/1161

Can you please revert this bit or rebase and then we can import this.

wincent:
Thanks @Globegitter. This looks pretty good to me. One comment inline though, otherwise looks good to import.

Globegitter:
@wincent Oh yeah sorry forgot to remove that again. Just rebased.

steveluscher:
@facebook-github-bot import

facebook-github-bot-0:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/110971109322117/int_phab) to review.

src/network-layer/default/__tests__/RelayDefaultNetworkLayer-test.js
```diff
@@ -229,6 +237,31 @@ describe('RelayDefaultNetworkLayer', () => {
...
+        errors: [{
+          message: 'Something went completely wrong.'
+        }],
+      };
+
+      expect(fetch).not.toBeCalled();
+      networkLayer.sendMutation(request);
+      expect(fetch).toBeCalled();
+      const failureResponse = genFailureResponse(response);
+      expect(failureResponse.status >= 300 || failureResponse.status < 200).toBeTruthy();
```
steveluscher: Since `failureResponse` isn't part of the network layer, let's not make assertions about it in the test body.

src/network-layer/default/RelayDefaultNetworkLayer.js
```diff
@@ -163,7 +163,11 @@ function throwOnServerError(response: any): any  {
   if (response.status >= 200 && response.status < 300) {
     return response;
   } else {
-    throw response;
+    return response.text().then(text => {
```
steveluscher: How would you feel about sharing the code that already handles mutation errors (see `if (payload.hasOwnProperty('errors')` of `sendMutation()`)? Perhaps we should trigger the same error handling code when `errors` is present in the payload or when the status is non-200.

steveluscher:
Ping! I left some comments in the code that came up after an internal code review.

Globegitter:
@steveluscher sorry about the delay, was a bit busy and then went on holiday (well still am). Just looked over sharing the code and we can definitely do something there, but they are also slightly different scenarios (i.e. a non-200 response is might to not return json).

Overall I think I am for it so I'll update the PR asap.

Globegitter:
Only typecheck is failing so good to review again @steveluscher

src/network-layer/default/RelayDefaultNetworkLayer.js
```diff
@@ -132,7 +122,7 @@ class RelayDefaultNetworkLayer {
         method: 'POST',
       };
     }
-    return fetch(this._uri, init).then(throwOnServerError);
+    return fetch(this._uri, init).then(throwOnServerError.bind(this, request));
```
steveluscher: Why `.bind(…)`? I couldn't see a reference to `this` inside that function body.

steveluscher: OH. The `request` argument. I see.

Globegitter: Yeah, it wasn't a 100% ideal but that seemed to me the best way to share the error creation in all three places (or of course an arrow function if that has any advantages).

PR #1167, opened 2016-05-25
If fetch error-ed but has a partial response populate the cache.

Globegitter:
Title says it all.

We had this case in one of our apps today where some deeply nested data had an issue (and threw an error) but we could have displayed 90% of the ui with just a local error message in the part of the ui where the error is relevant / the data would actually be missing.

So now, if the response has an error but also has some data both the `error` as well as the `props` will get populated (with whatever data it does have), making it possible to increase usability of our apps in certain cases.

Have done some rudimentary testing in our app and it seems to work.


Globegitter:
Merge in master and fixed tests, so is no ready for review.

josephsavona:
Thanks for the PR. This is tricky, as different producys may have different expectations about whether a query should resolve or error in the presence of `errors`. A workaround is to use a custom network layer that simply strips out the errors field and resolves the query normally.

Globegitter:
@josephsavona This PR should have allowed both use-cases you described. We have the same case in our app, sometimes we might want to show a partial UI and sometimes not (i.e. keep the old behaviour).

Why can there not be a behaviour to have both? It seems to me also the most correct thing to do (as this is graphql default), you got some data but also some error, figure out what you want to do with it . And right now with relay we have the data striped out so can not do anything with it.

The problem with stripping out the errors field would be that we would loose any specific error details.

leebyron:
I'm so sorry your PR has sat here this long. Unfortunately it does not merge well anymore and we're attempting to focus on Relay Modern going forward.

I'm going to close this PR, however if this is still important to you please rebase it and reopen

PR #1173, opened 2016-05-27
Fix erroneous warning about already injected network layer

LegNeato:
Issue likey caused by https://github.com/facebook/relay/commit/173ca148f07f5df069fe098d88fa642ec530f874.

### DIFF
--- a/src/__forks__/Relay.js
+++ b/src/__forks__/Relay.js
@@ -27,7 +27,7 @@ if (__DEV__) {
 
 // By default, assume that GraphQL is served at `/graphql` on the same domain.
 // To override, use `Relay.injectNetworkLayer`.
-RelayStore.injectDefaultNetworkLayer(new RelayDefaultNetworkLayer('/graphql'));
+RelayStore.injectDefaultNetworkLayer();
 
 module.exports = {
   ...RelayPublic,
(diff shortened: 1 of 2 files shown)


LegNeato:
This issue is that `injectDefaultNetworkLayer` and `injectNetworkLayer` keep track of if they have injected or not separately. If we use `injectNetworkLayer` to inject the default, relay sees a layer already injected when an end user user calls `injectNetworkLayer` and throws a warning (because it thinks it is overwriting instead of replacing the default).

wincent:
I don't think we ever intended any external users to use `injectDefaultNetworkLayer` (see ed8d28ee706b31). Like I said on that commit:

> I had previously merely refrained from documenting this method which I don't intend anybody but us to use, but we can go further and literally remove it from the public API.

What is your use case for wanting to override the default *and* have somebody else override your override?

LegNeato:
I don't have a use case for that 😄 .

In my `app.js` I am merely calling `injectNetworkLayer` with my custom networking layer **once** and seeing the warning. Debugging shows that the first call that sets the instance var is in relay code and the second call is in my code for my custom layer. This seems like a bug but perhaps I am misunderstanding?

wincent:
Unless you're running 0.9.0 (not released yet) or `master`, this overzealous warning is expected, because it got fixed in d1abc54d5ab6 which is only on `master`. Sorry about that!

LegNeato:
Ohhhh, I didn't see the fix and didn't catch the the line I patched was different than the one on master! You are totally right, already fixed on master. 👍 

My `instanceof` check might be good for hardening against future bugs but isn't critical. Closing.

PR #1184, opened 2016-05-28
Fix flow error by fixing the site scoping of suppression comments

gabelevi:
We use the `site=` part of the suppression comment to scope a suppression
comment to a specific root. So for example, we could add a comment like

    // $FlowFixMe(site=relay)

and it should only work when you're typechecking relay. Before this change,
however, relay was accepting suppression comments with any site. After, relay
will only accept suppression comments like

    // $FlowFixMe - This one has no site
    // $FlowFixMe(site=relay) - only relay
    // $FlowFixMe(site=bar,baz,relay,foo) - relay and some others

### DIFF
--- a/src/.flowconfig
+++ b/src/.flowconfig
@@ -26,8 +26,8 @@ suppress_type=$FlowIssue
 suppress_type=$FlowFixMe
 suppress_type=$FixMe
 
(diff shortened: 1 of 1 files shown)


gabelevi:
@facebook-github-bot shipit

PR #1193, opened 2016-05-31
Remove some more unnecessary indirection working with Deferred instances

wincent:
We don't need to call `getPromise()` on `Deferred` in order to invoke `catch()`; we can just invoke `catch()` directly.

I did some of this a while back in D3183390, but since then, [[ https://github.com/facebook/fbjs/commit/e3cd4970f5 | e3cd4970f5 ]] landed in fbjs, adding a `catch` method to `Deferred`, and as of [[ https://github.com/facebook/relay/commit/d653f0e5443 | d653f0e5443 ]] we're using the latest release of fbjs in Relay, so we can make use of it.

### DIFF
--- a/src/mutation/__tests__/RelayGraphQLMutation-test.js
+++ b/src/mutation/__tests__/RelayGraphQLMutation-test.js
@@ -83,7 +83,7 @@ describe('RelayGraphQLMutation', () => {
     queue = storeData.getMutationQueue();
     sendMutation = jest.fn(request => {
       requests.push(request);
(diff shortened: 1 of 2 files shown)


wincent:
@facebook-github-bot import

facebook-github-bot-7:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/476822405858499/int_phab) to review.

PR #1194, opened 2016-05-31
Update the contributors' guidelines to mention Stack Overflow and the Relay Playground

steveluscher:


steveluscher:
@facebook-github-bot import

PR #1216, opened 2016-06-15
Fix getRangeBehavior to allow non-string calls

### LINKED ISSUE #1208: rangeBehaviors are too strict
v0.9 introduces a way to declare `rangeBehaviors` as a function (56e52d38c6adc77bcc83926993ae39dec51bb100). This is super useful and I cannot thank you enough for this! However, they introduces a new restriction:

```js
// getRangeBehavior.js#L75
invariant(
  typeof behavior === 'string',
  'getRangeBehavior(): Expected range behavior for key `%s` to be a ' +
  'string, got `%s`.',
  call.name,
  behavior
);
```

This is an issue if your connection is defined like this:

```graphql
enum PostOrder {
  DATE_ASC
  DATE_DESC
  TITLE_ASC
  TITLE_DESC
}

type Blog implements Node {
  id: ID!
  title: String
  posts(
    first: Int
    after: ID
    order: PostOrder
    since: Int
  ): BlogPostConnection
}
```

The `invariant` test may fail on valid arguments. The following `calls` would raises an error:

```js
[
  // `order` holds an enum value.
  // Passing the value as a string is fine.
  // However, passing `null` would throws.
  {
    name: "order",
    value: null,
  },
  // `since` holds a timestamp.
  // Converting it to a string would result in an invalid query but numbers are allowed!
  {
    name: "since",
    value: 1465542134907,
  }
]
```

`rangeBehaviors` object notation isn't affected by this issue because `calls` values are serialized.

xuorig:
Fixes #1208

That `invariant` doesn't seem to make sense in that function, removing it will allow us to use `rangeBehavior` functions with other types than string.

@josephsavona

### DIFF
--- a/src/mutation/getRangeBehavior.js
+++ b/src/mutation/getRangeBehavior.js
@@ -70,15 +70,7 @@ function getObjectFromCalls(
 ): {[argName: string]: string} {
   const behaviors = {};
   calls.forEach(call => {
-    const behavior = call.value;
-    invariant(
-      typeof behavior === 'string',
-      'getRangeBehavior(): Expected range behavior for key `%s` to be a ' +
-      'string, got `%s`.',
-      call.name,
-      behavior
-    );
-    behaviors[call.name] = behavior;
+    behaviors[call.name] = call.value;
   });
   return behaviors;
 }
--- a/src/mutation/__tests__/getRangeBehavior-test.js
+++ b/src/mutation/__tests__/getRangeBehavior-test.js
@@ -28,6 +28,14 @@ describe('getRangeBehavior()', () => {
       const rangeBehavior = getRangeBehavior(rangeBehaviors, calls);
       expect(rangeBehavior).toBe('append');
     });
+
+    describe('when calls contain non-string values', () => {
+      it('returns the rangeBehavior for the calls', () => {
+        const calls = [{name: 'status', value: 1}];
+        const rangeBehavior = getRangeBehavior(rangeBehaviors, calls);
+        expect(rangeBehavior).toBe('refetch');
+      });
+    });
   });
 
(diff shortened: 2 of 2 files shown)


kassens:
Thanks for the fix! I checked the internal discussion and it indeed seems like an oversight.

Would you mind adding a test case to [`getRangeBehavior`](https://github.com/facebook/relay/blob/master/src/mutation/__tests__/getRangeBehavior-test.js)?

xuorig:
Added a regression test @kassens !

src/mutation/getRangeBehavior.js
```diff
@@ -70,15 +70,7 @@ function getObjectFromCalls(
...
-    const behavior = call.value;
-    invariant(
-      typeof behavior === 'string',
-      'getRangeBehavior(): Expected range behavior for key `%s` to be a ' +
-      'string, got `%s`.',
-      call.name,
-      behavior
-    );
-    behaviors[call.name] = behavior;
+    behaviors[call.name] = call.value;
```
josephsavona: the return type of this function has to be changed to `{[argName: string]: mixed}`, since that's the type of `call.value`. not sure why flow isn't finding this.

josephsavona:
@xuorig see comment about the return type

kassens:
I'll fix up the return type before merging this.

@facebook-github-bot import

facebook-github-bot-0:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1719125375028451/int_phab) to review.

xuorig:
Thanks @kassens !

keithpitt:
I applied this patch locally to my copy of Relay and it fixed the problem for me. Thanks for fixing @xuorig!

PR #1223, opened 2016-06-19
Print subscriptions

edvinerikson:
I am not sure that we want to have this support but I think it would be nice.

I replaced the `printMutation` function with a `printOperation` function that can print any instance of `RelayQuery.Operation`. I am not sure if that's the behavior we want or if we should check for `RelayQuery.Subscription` and `RelayQuery.Mutation`.

Also as it is right now we always have to supply a `$input` argument which may not be the case when using subscriptions.

### DIFF
--- a/src/traversal/printRelayOSSQuery.js
+++ b/src/traversal/printRelayOSSQuery.js
@@ -61,8 +61,8 @@ function printRelayOSSQuery(node: RelayQuery.Node): PrintedQuery {
   let queryText = null;
   if (node instanceof RelayQuery.Root) {
     queryText = printRoot(node, printerState);
-  } else if (node instanceof RelayQuery.Mutation) {
-    queryText = printMutation(node, printerState);
+  } else if (node instanceof RelayQuery.Operation) {
+    queryText = printOperation(node, printerState);
   } else if (node instanceof RelayQuery.Fragment) {
     queryText = printFragment(node, printerState);
   }
@@ -121,10 +121,8 @@ function printRoot(
     oneIndent + fieldName + children + newLine + '}';
 }
 
-function printMutation(
-  node: RelayQuery.Mutation,
-  printerState: PrinterState
-): string {
+function printOperation(node: RelayQuery.Operation, printerState: PrinterState): string {
+  const operationKind = node.getConcreteQueryNode().kind.toLowerCase();
   const call = node.getCall();
   const inputString = printArgument(
     node.getCallVariableName(),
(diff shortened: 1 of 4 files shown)


edvinerikson:
Tests seems to fail because of this:
```
[11:42:33] Starting 'dist:min'...
fs.js:887
  return binding.lstat(pathModule._makeLong(path));
                 ^
Error: ENOENT: no such file or directory, lstat '/home/travis/build/facebook/relay/relay.js'
```

However I ran `npm test` locally and all tests passed (Node 6.2.2).

src/traversal/printRelayOSSQuery.js
```diff
@@ -121,10 +121,8 @@ function printRoot(
     oneIndent + fieldName + children + newLine + '}';
 }
 
-function printMutation(
-  node: RelayQuery.Mutation,
-  printerState: PrinterState
-): string {
+function printOperation(node: RelayQuery.Operation, printerState: PrinterState): string {
+  const operationKind = node.getConcreteQueryNode().kind.toLowerCase();
```
josephsavona: This is reasonable and will work, but we prefer not to rely on the `kind` to correspond to a printable value (we might change it to be an integer/enum, for example). Maybe do an `instanceof` check and make the string value explicit for each subclass?

edvinerikson: Yeah, maybe add `operationKind` as an argument and call the `printOperation` explicitly for `Mutation` and `Subscription` (instead of calling it for every `Operation`)?

`printOperation('subscription', node, state)`?

josephsavona: The text to print is an implementation detail, so how about passing the RelayQueryOperatiom object and add a ternary for mutation/subscription?

edvinerikson: Yeah that might work, but then I would also want to change `node: RelayQuery.Operation` to `node: RelayQuery.Mutation | RelayQuery.Subscription` because I don't think it make any sense to send anything else into it since we only support Subscription and Mutation.

what do you think about that?

edvinerikson:
I have updated the PR. :+1:

wincent:
Looks good apart from one long line that needs to be wrapped; can fix that during the import. I think the Travis failures might be fixed by #1226, maybe? Not sure exactly what the nature of the build breakage that that one refers to is.

wincent:
@facebook-github-bot import

facebook-github-bot-3:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/295789987427863/int_phab) to review.

edvinerikson:
Thanks @wincent and @josephsavona for accepting the PR! :+1:

PR #1227, opened 2016-06-21
Remove examples

wincent:
We'll be moving these into another repo soon:

https://github.com/relayjs/relay-examples

At the time I write this, that repo contains an exact copy of the examples that I am removing in this commit, but it will be a few days(?) before I can actually flip the switch and make it public.

As such, in this commit I am removing the examples and updating all references to them to instead point to the last commit at which the examples existed. I will follow-up afterwards and update the links to point to the new repo.

Q: Why are we doing it in this order?

A: Because we want to decouple the examples from the main repo in order to make updating them independent of the release process for Relay itself. This will reduce the friction to shipping new versions of Relay and help us to ship more often, starting with the very next release that I am preparing right now.


wincent:
Whoops. Need to update Travis as well. My `ag` failed me.

wincent:
@facebook-github-bot import

facebook-github-bot-6:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1755868961292350/int_phab) to review.

cgarvis:
It's been 2 day and the examples are still not up.  They are some of the best examples for newbies to figure out how relay works.  This should have probably waited until you could make the other repo public 😦

cgarvis:
For those that find this, you can browse the code base before this commit here: https://github.com/facebook/relay/tree/081b4a3f17dcf37419155317ed97ca5e5ac97522

wincent:
Release is a multi-step, non-atomic process, and I did not want it to be blocked by something I don't control (the administrative/legal side of toggling the new repo from private to public).

Also, as you have discovered, Git history is immutable and the examples are all still in the repo. Additionally I link directly to the last commit where the examples exist in multiple places. See https://github.com/facebook/relay/commit/c73d91ce36819c0f9cafd36481ceeb628de25146.

wincent:
@cgarvis: the examples repo is public now. Sorry that it took as long as it did.

Issue #1061, opened 2016-04-18
Loosening object identification GraphQL spec


calebmer:
On a project I'm working on we are automatically generating a GraphQL server from a SQL server and many of our tables have an integer column named `id`. When serving this table through GraphQL we would like to make the `id` column queryable, however we also want our rows to implement a `Node` type in accordance to the Relay GraphQL spec. According to the [spec](https://facebook.github.io/relay/graphql/objectidentification.htm#sec-Node-root-field.Introspection) an `id` field that is globally unique is required for compliance. This puts us in a binding position. Do we not allow columns to be named `id`? Do we rename `id` to `rowId` in GraphQL? These solutions aren't ideal as we also have many more foreign key columns like `authorId` which reference an `id` primary key.

The optimal solution would be if Relay allowed the `id` field to be renamed, either to `_id` or maybe `gid` (for global ID). Would it be possible to update the specification and Relay to allow for this? Is there a solution I'm not seeing?

wincent:
I'd be reluctant to incur the internal complexity of making this kind of thing configurable, especially given that a work around is possible (exposing the underlying database's `id` field using another field name).

But leaking the underlying primary key sounds like a leaky abstraction to me anywa. I'd recommend that you only expose and use "global" IDs in your GraphQL schema (effectively `base64encode(type + ':' + databaseID)`). You shouldn't need to ever leak foreign keys like `authorID` in your schema because you'll typically be querying for them compositionally (ie. `viewer { books { author } }`) or via the `node` root field (ie. `node(id: "someGlobalID")`).

calebmer:
The difficulty comes when introducing compound keys. There is no way (in my case) to intelligently compact the fields into a single field name *and* hiding these fields could also cause problems.

Another layer of complexity is added when these compound primary keys are also foreign keys.

There are some consumers that would greatly benefit from being able to access the data in the same form it is stored and other consumers (*cough* Relay *cough*) which would benefit from the `Node` format.

wincent:
> The difficulty comes when introducing compound keys. There is no way (in my case) to intelligently compact the fields into a single field name and hiding these fields could also cause problems.
>
> Another layer of complexity is added when these compound primary keys are also foreign keys.

Can you provide a concrete example of these cases in a bit more detail, to ground the discussion? I want to make sure I fully understand the exact nature of the problem that you're facing.

calebmer:
[...]

PR #1232, opened 2016-06-23
Adds support for custom Node ID fields.

alloy:
As described [here](https://github.com/facebook/relay/issues/1061#issuecomment-225940487), we need to use a different field for Relay’s Global Object Identification purposes.

This patch adds that support which I have tested to work [as expected in our app](https://github.com/artsy/emission/issues/181#issuecomment-227737481). However, it should be fully backwards compatible, as can be judged from the test coverage.

When/if this is accepted, I will make a PR for [the graphql-relay utilities](https://github.com/alloy/graphql-relay-js/commit/efb893df0d52f59997c80a3e6a1cace08eb63580) to define the schema with a custom Node ID.


scripts/babel-relay-plugin/src/RelayQLAST.js
```diff
@@ -709,6 +709,37 @@ class RelayQLType {
...
+        let nodeInterface = this.schemaUnmodifiedType === node && this;
+        if (!nodeInterface) {
+          nodeInterface = find(this.getInterfaces(), (iface) => {
+            return iface.schemaUnmodifiedType === node;
+          });
+        }
+        if (!nodeInterface) {
+          console.warn(`Using ${field} of ${this.schemaUnmodifiedType.name} ` +
+                       'as Node ID, even though it does not explicitely ' +
+                       'implement the Node interface.');
```
alloy: The test suite currently includes this warning:

```
Using relayID of NodelessUser as Node ID, even though it does not explicitely implement the Node interface.
```

scripts/babel-relay-plugin/src/RelayQLAST.js
```diff
@@ -709,6 +709,37 @@ class RelayQLType {
     };
     return new RelayQLFragment(this.context, (generatedFragmentAST: any), this);
   }
+
+  getNodeIdField(): ?string {
+    const node = this.context.schema._typeMap['Node'];
+    if (node && node instanceof types.GraphQLInterfaceType) {
+      const fields = node._fields;
```
alloy: Could you point me to where I should look for the correct API to access this GraphQL data, instead of `_typeMap` and `_fields`?

scripts/babel-relay-plugin/src/RelayQLAST.js
```diff
@@ -709,6 +709,37 @@ class RelayQLType {
...
+      if (field && this.hasField(field)) {
+        let nodeInterface = this.schemaUnmodifiedType === node && this;
+        if (!nodeInterface) {
+          nodeInterface = find(this.getInterfaces(), (iface) => {
+            return iface.schemaUnmodifiedType === node;
+          });
+        }
+        if (!nodeInterface) {
+          console.warn(`Using ${field} of ${this.schemaUnmodifiedType.name} ` +
+                       'as Node ID, even though it does not explicitely ' +
```
wincent: nit: s/explicitely/explicitly/

scripts/babel-relay-plugin/src/RelayQLAST.js
alloy: Never mind. Found it and obvious in hindsight (I’m not up-to-date yet with all JS conventions).

josephsavona:
This basically looks good, but this is actually step two - shipping it now would allow people to configure the plugin in a way that wouldn't actually work with the runtime. Before we proceed with this, we'll need to change the runtime to always be aware of what name to use for the "id" field. Ideally we wouldn't have to add lots of metadata to every query (affects perf). We also destructure `RelayNodeInterface` in lots of modules (again, perf), so it isn't as simple as making a constant configurable.

More generally, getting `__id` into the spec seems like a far simpler and portable way to address the problem of `id` field collisions - rather than every client have to incur similar complexity to what's added here.

alloy:
@josephsavona Hmm, bear with me, when you refer to the runtime not supporting it, do you mean functionality like refetching? The de-duplication works for me, but as I have not yet been able to use the Node based functionality I haven't been able to test this yet.

alloy:
Regarding getting the __id field in the spec, I get that that makes things simpler and why you might not want something like this. I will have to make this work nonetheless, because I can't wait for such long-term changes, I would, however, consider naming our field __id for forward compatibility.

josephsavona:
> do you mean functionality like refetching? 

Yes - Relay uses the `id` field of a record as the value by which to refetch it, so this change (and a schema using an identifier field not named "id") will break refetching.

> The de-duplication works for me

Are you sure? Again, Relay uses `id` field for de-duplication. Your records may be getting normalized (flattened), but you will get collisions if your `id` values are type-specific ids that could overlap. 

> I will have to make this work nonetheless.

We understand that you need to make something work today. That said, this is a lot of complexity to introduce and we need to weigh that cost. For starters, we *definitely* cannot move forward without a concrete plan for dealing with refetching and normalization. This will likely involve some fairly invasive changes to Relay, which makes us even more wary of proceeding.

alloy:
Alright, based on your helpful feedback I now have a better understanding of the difference and working of the plugin and runtime parts. After some more testing it’s indeed clear that the runtime is not respecting my choice in ID field, it is atm simply ignoring _both_ the default `id` field and my custom ID field…

‘Sigh’ and ‘hah’, it’s pretty obvious in hindsight.

>> The de-duplication works for me
>
> Are you sure? Again, Relay uses id for de-duplication. Your records may be getting normalized (flattened), but you will get collisions if your id values are type-specific ids that could overlap.

This was the specific problem [I ran into](https://github.com/artsy/emission/issues/181) when I started the work to add Relay support to our GraphQL server and is what I was focussing on during testing. We had an ‘artist’ and an ‘artwork’ by the same `id`, after my changes the collision no longer occurred.

However, after your explanation I’ve set some more breakpoints and checked what Relay is using as `__dataID__` and came to the conclusion that it is not using the value of the custom ID that I was expecting it to. Relay generates values itself, e.g. `"client:-36268863541"` and because these are unique the problem was _seemingly_ solved.

It is now clear what piece I’m missing and obvious that that also applies to the re-fetching etc.

>> I will have to make this work nonetheless.
>
> We understand that you need to make something work today, but we definitely cannot move forward without a concrete plan for dealing with refetching and normalization. This will likely involve some fairly invasive changes to Relay, which makes us even more wary of proceeding.

I totally understand and agree with this from your side 👍 

It is not entirely clear to me if you would consider accepting customisability like this _at all_, though, or if you’re basically saying ‘no’ in lieu of the future `__id` field?

> Have you evaluated the work necessary to simply conform to the Node specification and use id as the fieldname?

Aye, we have. Alas this is definitely not an option in our case, I would have greatly preferred to just use the `id` field.

The issue is that our flagship iOS app has been using our GraphQL server since before we ever looked at using Relay, which means that [it now depends on the `id` field](https://github.com/artsy/eigen/blob/54584b69eedeb0cc145f2b7e9289f8c696520a31/Artsy/Networking/ARRouter.m#L1031) to hold a value that it can use against other APIs and there is no real way for us to control what versions of the app our users have installed. Thus changing this would basically break all installations of the app.

In the case of our web clients and APIs it would all theoretically have been possible, albeit it a highly complex and sensitive undertaking, because those are all under our control.

The other option would have been to not use Relay for [our new React Native work](https://github.com/artsy/emission) inside [our flagship app](https://itunes.apple.com/us/app/artsy-collect-bid-on-fine/id703796080?mt=8), but the possibilities for improved mobile network utilisation in this new paradigm is something that we really would like to use.

alloy:
> It is not entirely clear to me if you would consider accepting customisability like this at all, though, or if you’re basically saying ‘no’ in lieu of the future `__id` field?

To expand on this, if you prefer to not add any complexity at this moment in time, then that’s perfectly fine by me. Knowing now that you will have a new name for this field in the future, and what that name is, I can also just maintain a fork of Relay that changes the hardcoded name to `__id` and use that. In the future once you officially start using the new name I can then switch back.

The main reason for me to continue to have this officially be customisable _right now_ would be for community reasons.

alloy:
After spending some more time on this and giving it some more thought I have to agree with you that the added complexity would be quite a lot and maintaining a fork that uses `__id` instead seems like a saner route.

josephsavona:
Thanks for giving us more context on your situation. I didn't meant to be vague with my previous statement, it just wasn't yet clear what the best course of action is. I chatted with @leebyron a bit and `__id` is something that is being considered but not definite yet. Here's what I would recommend:

- Instead of this PR, let's go forward with a smaller change that moves the value of `id` to a constant in its own file/module (similar to how we have `RelayNodeInterface` for the runtime). 
- If there are any places in the runtime code that references `id` directly, let's change them to use `RelayNodeInterface.ID` (destructuring is ok).

We're okay with accepting PRs for the above. With these changes, you should be able to maintain a very minimal fork of Relay - just fork the plugin interface file and RelayNodeInterface.js, and manually run the build script to get a build that uses your own id fieldname. For the name itself, it sounds like `__id` is likely *enough* to become part of the spec (or at least a supplemental one) that it's overall probably best to use that name instead of some other name.

alloy:
> Thanks for giving us more context on your situation. I didn't meant to be vague with my previous statement, it just wasn't yet clear what the best course of action is.

I should have included the full context here from the get-go, instead of linking to it. You run a tight ship and I very much appreciate that.

> For the name itself, it sounds like `__id` is likely _enough_ to become part of the spec (or at least a supplemental one) that it's overall probably best to use that name instead of some other name.

Fair enough, I get that no hard promises can be made on this. Even if the name would end up being different, migrating away from one name to another shouldn’t be a real problem.

> We're okay with accepting PRs for the above. With these changes, you should be able to maintain a very minimal fork of Relay

This sounds like a great, usable, and simple path forward, I will send a PR for that ASAP.

Thanks a lot for your time and assistance! 👌👏

PR #1239, opened 2016-06-24
Tweak QuickStart-Tutorial.zh-CN.md metadata for consistency

wincent:
Make this one like our other translations:

- Use `.zh-CN` suffix instead of `-zh-CN`: this means that an appropriately configured web server can choose to serve up the right file based on the `Accept-Language` header.
- Use same `id` as canonical version. We might end up changing our minds about that (see https://github.com/facebook/relay/pull/1238), but in the meantime we should at least be consistent.

I think we likely still have some issues here, so I'm going to keep poking around after this. Basically, I think at this point that we've *unbroken* our English docs and our Chinese docs are at least consistently defined, but the build process for the translations isn't producing a useful/accessible result.

### DIFF
--- a/docs/QuickStart-Tutorial.zh-CN.md
+++ b/docs/QuickStart-Tutorial.zh-CN.md
@@ -1,9 +1,9 @@
 ---
-id: tutorial-zh-CN
+id: tutorial
 title: 教程
 layout: docs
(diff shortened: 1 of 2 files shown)


wincent:
@facebook-github-bot import

PR #1246, opened 2016-06-28
[babel-relay-plugin] Move Node ID fieldname definition into forkable file.

alloy:
As discussed in https://github.com/facebook/relay/pull/1232, this should make it easier to maintain a forked version of Relay and change the name of the field to use as Node ID.

@josephsavona Did I understand everything correct?


scripts/babel-relay-plugin/src/RelayQLAST.js
```diff
@@ -29,6 +29,7 @@ const find = require('./find');
 const invariant = require('./invariant');
 const util = require('util');
 const RelayTransformError = require('./RelayTransformError');
+const { ID } = require('./RelayQLNodeInterface');
```
josephsavona: nit: no spaces: `const {ID} = ...`

scripts/babel-relay-plugin/src/RelayQLPrinter.js
```diff
@@ -31,6 +31,7 @@ const find = require('./find');
 const invariant = require('./invariant');
 const util = require('util');
 const RelayTransformError = require('./RelayTransformError');
+const { ID } = require('./RelayQLNodeInterface');
```
josephsavona: same here

josephsavona:
@alloy Awesome, thanks! This is exactly what I was thinking.

alloy:
Related but off-topic, why are the compiled plugin sources checked-in but not the runtime ones? I’m wondering if I should do this in my fork, because from my limited experience with NPM it appears that if any `prepublish` work needs to be done you _have_ to push a package to the registry, you can’t use such a package directly from a git repo.

josephsavona:
The plugin sources are checked in compiled because they're used *within* our babel compilation toolchain. It's certainly possible for these to be compiled dynamically anyway (two instances of babel, one compiling plugins for the other), but I *suspect* we avoided that in the RN packager for perf reasons.

steveluscher:
@facebook-github-bot shipit

facebook-github-bot-3:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/267584053605947/int_phab) to review.

alloy:
@josephsavona Gotcha. Thanks for all the help 👍 👏

josephsavona:
@alloy thanks for this!

alloy:
Seems like this doesn’t actually yet makes it fully work at runtime. In the case of the root node in a query it works as expected, but otherwise the dataIDs are of the form `client:integer`. Looking into it further, if you have a hunch of where I should look, that would be great.

alloy:
Awww, that was silly. I missed the two places I had actually changed in my previous PR. Will send a PR for that.

Issue #822, opened 2016-02-10
Idea: Remove the `Route` concept?


ianstormtaylor:
One of the things I noticed while trying to understand “Routes” in the first place, was that they seemed very lightweight. So lightweight that I was wondering if they even needed to exist. We might instead be able to just combine their usage with Containers themselves.

For example, the first example in the docs uses a plain object as a Route, at which point it’s hardly anything special, it’s just a dictionary of options passed into a Root Container. The second example goes a step further, subclassing `Relay.Route`, but it seems like the only reason to do that is to get the error handling logic in `paramDefinitions`, which could be achieved in others ways.

The docs mention that `Routes must also define a string name.` but it’s unclear why this need exists, and I don’t think it’s ever mentioned again in the docs. I assume it’s for error handling too, but that too could be solved differently if it changing it allows for a simpler API?

Routes themselves seem like they'd rarely be re-used, and even if they were, why not just re-use the specific queries (via sharing them like regular functions) instead of needing an extra object to wrap them?

What if instead, containers could be declared with their own `queries` and `paramTypes` (renamed from `paramDefinitions` to match React’s naming scheme):

``` js
const Container = Relay.createContainer(Profile, {

  queries: {
    user: () => Relay.QL`
      query { user(id: $userId) }
    `,
  },

  paramTypes: {
    userId: Relay.ParamTypes.string.required // potentially unnecessary, but for consistency
  }

}
```

Instead of needing a `Route` object that throws without the proper parameters, that error handling logic can be taken on by the container itself, and the “Route” really is just a configuration dictionary at that point, hardly even needing a special name. Instead of passing in a `Route` to a `RootContainer`, the `RootContainer` could then just simply take `params={…}`:

``` jsx
<RootContainer Component={Thing} params={...} />
```

This is similar to the simplicity that libraries like [`react-router-relay`](https://github.com/relay-tools/react-router-relay) (which was very nice to get setup and remove the complexity of Relay) achieve for you by abstracting away the need to touch Routes or Root Containers.

wincent:
This is great feedback, @ianstormtaylor. We've already taken some steps towards de-emphasizing routes, but there's still more that we can do. When we first designed them they were much closer to what developers typically thing of as "routes", and that's why we named them that way.

[...]

Issue #503, opened 2015-10-22
Replace Relay.Route with RelayQueryConfig and Relay.RootContainer with RelayRenderer on homepage examples


steveluscher:

michaelchum:
@steveluscher The [Relay public API](https://github.com/facebook/relay/blob/master/src/RelayPublic.js) currently does not expose `RelayQueryConfig`. Are you planning on exposing it to subclass `Relay.RelayQueryConfig` as a standard instead of `Relay.Route`?

steveluscher:
Indeed. I just created this issue so that I wouldn't forget about the homepage when this happens!

josephsavona:
Closing; we'll update the examples when we release the new APIs from #1369.

PR #1254, opened 2016-07-01
Docs: Replace Relay.RootContainer guide with Relay.Renderer

washt:


docs/Guides-Renderer.md
```diff
@@ -0,0 +1,168 @@
...
+So far, we've covered two pieces that each contribute to declaring data:
+
+ - **Relay.Route** lets us declare query roots.
+ - **Relay.Container** lets components declare fragments.
+
+To use these pieces to construct a full-fledged GraphQL query that we can send to the server to fetch data, we need to use the **Relay.Renderer**.
+
+## Component and Route
+
+**Relay.Renderer** is a React component that, given a `Container` and a `route` or `queryConfig`, attempts to fulfill the data required in order to render an instance of `Container`.
```
steveluscher: Maybe don't mention routes. We're trying to get rid of them in favor of query configs.

docs/Guides-Renderer.md
```diff
@@ -0,0 +1,168 @@
...
+  />,
+  component
+);
+```
+
+When the **Relay.Renderer** above is rendered, Relay will construct a query and send it to the GraphQL server. As soon as all required data has been fetched, `ProfilePicture` will be rendered. Props with fragments will contain data that was fetched from the server.
+
+If either `Container` or `queryConfig` ever changes, **Relay.Renderer** will immediately start attempting to fulfill the new data requirements.
+
+**Relay.Renderer** renders the loading state whenever it cannot immediately fulfill data needed to render. This often happens on the initial render, but it can also happen if either `Container` or `queryConfig` changes.
```
steveluscher: “The loading state” here is a bit confusing, since that's not actually an entity in the system. As you've noted, the default is to render `null` or the previous view. Maybe merge these two paragraphs?

Also, I think you meant for this text to be below the heading.

docs/Guides-Renderer.md
```diff
@@ -0,0 +1,168 @@
...
+
+By default, nothing is rendered while loading data for the initial render. If a previous set of `Container` and `queryConfig` were fulfilled and rendered, the default behavior is to continue rendering the previous view.
+
+## render prop
```
steveluscher: Maybe “Render callback”?

docs/Guides-Renderer.md
```diff
@@ -0,0 +1,168 @@
...
+render({
+  props: ?{[propName: string]: mixed},
+  done: boolean,
+  error: ?Error,
+  retry: ?Function,
+  stale: boolean
+}): ?React$Element
+```
+
+### props
```
steveluscher: Wherever you're talking about a JavaScript identifier, wrap it in backticks.

```
### `props`
```

docs/Guides-Renderer.md
```diff
@@ -0,0 +1,168 @@
...
+
+### stale
+
+```
+stale: boolean
+```
+
+True when `forceFetch` is enabled and `props` is present due to available client data. Once the forced server request completes, `stale` will return to false.
+
+If a `render` callback is not supplied, the default behavior is to render the container if data is available, the existing view if one exists, or nothing.
```
steveluscher: Since this isn't part of the `stale` docs, we'll need to move this text around. Maybe it belongs above with the preamble under `## Render callback`?

docs/Guides-Renderer.md
```diff
@@ -0,0 +1,168 @@
...
+### Example
+
+```{4-6}
+// In this example, `ErrorComponent` and `LoadingComponent`
+// simply display a static error message / loading indicator.
+<Relay.Renderer
+  Container={ProfilePicture}
+  queryConfig={profileRoute}
+  render={({done, error, props, retry, stale}) => {
+        if (error) {
```
steveluscher: Two spaces of indentation always.

docs/Guides-Renderer.md
```diff
@@ -0,0 +1,168 @@
...
+        } else if (props) {
+          return <ProfilePicture {...props} />;
+        } else {
+          return <LoadingComponent />;
+        }
+      }}
+/>
+```
+
+### ref
```
steveluscher: I think this needs two hashes, since your back outside of the “Render callback” section:

```
## Refs
```

docs/Guides-Renderer.md
steveluscher: `environment` is required too.

docs/Guides-Routes.md
```diff
@@ -119,7 +119,7 @@ window.addEventListener('popstate', () => {
   var userID = getQueryParamFromURI('userID', document.location.href);
   var profileRoute = new ProfileRoute({userID: userID});
   ReactDOM.render(
-    <Relay.RootContainer
+    <Relay.Renderer
```
steveluscher: `environment` is also required.

```
environment={Relay.Store}
```

docs/Guides-Renderer.md
washt: I think this this was redundant, I removed it.

docs/Guides-Renderer.md
```diff
@@ -0,0 +1,168 @@
...
+id: guides-renderer
+layout: docs
+category: Guides
+permalink: docs/guides-renderer.html
+next: guides-ready-state
+---
+
+So far, we've covered two pieces that each contribute to declaring data:
+
+ - **Relay.queryConfig** lets us declare query roots.
```
steveluscher: This looks great! Only one thing left… we don't actually export `Relay.QueryConfig`.

We should, though, and we should deprecate `Relay.Route`. See https://github.com/facebook/relay/issues/822 and https://github.com/facebook/relay/issues/503.

The way I see it, there are three things to do here:

1. Expose `Relay.QueryConfig` through `Relay.js`
2. Write a github.com/relayjs/relay-codemod that people can use to convert their routes to query configs. Add a deprecation warning to `Relay.Route` that points people there and encourages them to migrate their routes over.
3. Replace the `Relay.Route` guide with one for `Relay.QueryConfig`

What do you say. Are you bored of this yet, or would you like to keep going? If you'd like to leave it at this, you can change this line back to `Relay.Route`, mention that routes implement the `RelayQueryConfig` interface, but that you can just as soon use a plain JS object having that shape, and we'll land this PR (we should probably do that anyway). If you'd like to keep going, I can file tasks for the above and you can chip away at them.

Thanks for all of your help!

washt: Thanks! I'll keep going, writing a codemod sounds like fun!

eugene1g:
Hi guys, RelayQueryConfig has now been exposed in [v0.9.3](https://github.com/facebook/relay/releases/tag/v0.9.3) so this issue can be merged & closed (also #503 I think)

josephsavona:
Thanks for taking the time to put together this PR. At this point we're close to releasing a successor  to RelayRenderer that removes the notion of routes or query config and further simplifies the API. Rather than documents RelayRenderer, lets just wait for the new API :-)

Issue #1256, opened 2016-07-03
toGraphQL gets null type for enum variable


papigers:
I'm using `isomorphic-relay`,  and I'm encountering some problem, which you can read about in more details [here](https://github.com/denvned/isomorphic-relay/issues/42). 

The relevant part for you, is that I saw the server's preloaded data looks like this:
```
...
...
{
          "children": [
            {
              "calls": [
                {
                  "kind": "Call",
                  "name": "orderBy",
                  "metadata": {
                    "type": null
                  },
                  "value": {
                    "kind": "CallValue",
                    "callValue": "TRENDING"
                  }
                },
                {
                  "kind": "Call",
                  "name": "first",
...
...
```

Here you can see that the type of `orderBy` is null, instead of PollSort, an enum type. 

The above snippet is, as mentioned, part of the preloaded data the server passed to the client. This is done in this file: [prepareData.js](https://github.com/denvned/isomorphic-relay/blob/master/src/prepareData.js). The relevant line is:
```
data.push({ query: toGraphQL.Query(request.getQuery()), response });
```

So the root of the problem is actually that `toGraphQL.Query` returns null for my input variable type. 

Any idea how can this be fixed?

josephsavona:
Thanks for filing such a detailed issue. It looks like `callsTooGraphQL` isn't passing the call type to `QueryBuilder.createCall`. Care to send a PR to fix?

papigers:
I'd be happy to, but I'm a bit of a newbie to this repo, and relay in general. How can I get the call type from `callsToGraphQL`?

papigers:
Oh it's simply an attribute of `Call`. I'd try to fix this as soon as I get home.

papigers:
Done. #1257, though there are some build errors...

PR #1257, opened 2016-07-03
Pass call type in from/toGraphQL

papigers:
Currently types are ignored and not getting passed in `callsFromGraphQL`
or `callsToGraphQL`.
This results in dangerous output from `toGraphQL`, in which all calls have
type null. This can cause problems when using non-scalar types (like
enum).
For more details see #1256.

### DIFF
--- a/src/query/callsFromGraphQL.js
+++ b/src/query/callsFromGraphQL.js
@@ -26,6 +26,7 @@ const invariant = require('invariant');
 type CallOrDirective = {
   name: string,
   value: ?ConcreteValue,
+  type?: ?string,
 };
 
 /**
@@ -43,6 +44,7 @@ function callsFromGraphQL(
   for (let ii = 0; ii < callsOrDirectives.length; ii++) {
     const callOrDirective = callsOrDirectives[ii];
     let {value} = callOrDirective;
+    let {type} = callOrDirective.metadata;
     if (value != null) {
       if (Array.isArray(value)) {
         value = value.map(arg => getCallValue(arg, variables));
@@ -53,7 +55,7 @@ function callsFromGraphQL(
         value = getCallValue(value, variables);
       }
     }
-    orderedCalls.push({name: callOrDirective.name, value});
+    orderedCalls.push({name: callOrDirective.name, value, type});
   }
   return orderedCalls;
 }
--- a/src/query/callsFromGraphQL.js
+++ b/src/query/callsFromGraphQL.js
@@ -23,10 +23,14 @@ import type {Variables} from 'RelayTypes';
 
 const invariant = require('invariant');
 
+type Metadata = {
+  type: ?string,
+};
+
 type CallOrDirective = {
   name: string,
   value: ?ConcreteValue,
-  type?: ?string,
+  metadata: ?Metadata,
 };
 
 /**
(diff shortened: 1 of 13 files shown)


facebook-github-bot-1:

facebook-github-bot-8:

src/query/callsFromGraphQL.js
```diff
@@ -42,7 +43,11 @@ function callsFromGraphQL(
   const orderedCalls = [];
   for (let ii = 0; ii < callsOrDirectives.length; ii++) {
     const callOrDirective = callsOrDirectives[ii];
-    let {value} = callOrDirective;
+    let {value, metadata} = callOrDirective;
+    let type = undefined;
```
josephsavona: `let type;`

src/query/callsFromGraphQL.js
```diff
@@ -26,6 +26,7 @@ const invariant = require('invariant');
 type CallOrDirective = {
   name: string,
   value: ?ConcreteValue,
+  type?: ?string,
```
josephsavona: Remove this? The code never accesses `type` on the `callsOrDirectives` array.

josephsavona:
@papigers Thanks for the PR. Did you run tests locally? The build is not failing, the tests are ;-)

src/query/callsFromGraphQL.js
papigers: True, my bad

src/query/callsFromGraphQL.js
papigers: :+1:

papigers:
@josephsavona Will do, working on that right now, sorry for this noobish PR.

Next time will be better, I hope ;)

papigers:
Sorry for the uninformatives commits

src/tools/RelayInternalTypes.js
```diff
@@ -23,7 +23,7 @@ import type RelayQuery from 'RelayQuery';
 
 export type Call = {
   name: string,
-  type?: string,
+  type?: ?string,
```
steveluscher: I haven't thought about this as deeply as you two have, but might we instead want `type` to either be a `string` or not supplied?

```
const orderedCall = {
  name: callOrDirective.name, 
  value,
};
if (metadata && metadata.type) {
  orderedCall.type = metadata.type;
}
orderedCalls.push(orderedCall);
```

That would cut down on the noise in the tests, among other things. If it's important to have `type: null` then, carry on!

josephsavona: Yeah, this already is marked as an optional field.

@papigers I apologize for the back and forth on this. Since this is already an optional field, how about reverting the addition of all the `type: null`s to focus this PR on just the logic & results that actually change?

src/query/callsFromGraphQL.js
```diff
@@ -23,9 +23,14 @@ import type {Variables} from 'RelayTypes';
 
 const invariant = require('invariant');
 
+type Metadata = {
+  type: ?string,
```
josephsavona: let's make this `type?: ?string`

josephsavona:
@papigers Thanks again for this PR. I commented to request some changes. Also, scanning through i don't see any tests that actually end up with `type` having a non-null value, which means it doesn't look like we're really testing the change itself. 

I apologize if I missed something, but if not can you add a test that has a non-null type? This would be an enum or object argument.

psalz:
I just ran into the same issue - any updates on this?

josephsavona:
@papigers are you able to update the PR?

papigers:
Not any time soon , I'm afraid. Too busy..

josephsavona:
closing in favor of #1336 - thanks @papigers!

PR #1268, opened 2016-07-07
Fix Renderer Example in Documentation

lukaswelte:
Fix the order of the parameters in the function so that it actually matches the implementation.


lukaswelte:
I actually read the doc wrong. Have not seen the object descructuring

PR #1276, opened 2016-07-11
Add more cases to the test schema

kassens:
This adds to the schema to test:

- Non nullable array of non nullable enum
- Array of enum


kassens:
@facebook-github-bot import

PR #1278, opened 2016-07-12
[Website] Update playground frame height to suggest line 20 is not the last line 

rpechayr:
At first it was quite difficult for me to understand the first 2 examples in the website homepage because they basically looked the same. 

I then realized it was because they both seemed to have 20 lines of code. The frames did not looke like they were scrollable. 

Before 
screen shot 2016-07-12 at 14 21 03

After
screen shot 2016-07-12 at 14 20 54

I've changed the height into 426. There might be a better magic number to suggest that line 20 is not the last line of the gist. 

I am pretty sure this would have helped me a lot when I first had a look at Relay website.


facebook-github-bot-7:

kassens:
Thanks!

@facebook-github-bot shipit

facebook-github-bot-9:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1038568506239420/int_phab) to review.

PR #1280, opened 2016-07-12
Use new eslint-config-fbjs, fix lint warnings

zpao:
I just shipped `eslint-config-fbjs` which much more closely matches the lint config we have internally, uses a few plugins, and has the globals we need for Flow. This also takes care of the really large list of Flow-related unused/undef var warnings. With this you're back down to 0 warnings.


kassens:
awesomesauce

@facebook-github-bot import

facebook-github-bot-4:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1115454901858536/int_phab) to review.

PR #1290, opened 2016-07-18
Allow for one-off query overrides for serverside rendering

toddtarsi:
First off, thank you guys for making this. This codebase is really awesome, and I love the approach to everything. Currently though, the main known open source serverside rendering solution (isomorphic-relay, another great codebase) is kind of in a bad spot when it comes to credential passing.

An example of this can be seen with isomorphic-relay, where the solution for passing httponly cookies in serverside rendering involves converting the endpoint to run sequentially, which is a major deoptimizer. By exposing overrides like this, we could skirt that approach by instead just passing in one off data into the sendQueries call.

Admittedly, I don't know this codebase all too well, and there are probably a lot of gotchas I'm overlooking, but I wanted to put this out there and just see what your guys thoughts were. Also, if you just want to see more polish and consistency, let me know and I can refactor / rebase this as needed.

Thanks as always!

### DIFF
--- a/src/network-layer/default/RelayDefaultNetworkLayer.js
+++ b/src/network-layer/default/RelayDefaultNetworkLayer.js
@@ -138,7 +138,7 @@ class RelayDefaultNetworkLayer {
   /**
    * Sends a POST request and retries if the request fails or times out.
    */
-  _sendQuery(request: RelayQueryRequest: Promise<any>, overrides?: Object = {}) {
+  _sendQuery(request: RelayQueryRequest, overrides?: Object = {}): Promise<any> {
     return fetchWithRetries(this._uri, {
       ...this._init,
       body: JSON.stringify({
--- a/src/network-layer/default/RelayDefaultNetworkLayer.js
+++ b/src/network-layer/default/RelayDefaultNetworkLayer.js
@@ -62,9 +62,9 @@ class RelayDefaultNetworkLayer {
     );
   }
 
-  sendQueries(requests: Array<RelayQueryRequest>): ?Promise<any> {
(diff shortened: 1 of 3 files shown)


facebook-github-bot-7:

josephsavona:
Thanks for the PR. We're open to considering this, but can you provide more detail as to why this is needed and how it would be used? It would be helpful to have a link to an example of how `isomorphic-relay` is forced to do serial fetching w/o this.

facebook-github-bot-2:

toddtarsi:
Absolutely, thank you for the prompt response:

This issue is closed in isomorphic-relay-router, but a quick read will help explain the issue:

https://github.com/denvned/isomorphic-relay-router/issues/13

By exposing this code in relay, and then adding an extra argument for overrides to the prepareData operation in isomorphic-relay-router and isomorphic-relay, we could pass through cookie data to the sendQueries call that these wrap without the overhead of the solution above.

toddtarsi:
Ha, on second thought, that issue tracker rambles:
TLDR: Since node-fetch proxies requests, and there isn't a one off way to override these, serverside rendering hits a challenge where it needs to override this using injectNetworkLayer. Since injectNetworkLayer is a global, the only current solution is to guard from async attack vectors by converting the stack to run sequentially.
By passing in an override function, it gives us a method to customize each request batch without injecting a new global network layer each time on the server.

toddtarsi:
Hey all, after talking with @robrichard, it sounds like my use case was resolved when Relay.environment was able to be used non-globally.
https://github.com/denvned/isomorphic-relay/pull/43
Thanks for taking the time to look this over, and if you'd like I can close this pull request now.
My use cases are resolved I believe.

toddtarsi:
Absolutely, thanks for the quick responses from everyone. 😄

PR #1292, opened 2016-07-20
RANGE_ADD mutation for plain GraphQLLists

### LINKED ISSUE #444: Enhancement: Append/Prepend/Remove for plain GraphQLLists
I know that in docs it says that RANGE_ADD add new edge to connection, so I assume its only for connections (with edges and node).

So want to basically ask if there is some other option how to add item to normal graphql list without refetching the whole list or if its planned.

Thanks.

Globegitter:
This is a minimal implementation to fix #444.

This PR in its current state is fully functional but still WIP, as it needs some cleaning up, possibly some minor refactors etc. but wanted to get it up to get the discussion on this functionality (and my implementation) started.

## API Additions

```js
class NewCommentMutation extends RelayMutation {
...
getConfigs() {
  return [
    {
      type: RelayMutationType.RANGE_ADD,
      parentID: feedbackID,
      listName: 'simpleTopLevelComments',
      elementName: 'comment',
      rangeBehaviors: () => GraphQLMutatorConstants.APPEND,
    },
  ];
}
```

`listName` matches the existing `connectionName` and has just been added for clarity and to be able to distinguish between a connection and a plain list mutation. `elementName` (which currently is still called `newElementName` in my code, but will rename shortly) is the field-name of the new element to add from the payload. 

range calls (arguments in the queries) are currently not supported, so rangeBehaviors will currently always be called without any arguments. So only very simple use-cases are currently supported (which satisfies a good chunk of our own internal use-cases), which in my opinion is a good start to get this PR in and any addition to this should come in separate PRs.

## Implementation

As I said, while all the functionality is there and has been tested (including in one of our live apps) I was not sure about a few parts of the implementation, mostly related to optimistic responses so I just took a few guesses that in the end got everything to work.

I was not sure how paths are supposed to work with an optimistic response of a new list element (that has no `id` provided and therefore creates a client-side id). 

Also for some reason when writing the optimistic response I had to copy over the existing records to the record writer, which I did not need for non-optimistic responses and also is not done for connection updates. So maybe there is some room for improvement in my code.

You might of course mention that in the code review, once my PR is ready of course, but just wanted to bring this up early-on.

## TODO
 - [ ] Code cleanup and fixing TODOs
 - [ ] Fix test for mutation config validation 
 - [ ] Add docs for the new functionality

## API improvements (off-topic)
This is not directly related to this PR an maybe not the best Platform for this discussion but when working over this API I had some ideas on how the general API could potentially be simplified and made more flexible. I suppose #538 is the place to discuss this, but wanted to use this chance to figure out if there is any room for writing some proposal/making some proof-of-concept PRs for that area? (I have also seen there has been work done on a new mutation API, but not sure how that differs to the current implementation)


Globegitter:
Oh yes, another thing that is still missing is the feature of `source { id }` that connections support and I quite like. So will probable add that as well.

josephsavona:
@Globegitter Thanks for such a detailed and extensive PR. We appreciate all the effort you put into this and other PRs. 

Unfortunately the timing of this is not ideal. The team is actively working on some major improvements to Relay core, and among these is a more flexible mutation API that will support effectively arbitrary modification of lists. We're still iterating on the details, but we'll be sharing more about this in the next few weeks.

Given the timing, the fact that the new mutations API would be even more flexible than simple append/prepend, and the complexity involved in testing & shipping this implementation, we're inclined to hold off on accepting this PR. If you have an immediate need for this, we'd be happy to work with you to come up with a workaround (perhaps by manually writing a new list into the store via low-level methods).

I know this can be frustrating, but we're trying to optimize for getting the new version out as soon as possible. What do you think about going with a workaround for now?

Globegitter:
@josephsavona Thank you for that detailed response and it does make sense to not merge this PR if the new and improved API is just a few weeks ahead. Looking forward to seeing arbitrary modifications to lists.

In terms of solutions to our problems, I will have to check with one of our developers first what we currently do there as a work-around before getting back here. But in one of the affected apps that I am directly working on (a non-consumer facing admin-panel) we just decided to force-fetch 100% of the time for now. This decision was also affected by the fact that this app often needs its cache updated with new data from the backend but we have no websockets yet. In the consumer facing app we fortunately could use ConnectionFields everywhere where we needed list mutations. The offer is very much appreciated.

Either way, I will close this PR.

josephsavona:
@Globegitter Thanks for understanding. An internal admin panel seems like a good use case for always-fresh data, so force-fetching almost seems like a good tradeoff there anyway. Definitely reach out though if you still need a workaround and we'd be happy to help!

mattecapu:
> perhaps by manually writing a new list into the store via low-level methods
Can you elaborate @josephsavona? It would be extremely handy to have a reference implementation of the workaround while waiting for the major improvements.

PR #1294, opened 2016-07-20
Parse serialized GraphQL range in getRangeMetadata

### LINKED ISSUE #1293: GraphQLRange is not properly deserialized
`RelayEnvironmentSerializer` just takes the `_storeData` and serializes it. When this happens, instances of `GraphQLRange` are not properly serialized, and `RelayRecordStore` does do any checks before calling `retrieveRangeInfoForQuery` on the range returned from `_getField`.

I understand that `RelayEnvironmentSerializer` is not even documented or released, but I wanted to see if this was being worked on, or if you would be open to accepting PRs fixing/expanding the current serialization behavior.

Aweary:
Resolves #1293 

Current if you try to de/serialize your RelayEnvironment you run into an error as `getRangeMetadata` tries calling a `GraphQLRange` method on the serialized array. To prevent this we check to see if we have a serialized instance (which is always an array) and parse it if we do.

### DIFF
--- a/src/store/RelayStoreData.js
+++ b/src/store/RelayStoreData.js
@@ -633,13 +633,9 @@ class RelayStoreData {
       nodeRangeMap,
     } = obj;
 
-    for (const key in records) {
-      const record = records[key];
-      const range = record.__range__;
-      if (range) {
-        record.__range__ = GraphQLRange.fromJSON(range);
-      }
-    }
+    deserializeRecordRanges(cachedRecords);
+    deserializeRecordRanges(queuedRecords);
+    deserializeRecordRanges(records);
 
     return new RelayStoreData(
       cachedRecords,
@@ -649,6 +645,20 @@ class RelayStoreData {
       rootCallMap,
       nodeRangeMap,
     );
+  }
+}
+
+/**
+ * A helper function which checks for serialized GraphQLRange
+ * instances and deserializes them in toJSON()
+ */
+function deserializeRecordRanges(records) {
+  for (const key in records) {
+    const record = records[key];
+    const range = record.__range__;
+    if (range) {
+      record.__range__ = GraphQLRange.fromJSON(range);
+    }
   }
 }
 
--- a/src/store/RelayStoreData.js
+++ b/src/store/RelayStoreData.js
@@ -44,6 +44,7 @@ import type {RecordMap} from 'RelayRecord';
 const RelayRecordStore = require('RelayRecordStore');
 const RelayRecordWriter = require('RelayRecordWriter');
 const RelayTaskQueue = require('RelayTaskQueue');
+const GraphQLRange = require('GraphQLRange');
 import type {TaskScheduler} from 'RelayTaskQueue';
 import type {Abortable, CacheManager, CacheProcessorCallbacks} from 'RelayTypes';
 
@@ -631,6 +632,14 @@ class RelayStoreData {
       rootCallMap,
       nodeRangeMap,
     } = obj;
+
+    for (const key in records) {
+      const record = records[key];
+      const range = record.__range__;
+      if (range) {
+        record.__range__ = GraphQLRange.fromJSON(range);
+      }
+    }
 
     return new RelayStoreData(
       cachedRecords,
--- a/src/store/RelayRecordStore.js
+++ b/src/store/RelayRecordStore.js
@@ -356,10 +356,7 @@ class RelayRecordStore {
     if (connectionID == null) {
       return connectionID;
     }
-    let range: ?GraphQLRange = (this._getField(connectionID, RANGE): any);
-    if (Array.isArray(range)) {
-      range = GraphQLRange.fromJSON(range);
-    }
+    const range: ?GraphQLRange = (this._getField(connectionID, RANGE): any);
     if (range == null) {
       if (range === null) {
         warning(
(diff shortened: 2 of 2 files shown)


josephsavona:
Thanks for this PR. We recently added a [utility for (de)serializing a RelayEnvironment](https://github.com/facebook/relay/blob/master/src/store/RelayEnvironmentSerializer.js) that is designed to solve this exact problem. Note that placing all of the serialization logic in `RelayEnvironmentSerializer` decouples values serialization from the logic that operates on those values. The PR as implemented introduces deserialization logic into a runtime module, introducing runtime and cognitive overhead.

Would you care to send a PR to export `RelayEnvironmentSerializer` as part of the public `Relay` API? (perhaps as `Relay.EnvironmentSerializer`).

Aweary:
@josephsavona I was actually using `RelayEnvironmentSerializer` when I ran into this issue. 

>Note that placing all of the serialization logic in RelayEnvironmentSerializer decouples values serialization from the logic that operates on those values.

I agree, that seems ideal. I didn't do it this way because that would require `RelayEnvironmentSerializer` to transverse the parsed `StoreData` instance, which it doesn't currently do. I could update this PR to do just that and also expose `RelayEnvironmentSerializer` if you'd like.

Aweary:
Alternatively, `RelayStoreData` and `GraphQLRange` already define their own serialization behavior, would it make sense to add this logic to `RelayStoreData.fromJSON()` instead? If `fromJSON` should return a consumable instance of that type then it seems to make more sense to have that logic live there.

josephsavona:
> would it make sense to add this logic to `RelayStoreData.fromJSON()` instead? If `fromJSON` should return a consumable instance of that type then it seems to make more sense to have that logic live there.

@Aweary I completely gree. Sorry for the misunderstanding; I thought that `RelayStoreData.fromJSON()` already was deserializing `GraphQLRange` instances. It looks like we missed that - we'd be happy to accept a PR fix this.

Aweary:
@josephsavona awesome, glad we are on the same page. I updated the PR so that `RelayStoreData` iterates through `records` and deserializes all `__range__` instances before it creates and returns the new `RelayStoreData` instance.

Let me know what you think about this approach, and any changes you'd like.

I can also add some tests for `RelayEnvironmentSerializer` and add it to `RelayPublic` if you'd like, or I can do that in another PR 😄

src/store/RelayStoreData.js
```diff
@@ -632,6 +633,14 @@ class RelayStoreData {
...
     } = obj;
 
+    for (const key in records) {
+      const record = records[key];
+      const range = record.__range__;
+      if (range) {
+        record.__range__ = GraphQLRange.fromJSON(range);
+      }
+    }
+
```
josephsavona: Note that we'll have to do this for all of `records`, `cachedRecords`, and `queuedRecords`

Aweary: I should probably break this logic out into a utility then, where would something like that live in the codebase?

josephsavona: Since it would only be used locally to the module, I would add it as a helper at the bottom of the file

josephsavona: Good question!

Aweary:
@josephsavona I implemented a `deserializeRecordRanges` helper at the bottom of the file and updated `toJSON` to call it for each set of records.

Aweary:
This is sort of tangentially related to this PR, but I was wondering you guys had considered a way to selectively pick which queries/data are serializable? I've found that our store ends up being rather large, making `JSON.stringify` and `JSON.parse` really expensive. It would be nice if I could select which fields/queries would be serialized, maybe with a directive or something

```graphql
fragment on Viewer {
  username @serialize,
  firstName,
  lastName
}
```

That way I could serialize data that was most important to rendering. Just a thought, I'd love to get some feedback on it.

josephsavona:
@facebook-github-bot import

facebook-github-bot-5:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1020088174766236/int_phab) to review internal test results.

josephsavona:
@Aweary Thanks for the PR. 

> This is sort of tangentially related to this PR, but I was wondering if you guys have considered a way to selectively pick which queries/data are serialized?

We have thought about this, but it sounds like you're using serialization for a different use case than we intended. For offline persistence, we have a [cache manager API](https://github.com/facebook/relay/blob/master/src/store/RelayStoreData.js#L183-L189) that allows persistence of individual records (see `RelayTypes` for more on the signature of the cache manager object). For server rendering, we recommend transferring payloads themselves from server to client and processing them there.

Aweary:
@josephsavona thanks for the guidance!

>We have thought about this, but it sounds like you're using serialization for a different use case than we intended. For offline persistence, we have a cache manager API that allows persistence of individual records (see RelayTypes for more on the signature of the cache manager object). 

I didn't realize that API existed, thanks for pointing it out. We're not using SSR just yet, I was just looking at offline persistence. Out of curiosity, what are the use cases you guys are implementing serialization for?


Also, do you want to get `RelayEnvironmentSerializer` into the public API? I'd be happy to do that here, or open another PR for it.

src/store/RelayStoreData.js
```diff
@@ -44,6 +44,7 @@ import type {RecordMap} from 'RelayRecord';
 const RelayRecordStore = require('RelayRecordStore');
 const RelayRecordWriter = require('RelayRecordWriter');
 const RelayTaskQueue = require('RelayTaskQueue');
+const GraphQLRange = require('GraphQLRange');
```
sampepose: nit: alphabetize imports

Issue #754, opened 2016-01-20
queryCache module-wide variables leaking memory on the server side


fabiosantoscode:
Using isomorphic-relay to get relay to work on the server, and using a load testing tool to create repeated requests and find a memory leak in an application I'm building, I observed that 2 maps called queryCache were growing unbounded.

These maps are in getRelayQueries.js, and in buildRQL.js, and they grow to thousands of keys over 600 requests to a page I have with no less than 2 containers.

Longer explanation in:

https://github.com/facebook/relay/pull/683#issuecomment-173021689

I'll do my best to provide an example if it's necessary to reproduce the issue.

denvned:
Hi @fabiosantoscode. Do your testing app recreates containers on each request using `Relay.createContainer(...)`? I can't see any other reason for such `queryCache.size` growth (well, probably except some bug in core-js polyfill for `Map`).

fabiosantoscode:
Hello! Yes it does. Since the app is wrapped in react-router and
react-router itself needs a history object which changes for every URL,
we've created an isomorphic function that creates the entire thing.

rodrigopr:
Hi, I'm facing the same behavior in our app. 
Is it reasonable to disable the `queryCache` and `fragmentCache` on the server?

I could try change how the react-router-relay create the routers if this makes more sense, but don't seem to be a simple change after a quick [look](https://github.com/relay-tools/react-router-relay/blob/master/src/QueryAggregator.js#L28-L118)

josephsavona:
@rodrigopr Does your app also redefine all the containers in every http request (by calling `createContainer`)? The face that the cache size is being exceeded is an indication that the application is doing unnecessary recomputation on each request; you could probably get a general performance boost by defining components/containers just once (they're meant to be reusable). 

We would consider adding some way to reset these caches, but I suspect that doing so in this case would mask a larger performance problem in your app.

rodrigopr:
I'm afraid react-router-relay and isomorphic-relay creates an QueryAggregator (similar to a RelayContainer) per render (see [react-router-relay/blob/master/src/QueryAggregator.js](https://github.com/relay-tools/react-router-relay/blob/master/src/QueryAggregator.js#L186-L207) and [isomorphic-relay-router/src/prepareData.js](https://github.com/denvned/isomorphic-relay-router/blob/master/src/prepareData.js#L6-L12))

josephsavona:
[...]

PR #1302, opened 2016-07-26
Add `Relay.disableQueryCache` static method

KCraw:
This allows for disabling the query cache in `createRelayQuery` and `buildRQL` by calling `Relay.disableQueryCache()`. 

Submitted in response to https://github.com/facebook/relay/issues/754#issuecomment-234995746

Because keys used to store queries include params, the query caches can grow to an unbounded size in certain circumstances (like text search input as a param). This allows the the cache to be disabled manually in circumstances where that is problematic, such as server side rendering. 

This PR does not disable the `fragmentCache` in `buildRQL`, which I believe only uses fragments and initial variables, and so has a bounded upper limit in terms of size.

### DIFF
--- a/src/RelayPublic.js
+++ b/src/RelayPublic.js
@@ -29,7 +29,6 @@ const createRelayQuery = require('createRelayQuery');
 const getRelayQueries = require('getRelayQueries');
 const isRelayContainer = require('isRelayContainer');
 
-
 if (typeof global.__REACT_DEVTOOLS_GLOBAL_HOOK__ !== 'undefined') {
   global.__REACT_DEVTOOLS_GLOBAL_HOOK__._relayInternals = RelayInternals;
 }
--- a/src/RelayPublic.js
+++ b/src/RelayPublic.js
@@ -23,10 +23,12 @@ const RelayRenderer = require('RelayRenderer');
 const RelayRootContainer = require('RelayRootContainer');
 const RelayRoute = require('RelayRoute');
 const RelayStore = require('RelayStore');
+const DisableRelayQueryCache = require('DisableRelayQueryCache');
 
 const createRelayQuery = require('createRelayQuery');
 const getRelayQueries = require('getRelayQueries');
 const isRelayContainer = require('isRelayContainer');
+
 
 if (typeof global.__REACT_DEVTOOLS_GLOBAL_HOOK__ !== 'undefined') {
   global.__REACT_DEVTOOLS_GLOBAL_HOOK__._relayInternals = RelayInternals;
@@ -50,6 +52,7 @@ const RelayPublic = {
   createContainer: RelayContainer.create,
   createQuery: createRelayQuery,
   getQueries: getRelayQueries,
+  disableQueryCache: DisableRelayQueryCache.disableCache,
   injectNetworkLayer: RelayStore.injectNetworkLayer.bind(RelayStore),
   injectTaskScheduler: RelayStore.injectTaskScheduler.bind(RelayStore),
   isContainer: isRelayContainer,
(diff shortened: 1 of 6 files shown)


facebook-github-bot-9:

facebook-github-bot-0:

src/container/getRelayQueries.js
```diff
@@ -36,14 +37,19 @@ function getRelayQueries(
...
+  if (queryCacheEnabled) {
+    cache = queryCache.get(Component);
+    if (!cache) {
+      cache = {};
+      queryCache.set(Component, cache);
+    }
+    cacheKey = route.name + ':' + stableStringify(route.params);
+    if (cache.hasOwnProperty(cacheKey)) {
+      return cache[cacheKey];
+    }
```
josephsavona: Let's break this into two methods: one that computes the query set (something like `buildQuerySet()`) and then a method (`getRelayQueries`) that calls it and knows about caching.

src/query/buildRQL.js
```diff
@@ -96,13 +97,17 @@ const buildRQL = {
...
+    let componentCache = {};
+    const queryCacheEnabled = DisableRelayQueryCache.getCacheEnabled();
+    if (queryCacheEnabled) {
+      componentCache = queryCache.get(queryBuilder);
+      if (!componentCache) {
+        componentCache = new Map();
+        queryCache.set(queryBuilder, componentCache);
+      } else {
+        node = componentCache.get(Component);
+      }
```
josephsavona: let's do the same thing here

josephsavona:
Thanks for the PR. Overall this looks good, but please see feedback about splitting the caching logic out from the main queryset/fragment building logic.

KCraw:
Code is now split based on what is being cached in each modile.

src/RelayPublic.js
```diff
@@ -23,11 +23,13 @@ const RelayRenderer = require('RelayRenderer');
 const RelayRootContainer = require('RelayRootContainer');
 const RelayRoute = require('RelayRoute');
 const RelayStore = require('RelayStore');
+const DisableRelayQueryCache = require('DisableRelayQueryCache');
 
 const createRelayQuery = require('createRelayQuery');
 const getRelayQueries = require('getRelayQueries');
 const isRelayContainer = require('isRelayContainer');
 
+
```
josephsavona: nit: remove extra blank lines

src/query/__tests__/buildRQL-test.js
```diff
@@ -190,6 +192,20 @@ describe('buildRQL', () => {
...
+      const builder = Component => Relay.QL`
+        query {
+          node(id:$id) {
+            ${Component.getFragment('foo')}
+          }
+        }
+      `;
+      const node1 = buildRQL.Query(builder, MockContainer, 'foo', {id: null});
+      const node2 = buildRQL.Query(builder, MockContainer, 'foo', {id: null});
+      expect(node1 === node2).toBe(false);
```
josephsavona: could do `expect(node1).not.toBe(node2)`, which would provide a slightly more useful error message (the value1 !== value2 instead of true !== false)

josephsavona:
@KCraw Thanks for the updates. I have a couple small nits and then this will be ready to ship :-)

KCraw:
Fixed

josephsavona:
Thanks!

wincent:
So, are we going to proceed with this? I understand from comments on #754 that there is some reticence to increase the API surface area by adding a new method, but on the flip side, it seems to me that a more sophisticated alternative involving LRU eviction would add non-trivial complexity, so I am on the balance fine with proceeding with this.

Thoughts? @josephsavona, @KCraw.

I'm preparing to cut a release in the next day or two, so it would be good to include this (or not).

josephsavona:
+1 for shipping this

wincent:
@facebook-github-bot import

facebook-github-bot-4:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3792224) to review internal test results.

wincent:
Landed this with some changes to the internal names to better follow our conventions. Thanks once again @KCraw.

PR #1069, opened 2016-04-20
Added pendingVariables to RelayContainer.

Globegitter:
This is still a WIP, but wanted to create the PR already, to get some eyes on it. Should be able to get it done and tested in our app by the end of the week. Ref #1049

PR #1310, opened 2016-07-30
RelayContainer shouldUpdate if there are pendingVariables

chollier:
This pull-request is a follow-up of https://github.com/facebook/relay/pull/1069

I was trying to use the feature but got confused because my decorated component wouldn't receive the updated `this.props.pendingVariables` when calling `setVariables`. (see discussion https://github.com/facebook/relay/commit/3b8f67fc027a683e0f2cac71cc2a47ece0f5ed20).

The changes here make sure RelayContainer's `shouldComponentUpdate` returns `true` when the nextState has some `pendingVariables`.

I looked at adding some tests but wasn't sure really where to put them ? `RelayContainer-test.js` / `RelayContainer_setVariables-test.js` ? What should I test ? I was thinking either the count of `render.mock.calls.length` or the output of shouldComponentUpdate

### DIFF
--- a/src/container/RelayContainer.js
+++ b/src/container/RelayContainer.js
@@ -771,6 +771,7 @@ function createContainerComponent(
 
       const fragmentPointers = this._fragmentPointers;
       return (
+        !!nextState.relayProp.pendingVariables ||
         !RelayContainerComparators.areNonQueryPropsEqual(
           fragments,
           this.props,
--- a/src/container/RelayContainer.js
+++ b/src/container/RelayContainer.js
@@ -771,7 +771,6 @@ function createContainerComponent(
 
       const fragmentPointers = this._fragmentPointers;
       return (
-        !!nextState.relayProp.pendingVariables ||
         !RelayContainerComparators.areNonQueryPropsEqual(
           fragments,
           this.props,
@@ -788,6 +787,10 @@ function createContainerComponent(
         !RelayContainerComparators.areQueryVariablesEqual(
           this.state.relayProp.variables,
           nextState.relayProp.variables
+        ) ||
+        !RelayContainerComparators.areQueryVariablesEqual(
+          this.state.relayProp.pendingVariables,
+          nextState.relayProp.pendingVariables
         )
       );
     }
--- a/src/container/__tests__/RelayContainer_setVariables-test.js
+++ b/src/container/__tests__/RelayContainer_setVariables-test.js
@@ -292,6 +292,15 @@ describe('RelayContainer.setVariables', function() {
       expect(mockInstance.state.queryData.entity).toBe(updatedQueryData);
     });
 
+    it('re-renders when there are pendingVariables', () => {
+      const before = render.mock.calls.length;
+      expect(render.mock.calls.length).toBe(1);
+      mockInstance.setVariables({site: 'www'});
+      jest.runAllTimers();
+      environment.primeCache.mock.requests[0].block();
+      expect(render.mock.calls.length - before).toBe(1);
+    });
+
     it('sets pendingVariables when request is in-flight', () => {
       mockInstance.setVariables({site: 'www'});
       jest.runAllTimers();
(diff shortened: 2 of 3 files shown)


facebook-github-bot-8:

Globegitter:
In my PR I previously had a `areQueryVariablesEqual` check: https://github.com/facebook/relay/pull/1069/files#diff-320b6df8cf530a681d201c75772401eaR779

chollier:
Oh interesting, @wincent is there a specific reason the check didn't make it to master?

facebook-github-bot-8:

src/container/RelayContainer.js
```diff
@@ -771,6 +771,7 @@ function createContainerComponent(
 
       const fragmentPointers = this._fragmentPointers;
       return (
+        !!nextState.relayProp.pendingVariables ||
```
josephsavona: This could cause unnecessary re-renders if props are received while a fetch is pending. Let's make this a comparison between the pending variables on the previous and next state.

josephsavona:
I'm not sure what happened to the variables comparison from @Globegitter's original PR, but I agree that we should do that. Also, can you add a test for this? There is a `RelayContainer_setVariables-test` that you can add to.

src/container/__tests__/RelayContainer_setVariables-test.js
```diff
@@ -292,6 +292,15 @@ describe('RelayContainer.setVariables', function() {
       expect(mockInstance.state.queryData.entity).toBe(updatedQueryData);
     });
 
+    it('re-renders when there are pendingVariables', () => {
+      const before = render.mock.calls.length;
+      expect(render.mock.calls.length).toBe(1);
+      mockInstance.setVariables({site: 'www'});
+      jest.runAllTimers();
+      environment.primeCache.mock.requests[0].block();
+      expect(render.mock.calls.length - before).toBe(1);
```
josephsavona: then
```
expect(render.mock.calls.length).toBe(1);
```

josephsavona: @chollier so this test passes even w/o the changes to RelayContainer? if you log the arguments to `render`, what are they? try logging with e.g `console.log(render.mock.calls)` and see what `props.relay` is for each render.

you could also try calling `jest.runAllTimers()` *after* the `block()` on line 300 and see what that does (might cause another render)

chollier:
@josephsavona just updated per your feedback regarding mockClear thanks! however I'm still a bit confused as to why when commenting the piece of code in `shouldComponentUpdate` this test doesn't fail

src/container/__tests__/RelayContainer_setVariables-test.js
chollier: I added some logging statements as well as a second `runAllTimers`:

```javascript
    it('re-renders when there are pendingVariables', () => {
      render.mockClear();
      console.log(render.mock)
      mockInstance.setVariables({site: 'www'});
      console.log(render.mock)
      jest.runAllTimers();
      console.log(render.mock)
      environment.primeCache.mock.requests[0].block();
      console.log(render.mock)
      console.log(render.mock.instances[0].props)
      expect(render.mock.calls.length).toBe(1);
    });
```

Here is the output 
```
{ calls: [], instances: [] }
{ calls: [], instances: [] }
{ calls: [], instances: [] }
{ calls: [ [] ],
  instances:
   [ { props: [Object],
       context: {},
       refs: {},
       updater: [Object],
       state: null,
       _reactInternalInstance: [Object] } ] }
{ entity:
   { __dataID__: '42',
     __fragments__: { '0::client': '42' },
     id: '42',
     url: '//url',
     profilePicture: { uri: '//url' } },
  relay:
   { applyUpdate: [Function: bound applyUpdate],
     commitUpdate: [Function: bound commitUpdate],
     forceFetch: [Function: bound forceFetch],
     getPendingTransactions: [Function: bound getPendingTransactions],
     hasFragmentData: [Function: bound hasFragmentData],
     hasOptimisticUpdate: [Function: bound hasOptimisticUpdate],
     hasPartialData: [Function: bound hasPartialData],
     pendingVariables: { site: 'www' },
     route: MockRoute { name: 'MockRoute', params: {}, queries: {}, uri: [Getter] },
     setVariables: [Function: bound setVariables],
     variables: { site: 'mobile' } } }
```

The output ends up being the same with or without the udpates in RelayContainer.js, I'm puzzled

chollier:
closing, I don't think it's relevant anymore with the rewrite.

PR #1313, opened 2016-08-01
Fix some syntax errors introduced by e2ba4a6ca07

chentsulin:
Obviously, there are some syntax errors introduced when code transforming. Try to fix it in this commit, or should we revert it directly?

https://github.com/facebook/relay/commit/e2ba4a6ca07f74cb90984302fb191e17d1e309bd

### DIFF
--- a/website/core/Header.js
+++ b/website/core/Header.js
@@ -15,7 +15,14 @@ const React = require('React');
 
 const Header = React.createClass({
   slug: function(string) {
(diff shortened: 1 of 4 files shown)


JenniferWang:
@spicyj Ben could you please take a look?

spicyj:
There's something broken with the github import (fb-internal link: https://fburl.com/398137359). I'll send a diff internally to fix this because I don't think importing this PR will actually fix it, unfortunately.

JenniferWang:
Will be fixed internally.

spicyj:
Fixed in fab36d9babe36d07e1619d59e2c451515e378c0f.

PR #1322, opened 2016-08-07
Wrap tracked fragments in their own types to support union types in mutations

### LINKED ISSUE #782: Fragment issue when committing Mutation

mattecapu:
Fixes #782 for my use case, when tracked fragments from a query on an union type leak in the intersected query of a mutation.

It breaks some tests for RelayQueryTracker and it needs new tests, please help me on this because I can't figure out how to do that.

### DIFF
--- a/src/store/RelayQueryTracker.js
+++ b/src/store/RelayQueryTracker.js
@@ -60,8 +60,16 @@ class RelayQueryTracker {
     const {isMerged, trackedNodes} = trackedNodesByID;
     if (!isMerged) {
       const trackedChildren = [];
+      let wrapperNode = null;
       trackedNodes.forEach(trackedQuery => {
-        trackedChildren.push(...trackedQuery.getChildren());
+        wrapperNode = new RelayQuery.Fragment.build(
+          'RelayQueryTracker',
+          trackedQuery.getType(),
+          trackedQuery.getChildren()
+        );
+        if (wrapperNode) {
+          trackedChildren.push(wrapperNode);
+        }
       });
       trackedNodes.length = 0;
       trackedNodesByID.isMerged = true;
--- a/src/store/RelayQueryTracker.js
+++ b/src/store/RelayQueryTracker.js
@@ -59,18 +59,13 @@ class RelayQueryTracker {
     }
     const {isMerged, trackedNodes} = trackedNodesByID;
     if (!isMerged) {
-      const trackedChildren = [];
-      let wrapperNode = null;
-      trackedNodes.forEach(trackedQuery => {
-        wrapperNode = new RelayQuery.Fragment.build(
-          'RelayQueryTracker',
+      const trackedChildren = trackedNodes.map(trackedQuery =>
+        new RelayQuery.Fragment.build(
+          'RelayQueryTrackerWrapperFragment',
           trackedQuery.getType(),
           trackedQuery.getChildren()
-        );
-        if (wrapperNode) {
-          trackedChildren.push(wrapperNode);
-        }
-      });
+        )
+      );
       trackedNodes.length = 0;
       trackedNodesByID.isMerged = true;
       let containerNode = RelayQuery.Fragment.build(
--- a/src/store/__tests__/RelayQueryTracker-test.js
+++ b/src/store/__tests__/RelayQueryTracker-test.js
@@ -57,7 +57,9 @@ describe('RelayQueryTracker', () => {
     tracker.trackNodeForID(query, 'client:1');
     const trackedChildren = tracker.getTrackedChildrenForID('client:1');
     expect(trackedChildren.length).toBe(1);
-    expect(trackedChildren[0])
+    expect(trackedChildren[0].getChildren().length).toBe(1);
+
+    expect(trackedChildren[0].getChildren()[0])
       .toEqualQueryNode(query.getFieldByStorageKey('actor'));
   });
 
@@ -75,7 +77,8 @@ describe('RelayQueryTracker', () => {
     const tracker = new RelayQueryTracker();
 
     tracker.trackNodeForID(query, nodeID);
-    const trackedChildren = sortChildren(tracker.getTrackedChildrenForID(nodeID));
+    const trackedChildren =
+        sortChildren(tracker.getTrackedChildrenForID(nodeID));
     const children = sortChildren(query.getChildren());
     expect(trackedChildren.length).toBe(3);
     children.forEach((child, ii) => {
@@ -102,7 +105,7 @@ describe('RelayQueryTracker', () => {
     tracker.trackNodeForID(actor, actorID);
     const children = sortChildren(actor.getChildren());
     const trackedChildren =
-      sortChildren(tracker.getTrackedChildrenForID(actorID));
+      sortChildren(tracker.getTrackedChildrenForID(actorID)[0].getChildren());
     children.forEach((child, ii) => {
       expect(trackedChildren[ii]).toEqualQueryNode(child);
     });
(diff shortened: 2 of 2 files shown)


facebook-github-bot-4:

facebook-github-bot-0:

src/store/RelayQueryTracker.js
```diff
@@ -60,8 +60,16 @@ class RelayQueryTracker {
     const {isMerged, trackedNodes} = trackedNodesByID;
     if (!isMerged) {
       const trackedChildren = [];
+      let wrapperNode = null;
       trackedNodes.forEach(trackedQuery => {
-        trackedChildren.push(...trackedQuery.getChildren());
+        wrapperNode = new RelayQuery.Fragment.build(
```
josephsavona: `const wrapperNode = ...`

src/store/RelayQueryTracker.js
```diff
@@ -60,8 +60,16 @@ class RelayQueryTracker {
...
+          'RelayQueryTracker',
+          trackedQuery.getType(),
+          trackedQuery.getChildren()
+        );
+        if (wrapperNode) {
```
josephsavona: Then you won't need this `if` check

josephsavona:
I suspect that the existing tests cover this, they simply need to be updated. This means going through the failing tests one by one, and adding the wrapping fragment to the test.

josephsavona:
Marking as needs revision bc of the test failures - feel free to comment on a specific test if you're unsure of how to update it!

src/store/RelayQueryTracker.js
josephsavona: Also I think `getType` *may* return null (I'm on mobile otherwise I'd check - look at RelayQuery.js). If so, then only add the wrapping fragment if there is a type.

src/store/RelayQueryTracker.js
mattecapu: Ok seemed useless to me too. I was mimicking [this](https://github.com/mattecapu/relay/blob/fae833a8e422206f7936a4b4363c07546cfea66d/src/store/RelayQueryTracker.js#L87) snippet of code.

src/store/RelayQueryTracker.js
mattecapu: This is the method
```js
getType(): string {
    return this.__concreteNode__.type;
  }
```
Thus it shouldn't return `null`?

src/store/__tests__/RelayQueryTracker-test.js
```diff
@@ -57,7 +57,8 @@ describe('RelayQueryTracker', () => {
     tracker.trackNodeForID(query, 'client:1');
     const trackedChildren = tracker.getTrackedChildrenForID('client:1');
     expect(trackedChildren.length).toBe(1);
-    expect(trackedChildren[0])
+    expect(trackedChildren[0].getChildren().length).toBe(1);
+    expect(trackedChildren[0].getChildren()[0])
       .toEqualQueryNode(query.getFieldByStorageKey('actor'));
```
mattecapu: This still doesn't pass. It seems the node I'm checking isn't of the right type (`Invariant Violation: printRelayOSSQuery(): Unsupported node type.`) but if I `console.log` the two objects are of the same type (and are actually the same).

mattecapu:
I updated some failing tests, but I suspect some of the passing ones are false positives.

src/store/__tests__/RelayQueryTracker-test.js
```diff
@@ -75,7 +77,8 @@ describe('RelayQueryTracker', () => {
     const tracker = new RelayQueryTracker();
 
     tracker.trackNodeForID(query, nodeID);
-    const trackedChildren = sortChildren(tracker.getTrackedChildrenForID(nodeID));
+    const trackedChildren =
+        sortChildren(tracker.getTrackedChildrenForID(nodeID));
```
mattecapu: Why does this test pass?

mattecapu:
up?

sibelius:
any progress on this?

josephsavona:
Thanks for the PR, for exploring this space, and for your patience. While this PR helps, we've found several difficulties with dynamically constructing mutations queries. For this and other reasons (predictability/simplicity) we've created a new, static mutation API: [`Relay.GraphQLMutation`](https://facebook.github.io/relay/docs/api-reference-relay-graphql-mutation.html) which we recommend. I'm going to close this PR, but we appreciate the time you put into this: seeing the problems with the current mutations helped us to develop the  new API.

mattecapu:
👍 thanks for the feedback!

PR #1336, opened 2016-08-13
Make `callsToGraphQL` and `callsFromGraphQL` preserve type (fixes #1256)

### LINKED ISSUE #1256: toGraphQL gets null type for enum variable

denvned:
As @papigers currently doesn't have time to finish https://github.com/facebook/relay/pull/1257 I decided to take over.

### DIFF
--- a/src/query/__tests__/RelayQueryRoot-test.js
+++ b/src/query/__tests__/RelayQueryRoot-test.js
@@ -159,7 +159,7 @@ describe('RelayQueryRoot', () => {
     expect(me.getIdentifyingArg()).toEqual(undefined);
 
     expect(usernames.getIdentifyingArg()).toEqual(
-      {name: 'names', value: 'mroch'}
+      {name: 'names', type: '[String!]!', value: 'mroch'}
     );
 
     expect(getNode(Relay.QL`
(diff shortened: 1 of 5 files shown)


josephsavona:
@facebook-github-bot shipit

josephsavona:
@papigers thanks for getting this started, and thanks @denvned for finishing! yay team ;-)

Issue #1337, opened 2016-08-13
"Using `null` as a rangeBehavior value is deprecated" when using "ignore"


miracle2k:
I implement `rangeBehaviours` as a function and return "ignore". Relay shows me this warning:

`Using `null` as a rangeBehavior value is deprecated. Use `ignore` to avoid refetching a range.
`
Callsite: https://github.com/facebook/relay/blob/712f5926cb67312aac284feef3ab3ed5e147c587/src/traversal/writeRelayUpdatePayload.js#L416

Is the condition there wrong? I'm thinking it might need to be `(!rangeBehavior && rangeBehavior !== IGNORE)`

xuorig:
Oops, good find @miracle2k. The condition was shared because we also want to early return when it's `IGNORE` so `(!rangeBehavior && rangeBehavior !== IGNORE)` would probably not be what we want here.

PR coming for this!

PR #1338, opened 2016-08-13
Range behaviours: Don't warn when value is IGNORE

xuorig:
Closes https://github.com/facebook/relay/issues/1337

`IGNORE` should early return but not warn :)

I believe existing tests should cover that. Could write a regression and check for no warnings but that feels weird.


wincent:
Thanks @xuorig. Sorry for the delay in getting to this. Aiming to cut a release this week so just going through and looking at PRs to merge now. Will queue this one up.

wincent:
@facebook-github-bot import

PR #1360, opened 2016-08-30
Guide Update

juangl:
Just erasing markdown inside the code, as it is not parsed.

If you don't like my way just please change that, because could be confused with `template strings`.

### DIFF
--- a/docs/APIReference-Container.md
+++ b/docs/APIReference-Container.md
@@ -155,10 +155,10 @@ module.exports = Relay.createContainer(ProfilePicture, {
   initialVariables: {size: 50},
   fragments: {
     user: () => Relay.QL`
-      # The variable defined above is available here as `$size`.
+      # The variable defined above is available here as '$size'.
       # Any variable referenced here is required to have been defined in initialVariables above.
-      # An `undefined` variable value will throw an `Invariant Violation` exception.
-      # Use `null` to initialize unknown values.
+      # An 'undefined' variable value will throw an 'Invariant Violation' exception.
+      # Use 'null' to initialize unknown values.
       fragment on User { profilePicture(size: $size) { ... } }
     `,
   },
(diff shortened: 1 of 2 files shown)


facebook-github-bot-4:

facebook-github-bot-2:

wincent:
Thanks for the PR, @juangl.

Using Markdown style formatting in comments is a convention that we follow widely (not just in samples in the documentation, but in the source code as well, in this and in other repos too) so this is not a change that we would want to make unless we were doing it consistently everywhere, which would require further discussion.

So I am going to close this for now. If you feel strongly about changing the convention please open an issue and we'll label it with the "discussion" label and we can talk about it further. I do worry that this would tend to be somewhat of a bikeshed issue, and I also think that in a case like this the value of sticking to a well-established convention outweighs the cost of conversion, and the cost of possible confusion that people might feel seeing backticks inside of comments (note: if you are concerned that people might see them as template literals, are you also worried that they might perceive single and double-quoted strings as string literals?).

Thanks once again for the submission though.

juangl:
@wincent Thanks for response. I will be more specific. It gets confusing particulary  when backticks  are inside a template literal. I won't open a issue because I also think that is not too relevant, but I make you know.

```
Relay.QL`
  # This fragment only applies to objects of type `User`.
  fragment on User {
    # Set the `size` argument to a GraphQL variable named `$size` so that we can
    # later change its value via the slider.
    profilePhoto(size: $size) {
      # Get the appropriate URI for the given size, for example on a CDN.
      uri,
    },
  }
`
```

wincent:
I see, thanks for clarifying. So it's not confusing; it's actually syntactically invalid. We should definitely fix those ones, but leave the others as they were before. Want to submit an update?

juangl:
@wikichen Sure! It's done. Hope your review.

wincent:
Thanks!

wincent:
@facebook-github-bot import

facebook-github-bot-5:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3793191) to review internal test results.

Issue #1357, opened 2016-08-29
Using Array type for fragment variable


josercruz01:
I've run into an issue where the `RelayContainer` resets its current fragment variables state and forgets variables I've already set via `this.props.relay.setVariables`. 

I've tracked down the problem to the method `RelayContainer#resetPropOverridesForVariables`, which is called inside the `RelayContainer#componentWillReceiveProps`. This method resets the container's fragment variables state to the initial variables if any of the new props has changed.

So basically the method checks all fragment variables and resets the state if at least one changed, but in my case it always gets reset because one of those variables is an `array` which never returns true for `===` comparison.

You can see below for my `prepareParams` method which converts the query param into an array. My question is, would it make sense to provide a hook for custom equality comparison (maybe check if the prop has an `equals` method, that way I could send in an Immutable List) or should I not be converting to array at all and send a string to the server?

I would rather not send a string if possible, just so I don't have to do extra work on the client and server to parse the string to array, any recommendations?

```
const prepareScreeningParams = function(params, { location }) {
  const { query } = location;
  const variables = Object.assign({}, params, {
    keywords: query.keywords,
    referrer: query.referrer,
    university: query.university,
    interval: parseInt(query.interval, 10),

    // This is now an array of ints
    currentStageIds: query.currentStageIds ?
      _.flatten([query.currentStageIds]).map(x => parseInt(x, 10)) : null
  });
  return _.mapValues(variables, x => x || null);
};
```

chirag04:
I think i have seen a similar issue in my setup. I don't have a resolution yet.

josephsavona:
Thanks for filing the issue. Its probably reasonable for the variable comparison here to use `areEqual` instead of `===` since variables can be arrays or objects. 

Feel free to send a PR for this!

josercruz01:
Thanks Joseph, opened up a PR.

PR #1362, opened 2016-08-30
Do not reset variables if outside variable props of type Object are the same

josercruz01:
Closes https://github.com/facebook/relay/issues/1357

When new props are received on a `RelayContainer` a verification step is executed with the new props and the current relay variables and if at least one prop variable has a different value than its corresponding relay variable then the current relay variables are reset to the initial state. But the problem was that `===` was used for comparison and this always return a mismatch for props of type `Object` (arrays, objects) resulting in relay variables always being reset to the initial state even if the prop variables are the same.

See github issue for a detailed example.

The fix was to use `areEquals` which does a deep comparison and can match `Object` javascript types.

### DIFF
--- a/src/container/__tests__/RelayContainer_setVariables-test.js
+++ b/src/container/__tests__/RelayContainer_setVariables-test.js
@@ -792,5 +792,99 @@ describe('RelayContainer.setVariables', function() {
         size: 48,
       });
     });
+
+    it('does not reset variables if outside props are the same', () => {
+      class MockInnerComponent extends React.Component {
+        render() {
+          return <div />;
+        }
+      }
+
+      const MockInnerContainer = Relay.createContainer(MockInnerComponent, {
+        fragments: {
+          entity: () => Relay.QL`  fragment on User {
+                        url(site:$site)
+                        storySearch(query:$query) {
+                          id
+                        }
+                        profilePicture(size:$size) {
+                          uri
+                        }
+                      }`,
+        },
+        initialVariables: {
+          site: 'mobile',
+          query: undefined, // <-- Object type
+          size: undefined // <-- Array type
+        },
+      });
+
+      class MockWrapperComponent extends React.Component {
+        render() {
+          return (
+            <MockInnerContainer
+              ref='inner'
+              query={this.props.relay.variables.query}
+              size={this.props.relay.variables.size}
+              entity={this.props.entity}
+            />
+          );
+        }
+      }
+
+      MockContainer = Relay.createContainer(MockWrapperComponent, {
+        fragments: {
+          entity: variables => Relay.QL`  fragment on User {
+            ${MockInnerContainer.getFragment('entity', {
+              query: variables.query,
+              size: variables.size,
+            })}
+          }`,
+        },
+        initialVariables: {
+          query: { text: 'recent' },
+          size: [32, 64]
+        },
+      });
+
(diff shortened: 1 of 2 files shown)


wincent:
Thanks @josercruz01. I have a couple of minor formatting nits but I'll just import this and patch them.

wincent:
@facebook-github-bot import

src/container/__tests__/RelayContainer_setVariables-test.js
```diff
@@ -792,5 +792,99 @@ describe('RelayContainer.setVariables', function() {
...
+        fragments: {
+          entity: () => Relay.QL`  fragment on User {
+                        url(site:$site)
+                        storySearch(query:$query) {
+                          id
+                        }
+                        profilePicture(size:$size) {
+                          uri
+                        }
+                      }`,
```
josephsavona: nit: fix the indentation, spaces between args/values (`size: $size`)

src/container/__tests__/RelayContainer_setVariables-test.js
```diff
@@ -792,5 +792,99 @@ describe('RelayContainer.setVariables', function() {
...
+              ref='inner'
+              query={this.props.relay.variables.query}
+              size={this.props.relay.variables.size}
+              entity={this.props.entity}
+            />
+          );
+        }
+      }
+
+      MockContainer = Relay.createContainer(MockWrapperComponent, {
```
josephsavona: nit: use a consistent naming pattern for the inner/outer components - maybe call this `MockWrapperContainer`?

src/container/__tests__/RelayContainer_setVariables-test.js
josercruz01: Done.

src/container/__tests__/RelayContainer_setVariables-test.js
josercruz01: Hey I made this change on the new test I created by creating a new variable and not using the `MockContainer` variable declared on the top level describe. I assume it was done this way for speed maybe? Let me know your thoughts.

Also, there are still some other test on this file that store the container of `MockWrapperComponent` in the top level `MockContainer` variable, should I go ahead and update those tests as well?

josercruz01:
Hey @wincent I pushed one commit to resolve some of the comments made by Joseph and while doing so I noticed a couple of linter warnings on the new test code. Are those the formatting bits you are taking care of in your patch or should I go ahead and push a commit for those?

wincent:
I was going to fix the nits that @josephsavona mentioned, but you already did that. Linter just exploded for me. What lints are you seeing?

josercruz01:
Oh sorry about that.

Here is the output of `npm run lint` (only showing files touched by my PR):

```
/Users/josecruz/src/relay/src/container/__tests__/RelayContainer_setVariables-test.js
  818:26  warning  Missing trailing comma           comma-dangle
  826:19  warning  Unexpected usage of singlequote  jsx-quotes
  846:25  warning  Missing trailing comma           comma-dangle
  856:23  warning  Missing trailing comma           comma-dangle
  861:20  warning  Missing trailing comma           comma-dangle
  868:23  warning  Missing trailing comma           comma-dangle
  873:23  warning  Missing trailing comma           comma-dangle
  880:23  warning  Missing trailing comma           comma-dangle
  886:23  warning  Missing trailing comma           comma-dangle
```

PR #1363, opened 2016-08-30
Make `enforceSchema` a plugin option

gabriel-laet:
Allowing `babel-relay-plugin` be configured with `enforceSchema` by default,
so you don't have to set it in your Babel's options.

This is specially useful if you want to ship a custom plugin that includes default settings for your Relay/GraphQL projects.


leebyron:
Sorry for not getting to this PR. We're deprecating the `babel-relay-plugin` package in favor of `babel-plugin-relay` and `relay-compiler`

Issue #1300, opened 2016-07-26
Pattern matching variables for queries in mutation.getFatQuery() ?


joonhocho:
As I commented on https://github.com/facebook/relay/issues/1046#issuecomment-235314338, I want to have a fatQuery that matches all variables in the store so that I do not have to specify all variable possibilities stored in the current relay store.
I tried to not specify variable, but it fails the query validation.

For example, 

``` javascript
getFatQuery() {
    return Relay.QL`
      fragment on Payload {
        viewer {
          fieldWithArguments(id: *, name: "a")
          // * as a wild card that matches all ids found in the current relay store.
        }
      }
    `;
  }
```

joonhocho:
I also noticed that @relay(pattern: true) only applies to connections, not other fields with arguments.

arypurnomoz:
is there anyway to declare variable for fatQuery?

josephsavona:
Fat queries are evaluated without looking at arguments, so in the example it should be sufficient to write just `fieldWithArguments` in the fat query instead of `fieldWithArguments(id: ...)`. If that isn't passing validation, we can change `@relay(pattern: true)` to apply to any fields within a fat query. The logic for handling this in the babel plugin - PRs welcome!

joonhocho:
@josephsavona Are you sure that `@relay(pattern: true)` works on any fields? It didn't work for me when I tried as it caused validation error. I think it only works for relay connection fields.

josephsavona:
> Are you sure that `@relay(pattern: true)` works on any fields?

No, I suspect it won't. As I mentioned, though, we'd welcome a PR to support this.

ddgromit:
I'm assuming in your example the arguments are required and you're (like me) expecting that `fieldWithArguments @relay(pattern: true)` will make those arguments not required.  This [errors out in GraphQL](https://github.com/graphql/graphql-js/blob/3083fc5743ef228326588f0d1c6b1ebcdf0c4a52/src/validation/rules/ProvidedNonNullArguments.js#L23) before it even gets to Relay.  

In a relay ConnectionType all the arguments `before`, `after`, etc are optional.  Omitting them passes GraphQL validation, but is caught at Relay validation time which checks that [at least one exists](https://github.com/facebook/relay/blob/master/scripts/babel-relay-plugin/src/RelayQLPrinter.js#L798) or the pattern directive was provided.   The validation happens in Relay so its pretty easy for Relay to get around it.

[...]

PR #1376, opened 2016-09-08
Allow @relay(pattern:true) directive on fields with required arguments.

ddgromit:
## Overview

Issue #1300. Allows using `@relay(pattern: true)` on fat queries for non-connection fields that have required arguments.  

Currently the pattern directive only can be used on connections and fields without arguments.  This prevents fat queries from being used to force the server to invalidate separate fields you know have changed but don't directly relate to the current object being mutated.
## Implementation

Disables the GraphQL rule for `ProvidedNonNullArguments`, instead relying on Relay to throw those errors.  
## Example

Full star wars example on [this branch](https://github.com/relayjs/relay-examples/compare/master...ddgromit:query-arg-patterns#diff-4826482ddf3139031cac42eff2e6706dR17)

You might have non-connection field that you query with arguments like `viewer { factions(names: ["empire", "rebels"]) }`.  Then run a mutation that updates the names of all the ships across all factions - not just the ones you've queried.  This change allows you to make a fat query that says all ships across all calls to `factions` are now invalid.

```
  fragments: {
    viewer: () => Relay.QL`
      fragment on Viewer {
        factions(names: ["empire", "rebels"]) {
          ships(first: 10) {
            edges {
              node {
                name
              }
            }
          }
        }
      }
    `,
  },
```

```
  getMutation() {
    return Relay.QL`mutation { changeAllShips }`;
  }

  getFatQuery() {
    return Relay.QL`
      fragment on ChangeAllShipsPayload @relay(pattern: true) {
        viewer {
          name
          factions {
            ships
          }
        }
      }
    `;
  }
```

### DIFF
--- a/scripts/babel-relay-plugin/src/RelayQLTransformer.js
+++ b/scripts/babel-relay-plugin/src/RelayQLTransformer.js
@@ -270,13 +270,6 @@ class RelayQLTransformer {
           'graphql/validation/rules/VariablesInAllowedPosition'
         ).VariablesInAllowedPosition,
       ];
-      if (!isMutation) {
-        rules.push(
-          require(
-            'graphql/validation/rules/ProvidedNonNullArguments'
-          ).ProvidedNonNullArguments
-        );
-      }
       validationErrors = validate(this.schema, document, rules);
     }
 
--- a/scripts/babel-relay-plugin/src/RelayQLTransformer.js
+++ b/scripts/babel-relay-plugin/src/RelayQLTransformer.js
@@ -28,6 +28,7 @@ const {
   RelayQLSubscription,
 } = require('./RelayQLAST');
 const RelayQLPrinter = require('./RelayQLPrinter');
+const GraphQLRelayDirective = require('./GraphQLRelayDirective');
 
 const invariant = require('./invariant');
 const util = require('util');
@@ -237,6 +238,16 @@ class RelayQLTransformer {
     const isMutation =
       definition.kind === 'OperationDefinition' &&
       definition.operation === 'mutation';
+    const isPattern =
+      definition.kind === 'FragmentDefinition' &&
+      (definition.directives || []).some(
+        directive => directive.name.value === GraphQLRelayDirective.name &&
+          (directive.arguments || []).some(
+            arg => arg.name.value === 'pattern' &&
+              arg.value.kind === 'BooleanValue' &&
+              arg.value.value
+          )
+      );
 
     const validator = this.options.validator;
     let validationErrors;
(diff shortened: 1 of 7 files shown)


facebook-github-bot-8:

facebook-github-bot-1:

scripts/babel-relay-plugin/src/RelayQLTransformer.js
```diff
@@ -270,13 +270,6 @@ class RelayQLTransformer {
           'graphql/validation/rules/VariablesInAllowedPosition'
         ).VariablesInAllowedPosition,
       ];
-      if (!isMutation) {
-        rules.push(
-          require(
-            'graphql/validation/rules/ProvidedNonNullArguments'
-          ).ProvidedNonNullArguments
-        );
-      }
```
josephsavona: This isn't safe to remove

scripts/babel-relay-plugin/src/__fixtures__/fragmentRequiredArgs.fixture
```diff
@@ -0,0 +1,17 @@
+Input:
+var Relay = require('react-relay');
+var x = Relay.QL`
+  fragment on Viewer {
+    actorByName() {
```
josephsavona: Is this meant as a test case that forgetting arguments without providing `pattern: true` still fails? If so, it is not actually testing that. This is failing because it is syntactically invalid GraphQL that does not *parse*. If you correct it to the syntactically correct `actorByName { name }`, it *will* parse and will also pass validation because of the removed "ProvidedNonNull" validation rule above.

We need to keep the non-null validation rule for queries/fragments.

ddgromit: Ah great catch.  In the fixture I updated the field call and moved the `relay(pattern: true)` directive to the `fragment on Viewer` level.

scripts/babel-relay-plugin/src/RelayQLTransformer.js
ddgromit: Rewrote the validator to only skip the `ProvidedNonNullArguments` if the top-level definition is a `fragment` with a `relay(pattern: true)` directive.

This doesn't support inline fragments but in a fat query I believe you'd only be using pattern on the top-level fragment.  Supporting inline fragments would require overriding GraphQL's `ProvidedNonNullArguments` rule.

ddgromit:
Ready for another review.  Changed to only skip NonNullArgument check if the top level document is a pattern.

proProbe:
Hello! I would love to have this! how is the PR actually doing?

ddgromit:
@josephsavona ^ would appreciate another look at this PR

josephsavona:
Thanks for the PR, for exploring this space, and for your patience. While this PR helps, we've found several difficulties with dynamically constructing mutations queries. For this and other reasons (predictability/simplicity) we've created a new, static mutation API: [`Relay.GraphQLMutation`](https://facebook.github.io/relay/docs/api-reference-relay-graphql-mutation.html) which we recommend. I'm going to close this PR, but we appreciate the time you put into this: seeing the problems with the current mutations helped us to develop the  new API.

PR #1396, opened 2016-09-27
Remove duplicate manual mock

flarnie:
We are adding a warning to Jest for duplicated manual mock files. It
detected this duplicate file when run internally. The remaining versino
of this mock is in
`src/legacy/mutation/__mocks__GraphQLMutatorConstants.js`
That one is in the `__mocks__` file colocated with
`GraphQLMutatorConstants.js` so it seems more correct than the
duplicate.

Test plan:
Ran unit tests, lint, and typechecks


kassens:
Seems like the file was removed some other time.

Issue #1242, opened 2016-06-26
Document @relay(plural:true) directive


GrigoryPtashko:
Hello.

I cannot seem to find the place in the docs (I've read them here https://facebook.github.io/relay/docs/getting-started.html#content) where it would be described the meaning of `@relay(plural:true)` and possibly the like structures.
These syntactic structures seem to be the only things that I don't understand in Relay so far.

Docs, source code, wiki?

Thank you.

wincent:
Thanks for asking.

I think this might be a gap in the docs. If you search Stack Overflow you will find a number of scattered references to the directive (eg. [this one](http://stackoverflow.com/questions/32491117/how-to-get-relayjs-to-understand-that-a-response-from-graphql-is-an-array-of-ite/32508505#32508505)), but I don't think it's explicitly spelled out in the documentation anywhere.

We'd welcome a PR for this. I expect it would most likely go somewhere under [this section](http://facebook.github.io/relay/docs/graphql-relay-specification.html).

josercruz01:
Hey @wincent just put up a PR for this. I placed it under the `Relay.QL` section ([here](http://facebook.github.io/relay/docs/api-reference-relay-ql.html#content)) for now since I saw some of the other directives there already (like `@include`) but I saw you mentioned it would most likely go in [this section](http://facebook.github.io/relay/docs/graphql-relay-specification.html). Is there a particular place you would recommend on that section for this documentation or is the `Relay.QL` ok?

wincent:
@josercruz01 : Thanks for the PR, I'll comment on that instead of here.

PR #1422, opened 2016-10-21
Adding documentation for the '@relay(plural: true)' directive

josercruz01:
# What's up

Resolves https://github.com/facebook/relay/issues/1242

Adds documentation for the `@relay(plural: true)` directive in the relay docs. This should help new adopters understand how to use it.

I placed it inside the `Relay.QL` section but if you guys feel it should be somewhere else just let me know.


wincent:
@facebook-github-bot shipit

PR #1425, opened 2016-10-23
Include Link to Star Wars Wiki Page

DanielRHayes:
The Star Wars reference may not be known universally, especially to an international audience, so a link to the wiki might help someone who is unfamiliar.

Completed CLA


wincent:
Thanks, @DanielRHayes! Will merge.

Issue #586, opened 2015-11-11
Uploading array of files


oguzbilgic:
Is there a way to upload array of files?

```js
getFiles(): ?FileMap

// Type of the FileMap object
type FileMap = {[key: string]: File};
```

Would it make sense to improve FileMap type to something like this: 

```js
type FileMap = {[key: string]: File | [File]};
```

devknoll:
@oguzbilgic Neat idea. Would you mind opening a PR?

You could work around this today by passing in an array to your mutations constructor and mapping each file to a unique key in `getFiles`, but it seems useful to be able to take advantage of the built-in support for arrays in the browser and web servers.

wincent:
Thanks for filing this issue @oguzbilgic!

This is important, so I'm going to fold it into #538, which is about overhauling the mutations API. (It will be easier to manage with the various interrelated mutations issues in a single place.)

eugenehp:
@oguzbilgic were you able to find a solution for this?

chrise86:
@oguzbilgic also interested to hear if you found a solution 👍

sibelius:
I think a file map makes more sense, should use the filename in the key and disregard it on server

PR #1431, opened 2016-10-27
Add support to multiple files upload

nossila:
A suggestion was made on https://github.com/facebook/relay/issues/586 but I thought using:

``` javascript
type FileMap = {[key: string]: File | FileList};
```

was a better idea since I would not need write a loop through the FileList (since its what is provided by the file input's API) to construct an array of files every time I need to send multiple files to a Mutation, this way Relay will handle it for me, I just need to write:

``` javascript
class AttachMultipleDocumentsMutation extends Relay.Mutation {
  getFiles() {
    return {
      files: this.props.files,
    };
  }
}
class FilesUploader extends React.Component {
  handleSubmit() {
    Relay.Store.commitUpdate(
      new AttachMultipleDocumentsMutation({file: this.refs.fileInput.files})
    );
  }
}
```

### DIFF
--- a/docs/APIReference-Mutation.md
+++ b/docs/APIReference-Mutation.md
@@ -387,10 +387,10 @@ class LikeStoryMutation extends Relay.Mutation {
 getFiles(): ?FileMap
 
 // Type of the FileMap object
-type FileMap = {[key: string]: File};
+type FileMap = {[key: string]: File | FileList};
 ```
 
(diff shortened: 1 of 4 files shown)


leebyron:
I'm so sorry your PR has sat here this long. Unfortunately it does not merge well anymore and we're attempting to focus on Relay Modern going forward.

I'm going to close this PR, however if this is still important to you please rebase it and reopen

PR #1436, opened 2016-11-03
Add `replaceSchema` method to the `babel-relay-plugin`

nodkz:
I'm using Webpack with babel-loader. So when I changing my GraphQL Schema, my `schema.graphql.json` file is rebuilt automatically. And babelRelayPlugin sends to console errors according to old schema.

But loaded instance of `babelRelayPlugin` in babel with webpack has no way to update the schema. Required full restart for webpack (and it takes minutes). This PR adds method `replaceSchema` to `babelRelayPlugin` for updating schema on the fly.

Right now my `./lib/babel-relay-plugin.js` for `babel-loader` in Webpack has following working code, which tracks changes and calls new `replaceSchema` method:
```js
const getBabelRelayPlugin = require('babel-relay-plugin');
const fs = require('fs');
const path = require('path');
const chalk = require('chalk');

// path to generated schema
const schemaPath = path.resolve(__dirname, '../../build/schema.graphql.json');

// errorless start if schema file does not exist
let schema;
try {
  fs.accessSync(schemaPath, fs.F_OK);
  schema = require(schemaPath);
} catch (e) {
  schema = null;
  console.log(chalk.white.bgRed.bold(`[BabelRelayPlugin] Cannot resolve GraphQL Schema from file ${schemaPath}`));
}
const emptyPlugin = () => ({});
const plugin = schema ? getBabelRelayPlugin(schema.data) : emptyPlugin;

// if `replaceSchema` exists then create unblocking event loop task which track changes 
// in the schema file, and reload schema in plugin Transformer
// PS. fs.watch blocks babel from exit, so using `setTimeout` with `unref`
// Checking changes every 2 second.
if (plugin.replaceSchema) {
  let prevMtime;
  const watcherFn = () => {
    try {
      const stats = fs.statSync(schemaPath);
      if (stats) {
        if (!prevMtime) prevMtime = stats.mtime;
        if (stats.mtime.getTime() !== prevMtime.getTime()) {
          prevMtime = stats.mtime;
          console.log(chalk.yellow.bold(`[BabelRelayPlugin] Reload GraphQL Schema from file ${schemaPath}`));
          delete require.cache[require.resolve(schemaPath)];
          schema = require(schemaPath);
          plugin.replaceSchema(schema.data);
        }
      }
      setTimeout(watcherFn, 2000).unref();
    } catch (e) {
      console.log(chalk.red.bold(`[BabelRelayPlugin] ${e}`));
    }
  };
  watcherFn();
}

module.exports = plugin;
```

For testing you may use this built package (which contains code from this PR):
```
npm i https://www.dropbox.com/s/jb0qjkkfvvr6vbe/babel-relay-plugin-0.9.3.tgz\?dl\=1
```

### DIFF
--- a/scripts/babel-relay-plugin/src/getBabelRelayPlugin.js
+++ b/scripts/babel-relay-plugin/src/getBabelRelayPlugin.js
@@ -48,15 +48,19 @@ function getBabelRelayPlugin(
     function() {} :
     console.warn.bind(console);
 
-  const schema = getSchema(schemaProvider);
-  const transformer = new RelayQLTransformer(schema, {
-    inputArgumentName: options.inputArgumentName,
-    snakeCase: !!options.snakeCase,
-    substituteVariables: !!options.substituteVariables,
-    validator: options.validator,
-  });
+  let transformer;
+  function initTransformer(schemaProvider) {
+    const schema = getSchema(schemaProvider);
+    transformer = new RelayQLTransformer(schema, {
+      inputArgumentName: options.inputArgumentName,
+      snakeCase: !!options.snakeCase,
+      substituteVariables: !!options.substituteVariables,
+      validator: options.validator,
+    });
+  }
+  initTransformer(schemaProvider);
 
-  return function({Plugin, types, version}) {
+  const plugin = function({Plugin, types, version}) {
     return babelAdapter(Plugin, types, version, 'relay-query', t => ({
       visitor: {
         /**
@@ -242,6 +246,9 @@ function getBabelRelayPlugin(
       },
     }));
   };
+
+  plugin.replaceSchema = initTransformer;
+  return plugin;
 }
 
 function getSchema(schemaProvider: GraphQLSchemaProvider): GraphQLSchema {


nodkz:
@steveluscher please review this, and if it's ok bump new version. Thanks.

nodkz:
@josephsavona can you review it and publish?

josephsavona:
@nodkz Thanks for the PR. I understand the use-case and that it is helpful to have the plugin reflect schema changes without completely restarting. However, this seems like something that should be achievable via plugin composition. My first thought would be to create a wrapper plugin that delegates all transform functions to an instance returned by `getBabelRelayPlugin`. Each time the schema changes, the wrapper would update that instance with a new Relay plugin.

nodkz:
@josephsavona yep, plugin composition is the better way. But I did not do that cause: 
- don't know how to write babel plugin wrapper (and have no time right now for reading docs)
- I saw code that moves this plugin to first place in babel's plugin list (so wrapper should has same functionality).
So both of this causes brings me to current more simple solution ;) Will be cool if you accept this PR as intermediate solution.

---
PS. In this week we should make a technical launch of our WebApp. But I found that our forms are sucks. So I decide to delay the launch and marry forms with Relay and GraphQL, has validation. Got such sweet HOC, that works like Relay.Container and its fragment can be used for normal Relay.Container. Still polishing it, but already got type validation on build and tons of invariant checks in the browser console.

screen shot 2016-11-10 at 18 33 52

josephsavona:
I'm going to close this. For now we recommend using composition, and we'll keep this use case in mind for future versions of the plugin.

nodkz:
😉 so nice and gently decline 💪

Don't worry i agree with you. Composition will be better, but right not i'm focused on another things and have not free time for improvement. Will restart server again for schema changes. 

So let's waiting my or somebody fatigue.

nodkz:
BabelRelayPlugin with hot reload 🔥
Reflect GraphQL Schema changes in json file without completely restarting dev server.
https://github.com/nodkz/babel-plugin-transform-relay-hot

PR #1440, opened 2016-11-09
Documentation: Configure plugins on top level in sample .babelrc

ctavan:
The plugins config in the example `.babelrc` must be on the top level, not inside the presets array.


wincent:
Thanks, @ctavan! Will merge.

PR #1441, opened 2016-11-09
Fix old links

lucasbento:


wincent:
Thanks @lucasbento, will merge.

PR #1446, opened 2016-11-15
Upgrade babel relay plugin to use GraphQL 0.8.1

gauravtiwari:
* Also fixes yarn.lock resolved module paths


josephsavona:
@gauravtiwari Thanks for the PR. Due to internal configuration this isn't *quite* as simple as just bumping the version number (and the plugin also needs changes to support things like null literals). I'm working on upgrading to 0.8.1 right now so this should be on master soon.

josephsavona:
I'm going to close this since I'm working on a change that includes this. Thanks for the contribution!

gauravtiwari:
Thanks @josephsavona 👍

PR #1471, opened 2016-12-07
Add yarn.lock and update yarn.lock of plugin

kassens:


kassens:
We actually need the offline-mirror variant internally, but that doesn't work in OSS since we don't have the files for the offline-mirror there.

PR #1499, opened 2017-01-12
corrects in the doc, parentId is optional when using NODE_DELETE

vnctaing:
According to https://github.com/facebook/relay/blob/ad5efb124cf455a42c0adb10fb1e977fe051811e/src/tools/RelayTypes.js#L157-L163 parentId seems to be optional.

The current documentation seems to not specify it : https://facebook.github.io/relay/docs/guides-mutations.html#node-delete


wincent:
You're right that it is optional. Thanks for the PR!

PR #1503, opened 2017-01-17
Fix #1494 - RelayContainer mutations act on outdated environment

### LINKED ISSUE #1494: Mutations after rendering with a new environment
After reading #233, I implemented "resetting the store" in my app like this:

    export const initApp = (token = '') => {
      const environment = new Relay.Environment()
      environment.injectNetworkLayer(new Relay.DefaultNetworkLayer('/graphql'))

      const queryConfig = new AppQueryConfig({ token })

      ReactDOM.render(
        <Relay.Renderer Container={AppContainer} queryConfig={queryConfig} environment={environment} />,
        document.getElementById('root')
      )
    }

But I've been having a problem with mutations after the store is reset.

My theory is that for containers that have been rendered at least once already, `this.props.relay.commitUpdate` doesn't get updated to reference the new environment's `commitUpdate` function. It seems that `this.state.relayProp.commitUpdate` in RelayContainer is set when the container is first rendered, and isn't updated when the environment changes. I think this could be causing issues in my `RANGE_ADD` mutations.

I've found 3 ways to make my issues disappear:
1. Forcing react to render the entire app again, using e.g. `ReactDOM.unmountComponentAtNode(document.getElementById('root'))`,
2. Creating a new `AppContainer` (i.e. calling `Relay.createContainer()` again) when updating the environment, and
3. Modifying the `Relay.Container` code

This change to `RelayContainer.prototype._initialize` fixes the issue for me, but I definitely don't understand the library enough to have any idea on whether it is a good or appropriate change:

	relayProp: this.state.relayProp.route === context.route && require('fbjs/lib/shallowEqual')(this.state.relayProp.variables, nextVariables) ? this.state.relayProp : (0, _extends3['default'])({}, this.state.relayProp, {
      route: context.route,
      variables: nextVariables,
      commitUpdate: context.relay.commitUpdate,
      applyUpdate: context.relay.applyUpdate
    })

(Please let me know if I shouldn't even be using the library in this way.)

colllin:
Fixes #1494 

Creates methods `commitUpdate` and `applyUpdate` on RelayContainer, which always delegate to *the current* RelayEnvironment (from `this.context.relay`), and passes these new functions to the underlying component as `props.relay.commitUpdate` and `props.relay.applyUpdate`, instead of direct references to the `RelayEnvironment` functions.

I also have this alternative solution ready, in case you prefer it:
https://github.com/facebook/relay/compare/master...colllin:relay-container-environment-sync
That solution just captures changes to the environment within `componentWillReceiveProps`, and updates `this.state.relayProp.commitUpdate` and `this.state.relayProp.applyUpdate`, which in turn update those props on the underlying component.

I also added tests, which, before my changes, failed with this output:
```
 FAIL  src/container/__tests__/RelayContainer-test.js (23.443s)
  ● RelayContainer › props.relay.applyUpdate › updates when the underlying RelayEnvironment changes

    expect(received).toBe(expected)
    
    Expected value to be (using ===):
      1
    Received:
      2
      
      at Object.<anonymous> (src/container/__tests__/RelayContainer-test.js:564:57)
      at Timeout.callback [as _onTimeout] (node_modules/jsdom/lib/jsdom/browser/Window.js:477:19)
      at ontimeout (timers.js:365:14)
      at tryOnTimeout (timers.js:237:5)
      at Timer.listOnTimeout (timers.js:207:5)

  ● RelayContainer › props.relay.commitUpdate › updates when the underlying RelayEnvironment changes

    expect(received).toBe(expected)
    
    Expected value to be (using ===):
      1
    Received:
      2
      
      at Object.<anonymous> (src/container/__tests__/RelayContainer-test.js:603:58)

```


leebyron:
Sorry for letting this sit - would you please rebase this PR so I can help merge it?

alloy:
While unfortunate that this never made it in and that the package still essentially ships with Classic, as per https://github.com/facebook/relay/issues/1494#issuecomment-304433108 I suppose this is no longer going to be considered much anymore, so maybe it’s time to close this?

/cc @jstejada

colllin:
Up to you what to do with this.  I updated just now, but now see what @alloy is saying, so, oops, there goes an hour.  It may not even be worth the time to review this again.

In case you do:
When I rebased on master, it looked like there had been a change where the contextual relay environment had been added to the RelayContainer's relayProp, so I made a quick attempt to try to patch that as well, but I doubt that there is a test directly covering the need for the environment on relayProp to be updated when the contextual environment changes.  The tests I added for `commitUpdate` and `applyUpdate` could probably be adapted to this case.

jstejada:
Yes, I believe that it is probably best to close this given that Modern is the path forward.

Issue #1507, opened 2017-01-25
Unable to attach files to GraphQLMutation


nicholas-l:
Hi,
I am unable to send files using GraphQLMutation using the below code. I am using GraphQLMutation both for future compatibility and I hit the #782 bug.

```
const files = {};
this.state.files.forEach((file, index) => (files[`file${index}`] = file));
const query = Relay.QL`mutation AddConsentMutation {
  PG_AddConsent(input: $input){
    clientMutationId
    ...
  }
}
const variables = {
  input: {...}
}
const callbacks = {
  onFailure: trans =>{...}
  onSuccess: () => {...},
};
const mutation = new Relay.GraphQLMutation(
        query,
        variables,
        files,
        Relay.Store,
        callbacks
      );
```

It causes a POST with content type application/json, not with a multipart. I have checked that mutation._files === files and it does but I cannot follow where it disappears.

PR #1514, opened 2017-01-30
RelayGraphQLMutation: Fix file attachment

apalm:
`this._files` was not being set in [`PendingGraphQLTransaction`'s constructor](https://github.com/facebook/relay/blob/5e4c38b23d1278c0fee02846995275f1e6ca5456/src/mutation/RelayGraphQLMutation.js#L256), which meant `files` was always `undefined` in [_sendMutation](https://github.com/facebook/relay/blob/5e4c38b23d1278c0fee02846995275f1e6ca5456/src/network-layer/default/RelayDefaultNetworkLayer.js#L94).

I believe this fixes https://github.com/facebook/relay/issues/1507.

### DIFF
--- a/src/mutation/RelayGraphQLMutation.js
+++ b/src/mutation/RelayGraphQLMutation.js
@@ -266,6 +266,7 @@ class PendingGraphQLTransaction {
     this._configs = [];
     this._query = query;
     this._variables = variables;
+    this._files = files;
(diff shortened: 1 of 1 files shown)


wincent:
Great catch. Thanks for this.

Ideally this would come with a regression test but I think this is worth merging straight away.

nicholas-l:
Thanks for fixing this!

PR #1519, opened 2017-01-31
RelayGraphQLMutation: Add test to ensure files get passed through

apalm:
This PR is a follow-up to https://github.com/facebook/relay/pull/1514. It adds a test to verify the fix made there.

Let me know if there's a more appropriate place to put the test, as I wasn't really sure.


wincent:
Awesome. Thanks for following up on this.

PR #1527, opened 2017-02-02
Process test files with babel-plugin-graphql

wincent:
Sneak preview of the kind of thing that is going to be necessary to unbreak the test suite.

Won't actually ship this yet (due to sync issues, I am going to have to move the plug-in) but just parking this here for now.


wincent:
Closing this before it gets too stale.

Issue #1501, opened 2017-01-13
Relay.GraphQLMutation doesn't output log


chollier:
I was using the new Relay.GraphQLMutation and came to notice that it doesn't output any log.
Hence errors such as an invalid config would get swallowed.

Example :
```js
const configs = [
    {
      type: "RANGE_ADD",
      parentName: "forumTopic",
      parentId: this.props.forumTopic.id, <--- it should be parentID
      connectionName: "responses",
      edgeName: "forumResponse",
      rangeBehaviors: (args) => {
        return "refetch";
      },
    },
]
```
fails silently with `Relay.GraphQLMutation` while with the old API we could see the error by expanding :
<img width="284" alt="screen shot 2017-01-13 at 10 42 28 am" src="https://cloud.githubusercontent.com/assets/1869/21941339/2c70fdbc-d97d-11e6-85bb-e440090bc0ea.png">

josephsavona:
hmm interesting. cc @yuzhi

yuzhi:
mhm It might just not be hooked up with Relay.GraphQLMutation. I can take a look when I have free cycles.

wincent:
With the release of Relay Modern, `GraphQLMutation` is no longer being worked on, so I'm going to close this issue. Please file any new issues that you might find in the new API, however. Thanks for reporting!

PR #1541, opened 2017-02-10
RelayGraphQLMutation: Add logging in dev

apalm:
This PR is an attempt to address https://github.com/facebook/relay/issues/1501.

The changes here are driven by my assumption that `RelayGraphQLMutation` should not rely on any RelayMutation* modules (let me know if this is not the case). Therefore, I basically inlined `RelayMutationDebugPrinter`'s `printMutation` and `printOptimisticMutation` into `getQuery` and `getOptimisticQuery`, respectively.

**Edit**: I just noticed `RelayGraphQLMutation` already uses `RelayMutationTransactionStatus`. So perhaps my assumption was false?

I did not add printing of the mutation configs, because there's kind of a lot going on in [buildQuery](https://github.com/facebook/relay/blob/master/src/mutation/RelayMutationQuery.js#L368). I can take a stab at adding this if it's critical, though.

The changes didn't cause any additional test failures when I ran the tests. I also ran the code locally and verified that:
1. Variables and query are printed when `commit` is called
2. Variables, query, and optimistic response are printed when `applyOptimistic` is called

If someone can give me pointers on the type of tests they'd like to see for this, I'm happy to add.

### DIFF
--- a/src/mutation/RelayGraphQLMutation.js
+++ b/src/mutation/RelayGraphQLMutation.js
@@ -323,6 +323,24 @@ class PendingGraphQLTransaction {
       );
       this._optimisticMutation =
         (mutation: any); // Cast RelayQuery.{Node -> Mutation}.
+      /* eslint-disable no-console */
+      if (__DEV__ && console.groupCollapsed && console.groupEnd) {
+        console.groupCollapsed(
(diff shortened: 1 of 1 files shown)


leebyron:
Sorry for not reviewing this sooner - we're focusing development on Relay Modern. If you would like to revisit this for the modern core, that would be great!

PR #1548, opened 2017-02-17
Fix deferred queries when root arguments are not of type ‘ID’

qgerome:
When building a deferred query, we explicitly set the type of the identifying argument to `RelayNodeInterface.ID_TYPE`. This causes the query to be invalid if the type is different.

```graphql
// Schema
type Query {
  story(slug: Int!): Story
}
```
Query to split:
```graphql
query Index($slug_0: Int!) {
  story(slug: $slug_0) {
    id
    ${Story.getFragment('story').defer()}
  }
}
```
Deferred query generated :
```graphql
query Index($slug_0:ID!) {   // wrong type
  story(slug:$slug_0) {
    id
   ...F0
  }
}
fragment F0 on Story {
  title
  version
  id
}
```


qgerome:
Of course, I'll do it this week.

leebyron:
Sorry this took forever to be reviewed! Thanks again for your help in making Relay better

PR #1552, opened 2017-02-24
Upgrade to Flow v0.40.0

gabelevi:
There were [20 errors](https://gist.github.com/73c2d13c221121fba682d3932f0a2fef) both before and after the upgrade in `src/`

PS why is `yarn.lock` in the `.gitignore`? Doesn't that defeat the point?


wincent:
> PS why is `yarn.lock` in the `.gitignore`? Doesn't that defeat the point?

It was ignored in 44b24f377a52ab415a0a with the rationale explained in a34867552d51c015f687:

> Unfortunately, we require internally at facebook, to have a yarn.lock without `http` resolved modules since we have an offline cache checked in. Since this cache is not availble in OSS, we the yarn.lock is currently useless for OSS.

kassens:
@facebook-github-bot shipit

PR #1562, opened 2017-03-04
Update to jest@19.0.2

kassens:
Verified that the same number of tests fail as did before.


kassens:
@facebook-github-bot import

Issue #1531, opened 2017-02-06
How to use Relay.Environment with Relay.GraphQLMutation?


nikhilag:
I am using Relay.GraphQLMutation so that it is easier for me to migrate to Relay 2 later. However I don't understand why I have to create Relay.Environment instance for using this? Can't we subclass Relay.GraphQLMutation just like we were subclassing Relay.Mutation? 

Also I am having to use `environment.injectNetworkLayer(...)` which I didn't have to use earlier. I am using a create-react-app setup based on https://github.com/Valentin-Seehausen/create-react-app. My GraphQL server is running on localhost:4000/graphql while my frontend is running on localhost:3000. When I have to create environment myself for using GraphQLMutation, the proxy setting in package.json  is not working and I don't want to manually set the network layer since it has to be different for dev and prod environments.

Here's my code for creating the mutation:-

```
        const environment = new Relay.Environment();
        const query = Relay.QL`mutation saveForm {
            saveFormMutation(input: $input) {
                ...            
            }
        }`;
        const variables = {
            input: {
               ...
            }
        };
        environment.injectNetworkLayer(
            new Relay.DefaultNetworkLayer('http://localhost:4000/graphql')
        );
        const mutation = Relay.GraphQLMutation.create(
            query,
            variables,
            environment
        );
```

wincent:
Thanks for the question. For future reference posting to Stack Overflow is probably your best bet in order to get a timely response (you may even find a pre-existing question there with an answer); we try to focus the issue tracker on bugs and feature requests.

You generally don't want to create a new environment for this but rather use an existing one. For example, if you are invoking this mutation from a Relay component, you could get the current environment from `this.context.relay.environment` and pass that in. Note that in order to be able to access this context you'll need something like this in your component:

```
const RelayPropTypes = require('react-relay/lib/RelayPropTypes');

class Component extends React.class {
  static contextTypes = {
    relay: RelayPropTypes.Relay,
  };
}
```

Even simpler, I think in most apps you can just relay on the default singleton instance of the environment, which is exposed as `Relay.Store`:

```
const mutation = Relay.GraphQLMutation.create(
  query,
  variables,
  Relay.Store
);
```

I'm going to close this now but please feel free to comment again if I've missed something.

chris-guidry:
Is there a working example that utilizes Relay.GraphQLMutation? The repo [Relay Examples](https://github.com/relayjs/relay-examples) is still using Relay.Mutation.

[...]

PR #1565, opened 2017-03-06
Clarify use of Relay.Store in GraphQLMutation

sambou:
I think it would be helpful to point out the most common use of this API.

Refers to the question in this issue: https://github.com/facebook/relay/issues/1531


wincent:
Looks fair enough. Thanks @sambou!

wincent:
@facebook-github-bot shipit

PR #1566, opened 2017-03-07
Update QuickStart-Tutorial.md

Gregoirevda:
English typo

### DIFF
--- a/docs/QuickStart-Tutorial.md
+++ b/docs/QuickStart-Tutorial.md
@@ -99,7 +99,7 @@ import {
 
 At this point, you can delete everything up until `queryType` in `./data/schema.js`.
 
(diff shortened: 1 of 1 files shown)


wincent:
Thanks for this @Gregoirevda. Note that "need only" is not a typo and is actually perfectly valid English:

> [I]n a rather formal style of English, there also exists an auxiliary verb ("helping verb") *need*, which is followed by a bare infinitive, and is only used in negative polarity contexts: with *not*, with *only*, in questions, and so on. Additionally, it only exists in the present tense. Aside from these two restrictions, it's very much like other auxiliary verbs (*can*, *may*, *should*, etc.): "We need not discuss it."

(via [quick Google search](http://english.stackexchange.com/a/297235)).

Having said that, I can see how this usage might confuse some readers, and we want the documentation to be as accessible as possible, so let's get this merged. Thanks once again.

wincent:
@facebook-github-bot shipit

PR #1579, opened 2017-03-30
Update Relay plugins in jest preprocessor

kassens:
This adds the new BabelPluginRelay and also fixes loading the schema which was missing some information when loaded from the json file.


kassens:
@facebook-github-bot import

kassens:
Will make this change from the internal copy.

PR #1581, opened 2017-03-31
Fix TypeError

apalm:
Currently, [this test](https://github.com/facebook/relay/blob/master/packages/react-relay/classic/traversal/__tests__/printRelayOSSQuery-test.js#L536) fails with a `TypeError: Converting circular structure to JSON`.

This PR fixes (or rather, works around) this by changing the message back to the way it was before a89f623ba69b7b38b06af274d67ac260a9e64bf7.

Also, the tests at https://github.com/relay-tools/react-router-relay and https://github.com/relay-tools/relay-local-schema no longer failed after this change when I ran locally (swapping out the `react-relay` imports for `react-relay/classic`).

### DIFF
--- a/packages/react-relay/classic/traversal/printRelayOSSQuery.js
+++ b/packages/react-relay/classic/traversal/printRelayOSSQuery.js
@@ -76,8 +76,7 @@ function printRelayOSSQuery(node: RelayQuery.Node): PrintedQuery {
   }
   invariant(
     queryText,
-    'printRelayOSSQuery(): Unsupported node type, got `%s`.',
-    JSON.stringify(node),
+    'printRelayOSSQuery(): Unsupported node type.'
   );
   const variables = {};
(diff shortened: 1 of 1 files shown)


YasserKaddour:
Thank you, I wanted to give relay2 a try and was blocked by that error after a switched `react-relay/classic`.

By the way congratulation to the team for this alpha release  :tada:, I am eager to make it work, thanks :v:

kassens:
👍 
There's a few more instances of this, I have a change coming to fix the rest. (Due to the restructure, I cannot land the fix from GitHub at the moment unfortunately)

apalm:
@kassens I see. Feel free to close this PR if it's not needed :)

kassens:
The fixes are now landed at fcfe1b8496702ff29116445ee2c7b8453377159e. Thanks for taking an early look at this release!

PR #1604, opened 2017-04-12
Update license headers to BSD style

kassens:
In the process of open sourcing, we forgot to update the license header in some
places. This fixes the headers.


kassens:
cc @taion :)

taion:
Oops, sorry – I was actually making fun of https://github.com/kassens/relay/blob/91edb3693805d3d8f53ccc9e726692b00deb814e/packages/relay-compiler/codegen/writeFlowFile.js#L29-L42

I'm not sure what the right thing is, but it'd feel a bit odd to have generated code in my source tree that was copyright Facebook.

kassens:
Oh, we'll need to make that injectable still.

PR #1609, opened 2017-04-14
No longer run tests on the old babel plugin, simplify run.

leebyron:


leebyron:
Abandoning in favor of #1650

Issue #1590, opened 2017-04-06
Adopt flat bundles?


robrichard:
Are there plans to use rollup and generate flat bundles for Relay (similar to what react did in https://github.com/facebook/react/pull/9327)? It would be nice to get smaller bundles and it looks like this would also address the performance regression in server side rendering (https://github.com/facebook/relay/issues/1321).

kassens:
We are a few days away from Making Travis Green Again and since the bundling/building is not used internally, this could be a great external contribution. I'll give @leebyron a chance to chime in who know the node landscape a bit better.

gaearon:
FWIW we still `require()` fbjs modules in the React flat bundle. Is there any downside to this? I’m not sure I understand the discussion in https://github.com/facebook/relay/issues/1321.

twhid:
Re: #1321 

Having the `require`s inline, particularly if you're using `require` hooks (eg babel/register), slows down a node server considerably. We saw a ~75% decrease in server rendering time when we switched to a custom relay build that hoists the `require` calls to the top of the scope.

gaearon:
Oh I see. We do hoist them in our React flat bundles.
They look like:

```js
var invariant = require('fbjs/lib/invariant');
var warning = require('fbjs/lib/warning');

// all flattened modules below are using those
```

leebyron:
This would be great! Right now we're using webpack to create these, but I agree that using Rollup would be better. I think this is a great task for someone looking to contribute to work on

robrichard:
I opened a PR for this here: https://github.com/facebook/relay/pull/1610

wincent:
> Right now we're using webpack to create these, but I agree that using Rollup would be better.

Haven't used it myself yet, but webpack (as of v2) also has [some support for tree-shaking features pioneered in Rollup](https://webpack.js.org/guides/tree-shaking/).

YasserKaddour:
@wincent unfortunately tree shaking is broken in webpack, see https://github.com/webpack/webpack/issues/2867

alloy:
Looks like webpack 4 is promising to fix this https://github.com/webpack/webpack/issues/2867#issuecomment-353714614. @robrichard is tree shaking the main motivation for this issue and linked PR?

gaearon:
I don't think so. This is beneficial for many reasons. In our case the main benefits were caching `process.env.NODE_ENV` access in SSR, being able to inline more helpers using GCC simple mode, protecting internals from being imported. I wrote a post about this:

[...]

PR #1610, opened 2017-04-14
Create flat bundles using rollup

robrichard:
Updated the gulpfile to use Rollup to generate commonjs and umd bundles as discussed here: https://github.com/facebook/relay/issues/1590. I followed the pattern used by the master branch of React for requiring the dev or prod enabled versions.

I'm being very explicit with regard to what modules are external vs included in the bundles. Let me know there is a different way you would like to handle it. 

Also what is the best way I can test that these bundles are built correctly and working as expected?

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -63,17 +63,18 @@ const babelOptions = require('./scripts/getBabelOptions')({
   ],
 });
 const del = require('del');
-const derequire = require('gulp-derequire');
 const es = require('event-stream');
 const flatten = require('gulp-flatten');
 const fs = require('fs');
 const gulp = require('gulp');
-const chmod = require('gulp-chmod');
 const gulpUtil = require('gulp-util');
-const header = require('gulp-header');
 const path = require('path');
 const runSequence = require('run-sequence');
-const webpackStream = require('webpack-stream');
+const rollup = require('rollup');
+const commonjs = require('rollup-plugin-commonjs');
+const enhancedResolve = require('enhanced-resolve');
+const uglify = require('rollup-plugin-uglify');
+const replace = require('rollup-plugin-replace');
 
 const SCRIPT_HASHBANG = '#!/usr/bin/env node\n';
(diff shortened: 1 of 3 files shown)


robrichard:
Hey @leebyron @kassens - are you still considering using Rollup to build the new relay packages as discussed in https://github.com/facebook/relay/issues/1590? If so it would be good to get it in before the final 1.0.0 release, otherwise it will be a breaking change. It would removes access to internal modules, e.g. `require('react-relay/lib/toGraphQL')`

kassens:
Interesting, hadn't thought of that. This would probably be a good thing though as we should expose all necessary things publicly.

alloy:
@kassens @jstejada Reading the history of this PR, it does seem like there was interest on FB’s side to merge such an external contribution. Does that still apply?

robrichard:
I did try to rebase that PR a few months ago and ran into some issues with the dynamic requires of the relay dev tools.

Also it would be much more beneficial if the relay source was updated to use ES modules as was done in the React source code. By using `rollup-plugin-commonjs` here, each module will still be wrapped in a closure.

alloy:
@robrichard Is that something that can be done iteratively?

robrichard:
@alloy hard for me to say as I don't know how it would mesh with FB's custom module system.

alloy:
I meant more, can this PR continue as-is and discuss ES6 modules as a next step or is blocked by that by now?

Also, out of curiosity, does React not use the Haste module system?

robrichard:
Yes this change would not be blocked by that.

React was converted to ES modules in these two PRs:
https://github.com/facebook/react/pull/11303
https://github.com/facebook/react/pull/11389

robrichard:
Going to close this as there are a few issues with this approach:
1. The rollup config is complicated by the way files are imported without relative file paths.
2. The source files use CommonJS, meaning that rollup still requires wrapping them in a closure function, meaning the final bundle is not truly flat.

PR #1611, opened 2017-04-14
Fix LeafType mistyping

leebyron:
RelayIR assumes the type of a ScalarField is `GraphQLLeafType`, however that does not account for the possibility of nullable or list variations. When `isLeafType` was updated in graphql@0.9.2 to match the definition of LeafType this mistyping started causing errors.

This proposes typing ScalarField's type as `GraphQLOutputType` which is a little broader than necessary, but importantly includes the possibilities of nullable and list variations. It also explicitly calls getNamedType before passing to isListType to ensure the check is on the unwrapped named type.

Also ups the dependency on graphql to the latest release to illustrate Travis test passing


packages/relay-compiler/core/RelayIR.js
```diff
@@ -158,7 +157,7 @@ export type ScalarField = {
   kind: 'ScalarField',
   metadata: ?{[key: string]: mixed},
   name: string,
-  type: GraphQLLeafType,
+  type: GraphQLOutputType,
```
josephsavona: This seems wrong: one of the major reasons we added LinkedField/ScalarField in the first place was to distinguish between values whose types were leaf-like and those that have subfields. Is there any reason not to be more explicit here?

```
type ScalarType = GraphQLLeafType | GraphQLNonNull<ScalarType> | GraphQLList<ScalarType>;

type ScalarField = {
   ...
  type: ScalarType,
}
```

josephsavona: My concern is that this change makes it too easy to construct invalid IR. With the previous type, Flow would help catch errors in which you accidentally assigned an object type to a linked field.

leebyron: I can add this explicit type, but it doesn't seem to be adding much direct value since there isn't any code leveraging this difference since the previous type wasn't correct and flow wasn't reporting any issues

leebyron:
Updated, cc @josephsavona

PR #1618, opened 2017-04-18
Better regexp to use for pre-compiled module names.

leebyron:
This regexp runs before transforming module names it seems, so avoid translating normal haste modules.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -135,7 +135,7 @@ const PRODUCTION_HEADER = [
 const buildDist = function(filename, opts, isProduction) {
   const webpackOpts = {
     debug: !isProduction,
-    externals: [/^[-/a-zA-Z0-9]+$/],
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -135,7 +135,7 @@ const PRODUCTION_HEADER = [
 const buildDist = function(filename, opts, isProduction) {
   const webpackOpts = {
     debug: !isProduction,
-    externals: [/^[-/a-zA-Z0-9]+$/],
+    externals: [/^[-a-z0-9]+(\/.+)?$/],
```
kassens: What is this intending to match in words? It is pretty close to just `/^[-a-z0-9]/`. If that's the intention, it'd be good to simplify to that. If not, it might be wrong.

leebyron:
I'm abandoning this actually - I noticed the tests failed and mistook this previous diff as a cause but it was another diff

PR #1619, opened 2017-04-18
Fix travis test run

leebyron:
Assuming this test was trying to test modern runtime mutations, not compat mutations.


kassens:
lgtm

PR #1622, opened 2017-04-18
Visual refersh of Relay site + more docs

leebyron:
This adds a bunch more docs, especially around the compiler and compat mode, and does a big pass of stylisitic cleanup.


nodkz:
shipit

PR #1623, opened 2017-04-18
Update PaginationContainer.md

nodkz:
Fix code typos

### DIFF
--- a/docs/modern/PaginationContainer.md
+++ b/docs/modern/PaginationContainer.md
@@ -85,7 +85,7 @@ class Feed extends React.Component {
           edge => <Story story={edge.node} key={edge.node.id} />
         )}
         <button
-          onPress={this._loadMore}
+          onPress={() => this._loadMore()}
           title="Load More"
         />
       </div>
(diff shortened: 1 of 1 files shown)


docs/modern/PaginationContainer.md
```diff
@@ -85,7 +85,7 @@ class Feed extends React.Component {
           edge => <Story story={edge.node} key={edge.node.id} />
         )}
         <button
-          onPress={this._loadMore}
+          onPress={() => this._loadMore()}
```
nodkz: Or maybe `.bind(this)` depends on your code style.

sibelius: this._loadMore works fine

nodkz: Should not. 
Cause `this._loadMore` does not autobind `Feed` instance context for getting its props.

kassens: You could use property initializer syntax, but that's an early feature. This fix seems good here.

PR #1636, opened 2017-04-19
Allow passing no callback to PaginationContainer loadMore

janicduplessis:
I think it makes sense to make this argument optional, it is already the case for RefetchContainer. Also some examples didn't pass a callback and would crash.


kassens:
Makes sense, thanks for making the change!

PR #1637, opened 2017-04-19
Allow non null types in connection validation

janicduplessis:
The current connection validation is too strict and doesn't allow non-null types for a lot of things. It also isn't compatible with the types generated by `graphql-relay-js`.

This uses `RelaySchemaUtils.getNullableType` before instanceof checks to make sure we get the actual type and not `GraphQLNonNull`. Also split some invariants into 2 different checks to keep it readable.

Tested in an app that uses a PaginationController with a schema generated with `graphql-relay-js`.

### DIFF
--- a/packages/relay-compiler/handlers/connection/RelayConnectionTransform.js
+++ b/packages/relay-compiler/handlers/connection/RelayConnectionTransform.js
@@ -378,12 +378,22 @@ function validateConnectionType(
     type,
     definitionName,
   );
+
   const node = edgeType.getFields()[NODE];
+  invariant(
+    node,
+    'RelayConnectionTransform: Expected type `%s` to have an %s.%s field in ' +
+    'document `%s`.',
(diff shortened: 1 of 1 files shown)


alangenfeld:
thanks!

gauravtiwari:
Hey, upgraded to rc2 and still getting this error due to strict validation

```graphql

# An edge in a connection.
type SomeEdge {
  # Not null still doesn't work
  cursor: String!
}
```
`Expected type `SomeConnection` to have an edges.cursor field for which the type is an scalar`

```graphql
type PageInfo {
  # Not-null isn't supported
  hasNextPage: Boolean!
  hasPreviousPage: Boolean!
}
```

`Expected type `PageInfo` to have an hasNextPage field for which the type is an scalar in document`

gauravtiwari:
Great thanks for fixing 👍

PR #1640, opened 2017-04-19
Fix typos in docs and better wording

tazsingh:
As I'm going through the docs and upgrading our app(s), I'll try to clarify things as much as possible. Thanks for the release!


leebyron:
Thanks!

PR #1642, opened 2017-04-19
Validate underlying types in Relay Compiler for connections

tazsingh:
When creating a Pagination Container for a connection, I noticed that `relay-compiler` was failing on a few types.

The underlying issue was that it was comparing against the `NonNull` type instead of the underlying type. This PR fixes that.


ekosz:
This may be a dup of https://github.com/facebook/relay/pull/1637 ?

tazsingh:
Ha ya it appears to be the same fix in #1637. Feel free to close this if it makes sense.

PR #1658, opened 2017-04-20
add extensions option to relay-compiler

robrichard:
Allows you to pass an extensions argument to relay compiler. A lot of people use alternative extensions like `.jsx` for React components. I followed the convention of `jscodeshift` of making the extensions comma separated.

```
relay-compiler --src ./src/ --schema schema.graphql --extensions=js,jsx
```

### DIFF
--- a/packages/relay-compiler/bin/RelayCompilerBin.js
+++ b/packages/relay-compiler/bin/RelayCompilerBin.js
@@ -33,21 +33,27 @@ const {
 import type {GraphQLSchema} from 'graphql';
 
 const SCRIPT_NAME = 'relay-compiler';
-const WATCH_EXPRESSION = [
-  'allof',
-  ['type', 'f'],
-  ['suffix', 'js'],
-  ['not', ['match', '**/node_modules/**', 'wholename']],
-  ['not', ['match', '**/__mocks__/**', 'wholename']],
-  ['not', ['match', '**/__tests__/**', 'wholename']],
-  ['not', ['match', '**/__generated__/**', 'wholename']],
-];
+
+function buildWatchExpression(options: {
+  extensions: Array<string>
+}) {
+  return [
+    'allof',
+    ['type', 'f'],
+    ['anyof', ...options.extensions.map(ext => ['suffix', ext])],
+    ['not', ['match', '**/node_modules/**', 'wholename']],
+    ['not', ['match', '**/__mocks__/**', 'wholename']],
+    ['not', ['match', '**/__tests__/**', 'wholename']],
+    ['not', ['match', '**/__generated__/**', 'wholename']],
+  ];
+}
 
 /* eslint-disable no-console-disallow */
 
 async function run(options: {
   schema: string,
   src: string,
+  extensions: Array<string>,
   watch?: ?boolean,
 }) {
   const schemaPath = path.resolve(process.cwd(), options.schema);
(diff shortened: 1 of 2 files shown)


robrichard:
Looks like the same test is failing on master

packages/relay-compiler/bin/RelayCompilerBin.js
```diff
@@ -66,6 +64,15 @@ const argv = yargs
   .help()
   .argv;
 
+const WATCH_EXPRESSION = [
```
leebyron: Mind moving this back to the top to make the diff more clear and avoid any possible rebase issues?

packages/relay-compiler/bin/RelayCompilerBin.js
```diff
@@ -58,6 +51,11 @@ const argv = yargs
       demandOption: true,
       type: 'string',
     },
+    'extensions': {
+        default: 'js',
+        describe: 'File extensions to compile',
+        type: 'string',
```
leebyron: Could you add yargs's `array` argument instead of accepting csv?

packages/relay-compiler/bin/RelayCompilerBin.js
```diff
@@ -58,6 +51,11 @@ const argv = yargs
       demandOption: true,
       type: 'string',
     },
+    'extensions': {
+        default: 'js',
+        describe: 'File extensions to compile',
```
leebyron: Perhaps provide an example of how to pass the args, like:

```
describe: 'File extensions to compile (--extensions js jsx)'
```

packages/relay-compiler/bin/RelayCompilerBin.js
robrichard: I moved it below to get access to the `extensions` value from `argv`.

robrichard:
@leebyron thanks for the feedback! Comments have been addressed.

packages/relay-compiler/bin/RelayCompilerBin.js
mjmahone: This should probably become a function, then: takes in an `options` argument with `{extensions: Array<string>}` as the only option for now. Outputs this watchman expression.

```
function buildWatchExpression(options: {extensions: Array<string>} {
  return [
    ...
    ['anyof', extensions.map(ext => ['suffix', ext])],
    ...
  ]
}
```

robrichard: Rebased and updated now to a buildWatchExpression function

robrichard:
@leebyron @mjmahone I rebased on top of master and comments have been addressed. Let me know if there are any other changes you would like me to make.

leebyron:
Nice work!

robrichard:
@leebyron pushed again to fix the merge conflict from https://github.com/facebook/relay/commit/9476ee3821528b8efeb9fb55f29a7c47cfd6cd6c.

jjoos:
I was playing around with this option and couldn't get it working with the example above, I needed to do pass `extensions` twice to relay compiler. 
```
node_modules/.bin/relay-compiler --src ./app/javascript/ --schema tmp/schema.graphql --extensions=jsx --extensions=js --watchman false
```
since this doesn't work:
```
node_modules/.bin/relay-compiler --src ./app/javascript/ --schema tmp/schema.graphql --extensions=jsx,js --watchman false
```

PR #1679, opened 2017-04-21
PaginationContainer docs are missing data

staylor:
ReactRelayPaginationContainer will throw an invariant:
`Unable to synthesize a getConnectionFromProps function.` 

... if you do not specify `getConnectionFromProps()` and `direction`


kassens:
Thanks

PR #1681, opened 2017-04-22
Add content-type header to network layer example

robrichard:
graphql-js will treat requests with no content-type as form-encoded. Since the example is sending JSON, add a content-type header to the request.


leebyron:
Good idea!

PR #1683, opened 2017-04-23
Add support for Json schema to relay-compiler

### LINKED ISSUE #1639: [Modern] Error with .json schema
##### Docs: 

> The `relay-compiler` script requires both the directory which holds your source files as well as a path to your GraphQL schema in **either** a .json or .graphql schema file.

##### Log:

```
> relay-compiler --src ./js --schema ./schema.json

Parsed default in 0.04s

Writing default
Error: Error loading schema. Expected the schema to be a .graphql file using the
GraphQL schema definition language.
```

##### JSON file:

```js
{
  "queryType": {
    "name": "Query"
  },
  "mutationType": {
    "name": "Mutation"
  },
  "subscriptionType": null,
  "types": [
    {
      "kind": "OBJECT",
      "name": "Query",
      "description": null,
      "fields": [

  // etc
```

YasserKaddour:
Fix #1639

### DIFF
--- a/packages/relay-compiler/bin/RelayCompilerBin.js
+++ b/packages/relay-compiler/bin/RelayCompilerBin.js
@@ -21,7 +21,13 @@ const fs = require('fs');
 const path = require('path');
 const yargs = require('yargs');
 
-const {buildASTSchema, parse} = require('graphql');
+const {
+  buildASTSchema,
+  buildClientSchema,
+  parse,
+  printSchema,
+} = require('graphql');
+
 const {
   codegenTransforms,
   fragmentTransforms,
@@ -122,6 +128,9 @@ function getRelayFileWriter(baseDir: string) {
 function getSchema(schemaPath: string): GraphQLSchema {
   try {
     let source = fs.readFileSync(schemaPath, 'utf8');
+    if (path.extname(schemaPath) === '.json') {
+      source = printSchema(buildClientSchema(JSON.parse(source).data));
+    }
     source = `
   directive @include(if: Boolean) on FRAGMENT | FIELD
   directive @skip(if: Boolean) on FRAGMENT | FIELD
@@ -132,8 +141,8 @@ function getSchema(schemaPath: string): GraphQLSchema {
     return buildASTSchema(parse(source));
   } catch (error) {
     throw new Error(`
-Error loading schema. Expected the schema to be a .graphql file using the
-GraphQL schema definition language. Error detail:
+Error loading schema. Expected the schema to be a .graphql or a .json
+file, describing your GraphQL server's API. Error detail:
 
 ${error.stack}
     `.trim());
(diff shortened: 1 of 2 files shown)


YasserKaddour:
Would you be open to a PR to retrieve the schema from a specified GraphQL endpoint or from the `GRAPHQL_ENDPOINT` environment variable. 

I would also like to tackle [`graphql-config`](https://github.com/graphcool/graphql-config) support, but only once the [GraphQL Configuration Protocol](https://github.com/graphcool/graphql-config/issues/20) is implemented, to avoid breaking changes.

cc @leebyron @kassens @schickling @asiandrummer

schickling:
@asiandrummer I think now is a great time to move forward with this!

packages/relay-compiler/bin/RelayCompilerBin.js
```diff
@@ -122,6 +128,9 @@ function getRelayFileWriter(baseDir: string) {
 function getSchema(schemaPath: string): GraphQLSchema {
   try {
     let source = fs.readFileSync(schemaPath, 'utf8');
+    if (path.extname(schemaPath) === '.json') {
+      source = printSchema(buildClientSchema(JSON.parse(source).data));
+    }
```
josephsavona: It seems inefficient to build a client schema, print it, and then parse it back to schema again just a few lines later. Lets refactor to avoid this extra conversion, as it could slow down the compiler for apps with larger schemes.

YasserKaddour: @josephsavona could you give me some pointers on how to incorpore those directives in the GraphQLSchema. I am not familiar with the inner working of graphql-js, Thanks.

leebyron: The correct fix to this is a much bigger refactoring. We shouldn't be doing this kind of directive patching in this file in the first place.

leebyron: I think this current solution is reasonable until we do that

YasserKaddour: Do you want me to leave a todo?
@leebyron What about my proposal to add support for GraphQL endpoint? Thanks

leebyron: I don't think the compiler should directly rely on having network access, especially since many private company graphql schema endpoints require authentication which we shouldn't be building here. 

However, I think it would be a great addition to have a simple script which allows you to provide a GraphQL endpoint and produce a schema.graphql file which you could later provide to the relay compiler

sibelius: https://github.com/graphcool/get-graphql-schema does it

it can generate a .json or a .graphql schema from a graphql endpoint

YasserKaddour: Ok. I knew about [`graphql-config`](https://github.com/graphcool/graphql-config) but not about get-graphql-schema. Thanks @sibelius

Issue #1695, opened 2017-04-26
[Modern] Updated variables in RefetchContainer not passed to child fragments


robrichard:
I have created a container using `createRefetchContainer` which updates variables used by child fragments. In Relay Modern the child components corresponding to those fragments are not re-rendered with the updated variables. I can see that query was executed with the updated variables, the store has been updated to hold these records but the components are still passed the old props. This works as expected with the compat environment.

Reproduction example here: https://github.com/relayjs/relay-examples/compare/master...robrichard:variables-example?expand=1

Running on http://localhost:3000/?useModern (uses modern environment)
Initial view
![image](https://cloud.githubusercontent.com/assets/3814096/25452042/29475454-2a92-11e7-9d1e-09d0b3972fc6.png)

click on "use b"
![image](https://cloud.githubusercontent.com/assets/3814096/25452051/3148c07a-2a92-11e7-9fde-30b60d74493f.png)

Running on http://localhost:3000/ (uses compat environment)
![image](https://cloud.githubusercontent.com/assets/3814096/25452083/5c3491c4-2a92-11e7-97fd-082257423453.png)

robrichard:
I did some more digging and it looks like `RelayModernFragmentSpecResolver` will resolve props for `FragmentContainers` based on the variables in context. `QueryRenderer` will add the variables for it's query to React context but `ReactRelayRefetchContainer` will not. Any fragment containers rendered as children of a refetch container will never be rendered with the data from the refetch query. It seems like the simplest solution is to implement `getChildContext` in `ReactRelayRefetchContainer`.

yuzhi:
Thanks @robrichard  for fixing this!

PR #1702, opened 2017-04-27
Update childContext to include new variables in RefetchContainer

robrichard:
See https://github.com/facebook/relay/issues/1695. 

```js
<QueryRenderer ... />
  <MyRefetchComponent ... />
    <MyFragmentComponent ... />
  </MyRefetchComponent>
<QueryRenderer>
```

In a structure like this, `MyFragmentComponent` will never be rendered with data fetched by `MyRefetchComponent`

### DIFF
--- a/packages/react-relay/modern/ReactRelayRefetchContainer.js
+++ b/packages/react-relay/modern/ReactRelayRefetchContainer.js
@@ -67,6 +67,7 @@ function createContainerWithFragments<TBase: ReactClass<*>>(
     _localVariables: ?Variables;
     _pendingRefetch: ?Disposable;
     _references: Array<Disposable>;
+    _relayContext: RelayContext;
     _resolver: FragmentSpecResolver;
 
     constructor(props, context) {
--- a/packages/react-relay/modern/ReactRelayRefetchContainer.js
+++ b/packages/react-relay/modern/ReactRelayRefetchContainer.js
@@ -271,6 +271,15 @@ function createContainerWithFragments<TBase: ReactClass<*>>(
       };
     };
 
+    getChildContext(): Object {
+      return {
+        relay: {
+          environment: this.context.relay.environment,
+          variables: this._localVariables || this.context.relay.variables,
+        },
+      };
+    }
+
     render() {
       if (ComponentClass) {
         return (
@@ -320,12 +329,14 @@ function createContainer<TBase: ReactClass<*>>(
   fragmentSpec: GraphQLTaggedNode | GeneratedNodeMap,
   taggedNode: GraphQLTaggedNode,
 ): TBase {
-  return buildReactRelayContainer(
+  const Container = buildReactRelayContainer(
     Component,
     fragmentSpec,
     (ComponentClass, fragments) =>
       createContainerWithFragments(ComponentClass, fragments, taggedNode),
   );
+  Container.childContextTypes = containerContextTypes;
+  return Container;
 }
 
 module.exports = {createContainer, createContainerWithFragments};
--- a/packages/react-relay/modern/ReactRelayRefetchContainer.js
+++ b/packages/react-relay/modern/ReactRelayRefetchContainer.js
@@ -121,6 +121,10 @@ function createContainerWithFragments<TBase: ReactClass<*>>(
       ) {
         this._release();
         this._localVariables = null;
+        this._relayContext = {
+          environment: relay.environment,
+          variables: relay.variables
+        };
         this._resolver = createFragmentSpecResolver(
           relay,
           containerName,
--- a/packages/react-relay/modern/ReactRelayRefetchContainer.js
+++ b/packages/react-relay/modern/ReactRelayRefetchContainer.js
@@ -83,6 +83,10 @@ function createContainerWithFragments<TBase: ReactClass<*>>(
         props,
         this._handleFragmentDataUpdate,
       );
+      this._relayContext = {
+        environment: this.context.relay.environment,
+        variables: this.context.relay.variables,
+      };
       this.state = {
         data: this._resolver.resolve(),
         relayProp: this._buildRelayProp(relay),
@@ -225,6 +229,10 @@ function createContainerWithFragments<TBase: ReactClass<*>>(
         }
         // TODO t15106389: add helper utility for fetching more data
         this._pendingRefetch = null;
+        this._relayContext = {
+          environment: this.context.relay.environment,
+          variables: fragmentVariables
+        };
         callback && callback();
         this._resolver.setVariables(fragmentVariables);
         this.setState({data: this._resolver.resolve()});
@@ -272,12 +280,7 @@ function createContainerWithFragments<TBase: ReactClass<*>>(
     };
 
     getChildContext(): Object {
-      return {
-        relay: {
-          environment: this.context.relay.environment,
-          variables: this._localVariables || this.context.relay.variables,
-        },
-      };
+      return {relay: this._relayContext};
     }
 
     render() {
--- a/packages/react-relay/compat/react/RelayCompatRefetchContainer.js
+++ b/packages/react-relay/compat/react/RelayCompatRefetchContainer.js
@@ -14,6 +14,7 @@
 'use strict';
 
 const ReactRelayRefetchContainer = require('ReactRelayRefetchContainer');
+const RelayPropTypes = require('RelayPropTypes');
 
 const {buildCompatContainer} = require('ReactRelayCompatContainerBuilder');
 
@@ -32,7 +33,7 @@ function createContainer<TBase: ReactClass<*>>(
   fragmentSpec: GraphQLTaggedNode | GeneratedNodeMap,
   taggedNode: GraphQLTaggedNode,
 ): TBase {
-  return buildCompatContainer(
+  const Container = buildCompatContainer(
     Component,
     (fragmentSpec: any),
     (ComponentClass, fragments) => {
@@ -43,6 +44,10 @@ function createContainer<TBase: ReactClass<*>>(
       );
     },
   );
+  Container.childContextTypes = {
+    relay: RelayPropTypes.Relay,
+  };
+  return Container;
 }
 
 module.exports = {createContainer};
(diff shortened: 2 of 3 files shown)


leebyron:
Assigned some folks who can give a more careful review - but looks generally right to me

chollier:
This is a pretty big blocker to converting our app, we rely heavily on component composition and can't refactor it to the compat API with QueryRenderer for now.
Thanks @robrichard for taking a stab at it.

packages/react-relay/modern/ReactRelayRefetchContainer.js
```diff
@@ -318,6 +327,8 @@ function createContainer<TBase: ReactClass<*>>(
       taggedNode,
     ),
   );
+  Container.childContextTypes = containerContextTypes;
+  return Container;
```
chollier: The same needs to be done in `RelayCompatRefetchContainer`

packages/react-relay/modern/ReactRelayRefetchContainer.js
```diff
@@ -260,6 +260,15 @@ function createContainerWithFragments<TBase: ReactClass<*>>(
       };
     };
 
+    getChildContext(): Object {
+      return {
+        relay: {
+          environment: this.context.relay.environment,
+          variables: this._localVariables || this.context.relay.variables,
+        },
```
josephsavona: We have an optimization in all of the containers that they ignore context changes whenever the `context.relay` is `===` to its previous value and forcibly update whenever it is `!==` to the previous value. To correspond with this `QueryRenderer` returns a new `relay` value whenever the environment or variables actually changes.

Let's apply similar logic here:

```
getChildContext() {
  return {relay: this._relayContext};
}
```

And then set `this._relayContext` as follows:
- initialize in the constructor
- overwrite with a new variable whenever a `refetch` completes succesfully.

josephsavona: Example from FragmentContainer: https://github.com/facebook/relay/blob/master/packages/react-relay/modern/ReactRelayFragmentContainer.js#L85

robrichard:
@josephsavona this has been updated with your requested changes.

@chollier I updated the childContextTypes in `RelayCompatRefetchContainer` and confirmed it is working as expected.

chollier:
Sorry, just noticed it would also be needed in RelayCompatContainer too (and the pagination one I guess, I don't use it though)

robrichard:
@chollier I don't see why this would be needed in `ReactRelayFragmentContainer`. It's built using `ReactRelayFragmentContainer` which does not have the ability to change variables.

josephsavona:
Agree, this isn't necessary in RelayCompatContainer or ReactRelayFragmentContainer

chollier:
yeah my bad

robrichard:
@josephsavona @leebyron @kassens is it possible to get this into the next rc?

chollier:
That would be nice

josephsavona:
This looks good. The team is pretty busy on some things, but we'll try to pull this in for the next RC.

chollier:
Any chance a release would be cut this week? This PR is the only remaining thing holding us from pushing the conversion PR forward and use Relay Modern in our app

yuzhi:
@robrichard I was about to work on something like this and noticed this PR. Thank you for contributing. Everything looks good, but can you add unit tests to ensure that the context for the children is updated properly after success refetch calls? I'll be happy to merge this in after that.

robrichard:
@yuzhi Thanks for reviewing. Unfortunately, the tests for this file are disabled. Can you please take a look at my other PR to re-enable them (https://github.com/facebook/relay/pull/1777). Once that is merged, I can rebase and add the tests.

robrichard:
@yuzhi rebased and added tests

yuzhi:
@robrichard  I think we might need to update` _relayContext` in `componentWillReceiveProps` when the variables and or the environment has changed to match what the component is reading out. Perhaps add a unit test that it will update props when the parent passes in a different set of context from before.

robrichard:
@yuzhi good catch. I updated the PR to update `_relayContext` in `componentWillReceiveProps` and added a test for it. Let me know if there is anything else needed.

yuzhi:
Thank you, this is awesome!

packages/react-relay/modern/ReactRelayRefetchContainer.js
```diff
@@ -225,6 +234,10 @@ function createContainerWithFragments<TBase: ReactClass<*>>(
         }
         // TODO t15106389: add helper utility for fetching more data
         this._pendingRefetch = null;
+        this._relayContext = {
+          environment: this.context.relay.environment,
+          variables: fragmentVariables
```
bochen2014: will this be a bit too expensive ? as in [`cwr`](https://github.com/facebook/relay/blob/master/packages/react-relay/modern/ReactRelayRefetchContainer.js#L114)  we will restart everything over again once context variable changes;

I may be wrong, but I think  [think line](https://github.com/facebook/relay/blob/master/packages/react-relay/modern/ReactRelayRefetchContainer.js#L128)  might be a better place to plug in the fix. 
my understanding is that : A child's `props` is part of its parents `state`
process is like:
1. parent call refetch
2. parent receive new data and call `setstate` with updated data
3. child gets new props 
4. child should be re-rendered 

but if child has done its own refetch , then *by definittion* it has the latest of everything that's why we had a `!this._localVariables` check. as @josephsavona  said before

bochen2014:
I tested the nested `refetchcontainer` scenario , something like below
```
<RefetchContainer#c1>  
   <RefetchContainer#c2/>  
</RefetchCotnainer#c1>
```
so far I've noticed that if I do a `refetch` on c1,  c2 will be `re-rendered`  *only if c1's refetch returns different data*  (because of [this](https://github.com/facebook/relay/blob/master/packages/react-relay/modern/ReactRelayFragmentContainer.js#L133). as you can see, it only cares about data, not `__localVariables`. so if you use different query vars but get the same data back, then a re-render will be skipped and your children won't get `re-rendered` either)  

on [this site](https://relay-modern.azurewebsites.net/), click the second refetch button and the 3rd button to see the difference in `console.log`

bochen2014:
I now agree that `context` is a better fix because
1. `parent` got the same data after a `refetch` doesn't mean that its child won't get any new data ( `parent` doesn't have access to children's data becuase child's fragment is opaque to parent). so we need to `re-render` the children nevertheless
2. adding `__localVariables` to `state` can solve the problem. but this will only take effect on its direct child. In the case where there is remotely nested `fragments`, we want them to be `re-rendered` as well
```
<RefetchContainer>
       <wrapper>
                <wrapper>
                         <fragmentContainer/>
</RefetchCotnainer>
```

so using `context` is a better solution than `props`

bochen2014:
right.... I can [reproduce the issue](https://relay-paginator.azurewebsites.net/) now...

yuzhi:
I had to revert this because it broke a couple of the components that was nesting a another container that has it's own context inside of a Relay container. Still trying to track down the issue. Will recommit this after issues are resolved.

packages/react-relay/modern/ReactRelayRefetchContainer.js
bochen2014: I could be wrong, but is it because here we are potentially exposing local variables into root variables?
we have `root argument` and `local argument`, only `root argument` should go to context; but here, the `fragmentVariables` is a mixed of both; I think we should filter it

bochen2014: A fragment container and pagination container only has visibility of ['RootArgument' ](https://github.com/facebook/relay/blob/master/packages/relay-runtime/store/RelayConcreteVariables.js#L44);  only refetch container has 'LocalArgument' and it should keep it to itself just in case it would clash with root ones

yuzhi: The issue wasn't related to what's in `relayContext`. Somehow childContextTypes were missing for certain compositions of containers.

albertstill: Just want to add that I was running this branch locally and it was the last [commit](https://github.com/facebook/relay/pull/1702/commits/d3c2acd4b589a76948c2657b0bec55175f017bdc) that assigned to `this._relayContext` that broke my application code. It caused my pagination component, that was wrapped by a refetch component, to receive no props after a successful GraphQL query on the initial render.

yuzhi: @albertstill is there minimal repro case that you can share?

yuzhi:
@bochen2014 
>A fragment container and pagination container only has visibility to 'RootArgument' ; only refetch container has 'LocalArgument' and it should keep it to itself just in case it would clash with root ones

I'm not sure that should matter, but I could be wrong. I think of a `refetchContainer` that has been refetched as a mini `QueryRenderer`. QueryRenderer passes down all the variables it used to fetch in the context, so should a refetched container. The new fragment variables in `refetchContainer` are effectively new RootArguments. As far as clashing goes, I can't really think of variables that shouldn't be passed down to the children after a refetch has happened. If the children uses something has a RootArgument, then it must have been part of the original query too and should be allowed and could be done on purpose. If it uses a fragmentLocalVariable then it needs to be passed down from the fragment pointers/references from it's parent.

bochen2014:
thanks @yuzhi for your comments.
yeah I think you are right.. both root args and local args live in the same namespace (operation variables) so they can't have clash; and `QueryRenderer` has already proved that..

robrichard:
@yuzhi were you able to resolve the issues that led you to reverting this change? Is there anything I can do to help this get merged again?

yuzhi:
@robrichard not yet. :/ I was busy with a couple of other things later that week. Will continue to dig in. The affected views seem to be a RelayContainer that wraps around another higher level component that also sets context that wraps around a plain React component.

packages/react-relay/modern/ReactRelayRefetchContainer.js
albertstill: Opps, I needed to be running the master version of `relay-compiler` and `relay-runtime` as well and then it worked. Cheers for fixing this guys!

PR #1703, opened 2017-04-27
Fix small errors in RelayCompiler docs

tgvashworth:
You need install watchman to get the `relay-compiler` example working and the example query contained the wrong fragment name, causing errors.


leebyron:
Thanks!

PR #1708, opened 2017-04-29
Remove nested Argument variable Invariant

jquense:
👋  I don't expect to have this merged as is, since it just flat out removes the validation error and doesn't move it. I did think this is probably the clearest way to continue from the twitter conversation yesterday with @KyleAMathews, @wincent and @josephsavona

For some context I have up a POC using RelayModern's compiler to handle query validation and fragment composition in Gatsby https://github.com/gatsbyjs/gatsby/pull/876 We don't use the relay runtime just the tooling to extract and compose queries.

We have this problem where many top-level queries accept complex inputs like 

```graphql
query template_blog_post_Query($slug; String!) {
   markdownRemark(slug: { eq: $slug }) {
      id
      html
    }
}
```

Normally you could simply work around this by doing `slug: $slug` but in this case. the actual values of the arguments are provided via a build/tooling step and can't specify the variables alongside the query. In other words `$slug` is just provided to Pages as variable, and it'd be fairly impractical to provide every possible variant of the inputs.

With regard to this PR i'm happy to move or toggle the validation but I didn't see a good place to do that. so some input is appreciated :)

Thanks ya'll!

_PS. While we are at it, it would also be nice to toggle/disable the naming convention requirements for operations if that isn't required by compiler_

### DIFF
--- a/packages/relay-compiler/core/RelayCodeGenerator.js
+++ b/packages/relay-compiler/core/RelayCodeGenerator.js
@@ -18,7 +18,6 @@ const RelaySchemaUtils = require('RelaySchemaUtils');
 
 const formatStorageKey = require('formatStorageKey');
 const invariant = require('invariant');
-const prettyStringify = require('prettyStringify');
 
 import type {
   ConcreteArgument,
@@ -207,15 +206,6 @@ const RelayCodeGenVisitor = {
     },
 
     Argument(node, key, parent, ancestors): ?ConcreteArgument {
-      invariant(
-        ['Variable', 'Literal'].indexOf(node.value.kind) >= 0,
-        'RelayCodeGenerator: Complex argument values (Lists or ' +
-        'InputObjects with nested variables) are not supported, argument ' +
-        '`%s` had value `%s`. Source: %s.',
-        node.name,
-        prettyStringify(node.value),
-        getErrorMessage(ancestors[0])
-      );
       return node.value.value !== null ? node.value : null;
     },
   },


josephsavona:
Thanks for creating the PR to start the discussion! The invariant is already in a Relay-runtime specific module: this is the code generator for artifacts used by the runtime. So ideally we'd leave this invariant as is. 

Instead we should be able to make it possible to use the compiler without executing Relay codegen, which it sounds like you don't need.

jquense:
Glad to get the ball rolling :) Your comment about bypassing codegen was helpful. I realized I was probably trying to use too much of the compiler code to do this. I've started an exploration of building a simpler Runner with the compiler that skips (as much as possible now) the codegen bits.

I've found tho that I can't actually get the compiler to run or add definitions to it with the public API, without using the IRWriter. In particular it seems that you need to use `ASTConvert`?

Here are the bits I've got so far. https://github.com/jquense/relay-test/blob/master/compile.js#L96-L100 Still exploring though, as I get a better handle on the codebase.

jquense:
Ok so Yeah after some more playing around It doesn't seem possible to instantiate the Compiler without depending on at least:

```
RelayCompilerContext
ASTConvert
RelayValidator // sort of
```

That is provided we add a toggle for disabling codeGen on the compiler. Otherwise if you want to say just get the printed queries, you need RelayPrinter as well.

MatthewHerbst:
We are running into this issue with the following schema code:

```
// relay-modern query via web client js
profile {
  someCoolField(profiles: [$compare]) @include(if: $fetchComparerProfile) {
    ...
  }
}
```

Where the `profiles` argument is defined as:

```
// schema definition
profiles: {
  type: new GraphQLList(GraphQLString),
},
```

We get the error: `RelayCodeGenerator: Complex argument values (Lists or InputObjects with nested variables) are not supported`

Not sure I fully follow the above conversation, but is it related to something that would fix the error we see, or are we just doing something plain wrong?

leebyron:
@MatthewHerbst - I think that is just a current limitation of Relay. Instead you could write:

`someCoolField(profiles: $compare)` and ensure the client provides an array value for the variable.

@jquense - sounds like there's consensus that removing the invariant isn't the right way to carry forward, and @wincent has been doing some exciting work lately that will allow the relay-compiler to be used in broader ways which hopefully will be helpful to you in the future.

I'll close this PR, thanks for the discussion!

PR #1716, opened 2017-05-02
filter null plural fields

chirag04:
Using NODE_DELETE, lists become [null] array with 1 item as per the discussion in https://github.com/facebook/relay/issues/693.

There was a PR for fixing this here: https://github.com/facebook/relay/pull/722 but it was closed and @josephsavona suggested `readRelayQueryData` is the right pace to fix this. Refer: https://github.com/facebook/relay/pull/722#issuecomment-181581707

This looks like a breaking change. Let me know if there is anything i can do to fix this.

cc @yuzhi @josephsavona @jquense @itajaja

### DIFF
--- a/packages/react-relay/classic/store/readRelayQueryData.js
+++ b/packages/react-relay/classic/store/readRelayQueryData.js
@@ -225,7 +225,8 @@ class RelayStoreReader extends RelayQueryVisitor<State> {
     if (dataIDs) {
       const applicationName = node.getApplicationName();
       const previousData = getDataValue(state, applicationName);
-      const nextData = dataIDs.map((dataID, ii) => {
+      const nextData = [];
+      dataIDs.forEach((dataID, ii) => {
         let data;
         if (previousData instanceof Object) {
           data = previousData[ii];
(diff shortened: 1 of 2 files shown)


leebyron:
Hey @kassens, @josephsavona, could either of you take a look at this? It's unclear to me what the right thing to do is

packages/react-relay/classic/store/readRelayQueryData.js
```diff
@@ -239,11 +240,13 @@ class RelayStoreReader extends RelayQueryVisitor<State> {
...
           storeDataID: dataID,
         });
+        // remove null plural fields
+        if (!nextState.data) {return;}
         node.getChildren().forEach(child => this.visit(child, nextState));
         if (nextState.isPartial) {
           state.isPartial = true;
         }
-        return nextState.data;
+        nextData.push(nextState.data);
```
josephsavona: Doing `foo.map(...).filter(...)` allocates two arrays instead of one: the intermediate result of map is thrown away. Functional APIs are great for readability, but unfortunately they don't have the optimizations applied that make them performant in other languages (often called fusion). We avoid this style throughout the codebase for performance reasons.

josephsavona:
Missed this the first time around. Note that Relay Modern treats null linked records consistently for singular and plural fields: by not filtering the null. For example if the object linked to by `user.address` is deleted, then `user.address === null`. Similarly, if a user has an array of addresses and the first is deleted, then `user.addresses[0] === null`. There are some cases where it's important to know how many items exist in a list, even if they are deleted or failed to load (null).

chirag04:
@josephsavona for now we are filtering out null values in our product codebase. I'm ok with leaving this behavior as is with a note somewhere in documentation.

Also It would be helpful to understand some of the cases you mentioned where it's important to know how many items exist in a list even if they are deleted. This data will mostly be stale right?

JCMais:
@chirag04 how did you solve this on your end?

PR #1743, opened 2017-05-11
[RFC] Don't delete invisible files in CodegenDirectory

wincent:
In a side-project, I have put an `.eslintrc` file in my `__generated__` directories in order to stop eslint complaining about the use of `'use strict';` in the generated files. Specifically, it contains:

```
{
  "rules": {
    "strict": [0]
  }
}
```

Running the compiler, however, causes those invisible files to be blown away:

```
Created:
 - FooQuery.graphql.js
Deleted:
 - .eslintrc
 - .eslintrc
```

This commit makes it so that we never delete invisible files (ie. files starting with a dot), which will allow people to stick `.eslintrc`, `.gitignore` etc in their `__generated__` directories without interference. We might want to consider a whitelist approach instead (ie. only delete files with a ".graphql.js" extension), but that would be a more invasive change.

### DIFF
--- a/packages/relay-compiler/codegen/CodegenDirectory.js
+++ b/packages/relay-compiler/codegen/CodegenDirectory.js
@@ -128,6 +128,10 @@ class CodegenDirectory {
     }
   }
 
+  /**
+   * Deletes all non-generated files, except for invisible "dot" files (ie.
+   * files with names starting with ".").
+   */
   deleteExtraFiles(): void {
     fs.readdirSync(this._dir).forEach(actualFile => {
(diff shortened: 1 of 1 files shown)


leebyron:
FWIW `.eslintignore` on `__generated__` would probably be a better fit for this sort of thing, yes?

wincent:
Yes, probably, but I still think this change is worthwhile: people may want to stick other dotfiles in these directories `.gitkeep`, `.gitignore`, `.randomcrap` etc.

Issue #1687, opened 2017-04-24
[Modern] "Dumb" QueryRenderer equivalent


taion:
I'm looking to build an equivalent to react-router-relay or Found Relay for use with Relay Modern, to handle fetching multiple "root" queries in parallel, despite those queries corresponding to nested routes and corresponding components.

I don't think the approach I've taken for either library quite works here. They're respectively:
- Aggregate a bunch of queries configs into a single query config, but Relay Modern doesn't have query configs
- Use `primeCache`, but Relay Modern doesn't do caching in all cases

The approach I plan to explore is pull out the logic up to `environment.streamQuery` and `environment.subscribe` in `<QueryRenderer>` into a separate function, and just hand off the snapshot to something like a "dumb query renderer". This seems analogous to the divide between `<Relay.Renderer>` and `<Relay.ReadyStateRenderer>` in Relay Classic.

The next-best option I could think of was to render a bunch of `<QueryRenderer>`s in parallel at the root level, then portal through the data via `context` or something, but that seemed pretty gross. Also, pulling the initial snapshot logic out of React components makes for a nicer server rendering API.

Given that, does the approach I've outlined above sound reasonable? If so, would it make sense for this "dumb query renderer" to live in Relay core? I believe it'd be something that could be factored out of the current `<QueryRenderer>`, and that equivalent split between `<Relay.Renderer>` and `<Relay.ReadyStateRenderer>` did exist in Relay Classic.

josephsavona:
Great question! I can imagine a bunch of different ways to go here. First some context, then suggestions.

> Aggregate a bunch of queries configs into a single query config, but Relay Modern doesn't have query configs

Query configs (routes, really) were primarily necessary in Relay classic to work around the fact that legacy, FB-internal GraphQL didn't support multiple root fields. Relay Modern is built around the OSS GraphQL spec, so query configs are replaced with just queries and variables. Static queries also means that it isn't really possible to aggregate multiple queries together beyond possibly batching multiple queries into a single HTTP request at the network layer, if your server supports unwrapping them and execute multiple queries.

> Use primeCache, but Relay Modern doesn't do caching in all cases

Relay Modern doesn't do caching at all *by default*. However all the tools are there. The main approaches are request/response caching at the network layer (we actually have a RelayQueryResponseCache class that should probably be exported) and to add new QueryRenderer-style APIs that use `lookup` and friends to try reading from local data first.

[...]

PR #1760, opened 2017-05-15
Add lookup prop to QueryRenderer

robrichard:
As was discussed in https://github.com/facebook/relay/issues/1687, this adds an optional `lookup` boolean prop to `QueryRenderer`. When set to `true`, `QueryRenderer` will `check` that data requirements can be fulfilled by the store, and if so will `lookup` the snapshot and render with the props immediately.

This is very useful for server side rendering. On the initial server renderer, we'll need to use `relay-runtime` to directly to make the GraphQL request, then render `QueryRenderer` to string with the new `lookup` prop so that it's own `render` prop is called with `props` populated right away. On the client, you'll follow the same patter. After hydrating the relay store, you will want to immediately render with the data.

There is discussion in https://github.com/facebook/relay/issues/1687 about building a custom `LookupRenderer` in user space, but all of the edge cases that `QueryRenderer` handles would need to be duplicated there (handle new environment, query, variables, etc in `componentWillReceiveProps`). For this reason, I think it makes sense to add this logic directly to `QueryRenderer`.


packages/react-relay/modern/ReactRelayQueryRenderer.js
```diff
@@ -96,10 +97,26 @@ class ReactRelayQueryRenderer extends React.Component {
...
+    if (operation) {
+      if (lookup && environment.check(operation.root)) {
+        // data is available in the store, render without making any requests
+        const snapshot = environment.lookup(operation.fragment);
+        this.state = {
+          readyState: {
+            error: null,
+            props: snapshot.data,
+            retry: () => {
+              this._fetch(operation, props.cacheConfig);
```
robrichard: I'm not sure why flow is failing here. I think it thinks that operation could potentially be `null`, but it is guarded by the `if (operation) {` above.

```
packages/react-relay/modern/ReactRelayQueryRenderer.js:110
110:               this._fetch(operation, props.cacheConfig);
                               ^^^^^^^^^ null. This type is incompatible with the expected param type of
220:   _fetch(operation: OperationSelector, cacheConfig: ?CacheConfig): void {
                         ^^^^^^^^^^^^^^^^^ object type
```

josephsavona:
Thanks for the PR! I think it's useful to be explicit about the use-cases that are being addressed, and I see at least two:
- Server-side rendering
- Client-side rendering from cached data

Let's take the second one first since it's easier. The use-case is that clients may have already fetched some data and would prefer to render with locally cached data rather than refetch from the server. All of the query-fetching APIs currently default to always refetching from the server. But the network layer is free to do its own caching, so we added `CacheConfig.force` as a way for products to disable that. Long term I think the answer here is to change `CacheConfig` to allow specifying a "fetched after" absolute timestamp. The environment and network layer would then be free to use cached data if either a request/response cached entry or locally cached data was fresh enough, and would refetch otherwise. Here, the use of `check` and `lookup` makes sense because they are likely to be faster than the alternative, a round trip to the server.

For server rendering, however, we know that the data is fresh (minus the time to transport it from sever to client and render it). Using `check` and `lookup` are overhead: we can assume that the data is correct. A better approach here would be something like [isomorphic-relay](https://www.npmjs.com/package/isomorphic-relay)'s approach: record the results of executing queries at the network layer on the server and then inject them into the network layer on the client side.

robrichard:
@josephsavona I tried the approach of caching at the network layer first for server rendering. The problem I ran into is that making calls to the network layer is asynchronous. In the initial render on the client, it's not possible to use `QueryRenderer` or even write your own component to query the network layer. You will have to render with `null` props first (loading spinner) before rendering with the actual data. This will not match the server rendered markup and fail the React checksum test. 

Instead you have to call the network layer before attempting to render and build a new component that can render from an `OperationSelector` or a `Snapshot`. This is the component I posted here: https://github.com/facebook/relay/issues/1687#issuecomment-297514032. As you posted in a follow up, I need to implement all the edge cases around changing environments, queries, variables, and disposing of subscriptions that `QueryRenderer` already handles.

When using the approach that's in this PR, I am passing the serialized Relay store from server to client (using `environment.getStore().getSource()`) instead of cached network responses. You are correct that a `check` is unneeded overhead for this specific case as we are certain the data is already there, but it seemed useful for the general case and to handle cases like subsequent renders triggered by changing the variables on the client.

I'm very interested to hear your advice though if there is another way to handle this by caching at the network layer and being able to render the data synchronously.

robrichard:
I added an [example isomorphic app](https://github.com/robrichard/relay-modern-isomorphic-example) based on todo-modern to demonstrate how I used this. I copied my modified version of `QueryRenderer` from this PR into that repo and called it `RelayLookupQueryRenderer`.

For the server render, I use `environment.sendQuery` and then render the component to string with the `lookup` prop set to true. I then add the serialized `RecordSource` to the html payload. See here: https://github.com/robrichard/relay-modern-isomorphic-example/blob/master/js/renderServer.js#L25-L42

On the client, I create a new environment and pass the records from the html payload to the `RecordSource` constructor and again render with `lookup` set to true. See here: https://github.com/robrichard/relay-modern-isomorphic-example/blob/master/js/app.js#L11-L28

@josephsavona Let me know if you think this is a bad or inefficient approach.

unirey:
Would it make sense to add a flag (or do it by default) to give `render` access to `props` we already have (if have any) and let us handle the missing data (or `nulls`) in our component logic?

This is extremely common scenario (_e.g_ `list view -> item view`) where you already have some data (from previous or some other view, _e.g_ title, author, avatar, points _etc_) which is enough to fill above the fold in many cases/devices and improves UX considerably. Changing views would be virtually instant. This is something I tried with classic and it also didn't expose `props` until everything was fulfilled.

packages/react-relay/modern/ReactRelayQueryRenderer.js
miracle2k: I suppose by the time retry() is called, Flow feels that it can no longer guarantee that the value of operation didn't somehow change.

miracle2k:
I very much liked the behaviour of Relay Classic: Render immediately the data that is has, if it has data, but (with forceFetch) start a query in the background to refresh it at the same time.

josephsavona:
Thanks again for the PR and apologies for not following up earlier. As I mentioned previously, I think it's important to separate the two use-cases being discussed. 

Rendering a view from previously cached data implies trade offs of performance vs data consistency (especially because mutations are more likely to only refetch part of the data they might change). There's also a further consideration which is that Relay Modern does not treat the `node` root field specially (because  it is not part of the GraphQL spec). So transitioning from e.g. A list view to a detail view will not necessarily be able to render from local data, even if all the objects are actually cached locally and consistent. If it's a root field/argument that hasn't been seen before, Relay (and any other spec-compliant client) has to treat the query as unfetched. So for this use-case, the proposed `lookup` property won't (yet) help: we should iterate in the spec or find a generalized workaround for defining custom behavior of root fields.

For sever rendering, lookup is a reasonable first pass but creates unnecessary runtime overhead. @robrichard I see what you mean about network layer caching; serializing the store works well. I created a proof of concept next.js & Relay Modern integration where I use that approach, with an alternate take on QueryRenderer. I'm hoping to open-source this at some point. Overall I think that this rough approach - create a custom variant of QueryRenderer for specific use-cases - is the best way forward for now. Thoughts?

packages/react-relay/modern/ReactRelayQueryRenderer.js
```diff
@@ -144,10 +157,21 @@ class ReactRelayQueryRenderer extends React.Component {
           environment,
           variables: operation.variables,
         };
-        this._fetch(operation, nextProps.cacheConfig);
-        this.setState({
-          readyState: getDefaultState(),
-        });
+        if (nextProps.lookup && environment.check(operation.root)) {
+          const snapshot = environment.lookup(operation.fragment);
+          this.state.readyState = {
```
miracle2k: This is wrong, it has to be a setState() call.

robrichard:
@josephsavona The overhead comes from calling `check` unnecessarily? Is there anything else that is inefficient in this approach?

For now I'll publish this module as a standalone package. Looking forward to seeing your next.js integration.

robrichard:
I published [relay-query-lookup-renderer](https://www.npmjs.com/package/relay-query-lookup-renderer) as a standalone package for now.

josephsavona:
@robrichard yup, the overhead is calling `check()` which takes times proportional to the query, where the worst case is that all data is available (because it returns early if anything is missing). 

Thanks for publishing as a standalone package! We'd like to encourage more of this type of experimentation in user space. If we add every option or variation that people need to the core API it could grow unwieldy. Instead it would be great if there was an ecosystem around Relay just like there is with React. I hope this doesn't come across as rejecting the value of your work. On the contrary, we definitely support exploring the idea. I'll reply later with a few api suggestions on your repo too :-)

robrichard:
@josephsavona thanks for your help and for taking a look! I have updated my [isomorphic example](https://github.com/robrichard/relay-modern-isomorphic-example) to use the standalone package.

taion:
The semantics are sort of tricky, though. If you have a streaming network layer, you don't want to skip sending the request just because you have results in your cache, necessarily – especially in the isomorphic case.

arthens:
> But the network layer is free to do its own caching, so we added CacheConfig.force as a way for products to disable that. Long term I think the answer here is to change CacheConfig to allow specifying a "fetched after" absolute timestamp. The environment and network layer would then be free to use cached data if either a request/response cached entry or locally cached data was fresh enough, and would refetch otherwise.

@josephsavona I'm looking into caching at the network layer, but I'm not clear how it would work with mutations. 

Example flow:
- user visits `/profile`, which fetches `static_query_a`; the result includes `user=1`
- user goes to `/edit/profile` and updates its own display name via `updateUserMutation`
- goes back to `/profile`

now:
- if the cache is implemented at the query rendered layer, it should be able to return the cached data - which will include the new display name (assuming that the mutation updated the store)
- if the cache is implemented at the network layer, if it uses cached data it will return the _old_ display name (which I assume wouldn't just display the old name in the component, but would also "rollback" the result of the previous mutation in the local store)

It seems to me that the network layer doesn't have enough context to implement a smart caching system. Is the implication that the network cache should be invalidated every time there's a mutation?  Am I missing something?

miracle2k:
A problem I had with this is that the query renderer would not update (used on the client side) when the store changed. It missed these calls ``this._rootSubscription = environment.subscribe(snapshot, this._onChange);``.

I also merged in more recent changes from the relay repo. The result is here:

https://gist.github.com/miracle2k/d8a9136d0613f7142a9b84378d17973c

graingert:
does this relate to `STORE_THEN_NETWORK`?

jstejada:
@graingert, yes --

@robrichard this behavior is currently available under an experimental/undocumented `dataFrom` prop, by passing a value of STORE_THEN_NETWORK. The main blocker for actually using this is garbage collection, which prevents the data from being available most of the time.

Do you think we should keep this PR open?

taion:
We'd really, really like to get something like `STORE_THEN_NETWORK` to get behavior similar to Relay Classic. This is close to being a blocker for us for moving entirely to Relay Modern, actually.

chadfurman:
@jstejada could you elaborate on garbarge collection and how it prevents this data from being available?  Are the consequences of that additional lookups?  What if we were just using this for SSR purposes?

taion:
I don't think you'd want `STORE_THEN_NETWORK` for SSR, unless you explicitly want your clients to immediately re-fetch all data from the back end. It's more like what you got from Relay Classic with `forceFetch` turned on.

oreqizer:
is this alive? WTB SSR

taion:
I think we should close this out, given that `STORE_THEN_NETWORK` implements this functionality.

PR #1766, opened 2017-05-16
Remove parent references from RANGE_DELETE in Docs

harunsmrkovic:
Arguments `parentName`, `parentID` and `connectionName` are not used anyhow for the `RANGE_DELETE` operation from the Mutation config.

They are inferred from `pathToConnection` array, as seen here https://github.com/facebook/relay/blob/master/packages/react-relay/classic/traversal/writeRelayUpdatePayload.js#L511

In whole `handleRangeDelete` function there is no usage of these 3 parameters, so they are superfluous and confusing in the documentation.


leebyron:
Thanks!

PR #1774, opened 2017-05-17
Add warning: must use QueryRenderer for dynamic initial query

lukecwilliams:
I'd suggest a line like this so that no one runs into the wall I did while trying to convert all my containers via Step 1 before setting up the QueryRenderer per Step 2.

Query variables that are only known at runtime must be set in QueryRenderer so in that case Steps 1 and 2 must be combined.

### DIFF
--- a/docs/modern/ConversionPlaybook.md
+++ b/docs/modern/ConversionPlaybook.md
@@ -17,6 +17,8 @@ Install the latest version of Relay from the [getting started guide](./relay-mod
 
 Start converting your components and mutations to use the Relay Modern APIs from the `'react-relay/compat'` module (`createFragmentContainer`, `createRefetchContainer`, `createPaginationContainer`, `commitMutation`). It will be easier to go from the leaf components up. The [conversion scripts](https://github.com/relayjs/relay-codemod) should make this step less tedious.
 
+Note that if your container requires variables that are determined at runtime, [you must use <QueryRenderer> to set the variables](https://github.com/facebook/relay/issues/1770#issuecomment-302175983).
+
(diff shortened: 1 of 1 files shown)


docs/modern/ConversionPlaybook.md
```diff
@@ -17,6 +17,8 @@ Install the latest version of Relay from the [getting started guide](./relay-mod
 
 Start converting your components and mutations to use the Relay Modern APIs from the `'react-relay/compat'` module (`createFragmentContainer`, `createRefetchContainer`, `createPaginationContainer`, `commitMutation`). It will be easier to go from the leaf components up. The [conversion scripts](https://github.com/relayjs/relay-codemod) should make this step less tedious.
 
+Note that if your container requires variables that are determined at runtime, [you must use <QueryRenderer> to set the variables](https://github.com/facebook/relay/issues/1770#issuecomment-302175983).
```
josephsavona: I agree that this is worth documenting, though I would make two changes:
- Move this to a separate section about query variables (toward the bottom, after introducing QueryRenderer for example).
- Inline that comment rather than link to it.

lukecwilliams:
Updated - comments?  I left one line in Step 2 since in many cases Step 2
cannot be completed without this information.

<http://www.blueseastudios.com/>

   - Luke Williams
   - Blue Sea Studios, LLC
   - www.BlueSeaStudios.com <http://www.blueseastudios.com/>

leebyron:
Thanks for your help! Sorry for the wait!

sibelius:
You can*

PR #1777, opened 2017-05-17
Re-enable tests for QueryRenderer, RefetchContainer, PaginationContainer

robrichard:
I wanted to write tests for the two PRs I have open (https://github.com/facebook/relay/pull/1702, https://github.com/facebook/relay/pull/1760) but noticed the tests for those files were disabled by @kassens in https://github.com/facebook/relay/pull/1588.

It seems like the only way they could pass was if a Promise polyfill that resolves synchronously was in place. I'm assuming that's how the FB internal testing environment is set up. 

I am able to make the disabled tests pass by modifying `RelayModernMockEnvironment`'s `resolve` and `reject` methods to return promises that wait for the handlers to be called first and then updating the failing tests to wait for that promise to resolve before making assertions.

Let me know if you think this is the wrong approach.

### DIFF
--- a/packages/react-relay/modern/__tests__/ReactRelayPaginationContainer-test.js
+++ b/packages/react-relay/modern/__tests__/ReactRelayPaginationContainer-test.js
@@ -646,8 +646,9 @@ describe('ReactRelayPaginationContainer', () => {
     });
 
     it('updates after pagination (if more results)', () => {
+      expect.assertions(1);
       loadMore(1, jest.fn());
-      environment.mock.resolve(UserQuery, {
+      return environment.mock.resolve(UserQuery, {
         data: {
           node: {
             __typename: 'User',
@@ -671,13 +672,15 @@ describe('ReactRelayPaginationContainer', () => {
             },
           },
         },
+      }).then(() => {
+        expect(hasMore()).toBe(true);
       });
-      expect(hasMore()).toBe(true);
     });
 
     it('updates after pagination (if no more results)', () => {
+      expect.assertions(1);
       loadMore(1, jest.fn());
-      environment.mock.resolve(UserQuery, {
+      return environment.mock.resolve(UserQuery, {
         data: {
           node: {
             __typename: 'User',
@@ -701,8 +704,9 @@ describe('ReactRelayPaginationContainer', () => {
             },
           },
         },
+      }).then(() => {
+        expect(hasMore()).toBe(false);
       });
-      expect(hasMore()).toBe(false);
     });
   });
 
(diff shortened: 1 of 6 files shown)


robrichard:
This is easier to review with whitespace disabled: https://github.com/facebook/relay/pull/1777/files?w=1

packages/react-relay/modern/__tests__/ReactRelayPaginationContainer-test.js
```diff
@@ -656,8 +656,9 @@ describe('ReactRelayPaginationContainer', () => {
     });
 
     it('updates after pagination (if more results)', () => {
```
leebyron: Can you make this an `async () =>`

And then use `await environment.mock.resolve(...`?

robrichard:
I did try using async functions but they were not working

```
TypeError: Cannot read property 'catch' of undefined
      
      at Object.<anonymous> (node_modules/jest-jasmine2/build/jasmine-async.js:45:33)
```

I'll dig in some more to see why.

packages/react-relay/modern/__tests__/ReactRelayQueryRenderer-test.js
```diff
@@ -604,8 +618,8 @@ describe('ReactRelayQueryRenderer', () => {
             __id: '4',
           },
         },
-        retry: null,
-      }).toBeRendered;
+        retry: jasmine.any(Function),
+      }).toBeRendered();
```
robrichard: Just want to call out this change here. It's the only test that was changed other than adding `expect.assertions` and `await`'s. It looked like an error that `toBeRendered` was not invoked and `retry` is always passed to `render` when props are successfully fetched.

robrichard:
@leebyron this PR is updated to use `async/await` instead of `return/.then()`. I had to remove `babel-plugin-transform-regenerator` from the test config for it to work. It should not be needed since the tests run on node 6 & 7 where generators are natively supported.

packages/react-relay/modern/__tests__/ReactRelayQueryRenderer-test.js
leebyron: Nice catch!

scripts/jest/preprocessor.js
```diff
@@ -39,7 +39,6 @@ const babelOptions = getBabelOptions({
       schema: SCHEMA_PATH,
     }],
     require('babel-plugin-transform-async-to-generator'),
-    require('babel-plugin-transform-regenerator'),
```
leebyron: Weird that this fixed the issue, but happy to see tests passing!

Issue #1797, opened 2017-05-23
[Modern] <QueryRenderer> is never rendered with props after successful fetch (Android React Native)


lukecwilliams:
`Fetch` succeeds, graphql response is properly formed, and `render()` is called after the fetch as expected - but `props` is always null.  So the Loading UI displays perpetually.

Note that *when we attach the debugger, props does get filled* - making this a pretty bad Heisenbug for us.  App also works perfectly in iOS, just not Android without a debugger.

How can we determine why `props` is not loaded?  Is there some kind of race condition that can occur, and attaching the debugger slows things down enough for props to be loaded?

lukecwilliams:
Verified that relayPayload.source is properly populated in RelayNetwork.js:70 (in the case of correct fetch, no render that we're trying to solve).

lukecwilliams:
Verified that partialState is *not* populated correctly in ReactComponent.setState - so between a successful fetch and setState, the retrieved data is failing to get propagated somewhere.

lukecwilliams:
Verified that there are no payloads to commit after a successful fetch, at the top of RelayPublishQueue.commitPayloads.

lukecwilliams:
Very weird behavior - this._pendingPayloads does get initialized, but in `commitPayload` and `_commitPayloads`, `this._pendingPayloads.size` returns undefined.

With the debugger attached, `this._pendingPayloads.size` returns 0 or 1 as expected.

lukecwilliams:
Verified that (only without debugger attached **and** on Android) _pendingPayloads is not initialized correctly.

This code prints "undefined" in that case, "0" in all others:

    this._pendingPayloads = new _set2['default']();
    console.log('STARTING size: ' + this._pendingPayloads.size);

lukecwilliams:
@josephsavona Joe, this line in RelayPublishQueue.js is creating different objects in the "RN Android with no debugger attached" case vs all other cases (iOS or no debugger).

    var _set2 = _interopRequireDefault(require('babel-runtime/core-js/set'));

In the correct case it creates an object with a _c field which is a Set.  In this case the .size getter works as expected.

In the incorrect case the _c field is an Object, `.size` returns `undefined`, and `_JSON.stringify(obj._c)` shows:

    _c: {"_map":{"_mapData":[],"_objectIndex":{},"_stringIndex":{},"_otherIndex":{}},"size":0}

[...]

Issue #1704, opened 2017-04-27
[Modern] RN - working only when debugging remotely


ukasiu:
Hi,
I'm playing with Relay Modern on React Native. It doesn't pass props down when running normally (though requests are made), it does only when I switch to debugging in browser (though request are still made from the device).

gusbicalho:
Not sure it's the same problem from what you've said, but probably related to this: https://github.com/zloirock/core-js/issues/297
At some point the runtime checks if `pendingQueries.size` is truthy, but the Set polyfill use by relay returns `undefined` on android (yeah, wtf). I've submitted a PR fixing it, but it hasn't been accepted yet.

Meanwhile, adding the code below fixed it for me:
```
(function(PolyfillSet) {
  if (!PolyfillSet) {
    return;
  }
  var testSet = new PolyfillSet();
  if (testSet.size === undefined) {
    if (testSet._c.size === 0) {
      Object.defineProperty(PolyfillSet.prototype, 'size', {
        get: function() {
          return this._c.size;
        },
      });
    }
  }
})(require('babel-runtime/core-js/set').default);
```

ukasiu:
thanks, works like a charm.

mortezaalizadeh:
I have the same problem here, everything works fine in iOS and Android when attached to ChromDevtool, but not in unattached or release mode on my phone. Could you please let me know where to apply the above fixes?

I looked at core-js source and latest version installed on my box is way older than what we currently have on the github repo. Looks like nothing has been released since a year ago. :( It is not an easy job to apply your changes. Only way is to replace the local installed core-js in node_modules.

gusbicalho:
Just add the snippet i posted to your index.js,before everything else.
Should work.
It seems that the core -js mantainer has been applying some fixes lately,
so I'm hoping for a new release

Em seg, 8 de mai de 2017 08:55, Morteza Alizadeh <notifications@github.com>
escreveu:

mortezaalizadeh:
Thanks heaps man, worked

petersuwara:
@gusbicalho, thanks. Your black magic fixed my Android QueryRenderer for React-Native. I wasn't getting the props through the results.

Could you please explain why/how this fixes the issue... It's frustrating that it doesn't make any sense whatsoever. 

Firstly 
1. Why is this broken on Android only ?
2. Will a working version without a hack get it's way to a working version...

gusbicalho:
[...]

PR #1799, opened 2017-05-23
Add warning: does not yet work with Android

lukecwilliams:
Relay just didn't work on Android for me.  My investigation here:
https://github.com/facebook/relay/issues/1797

Solution was posted here:
https://github.com/facebook/relay/issues/1704#issuecomment-297982006

I think this is an invaluable warning to save other early adopters hours trying to dig into Relay Modern source and debug this issue.  Thanks.

### DIFF
--- a/docs/modern/Introduction.md
+++ b/docs/modern/Introduction.md
@@ -102,4 +102,4 @@ Relay Modern enables a variety of new features. Some are available via the Compa
 
 ## React Native
 
-Note that on Android, Relay Modern may not run properly because of an outstanding core-js issue.  [This polyfill](https://github.com/facebook/relay/issues/1704#issuecomment-297982006) is a workaround until this issue is fixed.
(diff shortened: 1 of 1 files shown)


leebyron:
Thanks for your investigation into this!

Could you update this message to instead clarify how to set up the environment?

We use Relay with React Native on Android at Facebook for many applications, however it's possible that we have a polyfill of Set not defined in common environments.

leebyron:
Closing since the problematic Set polyfill will no longer be included in the next release.

lukecwilliams:
@leebyron I believe a minimal app with Relay and RN should repro - I was on Android SDK 23.  The key seems to be using core-js without [the fix](https://github.com/zloirock/core-js/pull/298); in Android runtime, possibly SDK 23 relevant; and without a polyfill.

Seems likely enough there's a polyfill internally at FB.  Won't pursue further given that you noted the dependency is going away.

Issue #1780, opened 2017-05-18
Missing optimisticResponse documentation (examples with collections, when is ID required)


lukecwilliams:
A couple pieces of documentation that would be very helpful and unblock us:

1. Can minimal examples of Create, Update, and Remove on a collection be provided?  I haven't found examples of these for either React Modern or the older GraphQLMutation, but this is a very common use case.  I'm unclear on whether these can be accomplished with optimisticResponse alone, without creating an updater function.

2. When is id required on an optimistic response?  E.g. [on this thread](https://github.com/facebook/relay/issues/1778) @josephsavona suggested using `id` in the optimisticResponse even though the query doesn't include id, and that worked; however in the [Relay documentation example](https://facebook.github.io/relay/docs/mutations.html), no ID is present:

```
const optimisticResponse = () => ({
  markReadNotification: {
    notification: {                               // <-- no ID
      seenState: SEEN,
    },
  },
});

commitMutation(
  environment,
  {
    mutation,
    optimisticResponse,
    variables,
  },
);
```

Thanks!

lukecwilliams:
Found several great examples in the [Relay Modern Todo app](https://github.com/relayjs/relay-examples/tree/master/todo-modern), we should be able to work off of those.

[Submitted Relay documentation update](https://github.com/facebook/relay/pull/1802) to link the Mutations docs to this example app, as I expect many others will be looking for how to do more complex optimistic updates.

nirvdrum:
The Relay Modern Todo app link in https://github.com/facebook/relay/issues/1780#issuecomment-303844606 no longer works. For anyone coming across this issue, as it ranks highly in search engines, you'll want to take a look at the renamed [Todo app](https://github.com/relayjs/relay-examples/tree/master/todo).

PR #1802, opened 2017-05-24
Add link to more optimistic examples in Todo app

lukecwilliams:
Add/remove/update within a list are essential operations for a wide variety of apps, and it's not yet obvious where to find those examples.

Adding a link to the Relay Modern Todo app which has several great examples.


lukecwilliams:
See poor souls thumbing up at https://github.com/facebook/relay/issues/1780.

PR #1810, opened 2017-05-26
Generate plural Flow types

jsteiner:
I'm pretty new to Relay/GraphQL, so I wanted to run a failing test by the team to make sure this is the expected behavior before I spend time fixing the actual issue. If somebody gives me the go ahead I can look into fixing this.

Currently, if you declare a fragment with the directive `@relay(plural: true)` the generated Flow type is an object of that type. I would expect this to be an array of that type. This commit adds a failing test to reproduce this scenario.

### DIFF
--- a/packages/relay-compiler/core/__tests__/fixtures/flow-generator/plural-fragment.golden.txt
+++ b/packages/relay-compiler/core/__tests__/fixtures/flow-generator/plural-fragment.golden.txt
@@ -1,0 +1,3 @@
+export type PluralFragment = $ReadOnlyArray<{|
+  +id: string;
+|}>;
--- a/packages/relay-compiler/core/RelayFlowGenerator.js
+++ b/packages/relay-compiler/core/RelayFlowGenerator.js
@@ -192,6 +192,18 @@ function mergeSelections(a, b) {
   return merged;
 }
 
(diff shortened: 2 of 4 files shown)


kassens:
Thanks, this was an oversight. A fix is welcome!

Internal task: https://fburl.com/h29jgn3w

packages/relay-compiler/core/__tests__/fixtures/flow-generator/plural-fragment.golden.txt
```diff
@@ -0,0 +1,3 @@
+export type PluralFragment = ReadOnlyArray<{|
```
kassens: This would be `$ReadOnlyArray`, but you'll see that if you work on the fix.

jsteiner: Ah, yes.

jsteiner:
@kassens thanks! Added the fix. I don't know the codebase that well, so let me know if there's an easier way to check if it's plural or not.

kassens:
We should probably a utility to pull a property out of the `@relay` directive, but I don't think we have one at the moment.

Thanks for the fix!

PR #1817, opened 2017-05-27
Fix scope error in Mutations.md

colepatrickturner:
Variables depends on `source` and `storyID` and must be embedded inside `markNotificationAsRead` in order to resolve properly.


kassens:
@HankMcCoy submitted the same fix which I merged today: https://github.com/facebook/relay/pull/1812

Thanks for the help anyway 👍

Issue #1825, opened 2017-05-30
relay-compiler: Add support for mutations without input


mgtitimoli:
Hi there,

Currently, as you can see [here](https://github.com/facebook/relay/blob/296e0b703d5bc8fd556b26dde7094d855367468f/packages/relay-compiler/codegen/flowtype/printFlowTypes.js#L48), there is an assumption that `selection.args[0]` is never empty, or in other words, mutations are expected to always require an input.

![image](https://cloud.githubusercontent.com/assets/4404683/26564042/527c4ba6-44b2-11e7-9c3f-a09122a3ebfe.png)
 
**FTR**: the mutation we have has no input, since the only thing it needs is the accessToken and this is obtained from the headers.

Are we missing something, or this is indeed a bug?

Thanks!

josephsavona:
Yeah, this is a bug in the [Flow type printer](https://github.com/facebook/relay/blob/296e0b703d5bc8fd556b26dde7094d855367468f/packages/relay-compiler/codegen/flowtype/printFlowTypes.js#L48), which only handles the Relay classic style of mutations, whereas the modern *runtime* supports arbitrary mutation arguments. This should probably map over `selection.args` instead of extracting just the first one. cc @kassens

mgtitimoli:
I guess something like this should work, what do you think @josephsavona?

```javascript
        const nodes = selection.args.map(transformInputObjectToIR);

        if (node.name !== RELAY_CLASSIC_MUTATION) {
          selection.name = `${node.name}Response`;
          nodes.push(selection);
        }

        return nodes
          .map(n => generate(normalize(transform(n))).code)
          .join('\n\n');
```

josephsavona:
I think so, though @kassens is much more familiar with the Flow type generation. Try it out, what does the output look like?

josephsavona:
Currently the exported Flow types have something like 

```
export type CreatePostInput = {
  title?: ?string;
  url?: ?string;
  votes?: ?number;
};
```

It would probably be preferable to have an object representing the variables that the mutation expects, something like:

```
export type CreatePostInput = {
  title?: ?string;
  url?: ?string;
  votes?: ?number;
};
export type CreatePostVariables = {
  input: CreatePostInput,
};
```

Exporting a type for each argument would maintain backwards compatibility, while the new merged "<Foo>MutationVariables" type would be useful for typing the total variables expected as input to the mutation. If we did this, it would also be useful for queries/subscriptions too.

[...]

PR #1827, opened 2017-05-30
Variadic mutations

mgtitimoli:
- Closes: #1825


leebyron:
Thanks @mgtitimoli! I'll look into merging this, however unfortunately this PR no longer merges cleanly. Would you mind rebasing?

alloy:
@mgtitimoli Is this still the case and, if so, would you be willing to take another stab at rebasing this?

alloy:
I took a quick look at rebasing this, but the code has drifted too far out of sync and the patch is not trivial enough for me to write it from scratch. So while very unfortunate, I think that by now we should close this PR and have someone else redo it when they run into this limitation, if it still exists at all.

@mgtitimoli Thanks for spending time on it, though!

PR #1836, opened 2017-05-30
Upgrade graphql to 0.10.1

kassens:


KyleAMathews:
Any idea when there'll be a new release with this? It's holding Gatsby back atm from updating GraphQL to 0.10.

wincent:
@leebyron was asking about cutting a release this week, but I don't know if he will get to it. As such, we don't have a specific date other than "soon". But it's good to know that there's interest in seeing a release cut: it feels about time to cut one.

KyleAMathews:
kk, it's not the biggest of deals but always nice to be on the latest and shiniest :-)

Issue #1828, opened 2017-05-30
relay-modern:get current fragment variables


bochen2014:
previously I can do a `this.props.relay.variabes` to get the current variables;
but now, it seems `relay-modern` doesn't expose variables to components.  what is the recommended way of retrieving current query variables?

the `todo-modern` is a bit too simple, can we have a more sophisticated example that can show case most new features?

bochen2014:
I extended the `todo-modern` example by adding a 'switch view' button, whereby user can click to switch to a different view `todo2`. The `todo2` component will query and render some extra data I added in schema.

here is the source code of `Todolist`, which is now a `refetchContainer`
```
export default createRefetchContainer(TodoList,
  graphql.experimental`
    fragment TodoList_viewer on User
        @argumentDefinitions( 
          isNormalView:{
            type:"Boolean!",
            defaultValue:true
           }
        )
          {
          todos(
            first: 2147483647  # max GraphQLInt
          ) @connection(key: "TodoList_todos") {
            edges {
              node {
                id,
                complete,
                ...Todo_todo  @include(if: $isNormalView),
                ...Todo2_todo @skip(if: $isNormalView),
              },
            },
          }
        }
  `,
  graphql.experimental`
  query TodoListViewRefetchQuery($isNormalView: Boolean!){
    viewer{
      user{
           ...TodoList_viewer @arguments(isNormalView: $isNormalView) 
        }
      }
  }`
);
```

now the problem is I don't know whether I'm in normalView or not...
```
  _onSwitchView = e => {
    this.props.relay.refetch(previousVars => ({ isNormalView: !previousVars.isNormalView }), null)
    this.setState({ isNormalView: !isNormalView })
  }

   render(){
        const isNormalView  = ???
       <div>
          { this.props.viewer.todos.edges.map(edge => isNormalView?<Todo... > :<Todo2 ...>)}
      </div>
  }
```
in my `render` function, how do I know whether I should render `todo` or `todo2`?

yes I can use `state` to store the variables. but then it will come with another problem:  you update your state and your component will get re-rendered immediately; **but** `relay` might be still busy with fetching data... (equivalent of `getpendingVariables` stuff which we don't have now)
so what's the best practice?

kassens:
@yuzhi and I had a discussion about this last week and Yuzhi actually has a PR to expose `props.relay.variables` ready. We had a concern that that added runtime overhead for adding a feature that most of the time has a better answer.

For example, if the top level query is
```
[...]

PR #1838, opened 2017-05-31
[relay-modern]implement getVariables() for `ReactRelayRefetchContainer`

bochen2014:
I created this PR to address issue https://github.com/facebook/relay/issues/1828

the main purpose is to expose `_localVariables` so that they can referred to by `this.props.relay.variables` in components.

main idea behind is that, 
1. `_getFragmentVariables` always return the `fragment` variables which only contains **default** values because `fragment` are static;
2. your component is per instance level and can be instantiated for multiple times. Each instance will have its own variables. so we shouldn't return fragment variable for a container, but rather should return `_localVariables`


bochen2014:
can @kassens  @yuzhi  @lukecwilliams  @josephsavona  have a look?

packages/react-relay/modern/ReactRelayRefetchContainer.js
```diff
@@ -112,15 +114,17 @@ function createContainerWithFragments<TBase: ReactClass<*>>(
...
           relay,
           fragments,
           nextProps,
           this._handleFragmentDataUpdate,
         );
         this.setState({relayProp: this._buildRelayProp(relay)});
-      } else if (!this._localVariables) {
+      } else if (true && !this._localVariables) {
+        // question: why did we only call `setProps` if contains hasn't refetched?
+        //           shell we always keep _resolver updated no matter whehter comtains has refetched or not?
```
josephsavona: Good question. `localVariables` is null when the container is using variables assigned from above, and non-null once data has been refetched. The prev/next ID check above ensures that if new props refer to different objects completely that we *will* update props. If the new props refer to the same record - this case - then by definition we have already updated the resolver to the correct ids & variables (when the refetch occurred).

bochen2014:
Hi @josephsavona 
thanks for reviewing!
can you put some comments on my PR statements (i.e. `fragment` and `containers` stuff...)
the challenge I can see is that, even we implement a lazy `getFragmentVariables` api, that won't return you the correct variables. because `getFragmentVariables` will always return the default value because fragment are statically built and are agnostic about containers
think about having 2 components rendered on the same screen. How can we find the correct variables?

yuzhi:
In addition to what Joe mentioned, I recommend avoiding overloading `localVariables` with this use case. It contains variables used for the refetch query, which is not always the same set of variables used to read data out. It may contain more variables and variables  that might be different from what's used to read.

packages/react-relay/modern/ReactRelayRefetchContainer.js
bochen2014: thanks for explanation. my understanding is that, `_resolver` knows about 
1. dataID 2. query node 3. variables

a change of `dataId` has already been catered for in `areEqual(prevIDs, nextIDs)` ;
query node won't change because they are statically constructed during building;
variables can change as a result of `refetch`

so what is the scenario, where a container hasn't called `refetch`, but needs to haves its `_resolver` updated?

bochen2014:
@yuzhi  you are absolutely right. but isn't the updated variables that we all care about?
e.g. in your application, you first read data using `{ showComments: false }` then you do a refetch with `{showComments: true}`.
now, refetch has completed and you component gets re-rendered, you need to make a decision on whether to show comment section or not (you only want to display comments section once data is ready), so you consult relay to get current variables.
say we have an lazy api called `getVariables()`, so what do you expect from `getVariables()` ? do you expect it to return `false` (fragment variables. used to read data. always false) or `true` which is refetch variables (live variables. per instance variables. container variables...)

packages/react-relay/modern/ReactRelayRefetchContainer.js
josephsavona: Good question. This case accounts for possible changes to variables in an ancestor container.

bochen2014:
hmm... I think we can do a data check to `reverse engineering` what variables we used to get them...

e.g. 
```
if(this.props.story.comment){
   // data re-fetched with {$showComments: true}

}else{
   // data initially read out with fragment variables { $showComments: false}
    
 }
```
then we don't need `variables` at all

the problem I can see is, a user clicks the `show comments` button, but there is no comments associated with that story. so , by `reverse engineering` data, the application might think that this is the initial read, while in fact it's a refetch

hope it all makes sense...

josephsavona:
> getFragmentVariables will always return the default value because fragment are statically built and are agnostic about containers think about having 2 components rendered on the same screen. How can we find the correct variables?

I don't follow. `getFragmentVariables()` accounts for defaults, root variables, and fragment arguments (if fragment-local variables are being used). It's possible that there's a bug, can you provide a failing case in https://github.com/facebook/relay/blob/master/packages/relay-runtime/store/__tests__/RelayConcreteVariables-test.js#L29 ?

josephsavona:
@bochen2014 Yeah this type of reverse engineering is error-prone. I agree that we should have some way of getting the variables used to render the current data, when necessary.

bochen2014:
ah.. we can check the `data-structure`  , not the actually data; 
as long as we can see the `comments` field, we always know that this is a re-fetch. even when the actually value is `undefined or null or '' or []`

bochen2014:
@josephsavona  spot on! I had the same question
1. check `relay\packages\relay-runtime\store\RelayConcreteVariables.js`, line 42
```
variables[definition.name] = definition.defaultValue; 
```
this will make sure that default value always get returned no matter what `argumentVariables` are

2.  check `relay\packages\relay-runtime\store\RelayModernSelector.js` line 79
```
 const argumentVariables = fragments[fragment.name];
 const fragmentVariables = getFragmentVariables(
      fragment,
      operationVariables,
      argumentVariables,
    );
```
the `argumentVariables` we passed in is `fragment`. nothing to do with `variables`. I was wondering if this is intended

bochen2014:
@josephsavona  I think the unit test passed because you didn't specify a default value in argument definition

https://github.com/facebook/relay/blob/master/packages/relay-runtime/store/__tests__/RelayConcreteVariables-test.js#L39

josephsavona:
@bochen2014 Nope, test passes with a default value too. [This line](https://github.com/facebook/relay/blob/master/packages/relay-runtime/store/RelayConcreteVariables.js#L36) ensures that an argument value is preferred over a default/root value. Is there another case that you are thinking of? You might be seeing an unrelated bug.

bochen2014:
right ..... can you look at 
https://github.com/facebook/relay/blob/master/packages/relay-runtime/store/RelayModernSelector.js#L79
and see if we have a problem there?

josephsavona:
That's intended, fragment-local variables are passed via an object that keys by fragment name. Can you provide a complete example of the behavior you're seeing and how you'd expect it to behave?

packages/react-relay/modern/ReactRelayRefetchContainer.js
bochen2014: cool. that makes sense. thanks

bochen2014:
does the screenshot sort of make sense to you?

image

josephsavona:
The screenshot doesn't provide enough context to debug. Can you describe the sequence of events and what data you get vs what you expect? Are you using fragment-local variables - `@argumentDefinitions` and `@arguments`?

bochen2014:
I expose `_getFragmentVariables` [here](https://github.com/bochen2014/frankel-relay/blob/forJoe/dist/react-relay/lib/ReactRelayRefetchContainer.js#L211)

I also have [this](https://github.com/bochen2014/frankel-relay/blob/forJoe/dist/react-relay/lib/ReactRelayRefetchContainer.js#L80) as an indicate that refetch has completed

[here i am getting runtime variables](https://github.com/bochen2014/learn-relay/blob/forJoe/relay-modern/js/components/TodoList.js#L42)

my `this.props.relay.getVariables()` always return default value

josephsavona:
Ah ok. Simply exporting `_getFragmentVariables()` will only return the variables determined from props and the ancestor root query's variables. Adding a `getVariables()` function will require using `_getFragmentVariables()` unless a refetch has occurred, in which case the local variables should be used (and as @yuzhi noted, those are a subset of _localVariables).

bochen2014:
I could be wrong. but it seems that my `@arguments` are never reflected in my `this.props`  even after a refetch -- but @josephsavona  has confirmed that it should be the case 
> That's intended, fragment-local variables are passed via an object that keys by fragment name

 so it either be a bug, or I'm doing something wrong

I'm now fully convenced that exposing  `getVariables` as a lazy api is a better solution comparing to exposing `_localVariables`.  totally agree on that.  look forward to seeing how it is implemented because it didn't work for me for the reason I mentioned above

bochen2014:
I was expecting my `@argments` be part of `viewer.__fragments.tolist_viewer` *after a refetch*
image

but it's always empty -- that made me think that `getFragmentVariables` always return default value. but later on you pointed out that we have 
```
 if (argumentVariables.hasOwnProperty(definition.name)) {
      return;
    }
```
which I missed...

josephsavona:
Yeah this is tricky: in your screenshot this.props are the props received from the refetch container's *parent*, which isn't affected by the refetch. The refetch container internally updates the props it provides to the component, using the expected variables.

josephsavona:
It's great that you're digging in and experimenting with the core functions! Care to try updating the getFragmentVariables() function with the logic I described?

bochen2014:
yep! I will have a try ;-) thanks @josephsavona . that has been extremely informative and helpful!

josephsavona:
Sure thing, thanks for getting involved!

packages/react-relay/modern/ReactRelayRefetchContainer.js
```diff
@@ -184,11 +185,11 @@ function createContainerWithFragments<TBase: ReactClass<*>>(
...
-    _getFragmentVariables(): Variables {
+    _getFragmentVariables = (): Variables => {
       const {
         getVariablesFromObject,
       } = this.context.relay.environment.unstable_internal;
-      return getVariablesFromObject(
+      return this._localVariables || getVariablesFromObject(
         this.context.relay.variables,
         fragments,
         this.props,
```
bochen2014: I noticed that we are not forward variables into descendants , which means, `_localVariables` is only visible to local;
so only root query's variables are visible to its children?

bochen2014:
done. a piece of cake after you explained how it works (or supposed to work) . thanks @josephsavona  again!

yuzhi:
Can you add some unit tests?

bochen2014:
yes I was expecting someone to ask ;-) haha

wincent:
Epic thread. 😄

bochen2014:
A challenge could be nested refetch containers.
say `stories` can refetch with `dateFilter` which is last week, last month or past 6 monts;
each `story` can also refetch with `showComments`;
it would be interesting to see what will happen when clicking the 'last month' button with 'show comments' button already clicked on stories of ' last week'
I think this would be better example than the official `todo-modern`

packages/react-relay/modern/ReactRelayRefetchContainer.js
```diff
@@ -188,15 +190,23 @@ function createContainerWithFragments<TBase: ReactClass<*>>(
...
+
+      const fragmentVariables = getVariablesFromObject(
         this.context.relay.variables,
         fragments,
         this.props,
       );
+
+      if (this._localVariables) {
+          forEachObject(fragmentVariables, (_, key) => {
+             fragmentVariables[key] = this._localVariables[key];
```
bochen2014: as @yuzhi mentioned,  `_localVariables` is a superset of `fragmentVariables`

bochen2014:
I addressed the `superset/subset` issue raised by @yuzhi  in the updated PR; I have also added a new unit test and refactored existing unit test. 

can someone have a review? thanks!

bochen2014:
1. the failed unit test is `Cannot find module 'RelayModern' from 'commitRelayModernMutation-test.js'` which I don't think is caused by my changes;
2.my previous change on `refetchContainer-test` was quite aggressive and hard to review. 
this time, I limit the changes to minimum required. So it should be very easy to reivew

@josephsavona @yuzhi  can you guys have a look ? thanks!

AGS-:
@bochen2014 Let me see what's up with the test failing tomorrow morning

AGS-:
@bochen2014 it should be fixed now

bochen2014:
thanks @AGS- . the unit tests now all passed..

can I ask someone to review my changes again and point out issues/problems if any?
our web site relies heavy on query variables and this PR is critical for our `relay-moder` migration. thanks!

bochen2014:
this PR is getting really messy. will create a new one instead

PR #1846, opened 2017-05-31
RFC: Teach compiler `--persist` option

wincent:
Sample usage:

    relay-compiler --src ./src --schema schema.graphql --persist scripts/persistQuery.js

Where `persistQuery.js` looks like this:

```
function persistQuery(text) {
  var id = someLogicToPersistQueryText(text);
  return Promise.resolve(id);
}
```

The alternative in the absence of this option is to copy-paste the entire `RelayCompilerBin` implementation just to override that bit.

Note that webpack makes this ghastly — necessitating the use of `eval` — because it wants to rewrite the `require` of the passed in module, which of course fails because it is dynamic. Google search turns up a bunch of results like this one lamenting the situation:

https://stackoverflow.com/questions/30575060/require-js-files-dynamically-on-runtime-using-webpack

Apparently it's better in webpack 2.

### DIFF
--- a/packages/relay-compiler/bin/RelayCompilerBin.js
+++ b/packages/relay-compiler/bin/RelayCompilerBin.js
@@ -60,6 +60,7 @@ async function run(options: {
   schema: string,
   src: string,
   extensions: Array<string>,
+  persist?: string,
   watch?: ?boolean,
 }) {
   const schemaPath = path.resolve(process.cwd(), options.schema);
@@ -70,6 +71,14 @@ async function run(options: {
   if (!fs.existsSync(srcDir)) {
     throw new Error(`--source path does not exist: ${srcDir}.`);
   }
+  const persistModule = options.persist && path.resolve(process.cwd(), options.persist);
+  if (persistModule && !fs.existsSync(persistModule)) {
+    throw new Error(`--persist path does not exist: ${persistModule}.`);
+  }
+  // Using eval here to prevent webpack from trying to rewrite the require and
+  // failing.
+  const persistQuery = persistModule &&
+    eval(`require(${JSON.stringify(persistModule)})`);
   if (options.watch && !hasWatchmanRootFile(srcDir)) {
     throw new Error(
       `
@@ -96,7 +105,7 @@ Ensure that one such file exists in ${srcDir} or its parents.
   };
   const writerConfigs = {
     default: {
-      getWriter: getRelayFileWriter(srcDir),
+      getWriter: getRelayFileWriter(srcDir, persistQuery),
       parser: 'default',
     },
   };
@@ -104,7 +113,6 @@ Ensure that one such file exists in ${srcDir} or its parents.
     parserConfigs,
     writerConfigs,
     onlyValidate: false,
-    skipPersist: true,
   });
   if (options.watch) {
     await codegenRunner.watchAll();
@@ -114,8 +122,12 @@ Ensure that one such file exists in ${srcDir} or its parents.
   }
 }
 
-function getRelayFileWriter(baseDir: string) {
+function getRelayFileWriter(
+  baseDir: string,
+  persistQuery: (text: string) => Promise<string>,
+) {
   return (onlyValidate, schema, documents, baseDocuments) =>
+
     new RelayFileWriter({
       config: {
         formatModule: formatGeneratedModule,
@@ -126,6 +138,7 @@ function getRelayFileWriter(baseDir: string) {
           queryTransforms,
         },
         baseDir,
+        persistQuery,
         schemaExtensions,
       },
       onlyValidate,
@@ -181,7 +194,7 @@ function hasWatchmanRootFile(testPath) {
 const argv = yargs
   .usage(
     'Create Relay generated files\n\n' +
-      '$0 --schema <path> --src <path> [--watch]',
+      '$0 --schema <path> --src <path> [--persist <module-path>] [--watch]',
   )
   .options({
     schema: {
@@ -192,6 +205,10 @@ const argv = yargs
     src: {
       describe: 'Root directory of application code',
       demandOption: true,
+      type: 'string',
+    },
+    persist: {
+      describe: 'Path to module exporting a `persistQuery` function',
       type: 'string',
     },
     extensions: {
--- a/packages/relay-compiler/bin/RelayCompilerBin.js
+++ b/packages/relay-compiler/bin/RelayCompilerBin.js
@@ -75,10 +75,9 @@ async function run(options: {
   if (persistModule && !fs.existsSync(persistModule)) {
     throw new Error(`--persist path does not exist: ${persistModule}.`);
   }
-  // Using eval here to prevent webpack from trying to rewrite the require and
-  // failing.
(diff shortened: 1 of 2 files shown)


wincent:
This is ugly as hell, but throwing it up there as an RFC anyway. In my own side-project, I went down the path of forking `RelayCompilerBin`, but it seems pretty limiting not to be able to use the `relay-compiler` executable for anything but very simple use cases.

packages/relay-compiler/bin/RelayCompilerBin.js
```diff
@@ -70,6 +71,14 @@ async function run(options: {
...
     throw new Error(`--source path does not exist: ${srcDir}.`);
   }
+  const persistModule = options.persist && path.resolve(process.cwd(), options.persist);
+  if (persistModule && !fs.existsSync(persistModule)) {
+    throw new Error(`--persist path does not exist: ${persistModule}.`);
+  }
+  // Using eval here to prevent webpack from trying to rewrite the require and
+  // failing.
+  const persistQuery = persistModule &&
+    eval(`require(${JSON.stringify(persistModule)})`);
```
wincent: Ugh.

josephsavona: Isn't there something like require.actual()? Or maybe alias require to something else?

josephsavona:
I was going to send a PR for this too, definitely something we should add.

packages/relay-compiler/bin/RelayCompilerBin.js
wincent: In vanilla node, there is not (that's a Jest thing). I looked on the interwebs, but I couldn't find anything. Moving to webpack 2 would provide us with some more options, I think, but that's obviously a bigger change.

josephsavona: What about aliasing require so webpack doesn't transform it? Or shell out to a separate process?

wincent: Good idea. Let me see if aliasing works.

wincent: Seems like it does work (at least, build runs without blowing up); webpack lets the line in question through as:

```
var persistQuery = persistModule && global['require'](persistModule);
```

in the built output.

wincent:
Gonna import this to run Flow tests internally (my local set-up is borked right now).

wincent:
> Gonna import this to run Flow tests internally (my local set-up is borked right now).

Heh, or I could just look at Travis...

kassens:
Webpack actually aliases `global`, I also couldn't find a way to do this and resorted to eval in a similar situation.
Do you have a pointer how this works in webpack 2? It might help in my situation.

wincent:
> Webpack actually aliases `global`, I also couldn't find a way to do this and resorted to eval in a similar situation.

I was hopeful that this would work because the built output looks like this:

```
            var persistModule = options.persist && __webpack_require__(5).resolve(process.cwd(), options.persist);
            if (persistModule && !__webpack_require__(6).existsSync(persistModule)) {
              throw new Error('--persist path does not exist: ' + persistModule + '.');
            }
            // Need to hide the `require` here to prevent webpack from rewriting
            // it and failing.
            var persistQuery = persistModule && global['require'](persistModule);
```

Note how it hasn't mangled `global['require']`. I also couldn't see any direct messing with `global`, at least inside the `relay-compiler` bin file, although be getting messed with somewhere else.
> Do you have a pointer how this works in webpack 2? It might help in my situation.

robrichard:
Do you need to run the `relay-compiler` bin through webpack at all? Since it's a node command line script, couldn't you leave all the `require`s intact?

wincent:
Cross-referencing: https://github.com/facebook/relay/pull/1710#pullrequestreview-42189528

@robrichard:

> Do you need to run the `relay-compiler` bin through webpack at all? Since it's a node command line script, couldn't you leave all the `require`s intact?

True.  My recollection is that we're doing this as an optimization to minimize cold-start time.

josephsavona:
Ship it!

unirey:
Any progress or new thoughts on this? Would love to see this feature added.

unirey:
Any progress or new ideas on this? Would love to see this feature added.

unirey:
What's holding this back? Maintaining my own version of Relay is not fun..

unirey:
What's holding this back?

josephsavona:
@unirey No idea. @leebyron, what *is* holding this back?

wincent:
I imported it but it has since grown stale. I need to dust it off and rebase it. A question was raised into whether the change as I had made it is would work everywhere, and I haven't had time to (re-)validate that myself yet.

unirey:
@wincent Thank you, I really appreciate it. Im planning to ship in 2-4 weeks & I would definitely love to persist but I would also prefer not take a slippery slope of creating, maintaining my own version of Relay.

wincent:
> I would also prefer not take a slippery slope of creating, maintaining my own version of Relay.

@unirey: not to dismiss your point (I'd like to see this option merged, obviously), but note that it's not really creating your own version of Relay: you just need to fork a single file (and in fact, internally at Facebook we do exactly that: we have a custom "compiler" executable that is configured to suit our specific needs). I did this for a side-project of mine, for instance: [`compile.js`](https://github.com/wincent/masochist/blob/92cfdd233b6e02509bcdb5ce755d80ebb6b74bd7/scripts/compile.js).

robrichard:
@wincent forking that file using a version of relay newer than 1.0.0 requires `formatGeneratedModule` which is not publicly exported. See https://github.com/facebook/relay/issues/1927

wincent:
> @wincent forking that file using a version of relay newer than 1.0.0 requires `formatGeneratedModule` which is not publicly exported.

I think it's probably reasonable to expose that, although we've historically been very conservative and minimal in terms of what we include in the public API. The main question is whether this stuff should be exposed at the `relay-compiler` level, or rather just the building blocks for it lower down, in the `graphql-compiler`.

GrigoryPtashko:
Hello, everyone. Is this going to be merged actually? That's a nice option I'd like to use instead of forking..

schickling:
:shipit:

yusinto:
I published my own package [relay-compiler-plus](https://github.com/yusinto/relay-compiler-plus/stargazers) which supports persisted queries and compiles graphql-js schema as well. Hope that helps for now.

jwb:
It would be really great to have support from the core to proceed with this!

sibelius:
@jstejada any plans for this feature?

devknoll:
I haven't spoken to the team yet, but my hunch is that if someone wanted to update this diff to make `persist` a command that is executed, rather than a JS module, we might be willing to merge it.

sibelius:
@yusinto can u send a PR making `persist` a command?

yusinto:
Yes I will submit a pr soon. Thanks for the suggestion @sibelius

yusinto:
Done. [Here](https://github.com/facebook/relay/pull/2354) it is.

alloy:
To avoid confusion, I’ll close this PR for now.

PR #1849, opened 2017-06-02
Remove write-only properties from QueryRenderer

taion:
These never get read from. Can they be removed?


josephsavona:
Seems reasonable

PR #1857, opened 2017-06-03
Make graphiQLPrinter optional

### LINKED ISSUE #1856: How do we use RelayNetworkLogger?
Regarding: https://github.com/facebook/relay/blob/v1.0.0/packages/relay-runtime/network/RelayNetworkLogger.js

There doesn't seem to be any documentation here. Is the idea to call:

```js
const network = Network.create(
  RelayNetworkLogger.wrapFetch(fetchQuery, () => ''),
);
```

Would it make sense to make the `graphiQLPrinter` argument optional?

taion:
Closes #1856

### DIFF
--- a/packages/relay-runtime/network/RelayNetworkLogger.js
+++ b/packages/relay-runtime/network/RelayNetworkLogger.js
@@ -39,7 +39,7 @@ let queryID = 1;
 const RelayNetworkLogger = {
   wrapFetch(
     fetch: FetchFunction,
-    graphiQLPrinter: GraphiQLPrinter,
+    graphiQLPrinter?: GraphiQLPrinter,
   ): FetchFunction {
     return (
       operation: ConcreteBatch,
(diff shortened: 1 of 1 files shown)


kassens:
Thanks!

packages/relay-runtime/network/RelayNetworkLogger.js
```diff
@@ -55,9 +55,13 @@ const RelayNetworkLogger = {
       console.time && console.time(idName);
 
       const onSettled = (error: ?Error, response: ?QueryPayload): void => {
-        console.groupCollapsed(`%c${idName}`, error ? 'color:red' : '');
+        if (console.groupCollapsed) {
+          console.groupCollapsed(`%c${idName}`, error ? 'color:red' : '');
```
taion: not defined in e.g. jest `CustomConsole`

kassens: I imported this to quickly. If `groupCollapsed` is missing, this might be a bit too verbose. Do you think we should just guard at the top and return the non-wrapped function so it works in environments like Jest or is the ignoring of grouping more helpful?

kassens:
What you're seeing seems to be the effect of 18f9d6103f615b622892b4217942db55cb24bdc6. I created #1858 to see what we want to do with `SyncOrPromise` before the next release.

taion:
It's nice to be able to get console debug output to diagnose failing tests.

I don't know offhand if it's just Jest that stubs out `console`, though. I guess if it's just Jest, it might be easiest to just fix Jest.

kassens:
React Native also has some sort of limited console implementation when not connected to a debugger going to `adb` or iOS logs. So checking for the capabilities definitely makes sense here. The question is only what to do if they're partially missing.

taion:
This is already sniffing there for `console.time` and `console.timeEnd`. I guess maybe the RN console limitation supported `groupCollapsed`?

Though on second thought, I probably do want to see that `idName`.

Issue #1818, opened 2017-05-28
[Modern] commitMutation's onCompleted firing with undefined response on Android only


thatwillwork:
I'm encountering a strange issue where `commitMutation` doesn't seem to be working properly on Android only. On my iOS simulator everything works as expected, but when I test on my Android device `onCompleted` seems to get called with the `response` undefined. 

Here is the `commitMutation` code:
```
    commitMutation(environment, {
      mutation,
      variables,
      updater: store => {
        console.log("this never gets printed");
      },
      onCompleted: (response) => {
        // response is undefined here
      },
      onError: err => {
        // error is also undefined here
      }
    });
```

Here is the `fetchQuery` function I pass to the `Network`:
```
function fetchQuery(operation, variables, cacheConfig, uploadables) {
  return fetch("https://XXXXXXX.com/graphql", {
    method: "POST",
    headers: {
      "content-type": "application/json",
      Accept: "application/json"
    },
    body: JSON.stringify({
      query: operation.text,
      variables
    })
  })
    .then(response => {
        return response.json();
    })
    .then(responseData => {
      console.error(responseData);
      return responseData;
    })
    .catch(error => {
      console.error(error);
    });
}
```

According to the `console.error(responseData)` call in the `fetch` function call above, the `responseData` is correct and consistent across the iOS simulator and my Android device. This leads me to believe that the response is getting lost between the `fetch` call and the callback set up in `commitMutation`.

Another observation is that when remote JS debugging is enabled on Android, the response returned is actually not undefined. Using the inspector the only difference between the network responses is that `responseType` is empty when remote JS debugging is off, and `responseType` is `blob` when remote JS debugging is on.

This is the full package.json used:
```
{
  "name": "foo",
  "version": "0.0.1",
  "private": true,
  "scripts": {
    "start": "node node_modules/react-native/local-cli/cli.js start",
    "flow": "flow check",
    "test": "jest",
    "relay": "relay-compiler --src ./src --schema schema.graphql",
    "get-schema": "node scripts/getSchema.js"
  },
  "dependencies": {
    "@shoutem/ui": "shoutem/ui#242/head",
    "react": "16.0.0-alpha.6",
    "react-native": "0.44.0",
    "react-native-gifted-chat": "^0.1.4",
    "react-native-vector-icons": "^4.0.1",
    "react-navigation": "^1.0.0-beta.11",
    "react-redux": "^5.0.2",
    "react-relay": "^1.0.0",
    "redux": "^3.6.0",
[...]

PR #1867, opened 2017-06-07
RFC: Do not include package-local polyfills.

leebyron:
This removes babel's core-js polyfills from the distributed NPM builds. This aids in further reducing app bundle size when targeting modern browsers, and allows flexibility in how these polyfills are provided if necessary, which can help fix React Native specific issues, as illustrated in #1799 and #1704 and #1818.

The drawback is that polyfills for these things are now something that client developers will need to think about directly. Shipping Relay to older browsers before this diff should work fine, but after this diff will require global polyfills. As such, this diff may constitute at least a minor-breaking change.


kassens:
Looks good, thanks for trimming this down 👍

leebyron:
Ok, so of course this ended up being way more complicated than originally anticipated.

Using native Promises instead of polyfilled Promises broke a bunch of tests since native Promises are implemented using `setTimeout()` which can be controlled via `jest.runAllTimers()` however native Promises use the micro-task queue which jest does not control. This means doing some test refactoring. So I've added to this patch some refactored tests using async/await.

leebyron:
Alas, because FB's internal environment still asserts a polyfilled Promise during test runs, these tests will fail in that environment even though they succeed in this one.

Closing this PR in favor of #1870, but hopefully this work is not meaningless if we ever change how this internal environment works and wish to support native promises in these tests.

PR #1877, opened 2017-06-10
Update doc block to use pageInfo field names

staylor:
`page_info` throws an error, as do the non-camelcase fields present in this block, because they are not present in the connection schema. Pagination containers are already hard to figure out without concrete examples, we should at least make the pseudo-code (mostly) runnable if copy-pasta'd.


kassens:
Thanks for the fix, this is caused by legacy field names used internally.

PR #1889, opened 2017-06-14
Make compatible with React Native 0.45

### LINKED ISSUE #1866: error on react native 45 and react 16-alpha.12
I'm using react-relay 1.0.0
I've upgraded my react native code to RN 45, and it is throwing the following error:

image

in this line: `require('./relayUnstableBatchedUpdates')`

```jsx
var mounted = _this3.mounted;
        if (mounted) {
          var updateProfiler = require('./RelayProfiler').profile('RelayContainer.update');
          require('./relayUnstableBatchedUpdates')(function () {
            _this3.setState(partialState, function () {
              updateProfiler.stop();
              if (isComplete) {
                completeProfiler.stop();
              }
            });
            if (callback) {
              callback.call(
              // eslint-disable-next-line react/no-string-refs
              _this3.refs.component || null, (0, _extends3['default'])({}, readyState, { mounted: mounted }));
            }
          });
        } else {
```

alloy:
Fixes #1866.

The `unstable_batchedUpdates` function is no longer available from the root module, but instead has to be fetched from the renderer shim, which in turn gets it from the correct renderer (stack vs fiber).

### DIFF
--- a/packages/react-relay/classic/tools/relayUnstableBatchedUpdates.native.js
+++ b/packages/react-relay/classic/tools/relayUnstableBatchedUpdates.native.js
@@ -13,5 +13,13 @@
 'use strict';
 
 const ReactNative = require('ReactNative');
+let batchedUpdates = undefined;
 
-module.exports = ReactNative.unstable_batchedUpdates;
+if (ReactNative.unstable_batchedUpdates) {
+  batchedUpdates = ReactNative.unstable_batchedUpdates;
+} else {
+  const Renderer = require('ReactNative/Libraries/Renderer/src/renderers/native/ReactNative');
+  batchedUpdates = Renderer.unstable_batchedUpdates;
+}
+
+module.exports = batchedUpdates;
--- a/packages/react-relay/classic/tools/relayUnstableBatchedUpdates.native.js
+++ b/packages/react-relay/classic/tools/relayUnstableBatchedUpdates.native.js
@@ -13,13 +13,8 @@
 'use strict';
 
 const ReactNative = require('ReactNative');
-let batchedUpdates = undefined;
 
-if (ReactNative.unstable_batchedUpdates) {
-  batchedUpdates = ReactNative.unstable_batchedUpdates;
-} else {
-  const Renderer = require('ReactNative/Libraries/Renderer/src/renderers/native/ReactNative');
-  batchedUpdates = Renderer.unstable_batchedUpdates;
-}
-
-module.exports = batchedUpdates;
+module.exports =
+  ReactNative.unstable_batchedUpdates ||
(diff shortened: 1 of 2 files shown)


leebyron:
Sweet, thanks!

packages/react-relay/classic/tools/relayUnstableBatchedUpdates.native.js
```diff
@@ -13,13 +13,8 @@
...
-} else {
-  const Renderer = require('ReactNative/Libraries/Renderer/src/renderers/native/ReactNative');
-  batchedUpdates = Renderer.unstable_batchedUpdates;
-}
-
-module.exports = batchedUpdates;
+module.exports =
+  ReactNative.unstable_batchedUpdates ||
+  require('ReactNative/Libraries/Renderer/src/renderers/native/ReactNative')
+    .unstable_batchedUpdates;
```
alloy: 👌

alloy:
Is the merge process stuck?

leebyron:
Yes. This broke in our Facebook-side test suite. I'll bring it up with the
React Native team for help

yungsters:
I don't think this is correct because `ReactNative` *should* have a function exported named `unstable_batchedUpdates`. Here is where that export is happening: https://github.com/facebook/react/blob/master/src/renderers/native/ReactNativeStackEntry.js#L48

alloy:
Hmm. I guess things may have shuffled around again since RN 0.45 was released?

yungsters:
@alloy Probably. I suspect one of the changes necessary to make React Native use flat bundles built from the React repository may have caused this to happen.

leebyron:
Ok, since this PR fails some integration tests and @yungsters thinks it may be a RN side issue, I'm going to close this PR, but we can keep track of this issue in #1866 and perhaps look forward to a RN release which fixes this.

Thanks again for your investigation, @alloy

alloy:
I get that this issue may have been an unfortunate result of files moving around, but I’d like to be sure that you’d prefer to skip compatibility with RN 0.45 completely?

(I haven’t tried RN 0.46 yet, I was panning to do that once the Relay migration was over, but I can give that a try sooner.)

yungsters:
@alloy Yes, definitely. Let me know if my suggested fix on the issue resolves the bug. If so, I will send a fix and try to get it released as a patch to 0.45.

yungsters:
Correction (because I misread): I would prefer to not skip compatibility in 0.45.

alloy:
As commented on #1866, your patch to RN fixes the issue. However, that does mean RN versions that did not have that patch applied won’t work. I think that in our case we’ll just migrate to the latest RN anyways, but let me know if you’d like me to revive this PR in some form to fix compatibility with older RN versions.

PR #1894, opened 2017-06-16
add error for missing string.startsWith

robrichard:
@leebyron Relay Runtime also requires a polyfill for String.prototype.startsWith/endsWith which is not supported in Safari <9 or IE11. This is required even for v1.0.0 which does not include https://github.com/facebook/relay/commit/3daedb2af1ab3baf67c61ad27afd7d8bd60aa59c.


leebyron:
Hmm maybe we should just not rely on these methods at all since even more modern browsers don't support them yet. Let me investigate quickly

Issue #1897, opened 2017-06-16
[relay-compiler] using __generated__ relay queries of shared module


rosskevin:
We have a backend ruby API/graphql engine, as well as a shared UI module that are reused in different applications (both server and client side).  Each client side app customizes the use and expands upon the `ui` module.

In the following case, our `dummy` client side app (used for testing the engine) uses the `ui` module which has the majority of the UI widgets and therefore relay code (from `dummy/src/routes.js`):

```js
// @flow
import React from 'react'
import Route, {makeRouteConfig} from 'found'
import AppFrame from '@alienfast/ui/App/AppFrame'
import {graphql} from 'react-relay'

export default makeRouteConfig (
  <Route
    component={AppFrame}
    query={graphql`
      query routes_DummyApp_Query {
        current_user {
          ...AppFrame_currentUser
        }
      }`
    }
  />
)
```

Error:

```
Command failed `relay-compiler --src ./src --schema ./schema.json`:
Invariant Violation: RelayApplyFragmentArgumentTransform: expected `AppFrame_currentUser` to be a fragment, got `undefined`.
  at invariant (/Users/kross/projects/af/spec/dummy/ui/node_modules/fbjs/lib/invariant.js:44:15)
at transformFragmentSpread (/Users/kross/projects/af/spec/dummy/ui/node_modules/relay-compiler/bin/relay-compiler:4874:26)
at /Users/kross/projects/af/spec/dummy/ui/node_modules/relay-compiler/bin/relay-compiler:4936:24
at Array.forEach (native)
at transformSelections (/Users/kross/projects/af/spec/dummy/ui/node_modules/relay-compiler/bin/relay-compiler:4931:15)
at transformField (/Users/kross/projects/af/spec/dummy/ui/node_modules/relay-compiler/bin/relay-compiler:4890:23)
at /Users/kross/projects/af/spec/dummy/ui/node_modules/relay-compiler/bin/relay-compiler:4946:24
at Array.forEach (native)
at transformSelections (/Users/kross/projects/af/spec/dummy/ui/node_modules/relay-compiler/bin/relay-compiler:4931:15)
at transformNode (/Users/kross/projects/af/spec/dummy/ui/node_modules/relay-compiler/bin/relay-compiler:4854:21)
```

---
Other links:
- #1790
- [My SO question](https://stackoverflow.com/questions/44593410/reuse-of-library-with-generated-relay-queries)

How can we reuse queries that are not in the source tree but contained in our shared module?

[...]

PR #1901, opened 2017-06-19
resolve module names for files named *.js.flow

rosskevin:
In the case where users custom compile with base documents parsed from node_module library shadow files (full context here: https://github.com/facebook/relay/issues/1897#issuecomment-309560870), we need to be able to resolve both standard one-part extensions e.g. `*.js` as well as `*.js.flow`.

This solves that problem, as module name resolution is otherwise broken, specifically indexes.


kassens:
Could you rebase this on 78ea60b8dd5758188d8bd54cffdc880aea2f61a0? I think it'll already fix your problem.

rosskevin:
I screwed up the rebase.  I'll re-submit.

PR #1905, opened 2017-06-21
Changes to must/may language for clientMutationId as per Relay Modern

object88:
The implementation of Relay Modern is such that for a mutation, `clientMutationId` is no longer required.  This should bring the docs in line with that reality.

See also [discussion](https://github.com/relayjs/relay-examples/pull/32) in a `relay-examples` PR.


wincent:
Looks good. Thanks!

PR #1907, opened 2017-06-21
Fix bug where RelayModernFragmentSpecResolver was not resolving data correctly

### LINKED ISSUE #1750: [Modern] SetLinkedRecords doesn't update the UI while last edges is deleted
I try to create my own custom deleteNode function.

```
function deleteByID(store, rootFieldName, deletedID) {
	const record = store.getRoot()
	const edges = record.getLinkedRecords(rootFieldName);
	if (!edges) {
		return;
	}
	let nextEdges;
	for (let ii = 0; ii < edges.length; ii++) {
		const edge = edges[ii];
		const node = edge;
		if (node != null && node.getDataID() === deletedID) {
			if (nextEdges === undefined) {
				nextEdges = edges.slice(0, ii);
			}
		} else if (nextEdges !== undefined) {
			nextEdges.push(edge);
		}
	}
	if (nextEdges !== undefined) {
		record.setLinkedRecords(nextEdges, rootFieldName);
	}
}
```

It works fine, except while the last edges is deleted, the UI doesn't updated.

### LINKED ISSUE #1904: [modern] Fragment containers not resolving data correctly after mutation
There is a chance that I'm using `relay-modern` incorrectly, but I believe I may have discovered a bug when using `RecordProxy#setLinkedRecords`. Please see the example repo thrown together at https://github.com/casparrolfe/relay-record-proxy-bug to reproduce the bug. `yarn start` is all you will need. Server responses have been mocked on the client.

To reproduce:
- Delete the last item in the displayed list.
- The response will be as expected and the in memory cache will be updated correctly, however the fragment container will fail to resolve the data correctly.

Delete any other item in the list and everything else will work as expected.

I have a hunch the problem may lie somewhere in here:
https://github.com/facebook/relay/blob/master/packages/relay-runtime/store/RelayModernFragmentSpecResolver.js#L92-L119

casparrolfe:
Fixes #1750
Fixes #1904 

The RelayModernFragmentSpecResolver was not resolving data correctly when the item at end of the array passed in props was removed.

i.e. given that my fragment is resolving data for `[zuck, beast]`, when this changes to `[zuck]`, the resolved data should reflect this. Previously it was returning the existing data as it was not checking if the length of the resolvers was not equal to the previous data length.

Does this need to be fixed in classic too?

### DIFF
--- a/packages/relay-runtime/store/RelayModernFragmentSpecResolver.js
+++ b/packages/relay-runtime/store/RelayModernFragmentSpecResolver.js
@@ -287,6 +287,9 @@ class SelectorListResolver {
           nextData.push(nextItem);
         }
       }
+      if (!nextData && this._resolvers.length !== prevData.length) {
+        nextData = prevData.slice(0, this._resolvers.length);
+      }
       this._data = nextData || prevData;
       this._stale = false;
     }
--- a/packages/relay-runtime/store/__tests__/RelayModernFragmentSpecResolver-test.js
+++ b/packages/relay-runtime/store/__tests__/RelayModernFragmentSpecResolver-test.js
@@ -465,6 +465,39 @@ describe('RelayModernFragmentSpecResolver', () => {
       });
     });
 
+    it('resolves fragment data when the item at the end of the array is removed', () => {
+      const resolver = new RelayModernFragmentSpecResolver(
+        context,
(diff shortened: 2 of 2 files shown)


kassens:
This is awesome. Thanks for including a solid test 👍

PR #1919, opened 2017-06-26
[relay-compiler] resolve module names for files named *.js.flow

rosskevin:
In the case where users custom compile with base documents parsed from node_module library shadow files (full context here: https://github.com/facebook/relay/issues/1897#issuecomment-309560870), we need to be able to resolve both standard one-part extensions e.g. `*.js` as well as `*.js.flow`.

Continuation of #1901


rosskevin:
@kassens I only added back the test assertions on master (here) and it is not fixed with your latest changes.  If you want me to look at a fix, please let me know.

rosskevin:
@kassens the fix is in and should be good to merge.

kassens:
Thanks for updating!

PR #1686, opened 2017-04-24
Ignore node_modules while compiling queries

MSNexploder:
This prevents nonintuitive errors like `'createRefetchContainer' expects a second argument of fragment definitions.` when using the project root (including node_modules) as source directory.

(fixes #1635)

PR #1921, opened 2017-06-27
add include/exclude options to compiler

robrichard:
It would be nice to be a little more flexible with the search paths for `relay-compiler`. As of https://github.com/facebook/relay/pull/1686 `relay-compiler` ignores node_modules. We have several Relay apps that share common components through node_modules and need a way to tell the compiler which ones to search. This change will let us add several more specific search paths and override the setting to ignore node_modules.

I was careful to do this in a way that is not a breaking change.

### DIFF
--- a/packages/relay-compiler/bin/RelayCompilerBin.js
+++ b/packages/relay-compiler/bin/RelayCompilerBin.js
@@ -43,15 +43,17 @@ const {
 
 import type {GraphQLSchema} from 'graphql';
 
-function buildWatchExpression(options: {extensions: Array<string>}) {
+function buildWatchExpression(options: {
+  extensions: Array<string>,
+  include: Array<string>,
+  exclude: Array<string>,
+}) {
   return [
     'allof',
     ['type', 'f'],
     ['anyof', ...options.extensions.map(ext => ['suffix', ext])],
-    ['not', ['match', '**/node_modules/**', 'wholename']],
-    ['not', ['match', '**/__mocks__/**', 'wholename']],
-    ['not', ['match', '**/__tests__/**', 'wholename']],
-    ['not', ['match', '**/__generated__/**', 'wholename']],
+    ['anyof', ...options.include.map(include => ['match', include, 'wholename'])],
+    ...options.exclude.map(exclude => ['not', ['match', exclude, 'wholename']])
   ];
 }
 
@@ -61,6 +63,8 @@ async function run(options: {
   schema: string,
   src: string,
   extensions: Array<string>,
+  include: Array<string>,
+  exclude: Array<string>,
   verbose: boolean,
   watch?: ?boolean,
 }) {
@@ -197,6 +201,23 @@ const argv = yargs
     src: {
       describe: 'Root directory of application code',
       demandOption: true,
+      type: 'string',
+    },
+    include: {
+      array: true,
+      default: ['**'],
+      describe: 'directories to include under src',
+      type: 'string',
+    },
+    exclude: {
+      array: true,
+      default: [
+        '**/node_modules/**',
+        '**/__mocks__/**',
+        '**/__tests__/**',
+        '**/__generated__/**',
+      ],
+      describe: 'directories to ignore under src',
       type: 'string',
     },
     extensions: {


robrichard:
@kassens @JenniferWang @yuzhi any thoughts on this?

kassens:
Seems good for now, but this is getting a bit complex. Hopefully we can move to `.graphqlconfig` or similar soon: https://github.com/graphcool/graphql-config/issues/16

rosskevin:
@robrichard - I'm confused, what is the correct way to use this so that shared modules' queries are included?  The moment I specify any `include`, this command fails to generate _anything_:

```
relay-compiler --src ./src --schema ./schema.json --include ./node_modules/my-shared-module
```

Do I also need to include the `./src` in the `include` list?

robrichard:
The options are globs so you need to include the `*`'s. Also every --include option needs to nested under the `src` option.
```
relay-compiler --schema ./schema.json --src . --include "src/**" "node_modules/my-shared-module/src/**" "node_modules/my-other-shared-module/src/**" --exclude "**/__generated__/**"
```

rosskevin:
Thanks @robrichard.
So this seems like a really hairy way to reuse a shared module, but it did work.  

My environment:
- transpiled ES5 files
- flow shadow files (`js.flow` files identical to source)

```
relay-compiler --src . \
--schema ./schema.json \
--verbose \
--include 'src/**' 'node_modules/my-module/**' \
--exclude '**/__generated__/**' 'node_modules/my-module/node_modules/**' \
--extensions 'js.flow' 'js'
```

Important points here for those that encounter this:
1. the `src` context needs to be `.` so that all directories fall under it.
1. the node module **must** include the original source.  In my case I have `js.flow`, which is original source, so I could use them by broadening extensions.
2. for every node module you include, you need to exclude it's `node_modules` folder so it doesn't find problems. One such would be a test in `react-relay`.

In the end, it seems like shared modules, while possible, is not very easy to get right.  Perhaps most don't want to use shared modules, but if you do, please let everyone know that @robrichard and I aren't the only two on the planet trying to reuse.

zth:
@rosskevin @robrichard is this still the way you're using shared packages with `relay-compiler`? I got it working using your elaborate instruction @rosskevin (thanks!) but I got to agree with you, it seems a bit... non-ergonomic ;).

rosskevin:
@zth we made the switch to apollo a few months ago.

zth:
@rosskevin Ahh I see! Any particular reason, if I may ask?

rosskevin:
@zth We had a significant investment in flow (type) and came to a dead end that caused us some pain.  We took a step back and examined facebook's management of open source projects including this one, and noted the ones that were less community driven e.g. open source but primarily only by facebook employees for facebook endeavors. We looked at github metrics and determined that it was in our interest to invest in open source in a place we were confident that we could have our issues heard, we had the skills to contribute/affect change, and our (reasonable) PRs accepted; even if it wasn't something that facebook would use internally.

While this is a criticism of facebook's management of _some_ open source projects, on the other hand, it is _entirely_ their prerogative.  I'm thankful for their pioneering work and their contribution to open source - without it, we wouldn't be using react and even apollo.

Simply put, we had to make a decision that was best for us, and I think facebook is making decisions that are best for them.  Things may have changed, they always do, so please evaluate the best fit for yourself.

robrichard:
@zth we are still using using relay with shared modules this way, with a few modifications. 
* we put all the code for the module in a top level `src/**` so we don’t need to explicitly exclude the nested node_modules directory. 
* we add an indicator to the modules’ package.json so we can automatically figure out which dependencies contain relay queries or fragments.

helielson:
@rosskevin 
> Perhaps most don't want to use shared modules, but if you do, please let everyone know that @robrichard and I aren't the only two on the planet trying to reuse.

You aren't alone, I'm also using shared modules. [Your answer](https://github.com/facebook/relay/pull/1921#issuecomment-326083890) worked like a charm. 
Thanks.

sorenhoyer:
We plan on using them as well for our new platform, where we'll have many react applications sharing relay components, so glad I found this.

@robrichard I'd be interested in hearing more about how you added _"an indicator to the modules’ package.json so we can automatically figure out which dependencies contain relay queries or fragments."_ and how you handle it.

robrichard:
@sorenhoyer the gist of it is

each package has a package.json field like
```
dibs: {
    relay: ["src/**"]
}
```

then a script reads every package.json in node_modules and builds up a list args to pass to --include:

```
[
    "node_modules/my-package-a/src/**,
    "node_modules/my-package-a/node_modules/my-package-b/src/**,
    "node_modules/my-package-c/src/**,
]
```

sorenhoyer:
Ahh okay. An alternative way I've used is just to prefix all my relay modules with `relay-`. Then you can just do `--include src/** node_modules/relay-*/**`.

But actually I have yet to find a good way of working with "shared relay modules". Because while developing the relay package, it's not really super handy having to compile your package from the consuming application.

So you probably want to have relay-compiler and react-relay as devDependencies anyways, to be able to render, test and build your package outside your application - at least I use the generated relay ts (same with flow) types in my components. Unfortunately this means you also need to have a subset of your schema in that package.

The only way around this issue with the partal schema is, yes, to compile your relay packages from your consuming application - also while developing that package. But really, why compile from the application and not the package itself? If the schema changes, you will need to change the components in the package anyways... so this whole compiling node_modules seems doesn't make a whole lot of sense for me in the end.

jahead:
How do we do this kind of work now that include is gone in v13/14? I'm finding it hard to find documentation on this? I seen @sibelius comment on the older issues. 

I'm looking to spread a fragment from a ui component from a module into a application's query

PR #1922, opened 2017-06-28
Update commitMutation docs to make `optimisticResponse` an object

Arthelon:
This change updates the documentation to reflect the change made in ed5f4f9d3a575236f503778915dcbbf19c8e2440 to the `optimisticResponse` field in the mutation config


leebyron:
Thanks!

Issue #1755, opened 2017-05-12
Relay schema path is read relative to process.cwd()


kassens:
If the schema path in the babel config is a simple string we load this path relative to the `process.cwd()`. This doesn't seem the best solution someone might execute babel from any random directory.

Ideally, the path would be relative to the config that defines it, but I couldn't find a way to figure that out since it could come from a `.babelrc`, the babel key in a `package.json` or passed directly to the babel API.

Is anyone aware of solutions to this?

PR #1933, opened 2017-07-02
Look for `compat` schema file relative to Babel source root

BinaryMuse:
Currently the schema specified in the Babel config is opened relative to the current working directory; this changes the behavior so that if the file is not found, we additionally look relative to Babel's ~module~ source root.

Fixes facebook/relay#1755


BinaryMuse:
I believe this is ready for some 👀

kassens:
Sorry for the delay, we had a busy week.

Not a babel expert, but this seems fine as it's just looking in an additional "fallback" directory. The only "risk" here would be to have the schema in both directories, but that seems fairly small.

PR #1952, opened 2017-07-08
[Relay-Compiler] Rename generic/core, add alias modules to prepare for split

Yan-J:
Rename modules in relay-compiler/generic as they're assumably used for general purpose, not limited to Relay. Add alias modules for some files as there's still some relay-spec part needed to be factor out.

The goal here is to split the current "relay-compiler" into two parts: a generic compiler and a relay-specific wrapper.

This is a change that we prepared in our internal SCM originally, but we wanted to test it out on the GitHub repo to be sure that Travis is still green with it.)


pranaygp:
@Yan-J Are you still using this to test against Travis? 

If so, you'll want to rebase onto master once #1993 gets merged.
If not, let's close it

Yan-J:
@pranaygp I'll close this PR. Thanks!

Issue #1667, opened 2017-04-21
[Modern] babel-plugin-relay not dealing w/ different query-type name


femesq:
I'm using [graphene](https://github.com/graphql-python/graphene) on Django, that can generate `schema.json`. Converting it to `schema.graphql` using [this script](https://github.com/facebook/relay/issues/1639#issuecomment-295314954) works fine.

Results:
```
schema {
  query: AllQueries
  mutation: AllMutations
}

type AllMutations {
  ...
}

type AllQueries {
  ...
}

```

Trying to convert to Relay-Modern rises this error:
```
-- Relay Transform Error -- AppPrivate --

File:  xxx/app/scripts/containers/AppPrivate.js
Error: TypeError: Cannot read property 'toString' of undefined
    at RelayQLType.getName (xxx/node_modules/babel-plugin-relay/lib/RelayQLAST.js:482:86)
    at RelayQLType.mayImplement (xxx/node_modules/babel-plugin-relay/lib/RelayQLAST.js:632:17)
    at shouldGenerateIdFragment (xxx/node_modules/babel-plugin-relay/lib/RelayQLPrinter.js:569:27)
...
```
Line `babel-plugin-relay/lib/RelayQLAST.js:482` corresponds to [this](https://github.com/facebook/relay/blob/master/packages/babel-plugin-relay/RelayQLAST.js#L525).

Changing `schema.graphql` by hand to this:
```
// Removed schema definition

type AllMutations {
  ...
}

type Query {  //  <= HERE
  ...
}

```

Solved the problem...

I'm using `babel-plugin-relay` w/ `compat: true` config.
All Relay packages are on `rc2`. The problem already occurred on `rc1`...

tlvenn:
Looks pretty similar with this https://github.com/facebook/relay/issues/1624

eugene1g:
Had the same issue, but only with `compat: true`

skosch:
To clarify, until this is fixed, you'll have to replace all instances of `AllQueries` in your schema.graphql (or `RootQueryType`, or whatever yours is called) with the `Query` from [here](https://github.com/facebook/relay/blob/master/packages/react-relay/classic/store/RelayStoreConstants.js#L25)

robrichard:
I opened a PR to address this issue: https://github.com/facebook/relay/pull/1962

Issue #1624, opened 2017-04-18
Relay Modern expects query type to be named Query


kassens:
We should also support other names such as `Root`.

leebyron:
Might blame to https://github.com/facebook/relay/blob/master/packages/react-relay/classic/store/RelayStoreConstants.js#L25

robrichard:
I opened a PR to address this issue: https://github.com/facebook/relay/pull/1962

leebyron:
Glad to see this fixed!

PR #1962, opened 2017-07-13
Don't break in compat mode when root field is not named Query

robrichard:
Fixes https://github.com/facebook/relay/issues/1667 and https://github.com/facebook/relay/issues/1624

I tested using this branch of the example repo: https://github.com/relayjs/relay-examples/compare/master...robrichard:custom-query-name?expand=1

@kassens


robrichard:
@kassens do you think it would be possible to get this into the next release?

leebyron:
Thanks for this excellent improvement, @robrichard! I'll work on landing this change and prepping it for the next release

Issue #1644, opened 2017-04-19
Make Watchman a non-hard dependency


kassens:
Not sure where a good place for this would be in the docs – probably the compiler section.

See #1626 for @unirey hitting this issue.

ivome:
It does not seem to be limited to windows, I get the same message on OS X 10.10.5

kassens:
Updated the title to reflect that.

igl:
Why is it required for a static build that does not `--watch`?
Jenkins is not happy.

wincent:
> Why is it required for a static build that does not `--watch`?

Good question. At the moment, I think the answer may be "because nobody has taken the taken the time to isolate the parts of the code that depend on Watchman". It probably would be interesting to do so, as having a small dependency footprint is a desirable property.

taion:
I think that's the bigger/more surprising thing. It's odd to require watchman for e.g. CI environments where:
- There's no watching going on, and
- watchman is a bit annoying to install

wincent:
Yep, I agree with you. Like I said [over here](https://github.com/relayjs/relay-examples/issues/23#issuecomment-305343527):

> I'd like to see us gracefully fallback to a non-watchman implementation when Watchman is not available (and especially when not running under --watch mode).

We'd definitely welcome a pull-request that addressed this.

wincent:
Retitled this to reflect the desired solution, so removing the "docs" label, although we'd still welcome clearer docs in the meantime.

ryancole:
Would this be considered rather simple to do, or are there a some gotchas involved? I have some free time and could try my best to provide a pull request. If it's going to be a bit complicated though I would probably just let somebody else with more knowledge of the internals take a shot at it.

wincent:
@ryancole: I don't think it is likely to be particularly difficult, but few things are ever as easy as you think they will...

If you wanted to look at the code to gauge the complexity of it I'd suggest starting here:

https://github.com/facebook/relay/blob/8f719e15a541dc8964892210716c329687cc4d7f/packages/relay-compiler/bin/RelayCompilerBin.js#L45-L55

That's the watch expression that the current system passes through to Watchman. To break the dependency, we need to implement simple code that knows how to identify all the files matched by that description. Doesn't look too bad, to be honest.

henryqdineen:
👍  but I am wondering what is the suggested way around this now. Are people just installing watchman as part of the CI build?

taion:
[...]

PR #1966, opened 2017-07-15
don’t use watchman when not watching

robrichard:
Fixes https://github.com/facebook/relay/issues/1644

This PR will update relay-compiler to not use watchman unless watching files.

### DIFF
--- a/packages/relay-compiler/bin/RelayCompilerBin.js
+++ b/packages/relay-compiler/bin/RelayCompilerBin.js
@@ -20,11 +20,13 @@ const RelayConsoleReporter = require('RelayConsoleReporter');
 const RelayFileIRParser = require('RelayFileIRParser');
 const RelayFileWriter = require('RelayFileWriter');
 const RelayIRTransforms = require('RelayIRTransforms');
+const RelayWatchmanClient = require('RelayWatchmanClient');
 
 const formatGeneratedModule = require('formatGeneratedModule');
 const fs = require('fs');
 const path = require('path');
 const yargs = require('yargs');
+const glob = require('fast-glob');
 
 const {
   buildASTSchema,
@@ -42,6 +44,7 @@ const {
 } = RelayIRTransforms;
 
 import type {GraphQLSchema} from 'graphql';
+import type {WatchmanChange} from 'RelayCodegenWatcher';
 
 function buildWatchExpression(options: {
   extensions: Array<string>,
@@ -60,6 +63,23 @@ function buildWatchExpression(options: {
   ];
 }
 
+async function getFilesFromGlob(baseDir, options: {
+  extensions: Array<string>,
+  include: Array<string>,
+  exclude: Array<string>
+}): Promise<Array<WatchmanChange>> {
+  const {extensions, include, exclude} = options;
+  const patterns = include.map(inc => `${inc}/*.+(${extensions.join('|')})`);
+
+  return await glob(patterns, {
+    cwd: baseDir,
+    bashNative: [],
+    onlyFiles: true,
+    ignore: exclude,
+    transform: name => ({name, exists: true}),
+  });
+}
+
 /* eslint-disable no-console-disallow */
 
 async function run(options: {
@@ -69,6 +89,7 @@ async function run(options: {
   include: Array<string>,
   exclude: Array<string>,
   verbose: boolean,
+  watchman: boolean,
   watch?: ?boolean,
 }) {
   const schemaPath = path.resolve(process.cwd(), options.schema);
@@ -78,6 +99,9 @@ async function run(options: {
   const srcDir = path.resolve(process.cwd(), options.src);
   if (!fs.existsSync(srcDir)) {
     throw new Error(`--source path does not exist: ${srcDir}.`);
+  }
+  if (options.watch && !options.watchman) {
+    throw new Error('Watchman is required to watch for changes.')
   }
   if (options.watch && !hasWatchmanRootFile(srcDir)) {
     throw new Error(
@@ -96,13 +120,20 @@ Ensure that one such file exists in ${srcDir} or its parents.
 
   const reporter = new RelayConsoleReporter({verbose: options.verbose});
 
+  let client = null;
+  if (options.watchman) {
+    client = await RelayWatchmanClient.createIfAvailable();
+  }
+
   const parserConfigs = {
     default: {
       baseDir: srcDir,
       getFileFilter: RelayFileIRParser.getFileFilter,
       getParser: RelayFileIRParser.getParser,
       getSchema: () => getSchema(schemaPath),
-      watchmanExpression: buildWatchExpression(options),
+      watchmanExpression: client ? buildWatchExpression(options) : null,
+      files: client ? null : await getFilesFromGlob(srcDir, options),
+      client,
     },
   };
   const writerConfigs = {
@@ -233,6 +264,11 @@ const argv = yargs
     verbose: {
       describe: 'More verbose logging',
       type: 'boolean',
+    },
+    'watchman': {
+      describe: 'Use watchman when not in watch mode',
+      type: 'boolean',
+      default: true
     },
     watch: {
       describe: 'If specified, watches files and regenerates on changes',
--- a/packages/relay-compiler/generic/codegen/RelayCodegenRunner.js
+++ b/packages/relay-compiler/generic/codegen/RelayCodegenRunner.js
@@ -25,7 +25,7 @@ import type FileParser from 'FileParser';
 import type {CompileResult} from 'RelayCodegenTypes';
 import type {File} from 'RelayCodegenTypes';
 import type {FileWriterInterface} from 'RelayCodegenTypes';
-import type {FileFilter, WatchmanExpression} from 'RelayCodegenWatcher';
+import type {FileFilter, WatchmanExpression, WatchmanChange} from 'RelayCodegenWatcher';
 import type {RelayReporter} from 'RelayReporter';
 import type {DocumentNode, GraphQLSchema} from 'graphql';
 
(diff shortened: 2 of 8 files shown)


robrichard:
@wincent @kassens let me know if you think this is the right approach?

robrichard:
The failing check here is also failing on master

wincent:
> let me know if you think this is the right approach?

Yeah this looks pretty good, although I would suggest first trying to use Watchman even in the non-watch code-path (it will be much faster on a very large repo) before falling back to a non-Watchman alternative.

robrichard:
I actually found `fast-glob` to be about 25% faster than watchman on my projects, but I can switch it if you prefer.

sibelius:
can we have this as an option?

so we can use `fast-glob` or `watchman` based on this

wincent:
> I actually found `fast-glob` to be about 25% faster than watchman on my projects, but I can switch it if you prefer.

How large are your projects? Watchman was built to support repos with well over a million files in them: in these large scenarios, querying an always-up-to-date, in-memory index that has been prepared ahead of time is always going to be faster than actually stating the filesystem, even when the OS's filesystem caches are hot.

robrichard:
@wincent @sibelius I pushed another commit that uses watchman by default to query files. It will fall back to `fast-glob` if either watchman is not installed or the `--noWatchman` option is passed.

js-n:
we got this working on Travis by adding:
```yml
before_install:
  # dirty ugly watchman hack https://github.com/facebook/relay/issues/1644#issuecomment-315998313
  - if [[ ! -e watchman ]]; then git clone https://github.com/facebook/watchman.git && cd watchman/ && git checkout v4.7.0 && ./autogen.sh && ./configure && make  && sudo make install; fi
  - sudo sysctl fs.inotify.max_user_watches=524288
  - sudo sysctl -p
  # end dirty watchman hack
```

packages/relay-compiler/bin/RelayCompilerBin.js
```diff
@@ -234,6 +221,10 @@ const argv = yargs
       describe: 'More verbose logging',
       type: 'boolean',
     },
+    'noWatchman': {
```
leebyron: Let's avoid double-negative options. How about just a `watchman` flag which defaults to true, so `--watchman=false` to disable?

packages/relay-compiler/generic/codegen/RelayCodegenWatcher.js
```diff
@@ -35,23 +44,52 @@ type WatchmanChanges = {
...
-  ]);
-  const resp = await client.command('query', watchResp.root, {
-    expression,
-    fields: fields,
-    relative_root: watchResp.relativePath,
-  });
-  client.end();
-  return updateFiles(new Set(), baseDir, filter, resp.files);
+  const {baseDir, include, extensions, exclude} = filterOptions;
+  if (
```
leebyron: Let's refactor this into two functions since this is getting a bit complicated. How about `queryFilesWithWatchman` and `queryFilesWithGlob` and you can call them from this condition?

packages/relay-compiler/generic/core/RelayWatchmanClient.js
```diff
@@ -17,6 +17,20 @@ const watchman = require('fb-watchman');
 class RelayWatchmanClient {
   _client: any;
 
+  static isAvailable(): Promise<boolean> {
+    return new Promise((resolve) => {
+      const client = new RelayWatchmanClient();
+      client.on('error', () => {
+        resolve(false);
+      });
```
leebyron: minor note: this could be syntactically simplified slightly to: 

```
client.on('error', () => resolve(false));
```

packages/relay-compiler/generic/core/RelayWatchmanClient.js
```diff
@@ -17,6 +17,20 @@ const watchman = require('fb-watchman');
...
+      client.hasCapability('relative_root')
+        .then(() => {
+          client.end();
```
leebyron: Should we really be creating a client, and then ending it like this? Only to create it again in this success case?

Perhaps a better API would be something like `createIfAvailable()` which promises to resolve to an instance of RelayWatchmanClient if "available" (has a relative_root) or otherwise returns null, which could then be detected in the code above

robrichard:
@leebyron thanks for taking the time to review. I pushed up another commit to address your comments. Please let me know if there is anything else you think I should change.

packages/relay-compiler/generic/core/RelayWatchmanClient.js
```diff
@@ -17,6 +17,15 @@ const watchman = require('fb-watchman');
 class RelayWatchmanClient {
   _client: any;
 
+  static createIfAvailable(): Promise<?RelayWatchmanClient> {
+    return new Promise((resolve) => {
+      const client = new RelayWatchmanClient();
+      client.on('error', () => resolve(null));
+      client.hasCapability('relative_root')
+        .then(() => resolve(client));
```
leebyron: Right now this is always resolving to the client, regardless of the return value of relative_root - is that intended? Do you mean for this to be

```js
.then(hasRelativeRoot => hasRelativeRoot ? client : null)
```

packages/relay-compiler/bin/RelayCompilerBin.js
```diff
@@ -221,9 +221,10 @@ const argv = yargs
       describe: 'More verbose logging',
       type: 'boolean',
     },
-    'noWatchman': {
-      describe: 'Do not use watchman when not in watch mode',
+    'watchman': {
+      describe: 'Use watchman when not in watch mode',
```
antmdvs: I'm confused. Is watchman still used when *not* in watch mode?

packages/relay-compiler/generic/core/RelayWatchmanClient.js
```diff
@@ -17,6 +17,15 @@ const watchman = require('fb-watchman');
 class RelayWatchmanClient {
   _client: any;
 
+  static createIfAvailable(): Promise<?RelayWatchmanClient> {
+    return new Promise((resolve) => {
+      const client = new RelayWatchmanClient();
+      client.on('error', () => resolve(null));
```
leebyron: I'm curious if there are certain kinds of errors which should cause this to be determined as "unavailable" as opposed to considering all errors? I'm interested in what you saw during investigating writing this PR that led you to this

packages/relay-compiler/bin/RelayCompilerBin.js
robrichard: Watchman will always be used in watch mode. Watchman will also still be used when not in watch mode unless `--watchman=false` is passed or we detect watchman is not installed.

leebyron: I agree that it's confusing that you could specify that watchman shouldn't be used but also that you want to use watch mode. I think in that scenario it's better to stop and report an error instead of disregarding wishes to not use watchman

packages/relay-compiler/generic/core/RelayWatchmanClient.js
robrichard: I based it on the [node watchman docs here](https://facebook.github.io/watchman/docs/nodejs.html#checking-for-watchman-availability) where they check capabilities for `relative_root`. 

Invoking `client.hasCapability` when watchman is not installed will cause the `on('error')` callback to be invoked (or a `Unhandled 'error' event` to be thrown if not implemented) and never execute the function in the `then` callback. I'm not able to catch this error by chaining a `.catch` or rewriting as an async function and wrapping in a `try/catch`. I'm investigating why that's the case now. 

I could check the error in `on('error')` to make sure it's `ENOENT` before resolving `null`.

packages/relay-compiler/bin/RelayCompilerBin.js
robrichard: @leebyron I added this check now

packages/relay-compiler/generic/core/RelayWatchmanClient.js
robrichard: @leebyron It seems like `fb-watchman` does not work as documented in the link above. The only way I found to handle the error from watchman not being installed is through the `on('error')` callback. I opened an issue on the watchman repo with a minimal example (https://github.com/facebook/watchman/issues/509). How do you think I should proceed?

leebyron: Let's try out your current implementation and adjust from there if anything goes wrong. Thanks for opening the issue on watchman, hopefully someone who knows more about it than I do can help us out.

packages/relay-compiler/generic/codegen/RelayCodegenRunner.js
```diff
@@ -36,7 +36,10 @@ export type ParserConfig = {|
   getFileFilter?: (baseDir: string) => FileFilter,
   getParser: (baseDir: string) => FileParser,
   getSchema: () => GraphQLSchema,
-  watchmanExpression: WatchmanExpression,
+  extensions: Array<string>,
+  include: Array<string>,
+  exclude: Array<string>,
+  watchman: boolean,
```
leebyron: I was working on merging this, but unfortunately this breaking change will prevent me from doing so.

This `RelayCodegenRunner` is public API, and it's used extensively by Relay projects internally at Facebook which provide complex `watchmanExpression` that cannot be replicated simply with include/exclude.

packages/relay-compiler/generic/codegen/RelayCodegenRunner.js
```diff
@@ -165,7 +168,7 @@ class RelayCodegenRunner {
         const query = {
           expression: [
             'allof',
-            config.watchmanExpression,
+            RelayCodegenWatcher.buildWatchExpression(config),
```
leebyron: I think this change here won't be enough - if watchman is disabled, then calling `getDirtyWriters` will fail here.

robrichard:
@leebyron can you take another look? I updated the PR to incorporate your suggestions. There are no longer any breaking changes to any modules.

leebyron:
This is really nice, thanks @robrichard - let me work on merging this

A-Tokyo:
I am facing this again as of relay-compiler@2.0.0 :/

Is there a plan to not require watchman if the flag is not passed with relay-compiler@2.0? 

Thanks a lot!

PR #1970, opened 2017-07-16
Add warning if still using buildCommand in RelayFileWriter config

AGS-:
@kassens Adding this since it got changed from 1.0.0 to 1.1.0

### DIFF
--- a/packages/relay-compiler/relay/codegen/RelayFileWriter.js
+++ b/packages/relay-compiler/relay/codegen/RelayFileWriter.js
@@ -77,11 +77,7 @@ class RelayFileWriter implements FileWriterInterface {
     this._documents = documents;
     this._onlyValidate = onlyValidate;
 
-    warning(
-      !this._config.buildCommand,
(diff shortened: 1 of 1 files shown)


AGS-:
I see that the test isn't passing because flow obviously doesn't like that I'm checking for a key that isn't defined in the type. Since buildCommand is deprecated it doesn't really make sense to add it as an optional in the type right?

AGS-:
I don't know what your thoughts are on maybe adding buildCommand back in as optional, warning the user that it's deprecated and replaced by formatModule, and maybe provide a sensible default FormatModule if they haven't provided it. This way it won't break if people upgrade.

kassens:
Maybe create a new `function validateConfig(config: Object): void { /* ... */ }` where you just type as Object and can now check any legacy fields.

AGS-:
Alright, sounds good. I'll update the PR in a bit

AGS-:
@kassens updated the PR

PR #1993, opened 2017-07-25
Upgrade fbjs and flow-bin and fix type errors

pranaygp:
We've been failing travis builds since 68241f7c9126b6c44646ffb078df63c2e3810d43 due to a bunch of flow errors. I'm making this PR so we can fix those errors :)


pranaygp:
https://github.com/facebook/fbjs/pull/258 should fix the remaining issues Travis is complaining about

Issue #2000, opened 2017-07-28
Use RefetchContainer to implement pull-to-refresh without variable change


jhalborg:
I'm trying to find out how/if it is possible to trigger a refresh in a Relay Modern RefreshContainer without passing (new) variables?

I’m looking for a way to implement the good ol’ pull-to-refresh on a React Native list, that should simply refetch the original query - no variables needed?

According to docs (https://facebook.github.io/relay/docs/api-cheatsheet.html) this should be possible using

`this.props.relay.refetch({}, callback, {force: true})`

but I get an error saying `"undefined is not an object ('evaluating taggedNode.modern')"` 

Is it not possible to force a refetch without a variable change? If not, consider this a feature request :-)

josephsavona:
It looks like Relay can't find the query for your refetch container. Did you define one?

jhalborg:
Yes :) I should have mentioned, of course, that the fragment and query works just fine if I use a fragmentContainer. 

`export default createFragmentContainer(MyComponent, graphql 'fragment MyComponent_entries on Viewer { ... } ' );`

If I change to refetch-container and use `this.props.relay.refetch({}, callback, {force: true})` when pull-to-refresh is activated, I get the error mentioned.

josephsavona:
Note that createRefetchContainer requires a third argument: a query to use when refetching the fragment data. See the docs at https://facebook.github.io/relay/docs/refetch-container.html

zetavg:
It seems that the arguments of `this.props.relay.refetch` has been change to `refetch(refetchVariables, renderVariables, callback, options)` (in https://facebook.github.io/relay/docs/refetch-container.html). I'm not sure that in which version this has change, but you can give it a try and change your code to `this.props.relay.refetch({}, null, callback, {force: true})`, it works for me :)

jhalborg:
Hey, 

Thanks for the suggestion :-) You are totally right, but unfortunately, it doesn't work for me :-/

Here's some more code for clarification. Note that I don't use variables in the query at all - does that matter?

```
class HistoryList extends React.PureComponent<void, Props, State> {
  state = { refreshing: false };

  _renderRow = ({ item }) => {
    return <HistoryListItem item={item.node} />;
  };

  _renderHeader = ({ section }) => {
    return (
      <Text style={[cs.breadText, _styles.sectionHeader]}>
        {section.title}
      </Text>
    );
  };

  _onRefresh = () => {
    this.setState({ refreshing: true });
[...]

PR #2005, opened 2017-08-01
Docs: Update APICheatsheet for this.props.relay.refetch

zetavg:
Seems that the arguments of `this.props.relay.refetch` in the API Cheatsheet is outdated and may confuse people (#2000). Just a small change on this 😃


leebyron:
Thanks!

PR #2010, opened 2017-08-02
relay-compiler hangs when watchman is installed

robrichard:
Hey @leebyron, I noticed after https://github.com/facebook/relay/pull/1966 was merged the code was amended to create a watchman client solely for purposes of testing if watchman is available. Unfortunately this client is never closed so the process will hang indefinitely.


leebyron:
Ah thank you for testing this further. Yeah, passing the client in had issues and I attempted to revert back to your previous version which just used the client to test. Let me work on merging this

robrichard:
rebased and resolved conflicts

PR #2015, opened 2017-08-04
[modern] ReactRelayQueryRenderer: Change readyState to getDefaultState value after retry

EuAndreh:
As exemplified in the [QueryRenderer documentation](https://github.com/facebook/relay/blob/b4b086a8ce0005d1d13af4a96112d5c6d87f3e4c/docs/modern/QueryRenderer.md#L33-L40), the `render` method receives `error` and `props` as arguments, and the component is supposed to show an error screen, a loading screen or the component based on those properties.

If `error` isn't `null`, show the error; if `props` isn't `null`, show the component; otherwise, show the loading.

However, after calling the `retry()` function, the `error` argument isn't passed again as `null`. This way, the component can't know if it is loading or the same error has just occurred again.

With this patch, `retry()` changes the value of `readyState` to the same as the initial value: all values set to `null`, telling the component that it is loading.

### DIFF
--- a/packages/react-relay/modern/ReactRelayQueryRenderer.js
+++ b/packages/react-relay/modern/ReactRelayQueryRenderer.js
@@ -205,6 +205,8 @@ class ReactRelayQueryRenderer extends React.Component {
         error,
         props: null,
         retry: () => {
+          const readyState = getDefaultState();
+          this.setState({readyState});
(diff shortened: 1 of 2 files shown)


EuAndreh:
I'm talking with my employer now, and the corporate CLA should be signed soon.

PR #2019, opened 2017-08-05
[Modern] Update refetchContainer example in documentation

scotmatson:
Perhaps I failed to see how refetchRender would be used for pagination, and you'll notice in this example I'm proposing that I did not use that here. 

The Relay documentation does not actively broadcast an easy solution to handling offset-based pagination as the PaginationContainer is specifically targeting the case of infinite cursor-based pagination. After much research and tinkering I've used the component state and a refetchContainer to implement offset-based pagination in Relay. For me this solution works very well.

I've taken the script I wrote for my application and modified it to utilize aspects of the existing example. The changes I've made simply make this a more complete implementation of a common case that others are likely to struggle to solve.

Additional changes possibly to consider would be not using ES6 syntax and changing naming conventions as FeedStories in itself does imply a more continuous pagination feed. But I won't get hung up on semantics here. I just hope the maintainers will consider this modification to the documentation.

### DIFF
--- a/docs/modern/RefetchContainer.md
+++ b/docs/modern/RefetchContainer.md
@@ -39,58 +39,97 @@ refetch: (
 ## Example
 
 ```javascript
-const {
-  createRefetchContainer,
-  graphql,
-} = require('react-relay');
+import React, { Component } from 'react';
+import PropType from 'prop-types';
+import { createRefetchContainer, graphql } from 'react-relay';
+import Stories from './Stories';
 
-class FeedStories extends React.Component {
+class FeedStories extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      offset: this.props.offset,
+      limit: this.props.limit,
+      total: this.props.viewer.stories.totalCount
+    };
+  }
+
+  _previousPage = () => {
+    if (this.state.offset > 0) {
+      this.setState({ offset: this.state.offset - this.state.limit }, () => {
+        this._loadMore();
+      });
+    }
+  }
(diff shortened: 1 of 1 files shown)


docs/modern/RefetchContainer.md
```diff
@@ -85,12 +125,12 @@ module.exports = createRefetchContainer(
...
-    `
+  `,
   },
   graphql.experimental`
-    query FeedStoriesRefetchQuery($count: Int) {
-      feed {
-        ...FeedStories_feed @arguments(count: $count)
+    query FeedStoriesQuery($offset: Int, $limit: Int) {
+      viewer {
+        ...FeedStories_viewer @arguments(offset: $offset, limit: $limit)
```
leebyron: We discourage using offset/limit style APIs, so I'd like to avoid using them in our documentation if possible.

alloy:
@leebyron’s suggestion does sound like a good one. @scotmatson would you perhaps still like to take a stab at that? Otherwise it might be time to close this as-is?

jstejada:
I'm going to close this one given that there's new docs for refetch container. @scotmatson if you want to improve those feel free to submit a new PR! thanks!

scotmatson:
@jstejada sounds good, thanks for the update. Will take a look when I have a chance (wife currently in labor!). Probably need to revisit some of my code to see how things have evolved since I originally implemented everything.

alloy:
@scotmatson Congrats on the baby! 🎉

PR #2020, opened 2017-08-05
Warning when multiple definitions found

robrichard:
@leebyron this is another issue that my team often runs into. Currently, if there is more than one fragment, query, or mutation with the same name, `relay-compiler` will continue silently and only use the last one it finds. This PR adds a warning message when duplicates are detected.

Tested with this branch: https://github.com/relayjs/relay-examples/compare/master...robrichard:multiple-definitions?expand=1

```
yarn run v0.24.5
$ relay-compiler --src ./js/ --schema ./data/schema.graphql 
Parsed default in 0.01s

Writing default
Warning: Multiple definitions found for AddTodoMutation
  mutations/AddTodoMutation.js
  a/AddTodoMutation.js
Warning: Multiple definitions found for TodoApp_viewer
  a/TodoApp.js
  components/TodoApp.js
Writer time: 0.23s [0.20s compiling, 0.03s generating, 0.00s extra]
Unchanged: 12 files
Written default in 0.25s
```

### DIFF
--- a/packages/relay-compiler/graphql-compiler/core/ASTConvert.js
+++ b/packages/relay-compiler/graphql-compiler/core/ASTConvert.js
@@ -37,6 +37,18 @@ type TransformFn = (
   definition: ASTDefinitionNode,
 ) => Root | Fragment;
 
+function throwMultipleDefinitionError(...definitions: Array<ASTDefinitionNode>) {
+  const msg = [
+    `Multiple definitions found`,
+  ];
+  definitions.forEach(definition => {
+    if (definition.name && definition.loc) {
+      msg.push(`  ${definition.name.value}: ${definition.loc.source.name}`);
+    }
+  });
+  throw new Error(msg.join('\n'));
+}
+
 function convertASTDocuments(
   schema: GraphQLSchema,
   documents: Array<DocumentNode>,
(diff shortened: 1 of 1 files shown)


leebyron:
This seems like it should be a validation step which causes the compile to fail rather than a file-writer warning, however this is better than nothing! I agree that being silent about this is not a good thing!

leebyron:
Actually, @robrichard, would you mind making this a fatal error? Builds should fail in this circumstances.

robrichard:
@leebyron updated to a fatal error.

leebyron:
Thanks for that, @robrichard 

I'm also curious how we the compiler even gets this far, considering that [validation should catch this](https://github.com/facebook/relay/blob/master/packages/relay-compiler/graphql-compiler/core/GraphQLValidator.js#L32). I'm a bit concerned that this is attempting to treat a symptom where there's a deeper problem at hand.

kassens:
@leebyron pretty sure we first collect a map of name => ast somewhere before it even makes it to the classic validation. At that point the duplicate is already dropped.

robrichard:
@leebyron, @kassens is correct. Each definition is added to a `Map` keyed on the name [here](https://github.com/facebook/relay/blob/master/packages/relay-compiler/graphql-compiler/core/ASTConvert.js#L93). The GraphQLValidator is invoked a few lines later in [convertASTDefinitions](https://github.com/facebook/relay/blob/master/packages/relay-compiler/graphql-compiler/core/ASTConvert.js#L110).

There is also similar validation in [GraphQLCompilerContext](https://github.com/facebook/relay/blob/master/packages/relay-compiler/graphql-compiler/core/GraphQLCompilerContext.js#L63-L68), but this is also called after fragments are de-duped.

leebyron:
Should this check occur within convertASTDefinitions then? That code should be looking for an existing key the Map before adding the same key - at that point you have both conflicts and you have the ASTs, which means you can theoretically print a nicer error message with filenames and lines.

I'm hesitant about this approach that waits until file-write time to throw errors when we could catch it much sooner

kassens:
Yeah, that seems better. Do you want to take a stab at this @robrichard? It's probably enough to error on the first conflict and might make the code also less complex.

robrichard:
@kassens yes, I'll update this PR later this week

robrichard:
@leebyron @kassens ready for another review

alloy:
@robrichard It would be nice to get this one in. Care to do another rebase?

robrichard:
@alloy I'm going to close this as it looks like @kassens already implemented it in https://github.com/facebook/relay/commit/29dbb55ea23713a9a360c7dc680715096dc89522

alloy:
👍

Issue #1790, opened 2017-05-22
[modern][relay-compiler] Unable to parse graphql literals when the call to the container factory is not in same module


casparrolfe:
An extremely minor issue and only tried this with `createFragmentContainer` so far:

`relay-compiler` can't cope with the the container factory function call not living alongside the graphql literal. For example, we attempted to create a wrapper to make the API a little more friendly when it came to HOC composition:

```javascript
function createFragmentContainer(fragmentSpec) {
  return (Component) => Relay.createFragmentContainer(Component, fragmentSpec)
}

const withFragmentContainer = createFragmentContainer(
  graphql`
    fragment Module_prop on SomeType {
      id
    }
  `
)

export default compose(
  withRouter,
  // ...
  withFragmentContainer,
)(SomeComponent)
```

This will work fine when the wrapped `createFragmentContainer` lives in the same module as the graphql literal, but as soon as its extracted for reuse, we see the following error:

```
TypeError: Cannot read property 'type' of undefined
    at CallExpression (/node_modules/relay-compiler/bin/relay-compiler:4581:21)
    at visit (/node_modules/relay-compiler/bin/relay-compiler:4748:6)
    at traverse (/node_modules/relay-compiler/bin/relay-compiler:4761:8)
    at visit (/node_modules/relay-compiler/bin/relay-compiler:4751:4)
    at /node_modules/relay-compiler/bin/relay-compiler:4765:12
    at Array.forEach (native)
    at traverse (/node_modules/relay-compiler/bin/relay-compiler:4763:13)
    at visit (/node_modules/relay-compiler/bin/relay-compiler:4751:4)
    at /node_modules/relay-compiler/bin/relay-compiler:4765:12
    at Array.forEach (native)
```

Thanks :)

`react-relay` v1.0.0
`relay-compiler` v1.0.0
`babel-plugin-relay` v1.0.1

rosskevin:
@casparrolfe is this the same or the same as my [SO question](https://stackoverflow.com/questions/44593410/ui-library-reuse-with-generated-relay-queries)?  I cannot relay-compile my apps that reuse a common ui library.

maikelrobier:
Same issue:
```js
import Relay from 'react-relay/compat'

const createFragmentContainer = (containerConfig: FragmentContainerConfig) =>
  (component: ReactClass<*>) => Relay.createFragmentContainer(component, containerConfig.fragments)

export default createFragmentContainer

// separate file

class MyComponent extends Component {...}

[...]

PR #2030, opened 2017-08-11
fixes decorators not working

### LINKED ISSUE #1910: [Modern] createFragmentContainer unable to detect fragment as variable
I was trying to do something like ...

```
const fragment = graphql`
  fragment Foo on Bar {
    ..some stuff
  }
`;

export default createFragmentContainer(NavbarAuthAuthenticated, fragment);
```

but that seems to fail. The relay compiler doesn't seem to detect that it's a fragment. Everything works when that fragment is passed directly to the function, though.

It'd be nice for cleanliness to be able to put fragments in variables. Although, I understand that this involve the babel plugin and such so perhaps there's some limitation there.

dsslimshaddy:
fixes #1910 #1790

### DIFF
--- a/packages/relay-compiler/codegen/FindGraphQLTags.js
+++ b/packages/relay-compiler/codegen/FindGraphQLTags.js
@@ -59,8 +59,6 @@ function find(
       const callee = node.callee;
       if (
         !(
-          (callee.type === 'Identifier' &&
-            CREATE_CONTAINER_FUNCTIONS[callee.name]) ||
           (callee.kind === 'MemberExpression' &&
             callee.object.type === 'Identifier' &&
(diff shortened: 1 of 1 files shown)


kassens:
Sorry for getting to this late. I'm not sure we should just remove the check. It might be safer to add a second condition for the decorator case. Would you mind adding a test case to https://github.com/facebook/relay/blob/master/packages/relay-compiler/codegen/__tests__/FindGraphQLTags-test.js?

alloy:
@jstejada Both issues that this refers to have been closed, so it might make sense to close this PR by now too.

jstejada:
sounds good, thanks

PR #2035, opened 2017-08-16
Add support for NullValue ast kind to RelayParser._transformValue()

nkohari:
Hi there! This patch is the result of debugging gatsbyjs/gatsby#1689. It seems like the switch statement inside of `RelayParser._transformValue()` was missing a case for `NullValue`, so queries which contained a `null` literal would throw an error.

Adding a case fixes the downstream issue that some folks have experienced with Gatsby, but I'm not sure if there's a better way to solve this problem.


nkohari:
Sure, here's a test that covers the change. Sorry about that. ;)

kassens:
Awesome, thanks!

Issue #2036, opened 2017-08-16
Relay compiler doesn't die with exit code 1 in case of error


steelbrain:
On the relay version 1.2.0-rc.1

If the error is outside of the reporter:
 - Process logs error and exit with non-zero code (correct)

If the error is inside the reporter:
  - Process logs and exits with with zero code (incorrect)

Relay should exit with code 1 even when the error is inside a reporter. This issue is making my testing scripts work incorrectly.

Fix is up in https://github.com/facebook/relay/pull/2037

steelbrain:
Fixed as of https://github.com/facebook/relay/commit/af35c307e16e613e4dee46dba040c0c990919b3c

PR #2037, opened 2017-08-16
Fix exit code for reporter errors

steelbrain:
Fixes https://github.com/facebook/relay/issues/2036

See bug for description of issue

This PR uses [`process.exitCode`](https://nodejs.org/api/process.html#process_process_exitcode) which doesn't immediately terminate the process but defines the code that it'll use when it exits at a later time

### DIFF
--- a/packages/relay-compiler/graphql-compiler/reporters/RelayConsoleReporter.js
+++ b/packages/relay-compiler/graphql-compiler/reporters/RelayConsoleReporter.js
@@ -25,6 +25,7 @@ class RelayConsoleReporter implements RelayReporter {
   }
 
   reportError(caughtLocation: string, error: Error): void {
+    process.exitCode = 1;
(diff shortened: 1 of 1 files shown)


kassens:
I don't think the console reporter should be responsible for setting the exit code. Instead we should make sure all errors are reported back up the call tree and the RelayCompilerBin exits with the expected return code.

Do you have a specific error case that's currently not exiting with an error code?

steelbrain:
I personally think that it's the **console** reporter, which means that direct APIs won't invoke it so it *can* change exit code. But I'll make the change

steelbrain:
Closing in favor of https://github.com/facebook/relay/commit/af35c307e16e613e4dee46dba040c0c990919b3c

kassens:
My thinking behind the reporters was that you could have a reporter with progress bar etc that might be used for interactive shells, another one  for verbose output to a log file, or one that logs statistics to a database. We have an instance of the latter to track build times over time for example.

PR #2038, opened 2017-08-17
Correct prop reference in PaginationContainer

mattleff:
If I'm understanding the query correctly:

```graphql
query: graphql`
    query FeedPaginationQuery(
    $count: Int!
    $cursor: String
    $orderby: String!
) {
    user {
        # You could reference the fragment defined previously.
        ...Feed_user
    }
}
```

...the prop should be `user` not `viewer`.


kassens:
Good catch, thanks for sending out the fix.

PR #2044, opened 2017-08-18
Don't hint about watch mode on CI

SimenB:
No reason to give hints about watch mode on CI


kassens:
Actually instead of taking on this extra (small) dependency, how about we'll hide the flag in `--validate` mode which is being added by #2050. This is what we do internally at Facebook.

SimenB:
👍

PR #2050, opened 2017-08-21
Add new CLI option: '--validate'

### LINKED ISSUE #2047: Expose onlyValidate option for relay-compiler
It's useful to be able to verify (in CI checks for instance) that relay-compiler's generated artifacts are up-to-date.  Ideally, this check is idempotent.

It seems like `RelayCodegenRunner` has the ability to behave this way, but no CLI option is exposed which supports it, yet.  I think that would be a great addition.

https://github.com/facebook/relay/blob/master/packages/relay-compiler/bin/RelayCompilerBin.js#L150

ajhyndman:
Fixes #2047

Usage looks like this:

```
$ relay-compiler --src ./src --schema ./schema.graphql --validate
Parsed default in 0.05s

Writing default
Writer time: 0.51s [0.46s compiling, 0.04s generating, 0.00s extra]
Out of date:
 - AppQuery.graphql.js
 - Module_data.graphql.js
 - AnotherModuleQuery.graphql.js
Written default in 0.54s
error Command failed with exit code 1.
```

I'd like to add a change to this PR to include documentation, but I'm not sure where to put it.  There are a handful of example usages of the CLI, but no enumeration of CLI arguments.

I guess I could put up a first version of such a section that includes only this option?

e.g.

```markdown
## CLI usage

`$ relay-compiler --src <source path> --schema <schema path> [...options]`

### Options

#### `--validate`

Looks for pending changes; if any are found, exits with 1 instead of writing to disk

```

https://facebook.github.io/relay/docs/relay-compiler.html

### DIFF
--- a/packages/relay-compiler/bin/RelayCompilerBin.js
+++ b/packages/relay-compiler/bin/RelayCompilerBin.js
@@ -93,6 +93,7 @@ async function run(options: {
   verbose: boolean,
   watchman: boolean,
   watch?: ?boolean,
+  validate: boolean,
 }) {
   const schemaPath = path.resolve(process.cwd(), options.schema);
   if (!fs.existsSync(schemaPath)) {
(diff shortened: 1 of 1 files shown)


kassens:
Thanks for exposing this on the OSS runner!

PR #2051, opened 2017-08-22
[relay-runtime] Unfreeze JSON object to allow reassignment

### LINKED ISSUE #2049: [Modern] bug when updating json scalar field
I am using  [JSON Scalar field](https://github.com/taion/graphql-type-json).

And getting error:
```
recycleNodesInto.js:48 Uncaught (in promise) TypeError: Cannot assign to read only property 'finishingGroupIds' of object '#<Object>'
    at recycleNodesInto.js:48
```

when relay store receives update from the mutation with new json scalar value, which has internally additional item in the array. When I reload the page - store interpret the json value just fine, so its just issue when replacing the old json value with new one.

This whole recycleNodesInto function already seems to be doing some magic for flow, so not sure where is actually the root of the problem. Maybe just some condition to avoid stepping into the json scalar field?

I could work on PR if you give me some hints how you think this should be addressed.

screenshot 2017-08-21 10 57 52

dbslone:
Fixes #2049 

The issue described in #2049 arises because the object being passed to recycleNodesInto is a frozen object. So `nextObject[key] = nextValue` will always fail.

This fix currently causes one of tests in `packages/react-relay/classic/legacy/store/__tests__/recycleNodesInto-test.js` to fail because it uses `===` comparison but the returned result is the same. Not sure if these tests are still being used since they don't appear to be included when running `npm test` from the root directory.


leebyron:
cc @kassens @josephsavona - can you help review this one? Looks like a real issue in Relay attempting to mutate a frozen record, however we should get a gut-check on whether this is the right fix.

dbslone:
@leebyron @kassens @josephsavona any updates on if this best strategy to resolve this issue?

josephsavona:
Is this still an issue on master? This has come up previously and we've already made several fixes to `recycleNodesInto` in order to avoid ending up mutating the previous input exactly because it could be frozen. I'm also concerned about the performance implications of this implementation - at a glance it would seem to incur quite a bit of copying items/properties.

jardakotesovec:
@josephsavona Just tried with 1.4.1 and also with master and still the same issue.

mike-marcacci:
I have another fix for this in #2193 which takes a different strategy.

dbslone:
Closing PR in favor of #2192

PR #2061, opened 2017-08-24
fix relay website build

Slava:
Got broken after React's createClass codemod


Slava:
@leebyron can you please regen the website and push it? I don't have commit bits to facebook/relay

PR #2062, opened 2017-08-26
Move parts of CONTRIBUTING to ISSUE_TEMPLATE

Gregoor:
I noticed that a lot of questions that are better suited for SO are coming in here. So maybe by using GitHub's `ISSUE_TEMPLATE` we can circumvent that.

Also I found it a bit tedious that one can't simply create a jsfiddle to recreate a problem with Relay, so I changed `relay-starter-kit` a bit to make it work with Glitch and included a link to it in the `ISSUE_TEMPLATE`.

Lmkwyt :)


leebyron:
Thanks!

PR #2064, opened 2017-08-27
Add @import directive support to remove restriction on globally unique fragment names

robrichard:
This adds support for a directive `@import` based off of @stubailo's proposal to the GraphQL spec https://github.com/facebook/graphql/issues/343#issuecomment-323367921.

```graphql
...MyOtherTodoApp_viewer @import(from: "./a/TodoApp" name: "TodoApp_viewer")
```

The approach I took was
* Append a hash of the file path to all definition names to ensure uniqueness among conflicting names. The full path and original name are added to the IR objects metadata.
* To avoid breaking changes, fragment spreads without an `@import` directive will be transformed to the generated name of the first matching fragment. Warnings are logged if there are more than one matching fragments.
* Fragment spreads with an `@import` directive will resolve the `from` argument using the node resolve algorithm and transformed to the generated name.
* There is no file crawling like webpack. An `@import` to a file outside of the provided watchman expression or file glob will not be resolved.

This still needs a lot of work, especially with tests and error handling but I would like to get some early feedback if possible. My main questions are
1. Would a change like this be considered by the Relay team?
1. Does this approach make sense?
1. What user provided options are needed (e.g. skip this entirely, customize generated fragment names, customize file resolution algorithm)?

I have been using this branch of todo-modern to test these changes: https://github.com/relayjs/relay-examples/compare/master...robrichard:multiple-definitions2?expand=1#diff-6a54bd9389f847a4ee61f63570aac714R64


packages/relay-compiler/graphql-compiler/transforms/DefinitionRenameTransform.js
```diff
@@ -0,0 +1,24 @@
...
+
+const shortHash = require('short-hash');
+const resolve = require('resolve');
+
+function transform(doc, filePath) {
+  return {
+    ...doc,
+    definitions: doc.definitions.map(def => ({
+      ...def,
+      metadata: {
```
robrichard: My first attempt was to do this with the rest of the `IRTransforms`, but there are assumptions of unique fragment names before those transforms are executed. This strategy is not ideal either because it is adding a new `metadata` property to objects that are typed externally by `graphql-js`. Do you think it would be better for this to happen in [`ASTConvert`](https://github.com/facebook/relay/blob/master/packages/relay-compiler/graphql-compiler/core/ASTConvert.js) or somewhere else?

leebyron:
Thanks for your investigation of this! You clearly put a lot of work into that, which is hugely appreciated.

However I'm very concerned about the costs of this approach. 

There are really strong benefits from having globally unique fragment names, many of which @wincent already mentioned in the discussion you linked to, and Relay benefits from many of them. Reducing the ability for our existing tools like graphql-language-service and graphiql to understand Relay's queries should be considered only with great caution. It goes against the philosophical change from Relay Classic to Relay Modern which attempts to write queries which much closer match the compiled output queries, which is important not just for tool integration, but for easing debugging and developer ergonomics overall. For example, being able to go directly from seeing query fragments in your server logs to grepping for references of those fragments in your client code has been a critical improvement for Relay's use at Facebook.

This also adds cost to the complexity of the compiler, and maintenance costs there are something we're taking very seriously.

I'd like to back up and first better understand the problem you're trying to solve, because I think we can discuss what other potential solutions we have available, and hopefully we can arrive at something that both solves the problem while minimizing costs.

If the sole problem here is maintaining unique filenames across a codebase, then I may suggest another possible approach: loosening or removing the current restriction that GraphQL fragment names mirror their filenames. That was added to ensure a successful transition via Relay Compat, but I'm not sure is necessary in a pure Relay Modern codebase.

robrichard:
@leebyron thanks for taking a look. I think there's benefits to discussing changes like this when there is a working code example so I opened this PR instead of an issue. I completely understand if this is ultimately not appropriate for Relay and is closed. 

I appreciate you listing the benefits of having globally unique fragment names and I understand that great caution should be taken when making trade-offs that diminish these benefits. 

To give you more insight into the problems I'm trying to solve - at @1stdibs our codebase is made up of many repositories that contain front-end applications that are deployed independently of each other. They all load data from the same GraphQL schema. We sometimes share UI components between these repositories by publishing them to a private `npm` server. We've run into issues when working on these shared components where fragment names are unintentionally duplicated between the shared component and a new consumer of that component. When adding a new fragment to a shared component, caution needs to be taken to ensure there are no conflicts with any consuming apps.

Removing the restriction on fragment names matching file names would definitely help but I feel like it would soon be necessary to adopt a naming convention with namespaces or something similar. This reminds me a lot of the issues many people run into with componentized UIs and CSS class names, leading to the adoption of `BEM` to avoid conflicts. The approach I took here was inspired by a combination of [CSS Modules](https://medium.com/seek-blog/the-end-of-global-css-90d2a4a06284) and the linked RFC thread. The trade-offs of CSS in JS solutions also parallel the trade-offs here, i.e. you can't just grep for a class name in your HTML anymore.

If a standard approach for importing fragments emerges from the RFC process, hopefully the tool integration would catch up and support jumping to definitions and rewriting imports the way WebStorm and others support the same for JavaScript modules.

Anyway, I'm interested in hearing your thoughts if you have a better way to solve this. Or maybe it's possible to refactor `relay-compiler` to make this sort of thing possible as a user provided transform? It's not currently possible as fragments get deduped before the transforms are executed.

tgvashworth:
@leebyron We're having a discussion about this over in https://github.com/facebook/graphql/issues/343 — could you take a look?

I'd like to explore all our options with your input available as early as possible.

leebyron:
> Or maybe it's possible to refactor relay-compiler to make this sort of thing possible as a user provided transform? 

That's potentially possible in the future. I know that @wincent has been doing some relay-compiler refactoring lately to allow for more flexibility for other use cases. I'm not sure if that would enable this specific use case though.

> Removing the restriction on fragment names matching file names would definitely help but I feel like it would soon be necessary to adopt a naming convention with namespaces or something similar. 

Let's start by working on relaxing the fragment name to file name restriction for relay-modern-only codebases. I think that will unlock a much simpler fix, albeit still not perfect.

I agree that a naming convention like a namespace is a great way to carve out an unlikely-to-conflict fragment name, while maintaining grepability. This pretty closely mirrors how Facebook's internal codebases appear, we tend towards longer but more descriptive file names to tend towards unique names to aid the ability to search our codebases.

I'm going to close this issue since there's convergence on the alternative approach, thanks again for investigating and thoughtful discussion!

wincent:
> That's potentially possible in the future. I know that @wincent has been doing some relay-compiler refactoring lately to allow for more flexibility for other use cases. I'm not sure if that would enable this specific use case though.

(Possibly straying off topic here, but giving an update on this anyway...)

At the moment we have two layers of abstraction:

- "relay-compiler": small, cohesive, special-purpose, tailored, not very extensible.
- "graphql-compiler": collection of low-level tools, flexible, few limitations, but extensive manual assembly required.

For now I've been focussing on getting the latter closer to being generally usable; not much done (or even really planned yet) to make the former more extensible. Nevertheless, I think as part of this process, there will be some degree of opening up in "relay-compiler" to make it easy to modify/augment/extend the internals. More details as this comes into focus.

PR #2079, opened 2017-09-01
Permit bidirectional PageInfo

eapache:
The existing specification for PageInfo requires that `hasPreviousPage`
be false when paginating forward, which is quite limiting. Adjust that
part of the specification to permit setting `hasPreviousPage` when
paginating forward, and symmetrically permit setting `hasNextPage` when
paginating backward. Implementations may still hard-code false if
calculating the correct value would be inefficient. This was suggested
by @dschafer in https://github.com/graphql/graphql-relay-js/issues/58#issuecomment-169542843.

Making this change exposed an inconsistency in handling unusual cases
such as when a client provides only `last` and `after` arguments. I
adjusted a few other parts of the algorithm to clarify how
best to handle these cases.

cc @swalkinshaw


website/graphql/Connections.md
```diff
@@ -310,29 +310,31 @@ The server must provide a type called `PageInfo`.
...
+`last`/`before`, then the server must return {true} if prior edges exist, 
+otherwise {false}. If the client is paginating with `first`/`after`, then the 
+client may return {true} if edges prior to `after` exist, if it can do so 
+efficiently, otherwise may return {false}. More formally:
 
 HasPreviousPage(allEdges, before, after, first, last):
   * If {last} is set:
     * Let {edges} be the result of calling {ApplyCursorsToEdges(allEdges, before, after)}.
-    * If {edges} contains more than {last} elements, return {true}.
+    * If {edges} contains more than {last} elements return {true}, otherwise {false}.
```
eapache: It's not a big deal either way (it doesn't affect our use case at all) but I'd originally left this to fall through intentionally. In the (admittedly bizarre) case where the client provides only `last` and `after`, and there are fewer than `last` nodes after `after`, I figured it made sense for the server to still be able to return `true` if it can efficiently determine that there are elements prior to `after`.

Issue #1740, opened 2017-05-10
[modern] Relay Compiler error: RangeError: Maximum call stack size exceeded


dmbrowne:
I previously wrote a comment in the thread [#1660](https://github.com/facebook/relay/issues/1660) about an error I am receiving when trying to build with the relay compiler. I have managed to work out why I am receiving an error, and I think it may be a bug with the compiler.

Take the following schema which has been extended / modified from the todo-modern relay example (the `+` is what I have added to the *todo-modern* schema for demo purposes):
```graphql
+ input AddUserInput {
+ 	username: String!
+ 	password: String!
+ 	todo: AddTodoInput
+ }

  input AddTodoInput {
    text: String!
+   user: AddUserInput
    clientMutationId: String
  }

+ type AddUserPayload {
+   changedUser: User
+   clientMutationId: String
+ }

  type Mutation {
+   addUser(input: AddUserInput!): AddUserPayload
    changeTodoStatus(input: ChangeTodoStatusInput!): ChangeTodoStatusPayload
    markAllTodos(input: MarkAllTodosInput!): MarkAllTodosPayload
    removeCompletedTodos(input: RemoveCompletedTodosInput!): RemoveCompletedTodosPayload
    removeTodo(input: RemoveTodoInput!): RemoveTodoPayload
    renameTodo(input: RenameTodoInput!): RenameTodoPayload
  }
```
Notice the dependancy of `AddTodoInput` on `type AddUserInput`, but also the dependency of `AddUserInput` on `type AddTodoInput`.

When I run the relay compiler, I get the error `RangeError: Maximum call stack size exceeded` and the compiler immediately quits its operation and does not write any \__generated\__ queries or mutations.

mjmahone:
Interesting. I don't know if I've ever though about recursive input types before, but they should work (i.e. it's a bug that they don't)! That would be a really good test case to add in to the compiler tests, but I'm curious where the error is actually happening (which module).

dmbrowne:
I'm using scaphold.io for my graphql server, so unfortunately I'm not in control of the schema generation, so avoiding this error is impossible.  My work around for now is to manually delete the recursive inputs before trying to compile.

dmbrowne:
@mjmahone i'm not sure I understand your question, or was it rhetorical?

josephsavona:
[...]

PR #2080, opened 2017-09-04
Add a recursive input type to compiler test schema

benjaminogles:
Adding a mutation with recursive inputs breaks two tests in the relay compiler: `RelayFlowGenerator-test` and `printFlowTypes-test`. It causes the call stack to exceed its max as the code tries to recursively generate flow types for the input objects. See issue https://github.com/facebook/relay/issues/1740 Where someone said that a pull request with a failing test case would be helpful

The two function calls exceeding the stack seem to be here.
https://github.com/facebook/relay/blob/95319d19c2027661efd400eea3137ba7b6885431/packages/relay-compiler/core/RelayFlowGenerator.js#L452
https://github.com/facebook/relay/blob/95319d19c2027661efd400eea3137ba7b6885431/packages/relay-compiler/codegen/flowtype/transformInputObjectToIR.js#L50

I was able to add a base case to the recursive function in the `RelayFlowGenerator` to get it to generate types for recursive inputs but it also resulted in huge, and unpredictable generated mutations. At first I thought, instead of just disallowing recursive definitions at certain levels it might be possible to label the recursive definition as a separate type to reflect what the actual schema looks like and so the generated mutation would look something like this
```
type CommentVariables = {|
  clientMutationId?: ?string;
  feedbackId?: ?string;
  feedback?: ?{|
    commentsIds: ?$ReadOnlyArray<string>;
    comments: ?$ReadOnlyArray<CommentVariables>
  |};
|};
export type CommentCreateMutationVariables = {|
  input: CommentVariables;
  first?: ?number;
  orderBy?: ?$ReadOnlyArray<string>;
|};
```
But I saw a commit in the log of the RelayFlowGenerator that explained generating types for inner objects resulted in naming conflicts.

### DIFF
--- a/packages/relay-compiler/core/RelayFlowGenerator.js
+++ b/packages/relay-compiler/core/RelayFlowGenerator.js
@@ -42,6 +42,12 @@ export type ScalarTypeMapping = {
 };
 
 const printBabel = ast => babelGenerator(ast).code;
+const {getInputObjectSCCs} = require('getSCCsFromFlowContextDocument');
+
+function whiteListRecursiveTypes(node: Root) {
+  const sccs = getInputObjectSCCs(node);
+  return Array.prototype.concat.apply([], sccs);
+}
 
 function generate(
   node: Root | Fragment,
@@ -51,10 +57,13 @@ function generate(
   const defaultedCustomScalars = customScalars || {};
   const output = [];
   if (node.kind === 'Root' && node.operation !== 'query') {
+    const whiteList = inputFieldWhiteList
+      ? [...inputFieldWhiteList, ...whiteListRecursiveTypes(node)]
+      : whiteListRecursiveTypes(node);
     const inputAST = generateInputVariablesType(
       node,
       defaultedCustomScalars,
-      inputFieldWhiteList,
+      whiteList,
     );
     output.push(printBabel(inputAST));
   }
--- a/packages/relay-compiler/core/getSCCsFromFlowContextDocument.js
+++ b/packages/relay-compiler/core/getSCCsFromFlowContextDocument.js
@@ -41,7 +41,7 @@ function baseInputObjects(type: GraphQLInputType) {
   return Object.keys(fields)
     .map(k => fields[k])
     .filter(
-      field => baseInputType(field.type) instanceof GraphQLInputObjectType,
+      field => getRawType(field.type) instanceof GraphQLInputObjectType,
     );
 }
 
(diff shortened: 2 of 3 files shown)


benjaminogles:
Actually, in focusing so much on the source of the infinite recursive loop, I may have over scoped the problem. The only difference in output from my fixed version of the compiler is in the generated flow types. The exported `batch` constant remains unchanged.

I was assuming the structure of the generated flow type had some other critical role in query validation, but if it is just for static strong typing, to be optionally imported by the developer then maybe we can compromise a less specific generated flow type to let the compiler handle these recursive definitions?

benjaminogles:
This doesn't fix the problem for legacy flow types (the `printFlowTypes-test`) and I think it might be better to put the code in `GraphQLSchemaUtils` and build the white list at a higher level, maybe when instantiating the `RelayFileWriter` in the first place. But I want to know if this is a viable solution before I continue to work on it

schickling:
Thanks a lot for creating test cases that reproduce this problem as @josephsavona suggested.

Hopefully someone of the Relay/GraphQL team can soon look into this as it seems to affect many Relay users as mentioned in #1740. 💪

kguller:
Attempted to implement @benjaminogles compiler fix.  Went through the file changes and the logic looks good and straight forward.  The visited array flag should capture the correct level and skip adding the field to the Type definition and halt the recursion.  Thanks @benjaminogles, I'm just having some problem with installation, but I can reproduce the expected output and I'm probably missing a simple step in my rebuild attempt.

Environment:
Yarn / Webpack / React' / Relay-Compiler / Graph.Cool + Schema

Installation Issues in case you are using Yarn:
Trouble with yarn installing the pull request due to security rights.
Trouble rebuilding package with yarn as the install is global and doesn't seem to target new code changes after I patch in the change files for rebuild.

I ended up having to manually update the production outputted JS directly, so I can get the code to run and it does skip the 2nd level test type that is calling back to the recursive type. It would be great to get this reviewed and refactored a little bit like @benjaminogles mentions:

**Before:**
input TestReferencetestsTest {
  testField: String!
  testReferencesIds: [ID!]
  testReferences: [TesttestReferencesTestReference!]
}

input TesttestReferencesTestReference {
  testsIds: [ID!]
  tests: [TestReferencetestsTest!]
}

**After:**
input TestReferencetestsTest {
  testField: String!
  testReferencesIds: [ID!]
  testReferences: [TesttestReferencesTestReference!]
}

input TesttestReferencesTestReference {
  testsIds: [ID!]
}

benjaminogles:
@wincent or @leebyron Has there been any discussion about whether white listing the recursive field names from the flow type generator would be a viable solution?

If Relay will eventually generate flow types for query variables as suggested in https://github.com/facebook/relay/issues/2105 then I think generalizing my logic a bit would be necessary.
I still think finding the strongly connected components in the schema is the most straight forward way of finding recursive fields.

benjaminogles:
Closing in favor of https://github.com/facebook/relay/pull/2142

PR #2091, opened 2017-09-10
fix link to example

latik:
fix link to example after https://github.com/relayjs/relay-examples/commit/72e0644c572ef4a5f74d113fc75bc8a27030edd3


kassens:
Good catch!

Issue #2093, opened 2017-09-11
Relax constraint of unique file names in Relay Modern


leebyron:
Relay Classic / Compat require file names to match fragment names to ensure Relay Compat mechanisms work correctly. However in Relay Modern this doesn't necessarily need to be true.

There have been a few issues opened in the past about this issue, so it would be great to fix this with a pull request.

Note that fragment names should still continue to be unique across a compiled app.

owcc:
can `relay-compiler` support `.jsx`?

josephsavona:
For anyone interested in sending a PR, we are supportive  of  adding  a config option  (via relay-config) that allows an application to disable the document name validation.

LaylBongers:
This is still an issue currently, especially with frameworks like NextJS where module names for pages are used to determine their routing. This results in unhelpful name restrictions like:
- `src/pages/index.tsx` -> `pagesQuery`
- `src/pages/items/[uuid].tsx` -> `uuidQuery`

Mikoz93:
And kebab-case naming convention https://github.com/facebook/relay/issues/3081

maraisr:
@LaylBongers not entirely related; but you don't want queries in the pages folder for NextJS — as that leads to routes generated like `/items/__generated__/uuidQuery.graphql`

PR #2109, opened 2017-09-19
Relax constraint of unique file names in Relay Modern

jkettmann:
See #2093 

- Allow fragment and operation names to be different from their modules names.
- Ensure that all fragment and operation names are unique.

Only the Relay Modern part seems to be missing, as these changes are not specific for Relay Modern. This is why I only commented the constraining parts out.

How can we use these rules only on Relay Modern?

### DIFF
--- a/packages/relay-compiler/codegen/FindGraphQLTags.js
+++ b/packages/relay-compiler/codegen/FindGraphQLTags.js
@@ -259,17 +259,17 @@ function validateTemplate(template, moduleName, keyName, filePath, loc) {
     );
     const definitionName = def.name.value;
     if (def.kind === 'OperationDefinition') {
-      const operationNameParts = definitionName.match(
-        /^(.*)(Mutation|Query|Subscription)$/,
-      );
-      invariant(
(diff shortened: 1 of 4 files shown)


jkettmann:
@leebyron Might you have a look at this and tell me if this is the right direction? I could continue spending time on this and fix the failing flow checks then.

alloy:
@leebyron bump

laverneleah:
Unsubcribe me from this thread!  You are overloading my emails!

Sent from my BlackBerry - the most secure mobile device

alloy:
@laverneleah You can unsubscribe yourself with the link you see in that email you got or by using the button here on this ticket with the button in the sidebar. It would also be nice if you could delete your comment here, the unnecessary noise it adds tot this thread does not help.

josephsavona:
Thanks for submitting a PR. However, our focus with Relay Compiler is primarily on supporting Relay (Modern) runtime. Can you clarify how you'd actually take advantage of this? An alternative is  to add a flag (via the new relay-config) that optionally disables the module-name-prefix  requirement altogether.

Issue #2111, opened 2017-09-20
[Modern] RelayNetworkLogger log shows YellowBox on ReactNative


felippepuhle:
Hello folks, how are you?

<img width="444" alt="screen shot 2017-09-20 at 10 38 36 am" src="https://user-images.githubusercontent.com/15961929/30646778-8163bec6-9df0-11e7-88e9-bc6fb7e30625.png">

I think we could verify if `console.groupCollapsed` and `console. groupEnd` exists here: https://github.com/facebook/relay/blob/master/packages/relay-runtime/network/RelayNetworkLoggerTransaction.js#L90

That would solve this problem. I can send a PR if you agree.

Thanks!

leebyron:
I suggest checking in the constructor, that way we can throw an early error if you're attempting to use the network logger on a platform that doesn't support it. A PR would be great

Also, I think these methods become available if you connect to a chrome debugger (Cmd+D from the simulator I believe)

felippepuhle:
Left a message on the PR, thanks @leebyron.

PR #2113, opened 2017-09-20
RelayNetworkLogger - Avoid YellowBox on ReactNative

felippepuhle:
Fix https://github.com/facebook/relay/issues/2111

I think it's better to keep it simple and only verify if method exist. Checking/throwing an error in the constructor will generate the same result here, since every query seems to create a new instance of `RelayNetworkLoggerTransaction` .


leebyron:
I think this is a reasonable compromise. The console.group statements were put there because otherwise the output is quite noisy and hard to read, though I agree that throwing an error wouldn't be the best experience and dealing with noisy output isn't the end of the world for platforms that don't support console.group

Issue #2103, opened 2017-09-15
[Modern] refetching with include(if: false) makes existing field undefined


jardakotesovec:
Using the RefetchContainer I have two different fields I want refetch - but not always both.. depending how variables are changing I need to get update only from one or both. So I figured I can control with `@include(if: @shouldFetch)` expression which one I would refetch.

Request and response to the server works as expected. If include is false the field is completely omitted in response.

It looks like this:

```
export default createRefetchContainer(User, {
    viewer: graphql.experimental`
        fragment User_viewer on Viewer
        @argumentDefinitions(
            foo: {type: "String!", defaultValue: ""},
            bar: {type: "String!", defaultValue: ""},
            fetchFirst: {type: "Boolean!", defaultValue: false},
            fetchSecond: {type: "Boolean!", defaultValue: false},

        ) {
            firstField (foo: $foo) @include(if: $fetchFirst) {
                data
            }
            
            secondField (foo: $foo, bar: $bar ) @include(if: $fetchSecond) {
                data
            }
        }

    `
}, graphql.experimental`
    query UserRefetchQuery(
        $one: String!, 
        $two: String!, 
        $fetchFirst: Boolean!,
        $fetchSecond: Boolean!
        ) {
        
        viewer {
            ...User_viewer @arguments(
                one: $one, 
                two: $two, 
                fetchFirst: $fetchFirst, 
                fetchSecond: $fetchSecond,
            )
        }    
    }
`);
```

But issue is how its interpreted. Because when for example `two` is not present in response at all relay makes `viewer.two` undefined even if there was value before. 

So it seems that its serializing the response to the store also based on the actual query, not just based on the response, which make sense. But I would think if the @include/@skip is used to omit the field - it should be considered not requested at all, therefore not updating the store with undefined?

My intention is just update one of the fields with new value, but not loose value from the second one.

If we you are agreeable on this, some hint where to start to adjust this behavior?

Thanks!

sibelius:
I was thinking in saving some data in state of the component as a workaround for this

leebyron:
This definitely sounds like a bug, thanks for the clear report!

I think the fix is likely to be found in RelayResponseNormalizer, which is responsible for converting a network response into normalized Relay store in the context of a request. Seems like that might not be properly differentiating between `null` and `undefined`

maikelrobier:
[...]

PR #2114, opened 2017-09-21
Field is accessible also when @include(if: false)

jardakotesovec:
Addressing #2103 

What happens is simply that for @include(if: false) field gets skipped while reading even though the previous value is still in the store.

This change solves my problem, but not sure if this is the right thing to do anymore. Depends on point of view. 

Argument for it is: @include(if: false) says that field should be skipped in the request, therefore just should not be refetched and old value should be still accessible.
Argument against is: having in sync `variables` and values getting from the store provides clear overview of current state, without considering past. Which is what we all kind of like on react.

I am bit inclined to keep it as it is and find some other approach for my use case. What you think?

### DIFF
--- a/packages/react-relay/modern/__tests__/ReactRelayRefetchContainer-test.js
+++ b/packages/react-relay/modern/__tests__/ReactRelayRefetchContainer-test.js
@@ -717,7 +717,7 @@ describe('ReactRelayRefetchContainer', () => {
         },
       });
       expect(render.mock.calls.length).toBe(2);
-      expect(render.mock.calls[1][0].user.name).toBe(undefined);
+      expect(render.mock.calls[1][0].user.name).toBe('Zuck');
     });
 
     it('updates context with the results of new variables', () => {
--- a/packages/relay-runtime/store/RelayReader.js
+++ b/packages/relay-runtime/store/RelayReader.js
@@ -116,10 +116,7 @@ class RelayReader {
           this._readLink(selection, record, data);
         }
       } else if (selection.kind === CONDITION) {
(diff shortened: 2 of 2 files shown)


leebyron:
Interesting. I'd like to hear @josephsavona's opinion for the intent behind the original behavior since it's not my expertise. Perhaps @kassens knows as well

alloy:
@kassens bump

jardakotesovec:
Hi @josephsavona , could you elaborate bit more? It sounds like you are in favor of this change but at the same time saying 'I can’t think of any particular reason to make fields available that don’t pass the condition', which is contradictory, is not it?

Use case why I proposed this change was having multiple skipped fields (lets say field1 and field2) in refetch container with the goal to refetch only some of them in time. But problem is that if I refetch with unskipped field1 and later refetch with unskipped field2 - I can't access the field1 data anymore even though they are still in store.

alloy:
@jardakotesovec I think that using `skip` seems a reasonable option, seeing as the queries need to remain static (I.e. what gets fetched can’t solely be determined at runtime), but as this changes the semantics of the `refetch` API, maybe it’s worth it to add an option to `refetch` that controls wether or not store data is always returned and default to `false`?

josephsavona:
Sorry, that was missing a “not”: I agree that field data should only be available when the condition passes.

jardakotesovec:
Realized yesterday (and tested today) that I can already achieve that nicely using renderVariables argument in refetch function. Which are basically variables for reading operation, which is exactly what I need - I set the skipping variables to false there.

Thanks a lot for getting back to me!

PR #2121, opened 2017-09-24
Add Caching docs

felippepuhle:
Found a lot of issues talking about `RelayQueryResponseCache`. Thought that including it to docs is a good idea. What do you think?

Thanks!

### DIFF
--- a/docs/Modern-NetworkLayer.md
+++ b/docs/Modern-NetworkLayer.md
@@ -53,5 +53,77 @@ export default environment;
 
 Note that this is a basic example to help you get started. This example could be extended with additional features such as request/response caching (enabled e.g. when `cacheConfig.force` is false) and uploading form data for mutations (the `uploadables` parameter).
 
+## Caching
+
+Relay modern makes no assumptions about what to cache and will garbage collect any data that is no longer referenced.
+
+You have to implement your own cache strategy. A simple solution is to use `RelayQueryResponseCache`(an in-memory cache):
+
+```javascript
+import {
+  Environment,
+  Network,
+  RecordSource,
+  Store,
+} from 'relay-runtime';
+import RelayQueryResponseCache from 'relay-runtime/lib/RelayQueryResponseCache';
+
+const oneMinute = 60 * 1000;
+const cache = new RelayQueryResponseCache({ size: 250, ttl: oneMinute });
+
+function fetchQuery(
+  operation,
+  variables,
+  cacheConfig,
+) {
+  const queryID = operation.text;
+  const isMutation = operation.operationKind === 'mutation';
+  const isQuery = operation.operationKind === 'query';
+  const forceFetch = cacheConfig && cacheConfig.force;
+
+  // Try to get data from cache on queries
+  const fromCache = cache.get(queryID, variables);
+  if (
+    isQuery &&
+    fromCache !== null &&
+    !forceFetch
+  ) {
+    return fromCache;
+  }
+
+  // Otherwise, fetch data from server
+  return fetch('/graphql', {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({
+      query: operation.text,
+      variables,
+    }),
+  }).then(response => {
+    return response.json();
+  }).then(json => {
+    // Update cache on queries
+    if (isQuery && json) {
+      cache.set(queryID, variables, json);
+    }
+    // Clear cache on mutations
+    if (isMutation) {
+      cache.clear();
+    }
+
+    return json;
+  });
+}
+
+const environment = new Environment({
+  network: Network.create(fetchQuery),
+  store: new Store(new RecordSource()),
+});
+
+export default environment;
+```
+
 ## Custom open-source implementations
 **[react-relay-network-modern](https://github.com/nodkz/react-relay-network-modern)** on [npm](https://www.npmjs.com/package/react-relay-network-modern) - is a Network Layer for Relay Modern which has built-in highly customizable middlewares for commonly used scenarios: batching query requests, caching, authentication, request retrying, logging. Moreover, you may write your own middlewares with custom logic.


amkoehler:
This documentation is extremely helpful for setting up basic response caching. Any chance we could get this merged soon?

felippepuhle:
cc @leebyron

tlvenn:
wow I am confused.. Relay modern has no normalised cache at all anymore ?

fernandocalsa:
@tlvenn Relay-runtime exposes a class `RelayQueryResponseCache` that you can use in your Environment to cache the payload of the queries that Relay creates. @felippepuhle explains in this PR how to configure a simple cache, but I'm wondering if there are other examples with a more complex cache system

felippepuhle:
We could talk a little more about this, ping me if you have questions... 😄

fernandocalsa:
I would like to share my context. Right now in my app I have a `loginMutation` which payload contains `user` information, then in the menu I have a fragment to show the userName, so Relay is trying to create the query to fetch the `user` data, which is already in the store. I want to avoid this query when the user has just login, but execute it when the user reload the page.

From here I have started to look into the cache system, and I am wondering why do we have to use a cache system when we have the store, which is the source of truth, so if Relay executes a query asking for some data that is already in the store, this data shouldn't be requested again, so we could modify the query at this stage.

I don't have much experience with Relay so maybe this idea doesn't make any sense. What do you think @felippepuhle ?

felippepuhle:
@fernandocalsa I think it's better to open an issue with this thread and put your source code there.. what do you think?

tlvenn:
FYI I have opened a an issue on apollo side to explore integrating apollo-cache with relay modern which seems pretty doable and would let us use more advanced cache sytems.

https://github.com/apollographql/apollo-client/issues/2558

tlvenn:
@fernandocalsa just to clarify, there is no in memory cache at all with relay modern any more ? The documentation https://facebook.github.io/relay/docs/relay-modern.html#terminology is full of mentions of internal cache system which leads me to believe otherwise.

fernandocalsa:
@tlvenn what I understand in this page is that when they talk about cache for me is the Relay store. In this PR we are talking about cache the requests to the server, which is implemented with @felippepuhle PR

felippepuhle:
I think you're right @fernandocalsa

tlvenn:
@fernandocalsa then this opening statement in this PR is super confusing imho `By default, Relay Modern doesn't do any type of cache at all`.

Also https://facebook.github.io/relay/docs/thinking-in-graphql.html#client-caching spent a good amount of time explaining why a response cache is not really ideal with Graphql in contrast to what you would normally do with a REST api.

Given the Relay store has already a normalized cache that can properly deal with a graph, what is the motivation on layering a response cache on top of it ?

tlvenn:
I guess this issue https://github.com/facebook/relay/issues/2143 sheds some light on why it might be interesting to layer a cache on top of the Relay Store. Quoting from the issue:

`Relay modern makes no assumptions about what you want cached and will garbage collect any data that is no longer referenced.`

fernandocalsa:
@tlvenn in #2143 @sibelius recommends this PR to implement a cache system

felippepuhle:
I think you got it wrong @tlvenn: you can implement your own cache strategy as you want...

In my case, I use the query text and variables to determine how data would be stored/recovered... So, if you request another field or even pass other variables to the same query, you'll get a fresh response from the server.

Also, we can clear the cache when any mutation happens,
 ensuring the consistency of data - https://github.com/facebook/relay/pull/2121/files#diff-81d9abbdd67b503ef956c589a25e88cbR61

We've been using something like this in production for a while, and that's awesome (in my opinion).

tlvenn:
Hi @felippepuhle , I am not denying you can do just that. Your PR forced me to get up to speed with Relay Modern internals and exactly how the store is working now, so thank you for that ;)

As far as I understand, Relay Modern does provide a graph cache in the form of the `Store` interface and its `RelayMarkSweepStore` default implementation which has a pretty agressive garbage collection strategy, ie, it will purge any data that is not referenced from your view hierarchy.

The store is entirely pluggable given it is passed to the environment and it has a well defined contract provided by the interface `Store`. Relay Runtime itself treat any record from the store as immutable and even enforce this in dev so the store is entirely in control over its records.

The [architecture](https://github.com/facebook/relay/blob/master/packages/relay-runtime/ARCHITECTURE.md) doc provides a pretty good overview of the store and how it interacts with the other pieces of Relay Runtime.

So if you are in need of a more durable cache that the provided default store, you have 2 options:
- Implement a new store and provide it to your environment.
- Add another layer of cache on top of the existing Relay Store by proxying the network request.

The first option is more involving but provides a lot more power and is already integrated within Relay. The second option is simpler but I believe much more limited in scope.

felippepuhle:
@tlvenn Great point! Could you generate a simple demo of a custom store implementation? It would be awesome! 😄

alloy:
> this opening statement in this PR is super confusing imho `By default, Relay Modern doesn't do any type of cache at all`.

> I guess this issue #2143 sheds some light on why it might be interesting to layer a cache on top of the Relay Store. Quoting from the issue:
>
> `Relay modern makes no assumptions about what you want cached and will garbage collect any data that is no longer referenced.`

@felippepuhle I agree with @tlvenn that it would read better (less confusing( if the document were to be updated to mention exactly what type of case this addresses.

And I guess that by now it might need some more work to be updated for the new website too?

alloy:
@tlvenn A document like that indeed sounds awesome! Would you be willing to write one? 🙏

leebyron:
@felippepuhle would you mind rebasing this to account for the conflicts?

felippepuhle:
@leebyron I'll do as soon as possible, thanks!

alloy:
@kassens This PR is red on CI due to [issues](https://travis-ci.org/facebook/relay/builds/338849207) that were introduced in https://github.com/facebook/relay/commit/b81b02c56e4a9871bb11d1d63783b1d30d7b78ce.

jstejada:
@felippepuhle do you think it would be a good idea to add this as an additional section under the existing Network Layer docs?

felippepuhle:
@jstejada it might be with this implementation, but I think we can explore a lot more about caching in the future... 

What do you guys think?

alloy:
@felippepuhle I’d really love to see this get in the public docs sooner rather than later, so maybe we should just revisit the need for a separate document when there are future doc PRs on caching?

alloy:
Additionally, maybe it’s time to export `RelayQueryResponseCache` properly?

felippepuhle:
Nice! I'll take a look later on

docs/Modern-Caching.md
```diff
@@ -0,0 +1,70 @@
...
+const cache = new RelayQueryResponseCache({ size: 250, ttl: oneMinute });
+
+function fetchQuery(
+  operation,
+  variables,
+  cacheConfig,
+) {
+  const queryID = operation.text;
+  const isMutation = operation.query.operation === 'mutation';
+  const isQuery = operation.query.operation === 'query';
```
alloy: In Relay 1.5.0 I had to change `operation.query.operation` to `operation.operationKind`. Not sure if and when this changed, though.

felippepuhle: You're absolutely right

alloy:
Awesome, thanks for updating!

@jstejada Wanna take another look?

docs/Modern-NetworkLayer.md
```diff
@@ -53,5 +53,77 @@ export default environment;
...
+
+function fetchQuery(
+  operation,
+  variables,
+  cacheConfig,
+) {
+  const queryID = operation.text;
+  const isMutation = operation.operationKind === 'mutation';
+  const isQuery = operation.operationKind === 'query';
+  const forceFetch = cacheConfig && cacheConfig.force;
```
alloy: Did you experience that `cacheConfig` can be `null` or `undefined`? Because the type says it won’t https://github.com/facebook/relay/blob/e6481d7eeb7da386c907728222fbb0cbf4352ce0/packages/relay-runtime/network/RelayNetworkTypes.js#L89

richardgirges:
Thanks very much for this writeup! This significantly reduced the complexity behind migrating from Apollo to Relay Modern for my team. These instructions gave us a _close enough_ caching experience to what react-apollo provided out of the box. Without this, we had to deal with a ton of unexpected loading states in areas of the app where we previously relied on react-apollo to give us data that we _knew_ was already cached.

Would be great to get this merged as I'm sure many other developers in similar shoes would benefit from these instructions.

jstejada:
thanks!!

PR #2122, opened 2017-09-24
Fix false warning when end/startCursor is null because no edges are returned

### LINKED ISSUE #1852: [Modern] ReactRelayPaginationContainer pageInfo warning
The ReactRelayPaginationContainer complains about missing pageInfo fields whenever no edges are returned from a connection. Specifically, I am seeing this warning whenever pageInfo.endCursor === null. I am able to silence the warning by returning an empty string for pageInfo.endCursor. Is an empty string the expected value in this situation or am I missing something else?

Example response:
```
{
  "data": {
    "user": {
      "postsConnection": {
        "edges": [],
        "pageInfo": {
          "hasNextPage": false,
          "endCursor": null,
          "hasPreviousPage": false,
          "startCursor": null
        }
      }
    }
  }
}
```

Warning:
```
Warning: ReactRelayPaginationContainer: Cannot paginate without pageInfo fields in `MyView`. Be sure to fetch hasNextPage (got `false`) and endCursor (got `null`).
```

janicduplessis:
This allows null (or other values, don't think we need to check for that) for start or end cursor when edges is an empty array.

`graphql-relay-js` returns null for endCursor and startCursor when creating a connection from an empty array so it makes sense we support this.

Fixes #1852

### DIFF
--- a/packages/react-relay/modern/ReactRelayPaginationContainer.js
+++ b/packages/react-relay/modern/ReactRelayPaginationContainer.js
@@ -512,6 +512,9 @@ function createContainerWithFragments<
           : pageInfo[HAS_PREV_PAGE];
       const cursor =
         direction === FORWARD ? pageInfo[END_CURSOR] : pageInfo[START_CURSOR];
+      if (edges.length === 0 && cursor == null) {
+        return null;
+      }
       if (typeof hasMore !== 'boolean' || typeof cursor !== 'string') {
         warning(
           false,


yuzhi:
This might not be the best place to fix #1852. Can you take a look in the RelayConnectionHandler to make sure when zero edges gets returned from the server that PAGE_INFO gets updated correctly and that we don't over write real cursors with nulls?

janicduplessis:
@yuzhi I think we already handle this correctly in RelayConnectionHandler, null is even used as the default value if the server doesn't return pageInfo https://github.com/facebook/relay/blob/master/packages/relay-runtime/handlers/connection/RelayConnectionHandler.js#L97. The issue looks more like the warning check should consider null as a valid cursor value but since it just uses a typeof check it fails for null.

Looked into it a bit more and I should probably not return null from `_getConnectionData ` however but `{ hasMore: false, cursor: null, edgeCount: 0 }` this way calling `loadMore` would still work (https://github.com/janicduplessis/relay/blob/master/packages/react-relay/modern/ReactRelayPaginationContainer.js).

Let me know if that makes sense since I'm not very familiar with this code.

yuzhi:
Sorry,  I was initially confused as to why this could happen in RelayModern, but i realized this can totally happen in compat mode.

yuzhi:
Actually, we still want to return the rest of the info about the connection and just pretend the cursor is null and suppress the warning. I'll put up a fix and link it back here.

plasticine:
> I'll put up a fix and link it back here.

Was this actually completed?

janicduplessis:
@yuzhi Right, I think that's what I was saying by

> Looked into it a bit more and I should probably not return null from _getConnectionData however but { hasMore: false, cursor: null, edgeCount: 0 } 

but didn't update the PR :)

janicduplessis:
@plasticine Looks like this is the commit https://github.com/facebook/relay/commit/a17b462b3ff7355df4858a42ddda75f58c161302

plasticine:
@janicduplessis Ahh—thankyou!

PR #2126, opened 2017-09-24
Remove deprecated forwarding flag for yarn --watch

iamclaytonray:
`warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.`

That's the message I got from Yarn. Anyone using Yarn v1 + doesn't need the additional `--` to forward to the `--watch` portion of the yarn relay script.


leebyron:
Thanks

Issue #1968, opened 2017-07-16
add support for babel-macros


kentcdodds:
Hi there :wave:

I'm helping to investigate ways that create-react-app can use babel plugins like babel-plugin-relay without having to change configuration (see https://github.com/facebookincubator/create-react-app/issues/2730 for more context). This has resulted in the creation of [`babel-macros`](https://github.com/kentcdodds/babel-macros). With `babel-macros`, users can install and configure this plugin and have access to use any `macro` available (or create their own) without configuring anything else. This is appealing to create-react-app users because so far they've had to eject if they wanted to use something like [`preval`](https://github.com/kentcdodds/babel-plugin-preval), [`emotion`](https://github.com/tkh44/emotion), or the babel plugin for relay.

Adding support should be pretty simple (mostly a slight extracting refactor around [here](https://github.com/facebook/relay/blob/c1738f8234226ddb2d05fac8668ee25a64a9b846/packages/babel-plugin-relay/BabelPluginRelay.js#L72-L104) to allow code sharing between the plugin and macro) and starting out with experimental support would be totally fine considering we face a catch 22 situation (CRA wants to see usage of babel-macros in the community before adopting it and you all probably feel the same).

What do you all think?

gaearon:
cc @josephsavona if you're still interested :-)

kassens:
I'm not very familiar with create-react-app, if someone want's to publicise something or work on a `create-relay-app` example, feel free to open a PR or publish a new repo

gaearon:
I don’t think this issue is about creating an example. It’s about supporting this: http://babeljs.io/blog/#zero-config-code-transformation-with-babel-macros

kentcdodds:
That's correct. The idea is that hopefully if we get a `babel-macros` export as part of this package, then CRA could add `babel-macros` to the babel config and folks using create-react-app would be able to use relay without having to configure babel at all. That'd be a significant win IMO :)

kassens:
Re-opening for now as there's clearly interest and potential value.

Trying to understand what this would involve. We could move the babel transform into the `babel-macro` framework, all the `babel-plugin-relay` does for Relay Modern is (pretty much) to replace
```
// input
graphql`fragment MyFragment on SomeType { ... }`
// output
require('./__generated__/MyFragment.graphql')
```
But we would still need to run the relay compiler to generate the files. I guess users of CRA could still install the `relay-compiler` to generate these files?

kentcdodds:
[...]

PR #2171, opened 2017-10-27
Support babel-plugin-macros

apalm:
This PR adds support for [babel-plugin-macros](https://github.com/kentcdodds/babel-plugin-macros).

Usage:
```js
import {graphql} from 'react-relay/macro';
// Or: 
// const {graphql} = require('react-relay/macro');
```

`graphql` is exported from both the `react-relay` and `relay-runtime` packages, but as I created this PR mostly in the interest of possible Create React App integration, I only added an export to `react-relay`.

This PR currently only adds support for Relay Modern. I'm not sure if Compat and/or Classic should be supported?

I currently only have a single basic test. I certainly could copy over the fixtures in `__tests__/fixtures-modern` and change the imports, although I'm not sure how beneficial that would be, as most of the logic is shared with `babel-plugin-relay`.

Closes https://github.com/facebook/relay/issues/1968

Related: https://github.com/facebookincubator/create-react-app/issues/2730#issuecomment-313554085

/cc @kentcdodds


packages/babel-plugin-relay/__tests__/__snapshots__/BabelPluginRelayMacro-test.js.snap
```diff
@@ -0,0 +1,12 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`BabelPluginRelayMacro works 1`] = `
+"
+'use strict';
+
```
apalm: Unlike in `babel-plugin-relay`, the import is not currently retained. I'm not sure if that's acceptable? AFAIK the import is retained with `babel-plugin-relay` only for the "unexpected invocation at runtime" warning (someone correct me if I'm wrong).

packages/react-relay/modern/ReactRelayGraphQL.macro.js
```diff
@@ -0,0 +1 @@
+module.exports = require('../../babel-plugin-relay/BabelPluginRelayMacro');
```
apalm: This module is named the way it is because `babel-macros` requires the source string to match the regex `/[./]macro(\.js)?$/`, and I import this module in the test suite.

gulpfile.js
```diff
@@ -162,7 +162,8 @@ const builds = [
     exports: {
       classic: 'ReactRelayClassicExports.js',
       compat: 'ReactRelayCompatPublic.js',
-      index: 'ReactRelayPublic.js'
+      index: 'ReactRelayPublic.js',
+      macro: 'ReactRelayGraphQL.macro.js'
```
apalm: This outputs `macro.js` in `dist/react-relay`.

packages/babel-plugin-relay/__tests__/__snapshots__/BabelPluginRelayMacro-test.js.snap
kentcdodds: If the import needs to be retained (or replaced with something else) then your macros can get the Program node from the `state` that's passed in and you can unshift an import yourself :+1:

packages/react-relay/modern/ReactRelayGraphQL.macro.js
kentcdodds: You might want to add that as a comment in the file :+1:

packages/babel-plugin-relay/__tests__/BabelPluginRelayMacro-test.js
```diff
@@ -0,0 +1,45 @@
...
+      \`;
+    `;
+    const {code} = babel.transform(basic, {
+      plugins: ['babel-macros'],
+      filename: __filename,
+      compact: false,
+      parserOpts: {plugins: ['jsx']},
+      babelrc: false,
+    });
+    expect(code).toMatchSnapshot();
```
kentcdodds: Have you considered using [babel-plugin-tester](https://github.com/babel-utils/babel-plugin-tester)? It reduces the amount of boilerplate and the snapshot is a bit more insightful. [Here's an example with a macro](https://github.com/kentcdodds/babel-plugin-codegen/blob/master/src/__tests__/macro.js) and [here are the snapshots](https://github.com/kentcdodds/babel-plugin-codegen/blob/master/src/__tests__/__snapshots__/macro.js.snap).

packages/babel-plugin-relay/BabelPluginRelayMacro.js
```diff
@@ -0,0 +1,31 @@
...
+      const path = reference.parentPath;
+      const ast = getValidGraphQLTag(path);
+      if (ast) {
+        compileGraphQLTag(t, path, state, ast);
+      }
+    });
+  });
+}
+
+module.exports = createMacro(BabelPluginRelayMacro);
```
kentcdodds: It looks like [`babel-plugin-relay`](https://facebook.github.io/relay/docs/babel-plugin-relay.html) is configurable. Part of the point of using the macro is so people don't need to update the babel configuration (so they can use it with create-react-app). However it is still handy to be able to configure specific plugins, so `babel-macros` has an experimental config feature. [Learn about config](https://github.com/kentcdodds/babel-macros/blob/master/other/docs/author.md#config-experimental).

docs/modern/BabelPluginRelay.md
```diff
@@ -100,3 +100,11 @@ When compiling code for production deployment, the plugin can be configured to i
   ]
 }
 ```
+
+### Alternatives
```
kentcdodds: If you want to, you could add the babel-macro badge:

[Babel Macro](https://github.com/kentcdodds/babel-macros)

```
[![Babel Macro](https://img.shields.io/badge/babel--macro-%F0%9F%8E%A3-f5da55.svg?style=flat-square)](https://github.com/kentcdodds/babel-macros)
```

packages/babel-plugin-relay/BabelPluginRelayMacro.js
apalm: Thanks for the pointer. I didn't address this yet, as most of the options only pertain to Compat and Classic. Maybe @gaearon can help answer whether Compat and Classic should be supported?

kentcdodds: If configuration is an edge case, I'd suggest omitting it for now. That's something that can be added later and I think it's better to not slow down the merge-ability of this PR for something that not many folks will need/use.

kassens: I don't think we would want to support classic and compat for anything new. The rest of the options are only to allow us to use the same plugin internally at FB and shouldn't be needed in OSS.

packages/babel-plugin-relay/__tests__/__snapshots__/BabelPluginRelayMacro-test.js.snap
kassens: The import is not needed in Modern (classic pulls some runtime helpers from it).

apalm:
@kassens Thanks for looking!

Since it sounds like only Modern needs to be supported, I'll update `BabelPluginRelay.md` to note that this only works in Modern. 

I presume your [comment](https://github.com/facebook/relay/pull/2171#discussion_r147507913) is referring to the `haste` option, and it doesn't need to be supported?

Did you have any opinions on testing? i.e. like using the fixtures in `__tests__/fixtures-modern`. Also, the current way of testing by importing from `../../react-relay/modern/ReactRelayGraphQL.macro` feels iffy, since consumers would import from `react-relay/macro`, but not sure if there's a better way.

DavidBabel:
Any status on this PR ?

apalm:
Sorry for dropping the ball; been busy. I plan to update the PR either later this week or early next week.

package.json
```diff
@@ -21,6 +21,7 @@
     "babel-core": "6.25.0",
     "babel-eslint": "7.2.3",
     "babel-generator": "^6.26.0",
+    "babel-macros": "1.2.0",
```
wtgtybhertgeghgtwtg: [`babel-macros` is now `babel-plugin-macros`](https://github.com/kentcdodds/babel-plugin-macros/pull/42).

kassens:
@apalm: I still think this would be awesome, do you want to continue this or have someone take over?

apalm:
@kassens: Yes, I still want to continue this. Sorry again for dropping the ball.

With the latest commits, I:

* Fixed merge conflicts
* Changed from `babel-macros` -> `babel-plugin-macros`
* Addressed a few review comments. I followed Kent's suggestion and used [babel-plugin-tester](https://github.com/babel-utils/babel-plugin-tester), but if you're not keen on the extra dependency, it's easy to revert, as the change is in a separate commit.

Let me know if you need anything else from me.

wtgtybhertgeghgtwtg:
Is it still a work in progress, then?

wtgtybhertgeghgtwtg:
`babel-plugin-macros` support has landed in latest alpha for `create-react-app`: https://github.com/facebookincubator/create-react-app/pull/3675#issuecomment-358533128

alloy:
@apalm Your last comment makes it sound like this PR is no longer a WIP, is that the case? If so, please update the PR title.

apalm:
@alloy Yes, that's correct. I've updated the title.

alloy:
@kassens @jstejada Good to go?

zachasme:
This is the last remaining blocker for using relay in CRA@next without ejecting.

kassens:
Finally found some time to figure out what was wrong in the gulpfile and got a working CRA with Relay 🎉

I couldn't figure out how to push to this branch, so put up my commit with the fixes at kassens@ef2b729f

The last things I'm not clear about: 

First, I want to avoid a dependency from `relay-runtime` on the babel plugin, so I moved the macro into the babel plugin:
```
import graphql from "babel-plugin-relay/macro";
```
Is this weird? We could also move to a new package, maybe `import graphql from 'relay.macro'`?

Second, should `babel-plugin-relay` have some sort of dependency on `babel-plugin-macros` as there's a require in there?

kentcdodds:
> Is this weird?

Nope, not weird at all. I have several packages that do this.

> We could also move to a new package, maybe `import graphql from 'relay.macro'`?

This is what I do. For example `codegen.macro`'s [entry file](https://unpkg.com/codegen.macro@3.0.0/index.js):

```javascript
// the entire point of this package is to make it easier to use
// babel-plugin-codegen/macro
module.exports = require('babel-plugin-codegen/macro')
```

> should babel-plugin-relay have some sort of dependency on babel-plugin-macros as there's a require in there?

Yes, if `babel-plugin-relay` publishes code that uses `babel-plugin-macros` then it should have `babel-plugin-macros` as a regular dependency.

kassens:
@apalm want to merge the changes from kassens@d7321f9e here? When I merge this it'll all be a single commit and you did most of the work :)

kassens:
Figured it out, hopefully not running into problems with the internal merge now!

apalm:
@kassens Thank you for picking this back up (and for fixing the issues with the original approach 😅).

I think the [docs](https://github.com/facebook/relay/pull/2171/files#diff-0cb87f120cd38e13531971532a077d97R41) still need updating though, per the changes you mentioned in https://github.com/facebook/relay/pull/2171#issuecomment-411459604.

I was also able to confirm a working CRA locally using `react-scripts@2.0.0-next.3e165448` and using `import graphql from "babel-plugin-relay/macro"` :)

Issue #91, opened 2015-08-17
Hydrate scalar fields on client


KyleAMathews:
E.g. a DateTime field could be automatically turned into a Moment.js object.

Is this possible already and I just missed it in the docs?

josephsavona:
This isn't currently supported, and you'd have to manually convert complex types e.g. in your components given the plain data from Relay. This is definitely a use-case worth considering.

I'll leave this open to gather feedback about the idea and examples of data types people may need.

fson:
The most common use case that comes to mind is indeed DateTime (that was the first custom scalar type we added in our own GraphQL server). Another example that would be useful in apps that with location data could be a geo point, for example hydrated to `google.maps.LatLng`.

Because these scalar types and their client-side representation are application specific, if this is going to be added to Relay it should probably be fully extendable, e.g. you could define a mapping from your GraphQL types to your JavaScript object builders. Transit-js might be a [good place](http://swannodette.github.io/2014/07/30/hijacking-json/) to look for an example on how to provide extension points for this.

I think this is could be an interesting idea. While it can also be implemented separately in the transport layer, e.g. with Transit, the GraphQL schema already has the necessary type information for hydrating types, meaning no extra metadata about the object type would need to be passed in the responses. This might save lots of bytes in apps that handle large amounts of data.

On the other hand this could add complexity to the framework, so maybe it would be worth it to consider if this could be done as a separate library outside Relay? A custom `RelayNetworkLayer` gets access to the runtime representation of each query in `sendQueries` and `sendMutation`. Maybe if the Babel plugin would store enough information about the types in the tree, the hydration of responses and serialization of requests could happen in the network layer using that information about the query?

KyleAMathews:
Yeah, Transit-js was what I was thinking of when I wrote this issue.

taion:
This would be useful for having fields that represent plain JSON objects as well.

josephsavona:
The transport layer is probably the wrong place to introduce this as it couples too much of the system together; in particular, the transpiler (to annotate types), the query representation (to expose this metadata), and any network implementation. It also increases the responsibility of a network layer, putting a burden on all implementors.

The best way to approach this is to change `babel-relay-plugin` to annotate the types of leaf fields, and add support in the payload processing phase (`RelayQueryWriter`) for deserializing "complex" values. For each leaf field, check if it has a non-scalar (Int, String, Boolean) type and if so, determine if there is an injected handler for that type. Pseudo code:

```
[...]

PR #2142, opened 2017-10-07
Bugfix flowtype recursion level

benjaminogles:
Fixes https://github.com/facebook/relay/issues/1740

I originally opened https://github.com/facebook/relay/pull/2080 but this fix is better. This one still finds which fields will lead to infinite recursion when generating flow types but instead of completely white listing them it allows them to be generated up to a certain level and after the limit it just uses the `any` type annotation for the most deeply nested objects.

Deeply nested objects that are not recursive are left unaffected. The level of flow types to generate could eventually be an external option for the compiler but is left hard coded in this PR. I am closing the old pull request in favor of this one.

Edit:

I have changed my approach as requested by @jstejada. Instead of computing recursive fields ahead of time and limiting the depth of recursion, the code now just generates a flat list of flow types. One for each inner object type referenced in the original input type. These flow types are not exported.

PR #2193, opened 2017-11-15
fix #2192 - dont recycle frozen objects

### LINKED ISSUE #2192: Exclude custom scalar types from recycleNodesInto
Right now, as mentioned in [this comment](https://github.com/facebook/relay/issues/91#issuecomment-306582952), `recycleNodesInto` recurses into custom scalars. This has the effect of breaking things whenever a custom scalar is serialized as (or transformed into) a non-scalar javascript type (anything except String or Number).

For example, I have a `GeoJSONPolygonCoordinates` custom scalar, which serializes as a nested JSON Array of the following (flow) shape:

```ts
type GeoJSONPolygonCoordinates = [number, number][][];
```

Because these objects are frozen, this means that I end up with an uncaught exceptions, and nodes that are not stored in the Relay store:

```
Uncaught TypeError: Cannot assign to read only property '0' of object '[object Array]'
    at recycleNodesInto.js:32
    at Array.reduce (<anonymous>)
    at recycleNodesInto (recycleNodesInto.js:28)
    at recycleNodesInto.js:30
    at Array.reduce (<anonymous>)
    at recycleNodesInto (recycleNodesInto.js:28)
    at recycleNodesInto.js:30
    at Array.reduce (<anonymous>)
    at recycleNodesInto (recycleNodesInto.js:28)
    at recycleNodesInto.js:44
```

mike-marcacci:
This is one possible fix for #2192, which simply checks if objects are frozen before attempting to modify them.

Do note that while this fixes my use case and any others that are serialized as simple JSON-compatible types, this won't fix cases where a custom scalar is [hydrated to a more complex type](https://github.com/facebook/relay/issues/91), such as a `Date` object. For this, you may want to consider adding a check to skip the recycle if the inputs aren't simple objects or arrays:

```es6
if (
  Object.getPrototypeOf(nextData) !== Object.getPrototypeOf(prevData) ||
  (
    Object.getPrototypeOf(nextData) !== Object.prototype &&
    Object.getPrototypeOf(nextData) !== Array.prototype
  )
) {
  return nextData;
}
```
But, since this may have more consequences, and I would really like my use case fixed ASAP 😜 I've omitted that "fix" from this PR. Please let me know if you'd like me to open a separate issue for that.


mike-marcacci:
This also fixes #2049 and is an alternative to #2051.

dbslone:
I'm curious what the implications of removing the freeze logic entirely from Relay Modern is since its appears to only be applied when `__DEV__` is true. Or I could modify #2051 to use this global flag to determine if it should thaw the frozen object. Just want to help in anyway I can so I can stop using my forked version.

mike-marcacci:
@dbslone - I hadn't realized that freezing was only done in dev (which makes sense). I think perhaps the right solution here is to flag the object as "scalar" when it is created, and check for that here instead. We could just use a `Symbol` defined in relay as the flagging key, and export it so people who (for example) hydrate a custom scalar to a `Date` object can flag it.

Out of a desire to get this fixed ASAP, I'll submit another PR for that approach right now to replace this one. Apologies to the relay team for the noise!

mike-marcacci:
Actually, I realized that:

1. `Symbol` probably isn't something we can use in Relay, since it needs to target older browsers
2. The "flagging" approach solves a problem that I'm not sure really exists
3. There's no reason we'd want to attempt to modify a frozen object in production anyhow!

So, I'm re-opening this :)

ivosabev:
Any status on this? Will it be merged soon?

mike-marcacci:
Ya, we're running a fork locally, since this pretty much blocks development (although it works fine in production). I can't imagine this change would be problematic...

ekosz:
@jstejada @kassens Would a Relay contributor be willing to look / merge this? This is a blocker for us as well (we use a JSON scalar).

Also PRs #2142 & #1868 are really important for me and my team. Some communication on why they haven't been commented in 3+ months would be really helpful. There hasn't been a lot of communication about FB's internal priorities & open source strategy when it comes to Relay.

ivosabev:
I agree with @ekosz, the lack of clear roadmap of the project has been a big issue since Relay was open sourced. There are no clear milestones or time dates of features being released. I am great proponent of Relay and it pains me to see such bad communication with the community. I hope this changes in the future.

ekosz:
@kassens @jstejada Any consideration for this in 1.5.0? If not, what would need to be added to be considered?

alloy:
@ekosz I don’t know the answer to that, but maybe some added test coverage could help reviewing?

alloy:
@dbslone @mike-marcacci I’m not entirely sure I followed everything correct, but should #2051 be closed in favour of this one?

mike-marcacci:
@alloy I certainly would, given that the only reason mutation exists here in the first place is to preserve the object identity, which is lost in #2051 by copying.

dbslone:
@alloy I will close my PR in favor of continuing the discussion in this thread

alloy:
Thanks @dbslone 🙏🏽

alloy:
@mike-marcacci Alright, so could you add a test that specifically demonstrates that?

mike-marcacci:
Sure thing, I’ll add a couple to the PR tonight.

mike-marcacci:
I've added 3 tests, all of which demonstrate the problem (they fail on the current master branch) and the fix (pass on the PR).

alloy:
Nice 👌

@jstejada This looks good to me.

packages/relay-runtime/util/recycleNodesInto.js
```diff
@@ -30,29 +30,31 @@ function recycleNodesInto<T>(prevData: T, nextData: T): T {
   const prevArray = Array.isArray(prevData) ? prevData : null;
   const nextArray = Array.isArray(nextData) ? nextData : null;
   if (prevArray && nextArray) {
+    const isFrozen = Object.isFrozen(nextArray);
     canRecycle =
```
jstejada: I might be missing something, but would it make sense to just assign `canRecycle` to `false` if it is frozen?

packages/relay-runtime/util/recycleNodesInto.js
```diff
@@ -30,29 +30,31 @@ function recycleNodesInto<T>(prevData: T, nextData: T): T {
...
+        return wasEqual && nextValue === prevArray[ii];
       }, true) && prevArray.length === nextArray.length;
   } else if (!prevArray && !nextArray) {
     // Assign local variables to preserve Flow type refinement.
     const prevObject = prevData;
     const nextObject = nextData;
     const prevKeys = Object.keys(prevObject);
     const nextKeys = Object.keys(nextObject);
+    const isFrozen = Object.isFrozen(nextObject);
     canRecycle =
```
jstejada: same here?

packages/relay-runtime/util/recycleNodesInto.js
mike-marcacci: Hey @jstejada, that's a good question. I would *not* want to prematurely decide that `canRecycle` is `false`, since it's still possible for `wasEqual && nextValue === prevArray[ii]` to be true, and we wouldn't want to lose out on the ability to recycle deep objects just because it is frozen.

jstejada: Ah I see, thanks for the explanation!

packages/relay-runtime/util/recycleNodesInto.js
```diff
@@ -30,29 +30,31 @@ function recycleNodesInto<T>(prevData: T, nextData: T): T {
   const prevArray = Array.isArray(prevData) ? prevData : null;
   const nextArray = Array.isArray(nextData) ? nextData : null;
   if (prevArray && nextArray) {
+    const isFrozen = Object.isFrozen(nextArray);
```
jstejada: Since we only freeze in DEV, perhaps we should only perform this isFrozen check in dev?
cc @leebyron

mike-marcacci: Ya, I thought about that too but figured isFrozen should be pretty darn quick, and there’s no reason to try to modify a frozen object ever. Happy to add that though.

jstejada: Sounds good, I can go ahead and add that tiny change before landing this PR if that sounds good to you!

mike-marcacci: 👍 Sure!

mike-marcacci: Hey, not sure if you already made the change over at FB, but I added the dev check here. Let me know if there's anything more I can do to help get this merged...

packages/react-relay/classic/legacy/store/__tests__/recycleNodesInto-test.js
```diff
@@ -142,6 +164,13 @@ describe('recycleNodesInto', () => {
       expect(recycleNodesInto(prevData, nextData)).toBe(prevData);
     });
 
+    it('recycles arrays with equal objects without mutating frozen `nextData`', () => {
+      const prevData = [{foo: 1}, {bar: 2}];
+      const nextData = [{foo: 1}, {bar: 2}];
+      Object.freeze(nextData);
+      expect(recycleNodesInto(prevData, nextData)).toBe(prevData);
```
jstejada: This test is definitely wrong, correct? If `nextData` is frozen and we don't expect to mutate it, then we would expect `.toBe(nextData)`

packages/react-relay/classic/legacy/store/__tests__/recycleNodesInto-test.js
```diff
@@ -142,6 +164,13 @@ describe('recycleNodesInto', () => {
...
+    });
```
jstejada: Can we add another test for recycling in objects as well?

packages/relay-runtime/util/recycleNodesInto.js
```diff
@@ -30,29 +30,33 @@ function recycleNodesInto<T>(prevData: T, nextData: T): T {
   const prevArray = Array.isArray(prevData) ? prevData : null;
   const nextArray = Array.isArray(nextData) ? nextData : null;
   if (prevArray && nextArray) {
+    const isFrozen =
```
jstejada: @mike-marcacci we circled back on this while reviewing internally. You previously mentioned why you make the `isFrozen` check once per value, but it seems that given that we only do this in dev mode its fine if we just make an early check here:

```
+  // In dev builds, values may be frozen, in which case they cannot be recycled into.
+  if (__DEV__) {
+    if (Object.isFrozen(nextData) {
+      return nextData;
+    }
+  }
```

packages/react-relay/classic/legacy/store/__tests__/recycleNodesInto-test.js
mike-marcacci: Hmm, I'm pretty sure this test is correct. It's testing that even if objects are frozen, we still recycle `prevData` object if it is deeply equal to `nextData`. This is important for preventing identical renders from triggering react diffing downstream.

The behavior of frozen objects differs from mutable ones only in that we can't preserve object equity in unchanged branches of the state tree that *does* have changes.

Given this `prevData`:
```
       A1
  B1        B2
C1  C2    C3  C4
```

we can recycle the entire object of `nextData` looks like this:
```
       A1
  B1        B2
C1  C2    C3  C4
```
...whether or not the anything is frozen.

However, if our `nextData` looks like this:
```
       A1
  B1        B2
C1  C2    C3  C5
```
...we can recycle A1 and B1 from `prevData`, but this requires mutating A1, which crashes the whole app if A1 is frozen. If we were really trying to optimize the frozen object case, it would be possible to create a new A1 to hold the `prevData` B1 and `nextData` B2, but since this only effects the dev environment, that's optimization that is not worth doing.
```

File: packages/react-relay/classic/legacy/store/__tests__/recycleNodesInto-test.js (hunk 5 above)
Comment (mike-marcacci) [reply packages/react-relay/classic/legacy/store/__tests__/recycleNodesInto-test.js]: Sure thing!

File: packages/relay-runtime/util/recycleNodesInto.js (hunk 6 above)
Comment (mike-marcacci) [reply packages/relay-runtime/util/recycleNodesInto.js]: Either way is fine by me! Am I correct in assuming that these dev flags get replaced with constants in the build process, and then all constant paths get compiled out? If so, then it really doesn't matter – you'll end up with no checks in production regardless of where they happen in dev.

**Comment by mike-marcacci** [comment]
I was thinking about this again (encountered this in another dev environment this morning) and just wanted to clarify one thing from my explanation above, which supports the argument that these isFrozen checks stay where I put them:

In the second example, we CAN and DO recycle B1 from `prevData` as long as `A2` is mutable. This way, if `B1` is frozen but unchanged, we have the same behavior on dev and production. Because this freezing is done on custom scalars, which are by definition leaf nodes from graphql's perspective, this by far the most common case, and should work correctly even in dev. Since it doesn't effect a production build at all (assuming these checks get compiled out), this definitely feels like the correct behavior.

File: packages/react-relay/classic/legacy/store/__tests__/recycleNodesInto-test.js (hunk 5 above)
Comment (mike-marcacci) [reply packages/react-relay/classic/legacy/store/__tests__/recycleNodesInto-test.js]: I went to add the requested test, but it looks like I already had one that does this:
`does not mutate frozen equal parent objects with equal leaf objects`

So this should be good to go.

**Comment by mike-marcacci** [comment]
As I commented on your review, the comment test you requested already exists, so I believe this PR is good to go. I've switched my `process.env` checks with `__DEV__` globals, to follow the style used throughout the rest of relay. I also confirmed that these checks are stripped from production builds, so I don't think there's any reason to move them.

I'm sorry to be pushy, but this is a tremendous headache for me. Every time anyone runs `yarn` in development, they have to apply a patch to relay. In the meantime, we're going to ditch the patches and run off a fork I've temporarily published to `@boltline/relay-runtime`. Until this or another fix get merged I'll keep this in sync with Facebook's `relay-runtime` releases.

Please let me know if there's anything else I can do to help.

**Comment by mike-marcacci** [comment]
I cleaned up 2 tests and rebased from master to clean things up, but it looks like that pulled in some breakage. I'll rebase again once master is passing.

**Comment by jquense** [comment]
Can we get another look here? This is super annoying to have to deal with.

**Comment by mike-marcacci** [comment]
Hey @jquense when you get a chance would you mind looking over this again in the context of my above comments?

**Comment by dbslone** [comment]
@jstejada Any updates on when this will be merged?

**Comment by jstejada** [comment]
hey @mike-marcacci, sorry for the delay in getting back to this. I'll take a look at your updated changes and comments soon. Thanks for taking the time to update this!

**Comment by mike-marcacci** [comment]
For anybody else using my fork, `@boltline/relay-runtime` has been updated to `1.6.0` in step with FB's `relay-runtime`.

**Comment by virtualirfan** [comment]
Just checking in … is there a plan to merge this PR? I'm hitting this for doing subscriptions of a Json scalar type. I imagine a lot of people are hitting this, no?

**Comment by virtualirfan** [comment]
@mike-marcacci Wanted to test your fork … is that repo public? Couldn't find it.

**Comment by mike-marcacci** [comment]
Hey @virtualirfan, I published it as `@boltline/relay-runtime` (since it was being used in the boltline project) but the repo I my personal one: https://github.com/mike-marcacci/relay

I didn't expect this to take so long, so I didn't bother to do a neat fork and change the readme/metadata/etc from relay's to mine/

**Comment by virtualirfan** [comment]
Just wanted to report that the latest `@boltline/relay-runtime` published NPM package is working well for me via aliasing. See below.

@jstejada a validation point that might help you pull this PR into the next release.

```js
module.exports = {
...
  resolve: {
...
    alias: {
...
      'relay-runtime': '@boltline/relay-runtime',
    }
  }
}
```

dbslone:
Hey @jstejada what is the status on this being merged?

ivosabev:
What is the status of this, it has been a year?

mike-marcacci:
@josephsavona there can't possibly be a performance impact in production builds because all the changes are stripped during the production build process, as mentioned above.

mike-marcacci:
@josephsavona, ah that makes sense. I didn't realize that FB utilized multiple build systems internally; that seems like quite a challenge to keep track of!

I will say that while I know this is a particularly hot piece of code, the fact that `__DEV__` is checked before `Object.isFrozen` means that the added cost in naive production is simply the evaluation of a boolean.

You made me curious about this, so I [wrote a quick benchmark](https://jsperf.com/relay-issue-2193/1) to make sure my suspicions were correct. Perhaps this will be helpful. :)

PR #2217, opened 2017-12-04
Update <OperationType> to inline code

jayjaycross:
Add back-ticks to `<OperationType>` so that is not interpreted as an HTML tag.

screen shot 2017-12-03 at 9 51 03 pm


kassens:
Thanks!

PR #2218, opened 2017-12-04
disable text transform on buttons

graingert:


graingert:
uppercase text is unreadable, especially when the text used to be camelcase.

kassens:
Sorry to have this linger a bit, but I was waiting for Docusaurus to go open source. I don't think we should start forking styles here and instead see if that's something that could be changed upstream at https://github.com/facebook/docusaurus

graingert:
Can we at least merge this for the fixed custom CSS 404?

kassens:
Good call, I'll update the this PR before landing than!

PR #2219, opened 2017-12-05
Added documentation about the disposable.

jasoncarty:


edvinerikson:
Would it make sense to add a similar example as what refetch have as well?
https://facebook.github.io/relay/docs/en/refetch-container.html

eg
```js
type Disposable = {
  dispose(): void,
};

requestSubscription(
  environment: Environment,
  config: {
    subscription: GraphQLTaggedNode,
    variables: Variables,
    onCompleted?: ?() => void,
    onError?: ?(error: Error) => void,
    onNext?: ?(response: ?Object) => void,
    updater?: ?(store: RecordSourceSelectorProxy, data: SelectorData) => void,
    configs?: Array<RelayMutationConfig>,
  },
) => Disposable;

jasoncarty:
That could be helpful!

alloy:
@jstejada This looks good to me.

jstejada:
thanks!

PR #2229, opened 2017-12-12
De-emphasize Relay Classic docs

kassens:
Small CSS hack to make the RelayClassic docs less emphasized.

If this sounds good, I'll update it after we re-organize the sections.

image


kassens:
We'll figure out how to remove them instead and keep them as an old version.

PR #2249, opened 2017-12-26
Inflect DataID field from Node interface.

alloy:
This PR makes it possible to inflect the `DataID` field to use from the user’s schema, rather than assuming it to be `id`.

## How

* The ID is either the single `ID` field of the `Node` interface in the user’s schema or otherwise falls back to `id`.
* The ID field is selected using the `__id` alias.
* At runtime the `__id` field in the response data is used as store `DataID`.

## Example

If your schema looks like this:

```graphql
interface Node {
  globalObjectID: ID!
}

type Artist implements Node {
  globalObjectID: ID!
  name: String!
}
```

…and you have a query like:

```graphql
{
  artist(slug: "banksy") {
    name
  }
}
```

…then Relay will transform that to:

```graphql
{
  artist(slug: "banksy") {
    __id: globalObjectID
    name
  }
}
```

…and use the value of `__id` as the key when writing the record to the store.


packages/relay-runtime/store/RelayResponseNormalizer.js
```diff
@@ -305,7 +305,7 @@ class RelayResponseNormalizer {
       );
 
       const nextID =
-        item.id ||
+        (field.idField && item[field.idField]) ||
```
alloy: These two changes are the only runtime changes.

packages/relay-compiler/codegen/RelayCodeGenerator.js
```diff
@@ -66,6 +66,13 @@ const RelayCodeGenVisitor = {
         node.metadata.idType,
       );
     }
+    if (node.kind === 'Fragment' || node.kind === 'LinkedField') {
+      const selections = flattenArray(node.selections);
+      const idFieldSelection = selections.find(
+        selection => selection.metadata && selection.metadata.isDataID,
+      );
+      node.idField = idFieldSelection && idFieldSelection.name;
```
josephsavona: since this gets toJSON'd we probably want to make sure it's always `null` if there is no idFieldSelection, so that all Fragments and LinkedFields have the same property set and same hidden class

josephsavona: Also, this logic should probably go in a transform and have its own test, not the code generator

alloy:
@josephsavona I got a couple of notification emails from GH with comments from you on this PR, but I don’t see them here. Is it a GH fluke or did you remove the comments?

josephsavona:
Sorry about that. I meant to start a code review and accidentally posted not fully thought out comments instead - will take another look when I have computer time (holidays, baby, etc).

alloy:
Oh no worries at all, I just got confused. Please enjoy your holidays and baby instead! :pray: This can wait.

alloy:
Rebased on master.

packages/relay-compiler/codegen/RelayCodeGenerator.js
```diff
@@ -53,6 +53,16 @@ function generate(node: Batch | Fragment): RequestNode | ConcreteFragment {
...
+  enter(node) {
+    if (node.kind === 'Fragment' || node.kind === 'LinkedField') {
+      const selections = flattenArray(node.selections);
+      const idFieldSelection = selections.find(
+        selection => selection.metadata && selection.metadata.isDataID,
+      );
+      node.idField = idFieldSelection && idFieldSelection.name;
+    }
+    return node;
+  },
```
alloy: @josephsavona I think in your retracted initial review you said that this should become its own transformer, right? I think that makes sense, it’s kinda awkward here.

josephsavona: Yeah this should be its own transform rather than happening at codegen time

alloy:
> adding an extra property (that is almost always the same redundant "id" value) to tons of fields increases the size of query modules

Ah yes, good point. I hadn’t considered bundle size enough, probably because 99% of our Relay code thus far exists in our mobile app.

If that turns out to be a major concern, then we can just special-case `id` and revert to the previous behaviour of not emitting `idField` and hardcoding checks for `id` at runtime.

> it also incurs an extra condition/branch at runtime when reading

Related to the previous item, I think that even when reverting to special-casing the `id` field, this extra branch would remain because we’d have to first determine if an `idField` is specified before falling back to the hardcoded `id` field, otherwise we end up selecting an `id` field that should not ben selected.

packages/relay-compiler/codegen/RelayCodeGenerator.js
alloy: Will do 👍

alloy:
Ok, I think I addressed all your feedback, I’ve updated the original explanation to reflect this. The overall implementation is definitely a lot simpler 👌 

There’s a fork of the TODO example app that uses this here https://github.com/alloy/relay-examples/tree/custom-node-id/todo and we’re pulling this into our app for further testing as well https://github.com/artsy/emission/pull/985.

alloy:
@leebyron Ok, I think I addressed all your feedback, I’ve updated the original explanation to reflect this. The overall implementation is definitely a lot simpler 👌 

There’s a fork of the TODO example app that uses this here https://github.com/alloy/relay-examples/tree/custom-node-id/todo and we’re pulling this into our app for further testing as well https://github.com/artsy/emission/pull/985.

alloy:
Rebased on master.

alloy:
@leebyron FYI We’re now running this updated code in production in both our web and iOS app. You can see an example query [here](http://metaphysics-staging.artsy.net/?operationName=GeneContentsArtistsQuery&query=query%20GeneContentsArtistsQuery(%0A%20%20%24geneID%3A%20String!%0A)%20%7B%0A%20%20gene(id%3A%20%24geneID)%20%7B%0A%20%20%20%20...Artists_gene%0A%20%20%20%20__id%0A%20%20%7D%0A%7D%0A%0Afragment%20Artists_gene%20on%20Gene%20%7B%0A%20%20__id%0A%20%20artists%3A%20artists_connection(first%3A%2010%2C%20after%3A%20%22%22)%20%7B%0A%20%20%20%20pageInfo%20%7B%0A%20%20%20%20%20%20hasNextPage%0A%20%20%20%20%20%20endCursor%0A%20%20%20%20%7D%0A%20%20%20%20edges%20%7B%0A%20%20%20%20%20%20node%20%7B%0A%20%20%20%20%20%20%20%20__id%0A%20%20%20%20%20%20%20%20...ArtistRow_artist%0A%20%20%20%20%20%20%20%20__typename%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20cursor%0A%20%20%20%20%7D%0A%20%20%7D%0A%20%20filter_aggregations%3A%20filtered_artworks(aggregations%3A%20%5BMEDIUM%2C%20TOTAL%2C%20PRICE_RANGE%2C%20DIMENSION_RANGE%5D%2C%20size%3A%200%2C%20include_medium_filter_in_aggregation%3A%20true)%20%7B%0A%20%20%20%20...TotalCount_filter_artworks%0A%20%20%20%20aggregations%20%7B%0A%20%20%20%20%20%20slice%0A%20%20%20%20%20%20...Dropdown_aggregation%0A%20%20%20%20%7D%0A%20%20%20%20__id%0A%20%20%7D%0A%7D%0A%0Afragment%20ArtistRow_artist%20on%20Artist%20%7B%0A%20%20name%0A%20%20href%0A%20%20...Follow_artist%0A%20%20artworks%3A%20artworks_connection(first%3A%206)%20%7B%0A%20%20%20%20...Fillwidth_artworks%0A%20%20%7D%0A%20%20__id%0A%7D%0A%0Afragment%20TotalCount_filter_artworks%20on%20FilterArtworks%20%7B%0A%20%20counts%20%7B%0A%20%20%20%20total%0A%20%20%7D%0A%20%20__id%0A%7D%0A%0Afragment%20Dropdown_aggregation%20on%20ArtworksAggregationResults%20%7B%0A%20%20slice%0A%20%20counts%20%7B%0A%20%20%20%20name%0A%20%20%20%20id%0A%20%20%20%20count%0A%20%20%20%20__id%0A%20%20%7D%0A%7D%0A%0Afragment%20Follow_artist%20on%20Artist%20%7B%0A%20%20__id%0A%20%20id%0A%20%20is_followed%0A%7D%0A%0Afragment%20Fillwidth_artworks%20on%20ArtworkConnection%20%7B%0A%20%20edges%20%7B%0A%20%20%20%20node%20%7B%0A%20%20%20%20%20%20__id%0A%20%20%20%20%20%20image%20%7B%0A%20%20%20%20%20%20%20%20aspect_ratio%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20...FillwidthItem_artwork%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A%0Afragment%20FillwidthItem_artwork%20on%20Artwork%20%7B%0A%20%20image%20%7B%0A%20%20%20%20placeholder%0A%20%20%20%20url(version%3A%20%22large%22)%0A%20%20%20%20aspect_ratio%0A%20%20%7D%0A%20%20href%0A%20%20...Metadata_artwork%0A%20%20...Save_artwork%0A%20%20__id%0A%7D%0A%0Afragment%20Metadata_artwork%20on%20Artwork%20%7B%0A%20%20...Details_artwork%0A%20%20...Contact_artwork%0A%20%20__id%0A%7D%0A%0Afragment%20Save_artwork%20on%20Artwork%20%7B%0A%20%20__id%0A%20%20id%0A%20%20is_saved%0A%7D%0A%0Afragment%20Details_artwork%20on%20Artwork%20%7B%0A%20%20href%0A%20%20title%0A%20%20date%0A%20%20sale_message%0A%20%20cultural_maker%0A%20%20artists(shallow%3A%20true)%20%7B%0A%20%20%20%20__id%0A%20%20%20%20href%0A%20%20%20%20name%0A%20%20%7D%0A%20%20collecting_institution%0A%20%20partner(shallow%3A%20true)%20%7B%0A%20%20%20%20name%0A%20%20%20%20href%0A%20%20%20%20__id%0A%20%20%7D%0A%20%20sale%20%7B%0A%20%20%20%20is_auction%0A%20%20%20%20is_live_open%0A%20%20%20%20is_open%0A%20%20%20%20is_closed%0A%20%20%20%20__id%0A%20%20%7D%0A%20%20__id%0A%7D%0A%0Afragment%20Contact_artwork%20on%20Artwork%20%7B%0A%20%20_id%0A%20%20href%0A%20%20is_inquireable%0A%20%20sale%20%7B%0A%20%20%20%20is_auction%0A%20%20%20%20is_live_open%0A%20%20%20%20is_open%0A%20%20%20%20is_closed%0A%20%20%20%20__id%0A%20%20%7D%0A%20%20partner(shallow%3A%20true)%20%7B%0A%20%20%20%20type%0A%20%20%20%20__id%0A%20%20%7D%0A%20%20sale_artwork%20%7B%0A%20%20%20%20highest_bid%20%7B%0A%20%20%20%20%20%20display%0A%20%20%20%20%20%20__id%3A%20id%0A%20%20%20%20%7D%0A%20%20%20%20opening_bid%20%7B%0A%20%20%20%20%20%20display%0A%20%20%20%20%7D%0A%20%20%20%20counts%20%7B%0A%20%20%20%20%20%20bidder_positions%0A%20%20%20%20%7D%0A%20%20%20%20__id%0A%20%20%7D%0A%20%20__id%0A%7D%0A&variables=%7B%22geneID%22%3A%20%22minimalism%22%7D), but note that our schema uses `__id` for the GUID and thus most selections are just `__id`, only on line 150 will you find a `__id: id` selection.

alloy:
@leebyron Could I get another review?

alloy:
I’ve gone ahead and squashed all the commits now to ease rebasing on my end. I’ve also rebased, but alas can’t run the tests at the moment due to the issue described in #2407

alloy:
Tests pass again 👍

docs/PrinciplesAndArchitecture-Runtime.md
```diff
@@ -17,7 +17,7 @@ The Relay runtime is a full-featured GraphQL client that is designed for high pe
...
+For users of classic Relay, note that the runtime makes as few assumptions as possible about GraphQL. Compared to earlier versions of Relay there is no concept of routes, there are no limitations on mutation input arguments or side-effects, arbitrary root fields just work, etc. Like classic Relay, modern Relay still uses the `Node` interface for object identification purposes; however, rather than always assuming the globally unique field to be `id`, modern Relay will inflect it from the `Node` interface by searching for either an `ID` or `ID!` field.
```
jstejada: nit: can we use a different term than `inflect`, or not use it at all? (same for the title of this PR) I might just be me as a non-native english speaker, but I didn't know what inflect means, and googling the definition didn't really help until I actually looked at what this PR was doing.

maybe something like: "Relay Modern will find a field of type ID or ID! defined in the Node interface and use it as a globally unique ID"

packages/relay-compiler/core/RelayIRTransforms.js
```diff
@@ -48,6 +48,7 @@ const relayCommonTransforms: Array<IRTransform> = [
   RelayRelayDirectiveTransform.transform,
   RelayMaskTransform.transform,
   RelayDeferrableFragmentTransform.transform,
+  RelayGenerateIDFieldTransform.transform,
```
jstejada: sorry I'm not super familiar with the compiler code. If we weren't using this transform here before, where were we using it?

packages/relay-compiler/transforms/RelayGenerateIDFieldTransform.js
```diff
@@ -11,104 +11,113 @@
...
 
-type State = {
-  idField: ScalarField,
-};
-
 /**
- * A transform that adds an `id` field on any type that has an id field but
- * where there is no unaliased `id` selection.
+ * A transform that adds a `__id` field on any type that has a `Node` or `id`
+ * field but where there is no unaliased `__id` selection.
```
jstejada: so we'll always add an `__id` selection, *unless* there's already an *unaliased* `__id` selection.

what happens when there's already an aliased `__id` selection?, will we add another `__id` selection too?

packages/relay-compiler/transforms/RelayGenerateIDFieldTransform.js
```diff
@@ -11,104 +11,113 @@
...
 
-function visitLinkedField(field: LinkedField, state: State): LinkedField {
-  const transformedNode = this.traverse(field, state);
+function visitNodeWithSelections<T: Fragment | LinkedField>(node: T): T {
+  const transformedNode = this.traverse(node);
 
-  // If the field already has an unaliased `id` field, do nothing
-  if (hasUnaliasedSelection(field, ID)) {
+  // If the field already has an `__id` selection, do nothing.
+  if (hasSelection(node, '__id')) {
```
jstejada: answering my own question from above, it seems that this won't add an `__id` selection if a selection for `__id` already exists, be it aliased or unaliased. Maybe update the comment above to reflect that?

packages/relay-compiler/transforms/RelayGenerateIDFieldTransform.js
```diff
@@ -11,104 +11,113 @@
...
-    name: ID,
-    type: idType,
-  };
-  const state = {
-    idField,
-  };
-  return IRTransformer.transform(
-    context,
-    {
-      LinkedField: visitLinkedField,
```
jstejada: sorry I'm probably missing some context, why did we previously only apply this transform to `LinkedFields`, and now we do so for `Fragment`s too?

packages/relay-compiler/transforms/RelayGenerateIDFieldTransform.js
```diff
@@ -11,104 +11,113 @@
...
     };
   }
 
-  // If the field type is abstract, then generate a `... on Node { id }`
-  // fragment if *any* concrete type implements Node. Then generate a
-  // `... on PossibleType { id }` for every concrete type that does *not*
-  // implement `Node`
+  // - If the field type is abstract, then generate a `... on Node { __id: id }`
+  //   fragment if *any* concrete type implements `Node`. Then generate a
+  //   `... on PossibleType { __id: id }` for every concrete type that does
```
jstejada: what if `PossibleType` doesn't have an id field? can we add test coverage for this?

alloy:
New TODOs

* `RelayDeferrableFragmentTransform` hardcodes `id`
* `RelayViewerHandleTransform` hardcodes `id`
* `RelayValidator` disallows `id` alias

alloy:
This rebased version has a couple of failing tests, that I have left in as a good reminder of what to work on and I also need to address @jstejada’s feedback, but for now I want to first see if it actually works in our apps.

clentfort:
What is the current progress on this issue? Is there any way one could help to advance this PR?

alloy:
This patch has served us well, but we’ve decided that we are going to sunset this PR. The reasons are:

* Redoing this PR for current `master` is non-trivial and was a constant source for us not being able to upgrade to latest Relay easily.
* Since the creation of this patch, the GraphQL spec has decided to reserve fields starting with a dunderscore for internal use only, so even when/if we’d ever land this we at Artsy would still be left with a schema that all of our tooling has started complaining about.

In the end, we decided to start experimentation with a v2 of our schema in which we transform v1 to rename all of our ID fields to conform to the Node spec. Things are looking good thus far and we’ve been able to upgrade to Relay 4 easily.

alloy:
@clentfort What a coincidence that you would ask that on the day that I was planning on sunsetting this PR 😞

If you are interested in taking a solution like this to the finish line, [some work has been done](https://github.com/facebook/relay/commit/311412b50cd1ae451d2e8b375fd5fe03522f41d2) to being able to provide a custom DataID value at runtime. However, unless your DataID would rely on [`id`](https://github.com/facebook/relay/blob/master/packages/relay-compiler/transforms/RelayGenerateIDFieldTransform.js) and [`__typename`](https://github.com/facebook/relay/blob/master/packages/relay-compiler/transforms/RelayGenerateTypeNameTransform.js), which are already being queried for by default by relay, you would have to do some work to provide your own query transform that, at compile time, inserts the fields you need selected.

PR #2254, opened 2018-01-03
Update Modern-Mutations.md

nodkz:
`commitMutation` imported from`react-relay` does not update store before calling `onCompleted` callback. But imported from `relay-runtime` works as described in docs:
```
  * `onCompleted`: Callback function executed when the request is completed 
and the in-memory Relay store is updated with the `updater` function. 
Takes a `response` object, which is the "raw" server response, and `errors`, 
an array containing any errors from the server.
```

### DIFF
--- a/docs/Modern-Mutations.md
+++ b/docs/Modern-Mutations.md
@@ -54,7 +54,7 @@ commitMutation(
 Example of a simple mutation:
 
 ```javascript
-import {commitMutation, graphql} from 'react-relay';
+import {commitMutation, graphql} from 'relay-runtime';
 
 const mutation = graphql`
   mutation MarkReadNotificationMutation(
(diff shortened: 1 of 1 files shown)


kassens:
I think for typical product code using react-relay. All the import from that package should be enough.

Looking at the source, `commitMutation` from `react-relay` is actually just a re-export from `RelayRuntime`: https://github.com/facebook/relay/blob/3173131eecae00c4dcf7e2c564d45c6ac3c97731/packages/react-relay/modern/ReactRelayPublic.js#L50

Am I missing something?

nodkz:
Yep, it also my common problem when I try to understood what happens - code on github in master branch shows new files, but last npm build contains some old version of code.

Sorry, It's too difficult for me to find exact changes among last 3 months from mobile. But in my wrapper of  relayEnv `commitMutation` imported from react-relay and relay-runtime had different behavior for v1.4.1.

kassens:
Sorry for the long gap between releases. 👎 

I think this should work the same now and I think having typical product code in React only require from a single module seems like a good thing?

nodkz:
Import from single module ofcource bett

nodkz:
Import from single module of course better.

Feel free to close this PR.

And we are waiting new versionof Relay 😉

PS. Espesially a new version of `graphql-js` where is fixed export of getDescription method. Cc @leebyron

jstejada:
@nodkz not sure if you meant to re-open, so just going to close this again

PR #2255, opened 2018-01-03
Add Cirrus CI as a user

fkorotkov:
We use Relay for our [front end](https://github.com/cirruslabs/cirrus-ci-web) and we love it!

PR #2257, opened 2018-01-04
Update USERS.md to include Cirrus CI

fkorotkov:
As mentioned in https://github.com/facebook/relay/pull/2255


kassens:
Cool, merging this for now, but we can probably deprecate USERS.md in the future.

Might be a good opportunity to only copy active users over as well.

Issue #1656, opened 2017-04-20
[Modern] Client schema docs and examples


ivosabev:
Any idea when docs and examples would be available?

> Client Schema Extensions 
> The Relay Modern Core adds support for client schema extensions. These allow Relay to conveniently store some extra information with data fetched from the server and be rendered like any other field fetched from the server. This should be able to replace some use cases that previously required a Flux/Redux store on the side.

https://facebook.github.io/relay/docs/new-in-relay-modern.html#client-schema-extensions

leethree:
👍 I was trying to find any reference to how to do "client schema extensions" but failed.

mjmahone:
My suspicion is it'll happen in an example in [RelayExamples](https://github.com/relayjs/relay-examples) before hitting official documentation here. I don't have a timeline, but that's your best bet for taking advantage of this feature. It already works (at least, it works on my machine :)), and you can find references to it via things like the [RelayFilterDirectivesTransform](https://github.com/facebook/relay/blob/master/packages/relay-compiler/transforms/RelayFilterDirectivesTransform.js), but getting this to work outside the FB environment hasn't yet happened.

abhishiv:
I would love to know more about this to as well.

This is what I figured out till now

1) `extendSchema` function to extend schema. Also it's interesting to note that transformers are using this function.

Still missing

1) How do I manually update the Relay store with data.

josephsavona:
I tried documenting this the other day and realized that there is an issue with the OSS compiler that prevents schema extensions from being defined. Once this is fixed we can document this feature. 

Cc @kassens

unirey:
Any news on this? 
Maybe you could show us how it's used? Maybe we could help with PRs or test in different environments.

josephsavona:
@unirey The idea is that you can do something like this:

```
// client-schema.graphql
extend type SomeType {
  client: Foo
}
type Foo {
  name: String
}

// compile script:
relay-compiler --schema schema.graphql --client-schema client-schema.graphql ...

// container fragment
graphql`
  fragment on SomeType {
    client
  }
`

// mutation updater
function updater(store) {
  const someValue = store.get(idOfValueOfSomeType);
  someValue.setValue("Foo", "client");  // value, field name
}
```

The missing part is that the oss compiler binary doesn't expose a way to tell the compiler about the client schema - all the primitives exist in the compiler to support client fields, and the runtime supports them as regular fields.

fsenart:
Hi there 
[...]

PR #2264, opened 2018-01-07
Add support for client schemas

edvinerikson:
Makes it possible to have a client schema. Not sure if it should be changed to only allow one schema file or if this is good enough.

Forked @kassens changes and made some fixes to get it to work.


sibelius:
@kassens @jstejada can we get this for 1.5.0?

jstejada:
addressing #1656

alloy:
@jstejada It looks like you tried to import this patch, but I guess it failed? Can you see what’s up?

jstejada:
@edvinerikson thanks!

PR #2271, opened 2018-01-08
Make graphql a peer dependency

jquense:
`graphql-js` doesn't work if there are multiple copies installed in a project, regardless of if they are the same version. Having graphql be direct dep makes it really hard to consume relay-compiler in additional tools, like GatsbyJs, because we need to depend on the package manager deduping any instances, which can fail to happen in plenty of situations.

CC @KyleAMathews


alloy:
@jstejada This should be a good time saver for those that run into the issue. Any FB considerations?

packages/relay-compiler/graphql-compiler/package.json
```diff
@@ -7,9 +7,11 @@
   "dependencies": {
     "chalk": "^1.1.1",
     "fb-watchman": "^2.0.0",
-    "graphql": "^0.12.3",
     "immutable": "~3.7.6"
   },
+  "peerDependency": {
```
mattkrick: should this be `peerDependencies`?

jquense: jeez, oops not my finest PR :P

alloy: Great catch @mattkrick!

@jstejada I don’t know if the bot is still importing, but this new addition will be important.

jquense:
🎉

Issue #2039, opened 2017-08-17
Flowtype generation for queries with Inline Fragments 


ajhyndman:
This is a proposed improvement to flowtype generation for the relay compiler.

When writing queries against a union type, we use Inline Fragments to select fields from one member of the union type.  In the generated flowtypes, I am seeing a single object type with optional fields.  Flow does support union types as well, however.  If we are able to map GraphQL union types to Flow union types, client code we can get improved static guarantees and reduce the amount of pessimistic runtime checks that we need to execute.

e.g.

With a schema that looks like this:

```graphql
type Query {
  users: [User]!
}

type User = StudentUser | TeacherUser

type StudentUser {
  name: String!
  classes: [String]!
  age: Int!
}

type TeacherUser {
  name: String!
  position: String!
  title: String!
}
```

If my query looks like this:

```graphql
query AppQuery {
  users {
    ... on StudentUser {
      classes
      age
    }
    ... on TeacherUser {
      position
      title
    }
  }
}
```

I expect the generated flowtypes to look like this:

```javascript
type AppQueryResponse = {|
  users: $ReadOnlyArray<?{|
    classes: $ReadOnlyArray<?string>;
    age: number;
  |} | {|
    position: string;
    title: string;
  |}>;
|}
```

But instead, I see this:
```javascript
type AppQueryResponse = {|
  users: $ReadOnlyArray<?{|
    classes?: $ReadOnlyArray<?string>;
    age?: number;
    position?: string;
    title?: string;
  |}>;
|}
```

kassens:
We currently create the first case when you include `__typename` in the fragment. This then allows flow to distinguish the cases in code like:
```
if (response.__typename === 'StudentUser') {
  console.log(users.age);
}
```
I found that always creating the unions sometimes makes for awkward code to actually get the values. Certainly not 100% satisfied here.

ajhyndman:
@kassens Thanks for the tip!  I'd love to see that documented somewhere.

Can you recall any scenarios in which the former case is more awkward?  It seems to me that in the worst case, you could the the first case as the second when consuming without code changes, couldn't you?

DennisKo:
We actually encountered that problem as well.

Query:

```
feed {
  edges {
    node {
      __typename
      ... on Resource1 {
        id
        __typename
      }
      ... on Resource2 {
        id
        __typename
      }
      ... on Resource3 {
        id
        __typename
      }
      ... on Resource4 {
        id
        __typename
      }
  }
}
```
[...]

PR #2273, opened 2018-01-09
Always generate union types for inline fragment selections

ajhyndman:
### Overview

Removes an undocumented behaviour which generated different compiled flowtypes based on the presence of the `__typename` key selection.

I opened this PR in relation to issue #2039, which I opened a few months ago.

### Details

My team is working with relay modern and generated flowtypes at some scale in a rewrite of our API layer.  We find that the merged, optional key format output is never more desirable than union types.  Merging keys always results in a large quantity of redundant null checks, and makes it much harder for developers to identify the real, possible failure conditions.

Thanks to @kassens we've been selecting the `__typename` key in every fragment spread, which can force our preferred output.

However, there are a few downsides with this approach:

1. It gets confusing
    * In some more complicated fragment compositions, it can get very confusing trying to work out why relay has decided to prefer one output format vs the other.
1. It's not documented
    * New developers on the team have no insight into why we are selecting this key, nor how to get the desired type output.
1. It has bugs
    * There are some conditions, which I can try to isolate if desired, under which it seems to be impossible to convince relay to output a union type.

### Open Questions

This will probably be a breaking change for any users relying on relay's generated flow types.  It might be good to think about a migration strategy, or an opt-in toggle here.  I'm open to suggestions.

### DIFF
--- a/packages/relay-compiler/core/RelayFlowGenerator.js
+++ b/packages/relay-compiler/core/RelayFlowGenerator.js
@@ -130,14 +130,7 @@ function selectionsToBabel(selections, state: State, refTypeName?: string) {
 
   const types = [];
 
-  if (
-    Object.keys(byConcreteType).length &&
-    onlySelectsTypename(Array.from(baseFields.values())) &&
-    (hasTypenameSelection(Array.from(baseFields.values())) ||
-      Object.keys(byConcreteType).every(type =>
-        hasTypenameSelection(byConcreteType[type]),
-      ))
-  ) {
+  if (Object.keys(byConcreteType).length > 0) {
     for (const concreteType in byConcreteType) {
       types.push(
         groupRefs([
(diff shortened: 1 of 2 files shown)


mike-marcacci:
This is a great improvement! I do occasionally find myself adding `null` checks for what I *know* are impossible conditions. Not a huge pain point, but this is definitely the behavior I initially expected.

ekosz:
@kassens @jstejada Any consideration for this in 1.5.0? It looks like the test failure was due to an intermittent registry issue.

alloy:
There’s still this unanswered question from @ajhyndman in response to @kassens [here](https://github.com/facebook/relay/issues/2039#issuecomment-328899882), which might be important to know about? 

> Can you recall any scenarios in which the former case is more awkward? It seems to me that in the worst case, you could the the first case as the second when consuming without code changes, couldn't you?

mull:
Type generation for an GraphQL schema is currently heavily broken. Selecting the field `id` (and __typename) will generate cases where only the first union type has the field ID and the others suddenly don't have it.

How can I help to get this thing fixed? It's a massive headache and makes the generated types for these queries completely useless 😞

josephsavona:
Thanks for sending a PR. Emitting union types without some clear distinguishing tag (like `__typename`) can cause result in difficult to diagnose type errors. As you noted we already produce unions if developers  fetch  `__typename`,  so maybe the  answer here is to add a compiler option (via `relay-config`) to disallow fetching unions w/o fetching  `__typename`. 

So we understand how the current behavior can be confusing and are open to fixes along the above lines.

PR #2274, opened 2018-01-09
Do not use util.promisify

kassens:
This isn't supported in node 6.

### DIFF
--- a/packages/relay-compiler/graphql-compiler/codegen/SourceControl.js
+++ b/packages/relay-compiler/graphql-compiler/codegen/SourceControl.js
@@ -12,7 +12,19 @@
 'use strict';
 
 const util = require('util');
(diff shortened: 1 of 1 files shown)


packages/relay-compiler/graphql-compiler/codegen/SourceControl.js
```diff
@@ -12,7 +12,19 @@
 'use strict';
 
 const util = require('util');
```
wtgtybhertgeghgtwtg: Shouldn't this be removed, then?

kassens: Yep, good catch!

kassens:
Will merge this into the commit re-applying the reverted commit which broke a merge driver internally.

PR #2283, opened 2018-01-16
[Bug Fix: Relay Compiler] Don’t break on functions sharing names with object prototype methods

robrichard:
Reopened from https://github.com/facebook/relay/pull/2060 with a test case to prevent regressions.


robrichard:
@leebyron @kassens @jstejada the [previous PR](https://github.com/facebook/relay/pull/2060) for this bugfix is abandoned. I have reopened it and added a test case. Is it possible to get this into the next release?

kassens:
Thanks for including a regression test. We'll get this included in the 1.5.0 release.

Issue #2281, opened 2018-01-16
Relay-compiler does not support union types


brad-decker:
```graphql
type HomePage {
  title: String,
  body: String,
}

type TestPage { 
  body: String,
  quote: String,
}

union Page = HomePage | TestPage;

type Query {
  pages: [Page]
}
```

Given the above schema i'd expect to be able to write something like this

```graphql
pages {
  ... on HomePage {
    title
    body
  }
}
```

However i keep getting ```Expected undefined to be a GraphQL leaf type.``` anytime i include any sort of union. There have been a couple related issues that were closed without resolution:

https://github.com/facebook/relay/issues/2245
https://github.com/facebook/relay/issues/1880

Am i doing something wrong or expecting it to work in a way it wasn't meant to?

brad-decker:
To anyone who is having this problem, the issue isn't union types its that one or more of your union'd types is missing an id field. Relay spec requires an id field in order to use the node query, and when it comes to union types that is much more of a requirement as it has to be able to discern the type of item from the id for future queries against the same object. 

To fix the issue in the example above:
```graphql
type HomePage {
  title: String,
  body: String,
}

type TestPage { 
  body: String,
  quote: String,
}

union Page = HomePage | TestPage;

type Query {
  pages: [Page]
}
```

add ID fields

```graphql
type HomePage {
  id: ID,
  title: String,
  body: String,
}

type TestPage {
  id: ID,
  body: String,
  quote: String,
}

union Page = HomePage | TestPage;

type Query {
  pages: [Page]
}
```

you'll also want to add a resolver for that field and use the graphql-relay library's toGlobalId method to convert numeric ids into global ids. 

The way that i diagnosed this issue is by adding console.log's periodically throughout the relay-compiler bin file. Discovered that the culprit method was ```generateIDSelections``` on line 7794 of node_modules/relay-compiler/bin/relay-compiler. 

then saw this comment in the file:
```
// Union or interface: concrete types may implement `Node` or have an `id`
// field
```

steida:
@kassens Can you please only update mysterious error explaining why ID everywhere is actually a good concept? https://medium.com/code-oil/relay-graphql-de-mystifying-node-id-38757121b9c

Issue #2285, opened 2018-01-17
relay-compiler Expected undefined to be a GraphQL leaf type.


graingert:
relay-compiler assumes ID type is available.

https://github.com/facebook/relay/issues/1880#issuecomment-307705428

sibelius:
fixed here https://github.com/facebook/relay/pull/2249

alloy:
Duplicate of #2281 for which PR #2290 exists, although if #2249 indeed fixes this case then it might make sense to go that route instead of slightly patching the symptom.

@jstejada In the meantime, this one can be closed in favour of the other ticket/PR.

jstejada:
thanks @alloy, closing in favor of #2281

PR #2290, opened 2018-01-19
Throw better error for missing idType

brad-decker:
Related to #2281 

A better error message for the situation i ran into would have helped me figure out my mistake much faster. This just checks to see if there is no id field defined in the schema (returns undefined) and if so throws a different error message instead of asserting that 'undefined' is a leaf type (Which is most certainly is not).

### DIFF
--- a/packages/relay-compiler/transforms/RelayGenerateIDFieldTransform.js
+++ b/packages/relay-compiler/transforms/RelayGenerateIDFieldTransform.js
@@ -17,6 +17,7 @@ const {
   assertCompositeType,
   assertLeafType,
 } = require('graphql');
+const invariant = require('invariant');
 const {
   CompilerContext,
   SchemaUtils,
@@ -49,7 +50,10 @@ type State = {
 function relayGenerateIDFieldTransform(
   context: CompilerContext,
 ): CompilerContext {
-  const idType = assertLeafType(context.serverSchema.getType(ID_TYPE));
+  let idType = context.serverSchema.getType(ID_TYPE);
+  invariant(idType, 'Schema does not contain an id field on any types');
+  // reassigning avoids flow error, would love to hear thoughts on this
+  idType = assertLeafType(idType);
   const idField: ScalarField = {
     kind: 'ScalarField',
     alias: (null: ?string),


brad-decker:
There are some flow errors and what not locally but does not appear to be related to my changes -- also if you'd like a test case written for this message i am all for it and would appreciate a little guidance on where that should be best placed. :+1: Thanks

packages/relay-compiler/transforms/RelayGenerateIDFieldTransform.js
```diff
@@ -49,7 +49,11 @@ type State = {
 function relayGenerateIDFieldTransform(
   context: CompilerContext,
 ): CompilerContext {
-  const idType = assertLeafType(context.serverSchema.getType(ID_TYPE));
+  const idType = context.serverSchema.getType(ID_TYPE);
+  if (!idType) {
+    throw new Error('Schema does not contain an id field on any types');
+  }
```
alloy: I’m not sure I understand this. The original code appears to try to get the `ID` type, shouldn’t that be a builtin type that should always exist?

Regardless, can you use `invariant` here instead, which appears to be the way assertions are performed throughout the existing code-base?

alloy: Ok, I should dive deeper into when the `ID` type actually exists, as there’s #2285 by @graingert that also mentions this issue.

Interestingly enough, @sibelius mentions that it would be fixed by my #2249 PR, although I didn’t knew of that issue at the time. @sibelius have you verified that? @brad-decker & @graingert would you mind verifying if #2249 fixes your issues and then please leave feedback on the PR? 🙏

brad-decker: @alloy -- will do. Thanks for checking this out.

anatrajkovska: Ve sakam srcki ste!! <333

sibelius: schema without ID that will throw this error https://github.com/freitasgst/weather-express-graphql/commit/c8ee408afc2e5ac4deaeb45d260800040199f57d

alloy: @sibelius You indicated [here](https://github.com/facebook/relay/issues/2285#issuecomment-358309351) that #2249 fixes it; is that indeed the case or not after all?

sibelius: Not after all, this is the right fix for that

alloy:
Seeing as according to @sibelius this still needs to go in, here’s my review:

* At runtime Relay will generate IDs for records that have none, but I’m not 100% sure off the top of my head if those need the `ID` type to exist in the schema; so do we even really need to throw an error in the case of no `ID` type existing?
* If we do indeed need to throw, then please use `invariant` rather than `throw`.

brad-decker:
@alloy in my experience without the ID type the compiler fails, i'm open to other input on whether we need to throw. I'll just go ahead and change to invariant and then we can decide after that whether its required.

brad-decker:
Checking back in, anything I can do to make this better?

packages/relay-compiler/transforms/RelayGenerateIDFieldTransform.js
```diff
@@ -49,7 +50,10 @@ type State = {
 function relayGenerateIDFieldTransform(
   context: CompilerContext,
 ): CompilerContext {
-  const idType = assertLeafType(context.serverSchema.getType(ID_TYPE));
+  let idType = context.serverSchema.getType(ID_TYPE);
+  invariant(idType, 'Schema does not contain an id field on any types');
+  // reassigning avoids flow error, would love to hear thoughts on this
+  idType = assertLeafType(idType);
```
jstejada: would splitting this into 2 variables help?

alloy:
Sorry for missing your question @brad-decker, @jstejada’s suggestion makes sense to me.

PR #2293, opened 2018-01-21
Language plugin support.

### LINKED ISSUE #2073: Generate Typescript types in addition to Flow
Internally at Facebook, we probably won't get to it, but it'd be awesome to see an external contribution to add Typescript type generation added.

We'd need to make sure it can live nicely alongside Flow, but the Flow type generation is pretty well isolated so that this shouldn't be an issue.
For typing the core Relay API with TS, I assume the easiest would be to add .ts files somewhere?

If there's any interest working on this, happy to chat or discuss here.

alloy:
Closes #2073.
Closes #1710.

In preparation for https://github.com/facebook/relay/issues/2073 and [as requested by @kassens](http://twitter.com/kassens/status/946811879204245504), this adds language plugin support to relay-compiler.

The whole plugin aspect is really just a matter of consolidating all language specific APIs in the [`RelayLanguagePluginInterface.js` file](https://github.com/alloy/relay/blob/d17bfdb18fe45bae531b13487ba5b9efa1c96171/packages/relay-compiler/language/RelayLanguagePluginInterface.js) and moving all of the pre-existing JS/Flow related code around to conform to that plugin interface. i.e. nothing should have changed in the existing JS/Flow handling.

The one actual additions are:
* making relay-runtime able to [load artifacts that are ES6 modules](https://github.com/facebook/relay/pull/2293/files#diff-2c18fc91f11394e307550103afff9722R56) (with default exports)
* the ability to store all artifacts in a single directory (`artifactDirectory`), such that fragment references can easily be imported from other artifacts without the need for the Haste module system. i.e. all imports assume the imported module is a sibling of the requesting module. This should also improve the experience of JS/Flow users.

With this plugin infrastructure in place, @kastermester and I have been able to build a TypeScript plugin https://github.com/kastermester/relay-compiler-language-typescript. (You can find example artifacts [here](https://github.com/kastermester/relay-compiler-language-typescript/tree/master/example/ts/__generated__).)

----

Things that are left to do:
- [x] Document all of the types in the plugin interface
- [x] Remove version changes from `package.json` files
- [x] Rebase on latest upstream master
- [x] Verify that before and after JS/Flow artefacts are identical

But these are not critical to being able to start reviewing and providing feedback.

## Test plan

* See existing JS plugin in action in this fork of the TODO example app https://github.com/alloy/relay-examples/tree/language-plugin-test/todo
* See TS plugin in action in the forked TODO example app in the TS plugin repo, by following these steps https://github.com/kastermester/relay-compiler-language-typescript#installation

----

Adding the TS plugin to the TODO example app would require the following:

1. `yarn add relay-compiler-language-typescript --dev`
2. Tell babel-plugin-relay where to find artifacts (assuming you want fragment references to be type checked) in `.babelrc`:
    ```json
    {
      "plugins": [
        ["relay", { "artifactDirectory": "./ts/__relay_artifacts__" }],
        // ...
      ],
      // ...
    }
    ```
3. Then tell relay-compiler to use the TS plugin and to compile artefacts to the same directory as above:
    ```
    relay-compiler --src ./ts/ --schema ./data/schema.graphql --language typescript --artifactDirectory ./ts/__relay_artifacts__
    ```
4. Which results in:
    ```
    tree ts/__relay_artifacts__/
    ts/__relay_artifacts__/
    ├── AddTodoMutation.graphql.ts
    ├── ChangeTodoStatusMutation.graphql.ts
    ├── MarkAllTodosMutation.graphql.ts
    ├── RemoveCompletedTodosMutation.graphql.ts
    ├── RemoveTodoMutation.graphql.ts
    ├── RenameTodoMutation.graphql.ts
    ├── TodoApp_viewer.graphql.ts
    ├── TodoListFooter_viewer.graphql.ts
    ├── TodoList_viewer.graphql.ts
    ├── Todo_todo.graphql.ts
    ├── Todo_viewer.graphql.ts
    └── appQuery.graphql.ts
    ```


voxmatt:
Badass

kastermester:
While working on the TypeScript equivalent of the babel plugin I noticed that (at least for ts-loader), the file name that gets passed into the TypeScript transformation context is not actually an absolute path to the file. I need to do some testing to verify whether or not that is the case for babel loader as well. If that is the case then there's some minor changes needed to the babel plugin in order to make it work properly when setting the `artifactDirectory` config. I will post my findings once I get to it :)

alloy:
@kassens @jstejada Alright, regardless of the webpack thing that @kastermester is still going to look into, this should now be in a good state to review.

packages/relay-compiler/RelayCompilerPublic.js
```diff
@@ -15,11 +15,12 @@ const GraphQLCompilerContext = require('./graphql-compiler/core/GraphQLCompilerC
 const RelayCodeGenerator = require('./codegen/RelayCodeGenerator');
 const RelayFileWriter = require('./codegen/RelayFileWriter');
 const RelayIRTransforms = require('./core/RelayIRTransforms');
```
alloy: Currently Relay only exports the predefined groups of transforms from https://github.com/facebook/relay/blob/a8af6d7dea1c087bd6c5ba1ca4a2f538d9989cff/packages/relay-compiler/core/RelayIRTransforms.js. However, other language plugins will want to re-use some of these just like the Flow generator does. We are currently doing that in our TS plugin by simply [plucking the ones that the Flow generator uses out from the pre-defined groups](https://github.com/kastermester/relay-compiler-language-typescript/blob/88854c6d57e98bede437c5c4534e46b6c6db7c10/src/TypeScriptGenerator.ts#L535-L539), but this is fragile and so a more stable solution would be appreciated.

What do you think would be the best way to do this? Export _all_ transformers or is it unlikely that a type generator would need any other builtin transforms than the ones that the Flow generator also uses and so we should just add a predefined group for type generators?

alloy: I’m currently leaning towards creating a predefined group for type generators.

alloy:
Rebased on master again.

packages/relay-compiler/core/RelayFindGraphQLTags.js
```diff
@@ -0,0 +1,125 @@
...
+    }
+  });
+}
+
+// TODO: Not sure why this is defined here rather than imported, is it so that it doesn’t get stripped in prod?
+function invariant(condition, msg, ...args) {
+  if (!condition) {
+    throw new Error(util.format(msg, ...args));
+  }
+}
```
alloy: Any idea why this needed to be defined in `FindGraphQLTags.js` rather than importing it?

alloy:
@kastermester Do you have an update re the web pack stuff that still needs to be done? Or otherwise some details about what I should look into?

kastermester:
@alloy I haven't had time to look into it yet. But I'd say if you're using the option to put all artifacts in a single directory with the babel plugin and it works - then there shouldn't be any issues :)

alloy:
@kastermester Re-reading your original comment about the web pack issue, it seems like you’re saying that the issue was around determining the path to the destination directory, is that correct? If so, when using the `artifactDirectory` option there’s no real determination done at all, all artifacts are simply assumed to be siblings of the file being generated. Does that address the issue well enough?

kastermester:
Without using the new artifactDirectory option puts all artifacts in a single directory, where previously it would be put in a directory residing inside the same directory of the file.

This meant requiring artifact files were simply `require('./__generated__/SomeFile.graphql.js')`.

With the artifactDirectory we need some way of finding the single directory where the artifacts are stored. My initial attempt (in the TS plugin) I attempted to use the fileName/filePath property in the TypeScript transformation context and resolve a relative require (ie `../../__generated__/SomeFile.graphql.js`). However I discovered that the fileName passed into the transformation context was relative, not absolute - and thus this didn't work.

What I have ended up with (so far) in the TypeScript case is to give the responsibility back to the user. Ie:

In webpack, define an alias `__generated__` that always points to the generated directory, do the same (using `paths` option in tsconfig) for TypeScript, and set the `artifactDirectory` option to `__generated__`. Now require's are simply turned into `__generated__/SomeFile.graphql` which is understood by both TS and Webpack (or similiar system).

What I haven't had time to test yet - is whether the babel loader is passing absolute file names to the file being generated as opposed to the TS loader, or if it exhibits the same issues.

alloy:
Oh I see now what you mean, you’re referring to the runtime requiring of the concrete node, whereas I was referring to the types. We’re using Babel in our React Native app and it works well in that case 👍

kastermester:
Ah yes, this relates to the PR relay-tools/relay-compiler-language-typescript#21 for the TS transformer and the babel relay plugin in Relay. Might not have made that clear enough.

packages/relay-compiler/core/RelayFindGraphQLTags.js
leebyron: I think it was to allow the compiler to be more restrictive in its dependencies, but I'm not sure

alloy:
CI is red on master, which is fixed by #2318.

alloy:
Alright, everything has been rebased and tested to work well in:

* The TypeScript plugin example: https://github.com/relay-tools/relay-compiler-language-typescript/tree/master/example
* A fork of the JS example app: https://github.com/alloy/relay-examples/tree/language-plugin-test/todo

Interesting note about the JS example is that with the addition of the `artifactDirectory` option in this PR, JS/Flow users now also get the opportunity to have safe fragment references _without_ using Haste: https://github.com/alloy/relay-examples/blob/f8105c243bf476ff0bb0e24de06469ab8e4d62e7/todo/js/__relay_artifacts__/TodoList_viewer.graphql.js#L11-L12

So everybody wins 🎉

packages/relay-compiler/RelayCompilerPublic.js
```diff
@@ -30,6 +31,8 @@ const {
 
 export type {CompileResult, ParserConfig, WriterConfig} from 'graphql-compiler';
 
+const RelayJSModuleParser = RelaySourceModuleParser(FindGraphQLTags.find);
+
 module.exports = {
```
jstejada: so this interface remains the same, without being parametrized by a language, correct? this means that the only way to use plugins is via the cli, correct?

packages/relay-compiler/bin/RelayCompilerBin.js
```diff
@@ -83,6 +84,32 @@ function getFilepathsFromGlob(
...
+        `relay-compiler-language-${options.language}`,
+      );
+      if (languagePlugin.default) {
+        languagePlugin = languagePlugin.default;
+      }
+      if (typeof languagePlugin === 'function') {
+        // For now a plugin doesn’t take any arguments, but may do so in the future.
+        return languagePlugin();
+      }
+    } catch (err) {}
```
jstejada: I see, so if it isn't webpack it just gets ignored here?

packages/relay-compiler/bin/RelayCompilerBin.js
```diff
@@ -94,6 +121,8 @@ async function run(options: {
   watch?: ?boolean,
   validate: boolean,
   quiet: boolean,
+  language: string,
```
jstejada: should we make this optional and default to `'javascript'`?

packages/relay-compiler/bin/RelayCompilerBin.js
```diff
@@ -132,14 +161,34 @@ Ensure that one such file exists in ${srcDir} or its parents.
...
+  const sourceModuleParser = RelaySourceModuleParser(
+    languagePlugin.findGraphQLTags,
+  );
+
+  const artifactDirectory = options.artifactDirectory
+    ? // $FlowFixMe artifactDirectory can’t be null/undefined at this point
+      path.resolve(process.cwd(), options.artifactDirectory)
+    : null;
+
+  const isGeneratedDirectory = artifactDirectory || '__generated__';
```
jstejada: nit: maybe this would be better named as `generatedDirectoryName`?

packages/relay-compiler/core/__tests__/RelayFindGraphQLTags-test.js
```diff
@@ -10,100 +10,25 @@
...
-
-    it('parses JS with functions sharing names with object prototype methods', () => {
-      expect(
-        find(`
-          toString();
-          foo(graphql\`fragment FindGraphQLTags on User { id }\`);
-        `),
-      ).toEqual(['fragment FindGraphQLTags on User { id }']);
-    });
-  });
```
jstejada: these tests got moved, correct?

packages/relay-compiler/language/RelayLanguagePluginInterface.js
```diff
@@ -0,0 +1,246 @@
...
+ * When reading, the plugin is expected to parse and return GraphQL tags; and
+ * when writing the plugin is responsible for generating type information about
+ * the GraphQL selections made as well as generating the contents of the
+ * artifact file.
+ *
+ * This interface describes the details relay-compiler requires to be able to
+ * use the plugin. The plugin is expected to have as its main default export a
+ * function that returns an object conforming to this interface.
+ */
+export type PluginInterface = {
```
jstejada: nit: how do you feel about typing the plugin interface as the function that returns this shape itself, rather than just the shape? That way we can also type any arguments to language plugins here, in case they are ever added

packages/relay-compiler/language/RelayLanguagePluginInterface.js
```diff
@@ -0,0 +1,246 @@
...
+   */
+  transforms: Array<IRTransform>,
+
+  /**
+   * Given GraphQL document IR, this function should generate type information
+   * for e.g. the selections made. It can, however, also generate any other
+   * content such as importing other files, including other artifacts.
+   */
+  generate: (node: Root | Fragment, options: TypeGeneratorOptions) => string,
+};
```
jstejada: all of these types just got moved here right? did they also have all these comments? if the didn't, thanks for adding them!

packages/relay-compiler/language/javascript/FindGraphQLTags.js
```diff
@@ -141,46 +114,18 @@ function find(
     },
     TaggedTemplateExpression: node => {
       if (isGraphQLTag(node.tag)) {
-        const template = getGraphQLText(node.quasi);
-        if (validateNames) {
```
jstejada: so validation now occurs at the level of the unparametrized RelayFindGraphQLTags, right?

packages/relay-compiler/language/javascript/RelayFlowGenerator.js
```diff
@@ -414,6 +409,13 @@ function getFragmentImports(state: State) {
           // TODO(T22653277) support non-haste environments when importing
           // fragments
           imports.push(importTypes([refTypeName], usedFragment + '.graphql'));
+        } else if (
+          state.useSingleArtifactDirectory &&
+          state.existingFragmentNames.has(usedFragment)
+        ) {
+          imports.push(
+            importTypes([refTypeName], './' + usedFragment + '.graphql'),
```
jstejada: hmmm, what exactly are the effects of this?

packages/relay-compiler/bin/RelayCompilerBin.js
```diff
@@ -83,6 +84,32 @@ function getFilepathsFromGlob(
...
 }
 
+function getLanguagePlugin(options: {language: string}): PluginInterface {
+  if (options.language === 'javascript') {
+    return RelayLanguagePluginJavaScript();
+  } else {
+    try {
+      // prettier-ignore
+      // $FlowFixMe
+      let languagePlugin = __non_webpack_require__( // eslint-disable-line no-undef
```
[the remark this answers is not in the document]
alloy: Correct, webpack transforms all plain `require` calls to refer to bundled code, whereas we want it to actually require external code. It only makes that assumption for this file and it _is_ my understanding that this file is only ever used in a web pack environment. Is that not the case?

packages/relay-compiler/RelayCompilerPublic.js
alloy: That is currently correct.

packages/relay-compiler/bin/RelayCompilerBin.js
alloy: No, this is meant to more gracefully handle a plugin name being misspelled or the plugin not conforming to the expectation of exporting a function.

packages/relay-compiler/bin/RelayCompilerBin.js
alloy: It defaults to `javascript` in the option parsing phase, is there a reason to do it here instead?

packages/relay-compiler/bin/RelayCompilerBin.js
alloy: 👍

packages/relay-compiler/core/__tests__/RelayFindGraphQLTags-test.js
alloy: Correct.

packages/relay-compiler/language/RelayLanguagePluginInterface.js
alloy: Good point 👍

packages/relay-compiler/language/RelayLanguagePluginInterface.js
alloy: These all moved here indeed, but they did not have these comments 😄

packages/relay-compiler/language/javascript/FindGraphQLTags.js
alloy: Correct.

packages/relay-compiler/language/javascript/RelayFlowGenerator.js
alloy: That’s what [my comment above](https://github.com/facebook/relay/pull/2293#issuecomment-364661877) was about. Previously, only users of Haste _and_ those that forked the compiler bin to override the hardcoded disablement of Haste in the OSS compiler bin would be able to benefit from fragment reference type checking, which requires the artefact to import fragment reference types from other artefacts. The assumption is that this was the case because otherwise imports would break if files got moved around. As a non-Haste solution, the alternative is to store all artefacts in the same directory, in which case imports will always be stable and siblings of the current artefact file.

You can see the result here https://github.com/alloy/relay-examples/blob/f8105c243bf476ff0bb0e24de06469ab8e4d62e7/todo/js/__relay_artifacts__/TodoList_viewer.graphql.js#L11-L12

packages/relay-compiler/bin/RelayCompilerBin.js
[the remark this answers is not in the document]
jstejada: I see. I'm actually not 100% sure, but I'll double check cc @kassens.

Still, to be on the safer side, could we check first before calling it?

packages/relay-compiler/bin/RelayCompilerBin.js
jstejada: Ah got it, sorry missed that.

jstejada:
Thanks for the replies @alloy. This looks good to me sans the minor nits and any comments from @kassens.

I'm going to test this internally and make sure everything is working.

Thanks!

alloy:
I’ll be updating the PR today to address feedback 👍

packages/relay-compiler/bin/RelayCompilerBin.js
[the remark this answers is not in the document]
alloy: Sure, I could maybe do `(__non_webpack_require__ != undefined ? __non_webpack_require__ : require)(…)` ?

packages/relay-compiler/bin/RelayCompilerBin.js
alloy: Improved this a little further.

packages/relay-compiler/language/javascript/RelayFlowGenerator.js
alloy: I realise that I did not add a test for this, so doing so now.

ntkme:
@alloy Does this PR help the use case in #2339, where a language (JavaScript or TypeScript) is embedded in another one (Vue component / HTML)?

I'm looking at the following use cases:
- `.js` source → relay-compiler → `.js` artifacts
- `.ts` source → ts plugin → relay-compiler → ts plugin → `.ts` artifacts
- `.vue` with `<script>` → pre-language processing → `.js` source → relay-compiler → `.js` artifacts
- `.vue` with `<script lang="ts">` → pre-language processing → `.ts` source → ts plugin → relay-compiler → ts plugin → `.ts` artifacts

Webpack handle similar use case via loaders, that it executes multiple loaders one after another. I think we don't necessarily need reproduce that. However, in term of code reusability I would like to see some pre-language interface.

alloy:
@ntkme It absolutely _could_. Right now the plugin support allows the first two cases you mention, for the others you would currently have to make a Vue plugin that **wraps** the JS or TS plugin.

While that is not the solution you were probably thinking of, I think that it would keep things simpler in Relay for the time being. When somebody else also has a need for this, for another system _besides_ Vue, it may be the better time to think of a general solution that works for all. What do you think?

----

In any case, whatever we decide on, let’s work on that iteratively instead of making it part of this PR, but I’d love to work with you to get to a solution.

For the wrapping solution, we probably only need to expose the JS plugin that comes bundled with Relay and think of a good way to make your Vue plugin able to use whatever language plugin the user decides to provide as a peer dependency. I guess the Vue plugin could inflect what language plugin to load based on the source containing `<script>` or `<script lang="ts">` ?

alloy:
@kastermester I’ve tried many things, but trying to be compatible with webpack and without it and referencing `require` anywhere leads to various issues. I’ve also tried to replace the `require` with `__non_webpack_require__` from the gulp task, but that doesn’t pass all files through so couldn’t do that either. In the end this was the most straightforward and easily understandable version 🤷‍♂️

alloy:
@jstejada Alright, I addressed all the feedback and added the requested test plan steps.

ntkme:
@alloy I'm good with your direction, but there is a problem I found after briefly reviewed this PR. 

``` js
export type PluginInterface = {
  inputExtensions: string[],
  outputExtension: string,
  formatModule: FormatModule,
  typeGenerator: TypeGenerator,
};
```

For `.js` and `.ts`, it's fine to hardcode `outputExtension` because it's known for sure based on the input extension. For `.vue`, the outputExtension must be dynamically resolved based on the content within the file. Unfortunately, we cannot use output extension `.vue` and put a `<script>` in it, because `vue-loader` in `webpack` expects the `<script>`'s default export to be a Vue component. Instead, we must explicitly write out `.js` or `.ts` base on the input found in the file.

That means with `outputExtension` hardcoded, a `.vue` language wrapper must assume user would only use one language in `.vue` files, so it can hard code one `outputExtension`.

alloy:
@ntkme Ah yes, good point. I think that something like [this](https://gist.github.com/alloy/2e00f45df5d815b546f036f0dec14b7b) could work (untested code, though), because really the file extension is not that important in testing if a file is an artefact.

ntkme:
@alloy I started to create a vue language plugin based on your patch.

Wrapping `findGraphQLTags` was easy enough with [vue-template-compiler](https://www.npmjs.com/package/vue-template-compiler), which provides the functionality of extracting the `<script>` section.

However, I realized just after a minute that not only the `outputExtension` but also`formatModule` and `typeGenerator` need to be dynamically resolved to the js / ts version respectively based on detected language. This would require a relatively larger change in your current patch as currently `formatModule` and `typeGenerator.generate` have no knowledge of language at all.

ntkme:
The following makes me think once again that a pre-language stage makes more sense than the wrapper approach:

```
const { script } = require('vue-template-compiler').parseComponent(text)
if (script != null) {
  const { content, lang = 'js' } = script
}
```

If we have a pre-language stage doing the above, we can simply pretend we got `lang` as input extension to pick up the right plugin, then pass `content` as `text` directly to the `findGraphQLTags` function for that extension. I think this would save lots of troubles in the wrapper approach.

alloy:
I don’t think that would work well, as relay-compiler is architected in such a way that the various parts are configured at the very start, i.e. we can’t easily change `formatModule` and `typeGenerator` based on files being parsed at a later stage.

I think that either:
1. we determine if dynamic (on a per file basis) is really needed at all; do you mix languages in one Vue project? If not, then maybe we indeed add the notion of a pre-language step _but_ have that be configured statically, e.g.
    ```
    $ yarn install relay-compiler-pre-processing-vue relay-compiler-language-typescript --dev
    $ relay-compiler --pre-processing vue --language typescript
    ```
2. or we determine if a wrapper plugin could keep state itself that maps filename/fragment key to language after invoking its `findGraphQLTags` function and uses that state in its `formatModule` and `typeGenerator` implementations to delegate to the right wrapped plugin.

While I’m really not a fan of adding more configurable options to tooling, in this case it does seem like that would be the simplest solution; and a user probably only needs to figure out the right options once at the start of a project and then hardcode those in a build script.

kastermester:
@alloy Sorry was on my phone, so had a bit of trouble writing some code up properly (just inserting backticks was kind of annoying). I meant doing something like this:

```js
const req = "require";

const non_webpack_require = process[req];

const myModule = non_webpack_require('someModule'):
```

I would guess that would not mess anything up wrt. webpack?

robrichard:
@kastermester I don't think that would work because `require` is not actually a global variable accessible on `process` or `global`, but a local variable passed to the module. See  https://nodejs.org/api/modules.html#modules_the_module_wrapper. Both webpack and node do this.

ntkme:
@alloy I would prefer option 1 as developers _should_ only use one language in one project. Even though we have to add another option to cli, nothing in this PR needs to change, and the vue use case can be addressed in a relatively easier way.

alloy:
@kastermester Yep, what @robrichard said. I’m definitely open to other solutions than using `eval` but I was spending a lot of time on it and things did not get any simpler 😄

alloy:
@ntkme Agreed, that does sound like the best way forward then 👍

ntkme:
@alloy I would be happy to.

kastermester:
@alloy There's probably not a better way of doing it then, it just hurts my eyes ;). I'll just avoid opening that file :p

enisdenjo:
I can not get a **RelayRecordProxy** by providing a unique ID to the **proxyStore** inside the commitMutation updater (optimistic and regular). I keep getting **undefined**. I use Artsy's relay-compiler/runtime as described in the README.md. Providing the exact same ID worked before (on a normal 1.5.0 compiler/runtime). Can you help me out, I don't know why is this happening?

alloy:
@enisdenjo Are you using a `v1.5.0-artsy.x` release or a `v1.5.0-plugin.x` release? You should use the latter; although in case you’re using the former I’d love to chat more about the issue you’re running into, this is just not the right place for that. Mind pinging me in the #relay channel on [the GraphQL Slack](https://graphql-slack.herokuapp.com/)?

enisdenjo:
@alloy Thank you very much! I solved it by switching from 1.5.0-artsy.x to 1.5.0-plugin.x.

avikalpa:
Please merge quickly! Really in need of this! Thanks.

alloy:
@avikalpa If you really need it now, you can use a forked build, like we at Artsy do and @enisdenjo above. Here’s the latest build we’re using https://github.com/alloy/relay/releases/tag/v1.5.0-plugin.3

avikalpa:
@alloy Thanks! Checking into that. So far no issues.

alloy:
@jstejada Rebased again.

alloy:
I made a small change to the labels that the source parser and writers use such that it reflects the input file extensions it’s watching and the output extension of the artifacts:

```
$ yarn build --watch
yarn run v1.5.1
$ relay-compiler --src ./ts/ --schema ./data/schema.graphql --language typescript --artifactDirectory ./ts/__relay_artifacts__ --watch

Writing ts
Unchanged: 12 files
Watching for changes to ts/tsx...
Watching for changes to graphql...

Writing ts
Updated:
 - TodoApp_viewer.graphql.ts
 - appQuery.graphql.ts
Unchanged: 10 files
Watching for changes to ts/tsx...
```

alloy:
Also pushed a new build for those using this already https://github.com/alloy/relay/releases/tag/v1.5.0-plugin.5

arieldf:
@alloy Thank you! This is just what we need. Is there any estimate on when we can expect this to be merged? Weeks/months?

arieldf:
@kastermester Thanks a lot for the quick feedback!

kastermester:
@arieldf You're welcome. Feel free to email me if you have any questions, I'll do my best to help :)

enisdenjo:
Hey @alloy, thanks for the new release! But after installing and running it I get the following error:

```shell
Writing ts
ERROR:
rule is not a function
```

Am I missing something?

Thanks in advance! 😃

alloy:
@enisdenjo Hey thanks for letting me know! Can you reach out to me on https://graphql.slack.com/messages/relay/ and we can try to pinpoint the issue there?

enisdenjo:
Together with @alloy, we've pinpointed the problem to not having the **graphql@^0.12.0 || ^0.13.0** dependency.

After adding it, the issue is solved!

alloy:
I’ve gone ahead and squashed all the commits now to ease rebasing on my end. I’ve also rebased, but alas can’t run the tests at the moment due to the issue described in #2407

alloy:
Tests pass again 👍

jeromecovington:
I've been using [relay-compiler-language-typescript](https://github.com/relay-tools/relay-compiler-language-typescript) on a react-native project with great success (but having to pin to patches of relay-things to do so). It would be so cool to see this PR land so that plugin architecture will be supported in core!

lszomoru:
Any update on completing this pull request? Thanks!

sciyoshi:
Thanks @alloy for this work. I'm currently maintaining https://github.com/sciyoshi/relay-coffee-compiler, this PR would definitely help that effort since a lot of the code is duplicated there.

TrySound:
@alloy @kassens  Can splitting `artifactDirectory` feature in separate PR speed up reviewing/merging process? I could take care of it if you don't mind.

alloy:
Just got back from vacation and catching up. Will try to perform a rebase asap.

@TrySound I honestly doubt it would make much of a difference in review/merge time (but please correct me if that’s incorrect @jstejada/@kassens) and I really would not want to get into a situation where the `artifactDirectory` change gets reviewed/merge much later as they are so related for our case.

alloy:
@jstejada Rebased

voxmatt:
@jstejada @kassens If there's any chance of getting this merged, the whole community would thank you. I've been keenly watching this one for months and the work by @alloy and others here is tremendous.

arieldf:
I am waiting with a lot of excitement to get this merged. It is a fantastic contribution by @alloy

waterfoul:
I can mirror that sentiment and would love to have this functionality without having to install off github

jstejada:
hey everyone, sorry for the delay on this! Just a quick update, I'm hitting an error when running this internally-- as soon as I can get that fixed I'll land this!

fameoflight:
Any update? Eagerly waiting for this.

hally9k:
We too are excited to see this one merged in, pwetty pweeease! 😄

waterfoul:
It's been almost a month since the last update. Any news?

jstejada:
I'm really sorry to keep everyone waiting on this PR. This has been a particularly time constrained month for me, and I haven't been able to easily fix the internal issue we're running into on this PR, but please know that I will get to this as soon as I can, hopefully very soon!

Once again, really sorry to keep everyone waiting. Thanks for your patience!!

kastermester:
@jstejada Better have it land properly and well integrated than a half assed merge :) Thank you so much for looking into this. I think everyone here knows how stuff have a habit of getting in the way.

Keep up the good work! :)

alloy:
@jstejada What @kastermester said! Thanks for all the effort thus far :heart:

enisdenjo:
Hey guys, any updates on this merge? We've been eagerly (but yet patiently) waiting for 7 months now...

alloy:
FYI I’ll be rebasing this one over the weekend and signal @jstejada that we can continue.

alloy:
Ok

* I’ve rebased the PR
* Uploaded a build of this PR https://github.com/alloy/relay/releases/tag/v1.6.1-plugin.1
* Pushed a new version (0.9.1) of the TS plugin that uses this https://www.npmjs.com/package/relay-compiler-language-typescript

Other than the example app in the plugin repo I have not yet had the time to verify this in our larger apps. If you do in the meantime, please let me know.

packages/relay-compiler/language/javascript/formatGeneratedModule.js
```diff
@@ -33,12 +36,12 @@ const formatGeneratedModule: FormatModule = ({
 'use strict';
 
 /*::
-import type { ${documentType} } from '${relayRuntimeModule}';
-${flowText || ''}
+${documentTypeImport}
```
jstejada: this would generate incorrect type if `documentTypeImport` is an empty string. In which cases is `documentType` null?

alloy: As discussed elsewhere, this was basically just following some of the existing code and would indeed generate incorrect code but only if the inputs were already incorrect anyway.

enisdenjo:
Hey @alloy,

I have been using your `v1.6.1-plugin.1` in production for 4 days now. Haven't faced any regressions yet!

D.

alloy:
Awww yeah! Thanks for all the assistance shepherding this through @jstejada 🙌 🍾

waterfoul:
Any idea when a release will be cut containing these changes?

soneymathew:
@alloy Will this work for non-ts projects?
Please let me know if this is supported usage
https://github.com/soneymathew/relay-examples/commit/18740d8c0df68eeb3f7425c84db235604de9bceb

kastermester:
@soneymathew The `artifactDirectory` does not depend on which language plugin you're using. It will simply redirect all the artifact files to a single directory.

Note: If you're using flow this should allow flow to typecheck the fragment types akin to how it works if one were to use the haste module system prior to this merge. In short typechecking is applied to the relay properties based on the fragments being spread (Parent component can't render Child component without having done `...Child_prop` in the corresponding fragment).

soneymathew:
@kastermester That matches my understanding but, in this fork(https://github.com/soneymathew/relay-examples/commit/18740d8c0df68eeb3f7425c84db235604de9bceb), I couldn't make it work.

soneymathew:
Thanks @kastermester 🤦‍♂️  I was looking forward for this one way too long and skipped to check that fact...
https://github.com/facebook/relay/compare/v1.6.2...master

alloy:
Heyyy @soneymathew, sorry just getting back online after being away for a bit. I don’t have builds of 1.6.2 with this code applied, the last I had was 1.6.1 just before 1.6.2 was released https://github.com/artsy/relay/releases/tag/v1.6.1-plugin.1

merrywhether:
In case no one has seen it, this is included in the relay 1.7.0-rc.1 release! 🎉 

https://github.com/facebook/relay/releases/tag/v1.7.0-rc.1

Thanks so much to the relay team and @alloy for all the work on this!

PR #2297, opened 2018-01-22
Fix example instructions

gaearon:
The proposed instructions don't work. You can try them :-)

Related issue (closed by author but not actually solved): https://github.com/relayjs/relay-examples/issues/44

This fixes the instructions to also run the compiler.


jstejada:
thank you!!

PR #2298, opened 2018-01-22
[Site] Adds Artsy to the list of relay users

orta:
Hey there, I added us to the list of users. I added us as `sticky` because the front page felt a little bit lonely ;) happy to change it, if you prefer.

I also ran the other logos through image compression, generally saved about 60-80% on their file sizes (so roughly 12k -> 3k per file) - also happy to undo that too


jstejada:
thanks!

PR #2299, opened 2018-01-23
Changing field name in ThinkingInGraphQL docs

gunnigylfa:
In the Data/View Consistency section of the Thinking in GraphQL page the query refers to a node field but the caching and previous query examples indicate that the field should be `story`. 

Here is the query before the change:
```graphql
query {
  node(id: "1") {
    text,
    author { name, photo },
    comments {
      text,
      author { name, photo }
    }
  }
}
```
And here is how the query should be: 
```graphql
query {
  story(id: "1") {
    text,
    author { name, photo },
    comments {
      text,
      author { name, photo }
    }
  }
}
```


jstejada:
thanks!

PR #2304, opened 2018-01-29
[Async React + Relay] Fix issue where 'setState' fires too early

flarnie:
**what is the change?:**

There is logic in the contructor to fetch data, and sets 'onDataChange'
as the callback for when the data is ready.

With React rendering asynchronously, the 'onDataChange' could fire
before the component is mounted. This adds protection for that case,
adding a flag to track whether the component is mounted. Only in cases
that the 'onDataChange' fires before the component is mounted, we will
defer the 'setState' call until 'componentDidMount'.

**why make this change?:**

With React in synchronous mode, we can assume the component will always
be mounted by the time we call `setState` inside `onDataChange`, like
so:
```
   inside
   ReactRelayQueryRenderer

   constructor
   +
   v try to get data - it's not there!
     +
     v
     data must be featched
     +
     |
     v ++React will sync. render
       ||a spinner.
       ||
       ||
       ++
     data returns
     +
     v
     call setState:
     +
     |
     v ++React will re-render with data
       ||
       ||
       ||
       ||
       ||
       ++
```

Once React is rendering asynchronously, if for some reason it takes
longer than the data fetch, then the `onDataChange` will fire before the
component is mounted, and call `setState`. This would throw an error,
and is not supported. It is not defined how this would behave in the
future, but it would not be good.
```
inside
ReactRelayQueryRenderer

constructor
+
v try to get data - it's not there!
  +
  v
  data must be featched
  +
  |
  v ++React will sync. render
    ||a spinner.
    ||
    ++


    ++
    ||
    ||
    +-----------> data returns
                  BEFORE React
    ++            is done with
    ||            initial render
    ++            +
                  v
                  call setState:
    ++
    ||            React throws! :(
    ++
```

**test plan:**
Ran existing tests and they passed.
I'd like to add a test which covers this specific situation, but that
will take a bit more effort and if this change doesn't hurt anything
we'd like to land it asap. Will make a follow-up task to add test coverage.

**issue:**
internal task t25062790

### DIFF
--- a/packages/react-relay/modern/ReactRelayQueryRenderer.js
+++ b/packages/react-relay/modern/ReactRelayQueryRenderer.js
@@ -76,6 +76,8 @@ type State = {
 class ReactRelayQueryRenderer extends React.Component<Props, State> {
   _queryFetcher: ReactRelayQueryFetcher = new ReactRelayQueryFetcher();
   _relayContext: RelayContext;
+  _isMounted: ?boolean;
+  _dataFetchedBeforeComponentWasMounted: {snapshot?: Snapshot, error?: Error};
 
   constructor(props: Props, context: Object) {
     super(props, context);
@@ -94,7 +96,17 @@ class ReactRelayQueryRenderer extends React.Component<Props, State> {
     }
   }
 
+  componentDidMount(): void {
+    this._isMounted = true;
+    if (this._dataFetchedBeforeComponentWasMounted) {
+      // use the data to call setState
+      const {error, snapshot} = this._dataFetchedBeforeComponentWasMounted;
+      this.setState({renderProps: this._getRenderProps({error, snapshot})});
+    }
+  }
+
   componentWillUnmount(): void {
+    this._isMounted = false;
     this._queryFetcher.dispose();
   }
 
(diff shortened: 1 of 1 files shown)


flarnie:
What about in cases where first render would not have finished? - it would be nice if it just queues a second render though. I think when @acdlite and I talked it sounded like this would throw.

acdlite:
If you call `setState` in the constructor, React will warn.

If you call `setState` in between the constructor and `componentDidMount`, there's no warning: it will add the update to the queue, but it won't be applied until the next render.

sophiebits:
So that means the behavior is correct even without this change, correct?

flarnie:
If that behavior is the same in async React then we don't need this fix - I had thought that in async mode it be a problem. I thought from our discussions that it was going to throw, based on the comments that calling 'setState' in a constructor is "not supported", I see now I misunderstood. It's correct that we warn when 'setState' gets called on a component which is not mounted.

If we start seeing this problem then at least it's clear how to fix it.

sophiebits:
Sounds great.

My thought process was: I figured that it makes sense for React to handle this properly since this seems like a very common use case in line with our recommendations (initiating a fetch in cWM/render) and the code required to handle this in user space is pretty clumsy. The warning is added to catch people making mistakes that indicate broken or inefficient code, but there's likely no pitfall if people are calling setState from an async process during the mount phase, so there's no reason to warn. We also don't have another recommendation that would be cleaner, and we don't gain anything from making the user code clumsy (unlike, say, isMounted, where we encourage people to cancel their async processes) so it must make sense to build this queueing into React.

PR #2317, opened 2018-02-09
[Website] Add friday to list of users

gauravtiwari:
Hey there, 

Adds Friday to the list of users. We’re using Relay modern to power our new web app [friday](https://friday.work/) 🎉

Thanks :)

PR #2318, opened 2018-02-09
Remove a trailing comma in the gulp file which breaks older node builds

orta:
See comments in https://github.com/facebook/relay/pull/2317#issuecomment-364453167

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+scripts/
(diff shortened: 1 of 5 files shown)


orta:
OK, the scope of this PR changed a bit.

I've added a `.prettierignore` file, which tells everyone's editors/prettier to ignore the JS files in `scripts`, then I ran `yarn prettier --write "scripts/*" --trailing-comma es5` over the files so that they'll run on older node runtimes.

Issue #2077, opened 2017-08-30
Remove clientMutationId


steida:
> We should figure out how to remove it. I'm pretty sure it's just a fb-ism. Can you open an issue?

https://twitter.com/kassens/status/903031923827654658

cc @kassens

steida:
It's somehow used by the code because when I tried to use constant, React key was the same...
https://www.graph.cool/forum/t/clientmutationid/426/3?u=daniel_steigerwald

adjourn:
If we're already talking about it..

What about [graphql-relay-js](https://github.com/graphql/graphql-relay-js) `mutationWithClientMutationId`? 
Add `mutationWithoutClientMutationId` or maybe this gem: `mutation`?

omerzach:
Is there any hack possible to avoid all the boilerplate of parsing this ID and returning it again in every mutation? E.g., can I just include a random string in my mutation responses?

steida:
https://github.com/este/este/blob/0dce17f0553a0ecb0df0c378695192cd0351e20b/components/withMutation.js#L8

steida:
I think it was caused by graph.cool

omerzach:
Why was this closed? `clientMutationId` is still all over Relay. If it's not necessary it would be really great to remove it.

steida:
@kassens Any plan to remove this PITA? clientMutationId breaks adding a node with the same text for no obvious reason. It really sucks socks.

steida:
@kassens told me, that it should be removed from Relay. Now it must be used, otherwise, Relay will generate wrong ID for an item with the same text for an unknown reason.

steida:
@kassens Please check this. This is Relay bug.

![screen shot 2018-04-07 at 05 40 38](https://user-images.githubusercontent.com/66249/38450905-41575694-3a26-11e8-843d-77d8b505952d.png)

taion:
this is gated on https://github.com/facebook/relay/pull/2349

taion:
it's essentially the same thing as that PR

jaredpalmer:
testing out relay-experimental

alloy:
This is fixed now by https://github.com/facebook/relay/pull/2349 right? (Re-open if not.)

earksiinni:
So is this line from the docs inaccurate?

> ## Mutations
>
[...]

PR #2349, opened 2018-02-26
Remove the `clientMutationId` requirement by creating a new root id for each executed mutation

### LINKED ISSUE #825: Idea: Remove the mutation GraphQL specification requirements?

### LINKED ISSUE #1734: How to avoid Dup Records when executing Create Mutation Commits that Insert New Edges into Connections
Having trouble implementing a simple Document Component that adds a document and updates a connection relation with Application.documents edges.  The Create works fine, but the Updater keeps adding the LinkedRecord reference to the same Edge every time I run InsertEdgeAfter on the Application.documents Connection.  This causes duplicate records since it keeps updating the old Edge records by reference.  I'm hoping I don't have to destroy the Component to create a separate reference for the Create Edge LinkedRecord.  How do we add multiple edges to Connections if we always receive the same referenced Edge?

Here is the mutation to create a document, it returns an associated Application.id as well as the expected edge.

```
const mutation = graphql
  mutation CreateDocumentMutation($input: CreateDocumentInput!) {
    createDocument(input: $input) {
      document {
        id
        data
      }
      edge {
        __typename
        cursor
        node {
          id
          application{
            id
          }
          data
        }
      }
      application {
        id        
      }
    }
  }
```

Here is my updater method that gets the payload, finds the Edge, then uses the application proxy to get the Connection to insert the Edge into.  The problem is the Edge is a linked record and every time I run a commit on this create it keeps updating the Edge reference and incrementing pointers to the latest Edge value.

```
updater: (store) => {
        const payload = store.getRootField('createDocument')
        const newEdge = payload.getLinkedRecord('edge')
        const proxy = store.get(application.id)
        const conn = ConnectionHandler.getConnection(proxy, 'application_documents')        
        ConnectionHandler.insertEdgeAfter(conn, newEdge)
      }
```

My connection is simply just a list of the documents for this application.  After I insert each document, the list starts creating duplicates simply because the references all point to the last Edge created.  The GraphQL data storage is fine, but my local update is failing due to this issue.

Application.documents = [
11111,  <- created in an earlier session
2222,  <- created in an earlier session
12345,  <- created in the current session
12345,  <- created in the current session
12345,  <- created in the current session, keeps repeating as I run the commit trigger
]

### LINKED ISSUE #2333: Status Code:204 No Content Remote
I don't know why my server API returns 204 for now, but if I add two items with the same text, Relay will generate the same key.

screen shot 2018-02-21 at 04 46 24

robrichard:
Fixes #1734
Fixes #2333 
Fixes #825

Related to https://github.com/relayjs/relay-examples/issues/30
When a mutation is executed, it's payload is added to the store using a key of the serialized arguments. In the Todo Modern example this looks like this:

```js
{
  'client:root:addTodo(input:{"text":"a"})': { ... } // added by mutation response
  'client:VXNlcjptZQ==:__TodoList_todos_connection': {
    edges: [
      __refs: [
        'client:VXNlcjptZQ==:__TodoList_todos_connection:edges:0',
        'client:VXNlcjptZQ==:__TodoList_todos_connection:edges:1',
        'client:root:addTodo(input:{"text":"a"}):todoEdge', // added by mutation imperative updater function
      ]
    ]
  }
}
```

If you execute a second `addTodo` mutation with the same arguments, the `client:root:addTodo(input:{"text":"a"})` field in the store will get overwritten with the second mutation and the store will look like this:

```js
{
  'client:root:addTodo(input:{"text":"a"})': { ... }
  'client:VXNlcjptZQ==:__TodoList_todos_connection': {
    edges: [
      __refs: [
        'client:VXNlcjptZQ==:__TodoList_todos_connection:edges:0',
        'client:VXNlcjptZQ==:__TodoList_todos_connection:edges:1',
        'client:root:addTodo(input:{"text":"a"}):todoEdge',
        'client:root:addTodo(input:{"text":"a"}):todoEdge',
      ]
    ]
  }
}
```

When React renders the data, you will have the latter `todo` in the connection array twice and will likely get an error along the lines of `Encountered two children with the same key, <id>. Child keys must be unique; when two children share a key, only the first child will be used`.

This was [fixed in relay-examples](https://github.com/relayjs/relay-examples/commit/ed955945ca5158ac34ed8ce703ae0cc7cb3e38e7#diff-4b42725e6eb503f56da8e3f637e9a4e3) by adding a `clientMutationId` to the mutation arguments. I don't think this is ideal since
1. it requires the developer to manually add this to each mutation
2. it places an additional constraint on the GraphQL schema to support this field.

The approach I took to fix this, was to use a new root `dataID` (instead of the static `client:root`) for results on each mutation. This will namespace results from each mutation so they can never overwrite one another. 

With these changes, the relay store now looks like this after completing the same series of mutations:

```js
{
  'client:mutationRoot0:addTodo(input:{"text":"a"})': { ... },
  'client:mutationRoot1:addTodo(input:{"text":"a"})': { ... },
  'client:VXNlcjptZQ==:__TodoList_todos_connection': {
    edges: [
      __refs: [
        'client:VXNlcjptZQ==:__TodoList_todos_connection:edges:0',
        'client:VXNlcjptZQ==:__TodoList_todos_connection:edges:1',
        'client:mutationRoot0:addTodo(input:{"text":"a"}):todoEdge',
        'client:mutationRoot1:addTodo(input:{"text":"a"}):todoEdge'
      ]
    ]
  }
}
```

I tested this and it works to solve this issue, but I'm unsure if there could be any unintended consequences by this change.

### DIFF
--- a/packages/react-relay/classic/environment/RelayCombinedEnvironmentTypes.js
+++ b/packages/react-relay/classic/environment/RelayCombinedEnvironmentTypes.js
@@ -231,6 +231,7 @@ export interface CUnstableEnvironmentCore<
     request: TRequest,
     variables: Variables,
     operation?: TOperation,
+    dataID?: DataID,
   ) => COperationSelector<TNode, TRequest>;
 
   /**
--- a/packages/relay-runtime/mutations/RelayRecordSourceProxy.js
+++ b/packages/relay-runtime/mutations/RelayRecordSourceProxy.js
@@ -139,10 +139,10 @@ class RelayRecordSourceProxy implements RecordSourceProxy {
     return this._proxies[dataID];
   }
 
-  getRoot(): RecordProxy {
-    let root = this.get(ROOT_ID);
+  getRoot(dataID: DataID = ROOT_ID): RecordProxy {
+    let root = this.get(dataID);
     if (!root) {
-      root = this.create(ROOT_ID, ROOT_TYPE);
+      root = this.create(dataID, ROOT_TYPE);
     }
     invariant(
       root && root.getType() === ROOT_TYPE,
--- a/packages/relay-runtime/mutations/RelayRecordSourceSelectorProxy.js
+++ b/packages/relay-runtime/mutations/RelayRecordSourceSelectorProxy.js
@@ -54,7 +54,7 @@ class RelayRecordSourceSelectorProxy implements RecordSourceSelectorProxy {
   }
 
   getRoot(): RecordProxy {
-    return this.__recordSource.getRoot();
+    return this.__recordSource.getRoot(this._readSelector.dataID);
   }
 
   _getRootField(
--- a/packages/relay-runtime/mutations/__tests__/commitRelayModernMutation-test.js
+++ b/packages/relay-runtime/mutations/__tests__/commitRelayModernMutation-test.js
@@ -106,14 +106,14 @@ describe('Configs: NODE_DELETE', () => {
     ];
     const optimisticUpdater = jest.fn();
     const updater = jest.fn();
+    const operationSelector = createOperationSelector(FeedbackCommentQuery, {});
+    environment.commitPayload(operationSelector, payload);
     const snapshot = store.lookup({
       dataID: ROOT_ID,
       node: FeedbackCommentQuery.fragment,
       variables: {},
     });
     const callback = jest.fn();
-    const operationSelector = createOperationSelector(FeedbackCommentQuery, {});
-    environment.commitPayload(operationSelector, payload);
     store.subscribe(snapshot, callback);
     commitRelayModernMutation(environment, {
       configs,
@@ -547,13 +547,13 @@ describe('Configs: RANGE_ADD', () => {
         edgeName: 'feedbackCommentEdge',
       },
     ];
+    const operationSelector = createOperationSelector(CommentQuery, {});
+    environment.commitPayload(operationSelector, payload);
     const snapshot = store.lookup({
       dataID: ROOT_ID,
       node: CommentQuery.fragment,
       variables: {},
     });
-    const operationSelector = createOperationSelector(CommentQuery, {});
-    environment.commitPayload(operationSelector, payload);
     store.subscribe(snapshot, callback);
     commitMutation(environment, {
       configs,
@@ -757,13 +757,13 @@ describe('Configs: RANGE_ADD', () => {
         edgeName: 'feedbackCommentEdge',
       },
     ];
+    const operationSelector = createOperationSelector(CommentQuery, {});
+    environment.commitPayload(operationSelector, payload);
     const snapshot = store.lookup({
       dataID: ROOT_ID,
       node: CommentQuery.fragment,
       variables: {},
     });
-    const operationSelector = createOperationSelector(CommentQuery, {});
-    environment.commitPayload(operationSelector, payload);
     store.subscribe(snapshot, callback);
     commitMutation(environment, {
       configs,
--- a/packages/relay-runtime/mutations/__tests__/commitRelayModernMutation-test.js
+++ b/packages/relay-runtime/mutations/__tests__/commitRelayModernMutation-test.js
@@ -17,6 +17,7 @@ const {createOperationSelector} = require('RelayModernOperationSelector');
 const {generateAndCompile} = require('RelayModernTestUtils');
 const {ROOT_ID} = require('RelayStoreUtils');
 const {commitMutation} = require('react-relay/modern/ReactRelayPublic');
+const ConnectionHandler = require('RelayConnectionHandler');
 
 describe('Configs: NODE_DELETE', () => {
   jest.resetModules();
@@ -417,7 +418,7 @@ describe('Configs: RANGE_DELETE', () => {
   });
 });
 
-describe('Configs: RANGE_ADD', () => {
+describe('Adding to connections', () => {
   let callback,
     CommentQuery,
     data,
@@ -532,7 +533,7 @@ describe('Configs: RANGE_ADD', () => {
     };
   });
 
-  it('appends new edge', () => {
+  it('Configs: RANGE_ADD appends new edge', () => {
     const configs = [
       {
         type: 'RANGE_ADD',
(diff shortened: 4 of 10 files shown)


alloy:
This PR addresses the issue described in https://github.com/facebook/relay/issues/2333. And while this does not remove the ID, it _is_ related to https://github.com/facebook/relay/issues/2077.

alloy:
@steida Mind testing this PR for your situation? (And is your situation a production app?)

steida:
@alloy My situation is https://github.com/este/este. Not a production 😔

steida:
@kassens What do you think? Should I test it? Is there any chance it will be merged in, say, next six months?

alloy:
@steida We should help the Relay team by testing them preemptively do that when they have time they will know it was already tested by at least somebody.

taion:
Is there a constructive reason for the old behavior, incidentally? It does seem odd to essentially assume idempotency for all mutations.

taion:
BTW, the same problem arises for subscriptions. In this case the fix needs to be applied in this manner, as there's no place to inject a payload sequence ID (as far as I'm aware).

That said, @AndrewIngram's workaround is pretty satisfactory in practice.

sibelius:
@jstejada @kassens can we try to import this again ?

we could also merge this one https://github.com/facebook/relay/pull/2401 when this has landed

jstejada:
thanks for the ping @sibelius! apologies again for the delay in getting to this. We still need to discuss internally a little bit what the best approach would be here. From a short discussion, it seems like we could address separately the problem of ensuring that we don't overwrite mutation payloads vs removing the requirement for a clientMutationID. cc @josephsavona  @kassens

steida:
This open source community PRs neglecting is why people can not trust Facebook open source stuff (DraftJS, Flow) anymore. If nothing change, Relay will die like Flow.

taion:
I just ran into something like this. We've switched our mutations to return unions between the payload type and various typed errors as their root types. When running a mutation with the same inputs leads to a different concrete return type (e.g. an error instead of a payload), we get an invalid record update due to the type mismatch, because the root field gets reused.

tlvenn:
@jstejada do you have any update on this particular issue ? Thanks in advance.

jstejada:
i think @tyao1 already fixed this

robrichard:
@jstejada @tyao1 I'm pretty sure this is still an issue. I just tested in relay-examples todo with relay v7 and was able to reproduce

kassens:
This has been a *long* time coming, thanks @robrichard  for the initial PR and @tyao1 for picking this back up!

earksiinni:
Can someone please comment with respect to the relevance of this PR to the following, taken from the "GraphQL Server Specification" section of the documentation?

> Relay uses a common pattern for mutations, where there are root fields on the mutation type with a single argument, input, and where the input and output both contain a client mutation identifier used to reconcile requests and responses.

Is it correct to say that this documentation is incorrect because client mutation identifiers no longer exist? Or do they exist but they're an internal API (e.g., auto-generated)? Or something else...?

sibelius:
It is still a common pattern, but not a requirement anymore

pvgoddijn:
any guidance on how to implement the mutation identifier when writing a compatible server?

taion:
If you're using JS, graphql-relay provides primitives for this. Otherwise, many other frameworks have their own equivalents. It just depends on which server you're using.

sibelius:
clientMutationId is like an idempotence key (https://stripe.com/docs/api/idempotent_requests)

so your server can validate that user hasn't sent the same mutation twice

taion:
not exactly – it was originally meant for _clients_ to be able to correlate mutation payloads with mutations, without assuming request/response semantics on the network transport

Issue #1913, opened 2017-06-23
GraphQL error handling in Relay Modern


taion:
I think error handling is a little bit weird.

Contra the example in https://facebook.github.io/relay/docs/network-layer.html, it looks like I need to actually `throw` an `Error` in the event that there are any query errors. I can't just pass through the `graphql` object with the `errors` property.

This is because in https://github.com/facebook/relay/blob/v1.1.0/packages/relay-runtime/network/RelayNetwork.js#L197-L208, `normalizePayload` only throws an error if `data` is nully.

But the forward `throw` branch there is the only code path that will trigger `onError` for the request, and correspondingly the only path that will hit `onError` in `<QueryRenderer>` per https://github.com/facebook/relay/blob/v1.1.0/packages/react-relay/modern/ReactRelayQueryRenderer.js#L205-L212.

Otherwise we hit `onNext` per https://github.com/facebook/relay/blob/v1.1.0/packages/react-relay/modern/ReactRelayQueryRenderer.js#L229, and will always set `error: null` on the `readyState` that gets passed to the `render` callback.

This is weird. Am I missing something, or are things supposed to work this way?

johntran:
Looks like there is an error param for the onCompleted callback:

https://github.com/facebook/relay/pull/1938/files

Don't need to throw in the relay network layer anymore.

taion:
Not quite the issue – the problem is actually in QueryRenderer.

maikelrobier:
Yeah, I have found that `QueryRenderer` swallows errors. For instance, If I add `throw new Error()` inside the `render` method of a component rendered with `QueryRenderer` the error won't be shown anywhere.

taion:
That's not what I'm talking about. Of course errors in `render` get swallowed in some sense – but that's just how React works.

cberkom:
Agreed, it would be very helpful to have a way to view the GraphQL errors within the QueryRenderer render method.

eliperkins:
> Otherwise we hit `onNext` per `/packages/react-relay/modern/ReactRelayQueryRenderer.js@v1.1.0#L229`, and will always set `error: null` on the `readyState` that gets passed to the render callback.

This definitely seems like a bug, but I'm not sure what the best resolution is here, API-wise. Should the `onNext` be passed the error? Or should `onError` be called on the existence of any data in the `errors`?

I don't mind submitting a PR to fix it either way, but I'd like some advice on which direction to go in.

DennisKo:
[...]

PR #2352, opened 2018-02-27
Add docs for fetchQuery

robrichard:


docs/Modern-fetchQuery.md
```diff
@@ -0,0 +1,38 @@
...
+  }
+`;
+
+const variables = {
+  pageID: '110798995619330',
+};
+
+fetchQuery(environment, query, variables)
+  .then(data => {
+    // access the graphql response
```
alloy: Does this contain just the fulfilled data or does this object consist of `{ data, errors }` ?

robrichard: just the fulfilled `data`. As far as I can tell, there's no way to get the `errors` from this API.

orta: Would it throw?

alloy: You can have both fulfilled data and errors, so I doubt that.

robrichard: The promise is rejected only if the GraphQL response returns `data: null`, similar to what's described in https://github.com/facebook/relay/issues/1913.

alloy: Ah right, makes sense.

jstejada:
Awesome, thanks for adding this! Were you able to run the website locally to test it?

robrichard:
Yes - once it was running I had to change the version to actually see the changes. I don’t think it will show up by default until the next release is cut.

jstejada:
cool, thanks! yes, that is correct, this won't show up until we cut a new version with docusaurus or manually update the versioned docs.

PR #2353, opened 2018-02-28
Update to QueryRenderer docs

voxmatt:
QueryRenderer props can include a cacheConfig. I'm not sure of the proper place to go into details about what can be included on the cahceConfig object, but it would be nice to include reference here.


jstejada:
thanks @voxmatt !

PR #2354, opened 2018-02-28
Persisted queries

yusinto:
Hi guys, here's my attempt at implementing persisted queries for relay modern. Please review and let me know if it's ok.

Thanks!

### DIFF
--- a/docs/Modern-GraphQLInRelay.md
+++ b/docs/Modern-GraphQLInRelay.md
@@ -144,7 +144,33 @@ Will cause a generated file to appear in `./__generated__/MyComponent.graphql`,
 with both runtime artifacts (which help to read and write from the Relay Store)
 and [Flow types](https://flow.org/) to help you write type-safe code.
 
-The Relay Compiler is responsible for generating code as part of a build step which, at runtime, can be used statically. By building the query ahead of time, the client's JS runtime is not responsible for generating a query string, and fields that are duplicated in the query can be merged during the build step, to improve parsing efficiency. If you have the ability to persist queries to your server, the compiler's code generation process provides a convenient time to convert a query or mutation's text into a unique identifier, which can greatly reduce the upload bytes required in some applications.
+The Relay Compiler is responsible for generating code as part of a build step which, at runtime, can be used statically. By building the query ahead of time, the client's JS runtime is not responsible for generating a query string, and fields that are duplicated in the query can be merged during the build step, to improve parsing efficiency.
+
+### Persisting queries
+The Relay Compiler can convert a query or mutation's text into a unique identifier during compilation. This can greatly reduce the upload bytes required in some applications.
+Using a unique identifier in place of the query text means you have a well-known list of queries that the client can send instead of any
+arbitrary query. This technique of allowing only known queries to be run on the server is called query whitelisting. Whitelisting prevents malicious attacks
+such as:
+
+* an internal dos attack where the query depth is high and exploits circular relationships in the schema
+* a full schema introspection which might not be desirable
+* a harmful pagination query that fetches a million objects
+
+For more information on whitelisting and securing your graphql api from attacks, please refer to this [excellent blog by Max Stoiber](https://dev-blog.apollodata.com/securing-your-graphql-api-from-malicious-queries-16130a324a6b).
+
+The Relay Compiler can persist your queries with the `--persist` flag:
+
+```js
+"scripts": {
+  "relay": "relay-compiler --src ./src --schema ./schema.graphql --persist"
+}
+```
+
+This will create a matching `./__generated__/MyComponent.queryMap.json` containing the query id and the operation text of the query in the same directory.
+The Relay Compiler aggregates all the generated `*.queryMap.json` files into a single complete query map file at `./src/complete.queryMap.json`. You can then use this complete
+json file in your server side to map query ids to operation text.
+
+For more details, refer to the [Persisted Queries section](./persisted-queries.html).
 
 ### Set up relay-compiler
 
@@ -222,6 +248,11 @@ This would produce three generated files, and two `__generated__` directories:
 * `src/Components/__generated__/DictionaryComponent_definition.graphql.js`
 * `src/Queries/__generated__/DictionaryQuery.graphql.js`
 
+If you use `--persist`, then an extra query map json file will also be generated:
+
+* `src/Queries/__generated__/DictionaryQuery.queryMap.json`
+
+Only one query map json file is generated in this instance because only concrete queries can be persisted. Fragments are not persisted.
 
 ### Importing generated definitions
 
--- a/docs/Modern-PersistedQueries.md
+++ b/docs/Modern-PersistedQueries.md
@@ -1,0 +1,164 @@
+---
+id: persisted-queries
+title: Persisted Queries
+---
+
+The relay compiler supports persisted queries which is useful because:
+
+* the client operation text becomes just an md5 hash which is usually shorter than the real
+query string. This saves upload bytes from the client to the server.
+
+* the server can now whitelist queries which improves security by restricting the operations
+that can be run from the client.
+
+## Usage on the client
+
+### The `--persist` flag
+In your `npm` script in `package.json`, run the relay compiler using the `--persist` flag:
+
+```js
+"scripts": {
+  "relay": "relay-compiler --src ./src --schema ./schema.graphql --persist"
+}
+```
+
+The `--persist` flag does 3 things:
+
+1. It converts all query and mutation operation texts to md5 hashes.
+
+    For example without `--persist`, a generated `ConcreteRequest` might look like below:
+
+    ```js
+    const node/*: ConcreteRequest*/ = (function(){
+    //... excluded for brevity
+    return {
+      "kind": "Request",
+      "operationKind": "query",
+      "name": "TodoItemRefetchQuery",
+      "id": null, // NOTE: id is null
+      "text": "query TodoItemRefetchQuery(\n  $itemID: ID!\n) {\n  node(id: $itemID) {\n    ...TodoItem_item_2FOrhs\n  }\n}\n\nfragment TodoItem_item_2FOrhs on Todo {\n    text\n    isComplete\n}\n",
+      //... excluded for brevity
+    };
+    })();
+    ```
+
+    With `--persist` this becomes:
+
+    ```js
+    const node/*: ConcreteRequest*/ = (function(){
+    //... excluded for brevity
+    return {
+      "kind": "Request",
+      "operationKind": "query",
+      "name": "TodoItemRefetchQuery",
+      "id": "3be4abb81fa595e25eb725b2c6a87508", // NOTE: id is now an md5 hash of the query text
+      "text": null, // NOTE: text is null now
+      //... excluded for brevity
+    };
+    })();
+    ```
+
+2. It generates a matching `.queryMap.json` file containing a map of the id and the operation text in the same `__generated__`
+directory as the `.graphql.js` file. In the example above, the `__generated__` directory will have these files:
+
+    * `./__generated__/TodoItemRefetchQuery.graphql.js`
+    * `./__generated__/TodoItemRefetchQuery.queryMap.json`
+
+    The `.queryMap.json` file looks something like this:
+
+    ```json
+    {
+      "3be4abb81fa595e25eb725b2c6a87508": "query TodoItemRefetchQuery(\n  $itemID: ID!\n) {\n  node(id: $itemID) {\n    ...TodoItem_item_2FOrhs\n  }\n}\n\nfragment TodoItem_item_2FOrhs on Todo {\n    text\n    isComplete\n}\n"
+    }
+    ```
+
+3. It also generates a complete query map file at `[your_src_dir]/complete.queryMap.json`. This file contains all the query ids
+and their operation texts. You can specify a custom file path for this file by using the `--persist-output` option:
+
+```js
+"scripts": {
+  "relay": "relay-compiler --src ./src --schema ./schema.graphql --persist --persist-output ./src/queryMaps/queryMap.desktop.queryMap.json"
+}
+```
+
+The example above writes the complete query map file to `./src/queryMaps/queryMap.desktop.queryMap.json`. You need to ensure all the directories
+leading to the `queryMap.json` file exist. Also note that the file extension has to be `.json`.
+
+### Network layer changes
+You'll need to modify your network layer fetch implementation to pass a documentId parameter in the POST body instead of a query parameter:
+
+```js
+function fetchQuery(operation, variables,) {
+  return fetch('/graphql', {
+    method: 'POST',
+    headers: {
+      'content-type': 'application/json'
+    },
+    body: JSON.stringify({
+      documentId: operation.id, // NOTE: pass md5 hash to the server
+      // query: operation.text, // this is now obsolete because text is null
+      variables,
+    }),
+  }).then(response => {
+    return response.json();
+  });
+}
+```
+
+## Usage on the server
+On the server, you'll need to map the query id in the POST body to the real operation text. You can utilise the
+`complete.queryMap.json` file to do this so you'll need a copy of it on your server.
+
+For universal applications where the client and server code are in one project, this is not an issue since you can place
+the query map file in a common location accessible to both the client and the server.
+
+### Compile time push
+For applications where the client and server projects are separate, one option is to have an additional npm run script
+to push the query map at compile time to a location accessible by your server:
+
+```js
+"scripts": {
+  "push-queries": "node ./pushQueries.js",
+  "relay": "relay-compiler --src ./src --schema ./schema.graphql --persist && npm run push-queries"
+}
+```
+
+Some possibilities of what you can do in `./pushQueries.js`:
+
+* `git push` to your server repo
+
+* save the query maps to a database
+
+### Run time push
+A second more complex option is to push your query maps to the server at runtime, without the server knowing the query ids at the start.
+The client optimistically sends a query id to the server, which does not have the query map. The server then in turn requests
+for the full query text from the client so it can cache the query map for subsequent requests. This is a more complex approach
+requiring the client and server to interact to exchange the query maps.
+
+### Simple server example
+Once your server has access to the query map, you can perform the mapping. The solution varies depending on the server and
+database technologies you use, so we'll just cover the most common and basic example here.
+
+If you use `express-graphql` and have access to the query map file, you can import the `complete.queryMap.json` file directly and
+perform the matching using the `matchQueryMiddleware` from [relay-compiler-plus](https://github.com/yusinto/relay-compiler-plus).
+
+```js
+import Express from 'express';
+import expressGraphql from 'express-graphql';
+import {matchQueryMiddleware} from 'relay-compiler-plus';
+import queryMapJson from './complete.queryMap.json';
+
+const app = Express();
+
+app.use('/graphql',
+  matchQueryMiddleware(queryMapJson),
+  expressGraphl({schema}));
+```
+
+## Using `--persist` and `--watch`
+It is possible to continuously generate the query map files by using the `--persist` and `--watch` options simultaneously.
+This only makes sense for universal applications i.e. if your client and server code are in a single project
+and you run them both together on localhost during development. Furthermore, in order for the server to pick up changes
+to the `queryMap.json`, you'll need to have server side hot-reloading set up. The details on how to set this up
+is out of the scope of this document.
+
--- a/packages/relay-compiler/bin/RelayCompilerBin.js
+++ b/packages/relay-compiler/bin/RelayCompilerBin.js
@@ -22,6 +22,7 @@ const {
 const RelaySourceModuleParser = require('../core/RelaySourceModuleParser');
 const RelayFileWriter = require('../codegen/RelayFileWriter');
 const RelayIRTransforms = require('../core/RelayIRTransforms');
+const persistQuery = require('../codegen/persistQuery');
 const RelayLanguagePluginJavaScript = require('../language/javascript/RelayLanguagePluginJavaScript');
 
 const fs = require('fs');
@@ -139,6 +140,8 @@ async function run(options: {
   watch?: ?boolean,
   validate: boolean,
   quiet: boolean,
+  persist: boolean,
+  'persist-output': string,
   noFutureProofEnums: boolean,
   language: string,
   artifactDirectory: ?string,
@@ -149,7 +152,26 @@ async function run(options: {
   }
   const srcDir = path.resolve(process.cwd(), options.src);
   if (!fs.existsSync(srcDir)) {
-    throw new Error(`--source path does not exist: ${srcDir}.`);
+    throw new Error(`--src path does not exist: ${srcDir}.`);
+  }
+
+  const persist = options.persist;
+  let persistOutput = options['persist-output'];
+  if (persistOutput) {
+    persistOutput = path.resolve(process.cwd(), persistOutput);
+    const persistOutputDir = path.dirname(persistOutput);
+    if (!fs.existsSync(persistOutputDir)) {
+      throw new Error(
+        `--persist-output path does not exist: ${persistOutputDir}.`,
+      );
+    }
+
+    const persistOutputFileExtension = path.extname(persistOutput);
+    if (persistOutputFileExtension !== '.json') {
+      throw new Error(
+        `--persist-output must be a path to a .json file: ${persistOutput}.`,
+      );
+    }
   }
   if (options.watch && !options.watchman) {
     throw new Error('Watchman is required to watch for changes.');
@@ -244,9 +266,12 @@ Ensure that one such file exists in ${srcDir} or its parents.
         languagePlugin,
         options.noFutureProofEnums,
         artifactDirectory,
+        persist,
+        persistOutput,
       ),
       isGeneratedFile: (filePath: string) =>
-        filePath.endsWith('.graphql.' + outputExtension) &&
+        (filePath.endsWith('.graphql.' + outputExtension) ||
+          filePath.endsWith('.queryMap.json')) &&
         filePath.includes(generatedDirectoryName),
       parser: sourceParserName,
       baseParsers: ['graphql'],
@@ -281,6 +306,8 @@ function getRelayFileWriter(
   languagePlugin: PluginInterface,
   noFutureProofEnums: boolean,
   outputDir?: ?string,
+  persist: boolean,
+  persistOutput: string,
 ) {
   return ({
     onlyValidate,
@@ -305,6 +332,8 @@ function getRelayFileWriter(
         optionalInputFieldsForFlow: [],
         schemaExtensions,
         useHaste: false,
+        persistQuery: persist ? persistQuery : undefined,
+        persistOutput,
         noFutureProofEnums,
         extension: languagePlugin.outputExtension,
         typeGenerator: languagePlugin.typeGenerator,
@@ -425,6 +454,14 @@ const argv = yargs
         'writing to disk',
       type: 'boolean',
       default: false,
+    },
+    persist: {
+      describe: 'Use an md5 hash as query id to replace operation text',
+      type: 'boolean',
+    },
+    'persist-output': {
+      describe:
+        'The json filepath where the complete query map file will be written to',
     },
     noFutureProofEnums: {
       describe:
--- a/packages/relay-compiler/bin/__tests__/RelayCompilerBin-test.js
+++ b/packages/relay-compiler/bin/__tests__/RelayCompilerBin-test.js
@@ -1,0 +1,89 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ * @emails oncall+relay
+ */
+
+jest.mock('graphql-compiler');
+jest.mock('yargs');
+jest.mock('@babel/polyfill');
+
+require('graphql-compiler');
+const yargs = require('yargs');
+
+describe('RelayCompilerBin', () => {
+  let originalProcessExit;
+  let originalConsoleError;
+
+  const mockCliArguments = (schema = './', src = './', persistOutput = './') => {
+    yargs.usage.mockImplementation(() => ({
+        options: () => ({
+          help: () => ({
+            argv: {schema, src, 'persist-output': persistOutput},
+          }),
+        }),
+      })
+    );
+  };
+
+  beforeEach(() => {
+    jest.resetModules();
+    originalProcessExit = process.exit;
+    originalConsoleError = console.error;
+    process.exit = jest.fn();
+    console.error = jest.fn();
+  });
+
+  afterEach(() => {
+    process.exit = originalProcessExit;
+    console.error = originalConsoleError;
+  });
+
+  test('should throw error when schema path does not exist', async () => {
+    mockCliArguments('./some/path/schema.graphql');
+
+    await require('../RelayCompilerBin');
+
+    expect(console.error).toBeCalled();
+    expect(console.error.mock.calls[0][0]).toEqual(expect.stringContaining('Error: --schema path does not exist:'));
+    expect(process.exit).toBeCalled();
+    expect(process.exit.mock.calls[0][0]).toEqual(1);
+  });
+
+  test('should throw error when src path does not exist', async () => {
+    mockCliArguments('./', './some/path/src');
+
+    await require('../RelayCompilerBin');
+
+    expect(console.error).toBeCalled();
+    expect(console.error.mock.calls[0][0]).toEqual(expect.stringContaining('Error: --src path does not exist:'));
+    expect(process.exit).toBeCalled();
+    expect(process.exit.mock.calls[0][0]).toEqual(1);
+  });
+
+  test('should throw error when persist-output path does not exist', async () => {
+    mockCliArguments('./', './', 'some/path/complete.queryMap.json');
+
+    await require('../RelayCompilerBin');
+
+    expect(console.error).toBeCalled();
+    expect(console.error.mock.calls[0][0]).toEqual(expect.stringContaining('Error: --persist-output path does not exist:'));
+    expect(process.exit).toBeCalled();
+    expect(process.exit.mock.calls[0][0]).toEqual(1);
+  });
+
+  test('should throw error when persist-output path does not end with a .json extension', async () => {
+    mockCliArguments('./', './', './queryMap.graphql.js');
+
+    await require('../RelayCompilerBin');
+
+    expect(console.error).toBeCalled();
+    expect(console.error.mock.calls[0][0]).toEqual(expect.stringContaining('Error: --persist-output must be a path to a .json file'));
+    expect(process.exit).toBeCalled();
+    expect(process.exit.mock.calls[0][0]).toEqual(1);
+  });
+});
--- a/packages/relay-compiler/codegen/RelayFileWriter.js
+++ b/packages/relay-compiler/codegen/RelayFileWriter.js
@@ -14,10 +14,12 @@ const RelayParser = require('../core/RelayParser');
 const RelayValidator = require('../core/RelayValidator');
 
 const compileRelayArtifacts = require('./compileRelayArtifacts');
-const crypto = require('crypto');
 const graphql = require('graphql');
 const invariant = require('invariant');
 const path = require('path');
+const fs = require('fs');
+const md5 = require('../util/md5');
+
 const writeRelayGeneratedFile = require('./writeRelayGeneratedFile');
 
 const {
@@ -58,6 +60,7 @@ export type WriterConfig = {
   outputDir?: ?string,
   generatedDirectories?: Array<string>,
   persistQuery?: (text: string) => Promise<string>,
+  persistOutput?: string,
   platform?: string,
   schemaExtensions: Array<string>,
   noFutureProofEnums: boolean,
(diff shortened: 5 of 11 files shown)


packages/relay-compiler/codegen/RelayFileWriter.js
```diff
@@ -353,13 +360,39 @@ class RelayFileWriter implements FileWriterInterface {
...
+              ...singleQueryMap,
+            };
+          }
+        });
+      });
+
+      fs.writeFileSync(queryMapFilePath, JSON.stringify(queryMapJson));
+      this._reporter.reportMessage(
+        `Written complete queryMap file to ${queryMapFilePath}`,
+      );
```
alloy: I haven’t looked into how other files are written to disk by Relay, but is it normal to use synchronous access here instead of async?

packages/relay-compiler/codegen/persistQuery.js
```diff
@@ -0,0 +1,7 @@
+import md5 from '../util/md5';
```
alloy: This file probably needs a FB header like all other files.

packages/relay-compiler/util/md5.js
```diff
@@ -0,0 +1,11 @@
+//@flow
```
alloy: This file probably needs a FB header like all other files.

packages/relay-compiler/codegen/RelayFileWriter.js
```diff
@@ -353,13 +360,39 @@ class RelayFileWriter implements FileWriterInterface {
...
-    .update(x, 'utf8')
-    .digest('hex');
+  /**
+   * Find all *.queryMap.json and write it into a single file.
+   * @param allOutputDirectories
+   */
+  writeCompleteQueryMap(
+    allOutputDirectories: Map<string, CodegenDirectory>,
+  ): void {
+    const queryMapFilePath = `${this._config.baseDir}/queryMap.json`;
```
alloy: I’m wondering if it would make sense to take an optional path value for the `--persist` option to change this location?

packages/relay-compiler/bin/RelayCompilerBin.js
```diff
@@ -163,9 +165,10 @@ Ensure that one such file exists in ${srcDir} or its parents.
   };
   const writerConfigs = {
     js: {
-      getWriter: getRelayFileWriter(srcDir),
+      getWriter: getRelayFileWriter(srcDir, options.persist),
       isGeneratedFile: (filePath: string) =>
-        filePath.endsWith('.js') && filePath.includes('__generated__'),
+        (filePath.endsWith('.js') || filePath.endsWith('.queryMap.json')) &&
```
alloy: Wondering if it would make sense to follow the `*.graphql.extname` filename pattern used by the other artifacts? Because as shown in the patch in [this comment](https://github.com/facebook/relay/pull/2293#issuecomment-369036035) it wouldn’t be unthinkable to simplify artefact detection somewhat by not matching a specific extension.

alloy:
FYI This supersedes PR #1846.

robrichard:
The downside with this approach over https://github.com/facebook/relay/pull/1846 is that the query ids are not configurable. Your GraphQL server has to be able to lookup the query by the MD5 hash. 

This may not be ideal depending on the storage mechanism you are using to persist the queries server side.

packages/relay-compiler/codegen/RelayFileWriter.js
edvinerikson: Would be nice, in our app we have multiple compilations going on... so I guess without that, it would overwrite existing compilation.

alloy:
@robrichard With the language plugin support in https://github.com/facebook/relay/pull/2293 and the proposed pre-processing plugin support here https://github.com/facebook/relay/pull/2293#issuecomment-369212223, it’s not a stretch imo to imagine being able to do the same for query IDs and have MD5 hashing being a sane builtin default.

robrichard:
@alloy Yes, so it seems there is precedence now for Relay compiler allowing you to run a custom JS module. You can add another to get custom IDs in the `queryMap.json`, but then you still need another script to process this file to actually persist the queries.

#1846 seemed a lot cleaner because you just had to write one function to persist the query and return the ID.

alloy:
On my phone now, so will look more closely to code again later, but I don’t see why you would be able to do the same in this situation. Once you have the query to e.g. hash it and return an ID, can’t you also take that moment to persist it?

alloy:
…or did you mean when using the default builtin md5 one?

robrichard:
> Once you have the query to e.g. hash it and return an ID, can’t you also take that moment to persist it?

Yes you can, but then the logic to write `queryMap.json` is superfluous.

Just passing a persist function from a user supplied module directly to `RelayFileWriter` like in https://github.com/facebook/relay/pull/1846 would give you the flexibility to implement persisting however you see fit.

packages/relay-compiler/codegen/RelayFileWriter.js
yusinto: Other areas in Relay are using sync as well for example `RelayCompilerCache` line 55 and `CodegenDirectory` line 213. I think it's possible to change this to async, although I think it won't really make a significant impact. Generally I tend to use sync for file operations than async just to avoid file locking and concurrency issues. Imho.

yusinto:
@robrichard in [this pr](https://github.com/facebook/relay/pull/1846), @devknoll specified that `--persist` should be a command rather than a module. It's not hard to code, but we just need some concensus. Personally I've been using relay-compiler-plus for a while and I don't really change the persist function very much. You can use the resultant complete`./src/queryMap.json` file from the compilation process to say perform bulk inserts into your database at the end of the compilation rather than making many round trips during compilation which I believe will speed up your compile time.

packages/relay-compiler/codegen/persistQuery.js
yusinto: Done

packages/relay-compiler/util/md5.js
yusinto: Done

packages/relay-compiler/bin/RelayCompilerBin.js
yusinto: This is an interesting one. I'm not sure the query map file qualifies as a `graphql` file. It's a plain json file and naming it `graphql` could be misleading.

alloy: Agreed that it’s a tricky question. The JS artefacts also aren’t strictly GraphQL files but also derivatives, so figured that applies to these as well.

(Just to be clear, I meant something like `Foo.graphql.json`)

yusinto: That's a good point actually I never thought about that. I'll go ahead with the rename. Thanks for the response!

docs/Modern-GraphQLInRelay.md
```diff
@@ -142,7 +142,20 @@ Will cause a generated file to appear in `./__generated__/MyComponent.graphql`,
...
+
+```js
+"scripts": {
+  "relay": "relay-compiler --src ./src --schema ./schema.graphql --persist"
+}
+```
+
+This will create a matching `./__generated__/MyComponent.queryMap.json` containing the query id and the operation text of the query in the same directory.
+The Relay Compiler aggregates all the generated `*.queryMap.json` into a single complete query map file at `./src/queryMap.json`. You can then use this complete
+json file in your server side to map query ids to operation text.
```
alloy: Could you maybe add some details about where this would typically sit in one’s workflow?

For instance, I’m naively thinking that you’d _only_ use the `--persist` option once you decide to cut a release/perform a deploy, seeing as the server needs to know about these queries and you probably don’t want to continuously do that, is that correct?

It would also be nice to have an example of what you do with the resulting query map file, i.e. describe uploading to server and serving requests that use query IDs. This could maybe even be added to the canonical [example app](https://github.com/relayjs/relay-examples)?

alloy:
> @devknoll specified that `--persist` should be a command rather than a module. It's not hard to code, but we just need some concensus.

Agreed. Apologies for going off on a tangent regarding plugins.

> Personally I've been using relay-compiler-plus for a while and I don't really change the persist function very much.

Can you expand a bit on what you mean by not changing it “much”? Do you still make customisations and, if so, what are they?

> You can use the resultant complete`./src/queryMap.json` file from the compilation process to say perform bulk inserts into your database at the end of the compilation rather than making many round trips during compilation which I believe will speed up your compile time.

This makes sense to me, as in my naive thought you’d only run this on release time. Maybe some more information about _how_ one would use this in practice is what’s missing? (Left a comment asking for that [here](https://github.com/facebook/relay/pull/2354/files#r171505157).)

docs/Modern-GraphQLInRelay.md
yusinto: My use case is explained in my blog post [here](http://www.reactjunkie.com/relay-modern-persisted-queries/). My app is universal and I have hot reload setup on both the client and server. The express server imports the complete queryMap.json file and uses a middleware to match queries from graphql requests.

During development, I change my graphql queries a lot (in relay `graphql` tags). Each time this happens the operation text changes and hence the query id changes because it is an md5 hash of the operation text. The graphql.js and queryMap.json files get recompiled, and my express server reloads the queryMap.json and is able to continue mapping queries correctly without having to stop/start. 

The blog (http://www.reactjunkie.com/relay-modern-persisted-queries/) contains details about the express middleware and a complete working example. Hopefully the example can shed more light into this.

yusinto:
@alloy apologies by being ambiguous. I should have said my persist function does not change at all. I just use the md5 hash.

docs/Modern-GraphQLInRelay.md
MatthewHerbst: Have to agree with @alloy here. We have three client repos that each have their own queries, and then a 4th repo for our gql server. We'd likely set up a 5th repo to store persisted queries in.

docs/Modern-GraphQLInRelay.md
```diff
@@ -142,7 +142,20 @@ Will cause a generated file to appear in `./__generated__/MyComponent.graphql`,
 with both runtime artifacts (which help to read and write from the Relay Store)
 and [Flow types](https://flow.org/) to help you write type-safe code.
 
-The Relay Compiler is responsible for generating code as part of a build step which, at runtime, can be used statically. By building the query ahead of time, the client's JS runtime is not responsible for generating a query string, and fields that are duplicated in the query can be merged during the build step, to improve parsing efficiency. If you have the ability to persist queries to your server, the compiler's code generation process provides a convenient time to convert a query or mutation's text into a unique identifier, which can greatly reduce the upload bytes required in some applications.
+The Relay Compiler is responsible for generating code as part of a build step which, at runtime, can be used statically. By building the query ahead of time, the client's JS runtime is not responsible for generating a query string, and fields that are duplicated in the query can be merged during the build step, to improve parsing efficiency. 
+
+### Persisting queries
```
MatthewHerbst: It would be nice to add a paragraph maybe about _why_ someone would want to persist queries. There is both the performance benefits, and, security benefits if you use the persisted queries as a query whitelist.

packages/relay-compiler/codegen/RelayFileWriter.js
```diff
@@ -353,13 +360,39 @@ class RelayFileWriter implements FileWriterInterface {
...
+              ...queryMapJson,
+              ...singleQueryMap,
+            };
+          }
+        });
+      });
+
+      fs.writeFileSync(queryMapFilePath, JSON.stringify(queryMapJson, null, 2));
+      this._reporter.reportMessage(
+        `Written complete queryMap file to ${queryMapFilePath}`,
```
MatthewHerbst: Grammar nit: "Wrote"
`Wrote complete queryMap file to ${queryMapFilePath}`

Or, word order change:
`Complete queryMap written to file $${queryMapFilePath}`

packages/relay-compiler/codegen/RelayFileWriter.js
yusinto: @edvinerikson each one of your compilation should have a `--src` specified, which means a separate `queryMap.json` file will be created for each src. Is that your use case? If not can you please provide more details on what you mean by multiple compilations?

docs/Modern-GraphQLInRelay.md
yusinto: > I understand that in such a situation you basically have the JSON file in the same project and so it’s easy to continuously write the persisted queries, but in setups where the server is a different project altogether it doesn’t seem necessary to do this continuously. What’s more, in the situation I’m describing there isn’t even a good way to persist the queries on the server during development and running relay-compiler --watch.

Right I got your point. So there are 2 ways to use `--persist`:

1. As a single run with `--persist` will produce a master queryMap.json file. This does a one time compilation and stops. It doesn't do any continuous watching or compiling.

2. In combination with `--watch`. If you specify `--persist` and `--watch`, then queryMap.json will continuously update with the changes you make to your queries.

If you don't need continuous updates to queryMap.json, you can just do a single run like step 1. The watch mode is useful if like me you have the luxury of having a universal app where the server and client is in one project.

In the case where the server is a separate project and you run `--persist` either as a single run or in watch mode in the client, you'll still need to somehow deploy the query map file to your server. In the simplest case, this can be a git push to your server repo to update the server query map file. In more complex scenarios this can be a build step which does  database updates.

>As such, I think it would be great if persisted queries gets its own doc page where you could both include the example from your blog post for a universal app that includes the server and has a more agnostic example of a separate server that anybody can adjust to their situation.

That's a great suggestion. I'll document what I've said here and in my blog post in a standalone doc page specifically for persisted queries.

packages/relay-compiler/codegen/RelayFileWriter.js
yusinto: Thanks! Done.

docs/Modern-GraphQLInRelay.md
yusinto: Good point. Done.

packages/relay-compiler/bin/RelayCompilerBin.js
yusinto: @alloy
>Wondering if it would make sense to follow the *.graphql.extname filename pattern used by the other artifacts? Because as shown in the patch in this comment it wouldn’t be unthinkable to simplify artefact detection somewhat by not matching a specific extension.

Done

packages/relay-compiler/codegen/RelayFileWriter.js
edvinerikson: I am executing them like this
```
    "compile-gql-documents:desktop": "relay-compiler --schema ./node_modules/@leogears/leo-gateway-schema/schema.json --src ../../ --include src/common/** --include src/web/common/** --include src/web/desktop/**",
    "compile-gql-documents:mobile": "relay-compiler --schema ./node_modules/@leogears/leo-gateway-schema/schema.json --src ../../ --include src/common/** --include src/web/common/** --include src/web/mobile/**",
```

yusinto: Gotcha. You are right in your case the query map files will override each other. I will add an option to specify an output location. Thanks for the feedback!

packages/relay-compiler/bin/RelayCompilerBin.js
alloy: Ace 👍

docs/Modern-GraphQLInRelay.md
yusinto: >As such, I think it would be great if persisted queries gets its own doc page where you could both include the example from your blog post for a universal app that includes the server and has a more agnostic example of a separate server that anybody can adjust to their situation.

Done.

packages/relay-compiler/codegen/RelayFileWriter.js
yusinto: >I’m wondering if it would make sense to take an optional path value for the --persist option to change this location?
Done. @edvinerikson @alloy

yusinto:
@edvinerikson @alloy @MatthewHerbst I think I've done all outstanding items in this PR. Are you guys able to take another look and confirm please? Will be good to have this merged at some point soon, so the rest of the community can finally use it. It's been a long time coming. Thank you.

docs/Modern-PersistedQueries.md
```diff
@@ -0,0 +1,149 @@
...
+```js
+function fetchQuery(operation, variables,) {
+  return fetch('/graphql', {
+    method: 'POST',
+    headers: {
+      'content-type': 'application/json'
+    },
+    body: JSON.stringify({
+      queryId: operation.id, // NOTE: pass md5 hash to the server
+      // query: operation.text, // this is now obsolete because text is null
```
MatthewHerbst: It should be noted there is a 3rd, but much more complex, way to get the queries to the server. If all you care about is performance, then you can optimistically send the ID to the server. If the server responds that it doesn't know the ID, then you can send the whole query text, which the server can then cache. The next time you send the query ID to the server, it will have it. This is how [apollo-link-persisted-queries](https://www.npmjs.com/package/apollo-link-persisted-queries) works. I don't think we need to write an implementation of this, but I think we should note in the docs here that it is possible.

packages/relay-compiler/bin/RelayCompilerBin.js
```diff
@@ -94,15 +95,37 @@ async function run(options: {
...
   }
+
+  const persist = options.persist;
+  let persistOutput = options['persist-output'];
+  if (persistOutput) {
+    persistOutput = path.resolve(process.cwd(), persistOutput);
+    const persistOutputDir = path.dirname(persistOutput);
+    if (!fs.existsSync(persistOutputDir)) {
+      throw new Error(
+        `--persist-output path does not exist: ${persistOutputDir}.`,
```
MatthewHerbst: It would be nice to mention this in the docs with/near `--persist`

yusinto: Good point. Will update the docs. Thank you.

docs/Modern-PersistedQueries.md
yusinto: That's an interesting approach. It seems to conflict with whitelisting though? You can't optimistically send queries if you are also whitelisting queries. Perhaps I should include this point in the docs as well. Will update.

MatthewHerbst: Yeah, that's true, but it is an option of all you're looking to do is optimization. I lot of people won't be able to whitelist because they allow 3rd party queries, but this option allows them (or people who don't want to/can't update a central repository) to still cache queries.

yusinto: Gotcha. Will include in the docs. Thanks for the explanation!

packages/relay-compiler/bin/RelayCompilerBin.js
```diff
@@ -339,9 +366,15 @@ const argv = yargs
...
     persist: {
-      describe: 'Persist queries to srcdir/queryMap.graphql.json',
+      describe: 'Use an md5 hash as query id to replace operation text',
       type: 'boolean',
     },
+    'persist-output': {
+      describe:
+        'The json filepath where the complete query map file will be written to',
+      type: 'string',
+      default: null,
```
alloy: Is it not possible with yargs to use a single option (`—persist`) both with and without value?

yusinto: >Is it not possible with yargs to use a single option (—persist) both with and without value?

@alloy I spent some time looking into this. If we use just a single `--persist` option then the param type will have to change to a `string` with a default of `do-not-persist`. Like so:
```js
    persist: {
      describe: 'Use an md5 hash as query id to replace operation text',
      type: 'string',
      default: 'do-not-persist',
    },
```

Then the logic in the `run` method to detect persistence will change to this:
```js
  const persist = options.persist;
  if (persist === 'do-not-persist') {
    console.log('Not persisting');
  } else if (!persist.trim()) {
    console.log(`Persisting to default output location: ${srcDir}/queryMap.graphql.json`);
  } else {
    console.log(`Persisting to custom output location ${persist}`);
  }
```

The `--help` documentation on the console will display:
```
Options:
  --persist     Use an md5 hash as query id to replace operation text
                                            [string] [default: "do-not-persist"]
// other params...
```
So it is possible to just use a single `--persist` flag, however I feel that it's doing a bit too much and is not as clean as separating the two logical actions as separate flags. The `--help` documentation in particular is confusing, because a default string of `do-not-persist` doesn't really make sense to users unless they analyse the relay-compiler code. We can change the default string to something else, but it's not really going to help because this default string is only used in the code to detect if the `--persist` flag is completely absent from the command line. Hopefully that makes sense?

packages/relay-compiler/bin/RelayCompilerBin.js
yusinto: Done @MatthewHerbst

docs/Modern-PersistedQueries.md
yusinto: Done @MatthewHerbst

yusinto:
>LGTM. Just my two comments on docs. If you wanted slightly more test coverage it would be nice to see tests that tested that the new errors are thrown when you expected them to be.

Done @MatthewHerbst. Just waiting for @alloy's response now re using a single`--persist` command.

packages/relay-compiler/bin/RelayCompilerBin.js
alloy: I find it unfortunate that the tooling doesn’t nicely allow for the best UI, but your explanation makes sense and I agree with your conclusion 👍

alloy:
Thanks again for working on this @yusinto! I too can’t wait to use this.

@jstejada 👍

docs/Modern-GraphQLInRelay.md
```diff
@@ -142,7 +142,25 @@ Will cause a generated file to appear in `./__generated__/MyComponent.graphql`,
...
+The Relay Compiler can persist your queries with the `--persist` flag:
+
+```js
+"scripts": {
+  "relay": "relay-compiler --src ./src --schema ./schema.graphql --persist"
+}
+```
+
+This will create a matching `./__generated__/MyComponent.graphql.json` containing the query id and the operation text of the query in the same directory.
+The Relay Compiler aggregates all the generated `*.graphql.json` into a single complete query map file at `./src/queryMap.graphql.json`. You can then use this complete
```
alloy: > all the generated `*.graphql.json` **files**

docs/Modern-GraphQLInRelay.md
```diff
@@ -142,7 +142,25 @@ Will cause a generated file to appear in `./__generated__/MyComponent.graphql`,
...
+json file in your server side to map query ids to operation text.
+
+Fore more details, refer to the [Persisted Queries section](./persisted-queries.html).
```
alloy: > For more details

(drop the e after For)

docs/Modern-GraphQLInRelay.md
```diff
@@ -142,7 +142,25 @@ Will cause a generated file to appear in `./__generated__/MyComponent.graphql`,
...
+The Relay Compiler is responsible for generating code as part of a build step which, at runtime, can be used statically. By building the query ahead of time, the client's JS runtime is not responsible for generating a query string, and fields that are duplicated in the query can be merged during the build step, to improve parsing efficiency. 
+
+### Persisting queries
+The Relay Compiler can convert a query or mutation's text into a unique identifier during compilation. This can greatly reduce the upload bytes required in some applications.
+Using a unique identifier in place of your query or mutation also allows you to whitelist operations that are allowed on your server which improves security.
```
alloy: Should this expand a little more on _how_ it may improve security so people can decide if that applies to them? (Possibly by linking to a good article on query depth as a possible DOS vector.)

docs/Modern-GraphQLInRelay.md
```diff
@@ -220,6 +238,11 @@ This would produce three generated files, and two `__generated__` directories:
 * `src/Components/__generated__/DictionaryComponent_definition.graphql.js`
 * `src/Queries/__generated__/DictionaryQuery.graphql.js`
 
+If you use `--persist`, then an extra query map json file will also be generated:
+
+* `src/Queries/__generated__/DictionaryQuery.graphql.json`
+
+Only one query map json is generated in this instance because only concrete queries can be persisted. Fragments are not persisted.
```
alloy: > query map json **file**

Not sure if ‘a json’ is used as equivalent of a JSON file nowadays, just reads a little nicer to me, but I’m also a non-native reader/writer so 🤷‍♂️

packages/relay-compiler/bin/__tests__/RelayCompilerBin-test.js
```diff
@@ -0,0 +1,79 @@
...
+    mockCliArguments('./', './', './queryMap.graphql.js');
+
+    await require('../RelayCompilerBin');
+
+    expect(console.error).toBeCalled();
+    expect(console.error.mock.calls[0][0]).toEqual(expect.stringContaining('Error: --persist-output must be a path to a .json file'));
+    expect(process.exit).toBeCalled();
+    expect(process.exit.mock.calls[0][0]).toEqual(1);
+  });
+});
```
alloy: Only just noticed that you added the first bin tests, awesome! 👏

docs/Modern-PersistedQueries.md
```diff
@@ -0,0 +1,162 @@
...
+Some possibilities of what you can do in `./pushQueries.js`:
+
+* `git push` to your server repo
+
+* save the query maps to a database
+
+It is also possible to push your query maps to the server at runtime, without the server knowing the query ids at the start.
+The client optimistically sends a query id to the server, which does not have the query map. The server then in turn requests
+for the full query text from the client so it can cache the query map for subsequent requests. This is a more complex approach
+requiring the client and server to interact to exchange the query maps.
```
alloy: Should this be another entry in the list of options, rather than a separate paragraph?

docs/Modern-PersistedQueries.md
```diff
@@ -0,0 +1,162 @@
...
+
+### Simple server example
+Once your server has access to the query map, you can perform the mapping. The solution varies depending on the server and
+database technologies you use, so we'll just cover the most common and basic example here.
+ 
+If you use `express-graphql` and have access to the query map file, you can import the `queryMap.graphql.json` file directly and
+perform the matching using the `matchQueryMiddleware` from [relay-compiler-plus](https://github.com/yusinto/relay-compiler-plus).
```
alloy: Is this repo going to continue to exist once this feature would be merged and released?

alloy:
@yusinto FYI I had another chat with @robrichard about his specific needs and we concluded that it would make sense in a next PR to make persistence pluggable and make this the default built-in plugin.

packages/relay-compiler/bin/RelayCompilerBin.js
robrichard: You can remove the `type` option and then `yargs` will treat it as either a boolean if used alone or a string if passed with a value. However I agree that they should still be separate options to not overload the meaning of `--persist`.

docs/Modern-GraphQLInRelay.md
MatthewHerbst: There's two parts to that:

* Security against internal DOS
* Security against full schema introspection (which may or may not be undesirable)

docs/Modern-PersistedQueries.md
yusinto: Yes. I am actively maintaining this repo and plan to do so indefinitely.

docs/Modern-GraphQLInRelay.md
yusinto: Done.

docs/Modern-GraphQLInRelay.md
yusinto: Done.

docs/Modern-GraphQLInRelay.md
yusinto: Good point and done.

docs/Modern-GraphQLInRelay.md
yusinto: Done.

docs/Modern-PersistedQueries.md
yusinto: I have re-organised the server push options to have a compile time and a run time section. Optimistic query is a run-time solution and so is separate from the `npm run` option which is a compile time solution. Please have a look and let me know if this looks ok. Thanks @alloy.

yusinto:
@alloy I think I have done all the outstanding tasks from the last round of reviews. Are you able to take another look please? Thanks.

yusinto:
@robrichard @MatthewHerbst @edvinerikson @alloy Hi guys, I've addressed all issues in this PR and it's been sitting unmoved for 3 days now. Can you please advise what the next step is to get this merged? It will be good for the relay community to finally be able to use persisted queries.

Thank you.

MatthewHerbst:
@yusinto when FB hires me I'll make sure it get's merged :) until then @alloy is probably your best bet

robrichard:
@yusinto, @alloy and I are also not FB employees who can merge here. I think @jstejada is your best bet.

alloy:
@yusinto Hey, sorry for delay in response, I can def take another quick look! Like @robrichard mentions, I’m also no FB employee, the best we as community members can do is try to take away any obvious issues that would prevent the Relay team from merging it and thus try to optimise the very limited time the Relay team has to review PRs. Having said that, you should probably expect that it will take a little while for PRs that are sizeable to be merged (take a look at mine, for instance) and so it’s probably good practice to keep an eye out for your PR getting into a merge-conflict state and preemptively rebase etc.

packages/relay-compiler/util/md5.js
```diff
@@ -0,0 +1,23 @@
...
+import crypto from 'crypto';
+
+const md5 = (x: string): string => {
+  return crypto
+    .createHash('md5')
+    .update(x, 'utf8')
+    .digest('hex');
+};
+
+module.exports = md5;
```
alloy: Btw, just out of curiosity, I was wondering why you chose md5 over e.g. sha-256? Is it because the hash is shorter and the risk of collisions is very low when the user doesn’t maliciously _try_ to craft a document that would generate a collision (which I _think_ is the status quo, but I’m no expert)?

yusinto: It's an existing hash function from the existing code base, I merely refactored it into a separate file for re-use. I'm no expert as well, but there shouldn't be any problem changing it to sha-256 if facebook wants to. I think that's a little out of scope of what this PR is about though, so I will just leave this as is for now.

Re merging and time, yes ok I understand it will take some time. However is it possible for you to approve to indicate that everything looks ok from the reviewer's perspective? Of course only do that if you do think that everything looks ok. Happy to make more changes if you see the need. @alloy . Thanks!

packages/relay-compiler/bin/RelayCompilerBin.js
alloy: @yusinto Hrmm, I’ve been trying to integrate this PR into the build we use at Artsy and I’m finding that in our case the (single query map) json file gets loaded at runtime _before_ the source artefact file (a `.tsx` file in our case). Have you been able to use this exact code in production?

Changing these filenames back to `*.queryMap.json` makes it work well.

alloy:
@yusinto Btw, the output looks slightly messy:

```
Complete queryMap written to /Users/eloy/Code/Artsy/emission/src/__generated__/complete.queryMap.json
Updated:
 - ForYou_forYou.graphql.ts
 - QueryRenderersForYouQuery.graphql.ts
 - QueryRenderersForYouQuery.queryMap.json
Deleted:
 - complete.queryMap.json
```

…compared to the existing pattern, where I’d expect something like:

```
Updated:
 - ForYou_forYou.graphql.ts
 - QueryRenderersForYouQuery.graphql.ts
 - QueryRenderersForYouQuery.queryMap.json
 - complete.queryMap.json
```

Did you look into that previously?

alloy:
@yusinto bump

yusinto:
@alloy merged

> Made a PR to fix that which you can merge into this PR yusinto#1

yusinto:
@alloy the complete query map file is always regenerated and never updated that's why it appears under 'updated' in the output. Is this what you mean by messy? It's difficult to update query maps because it involved searching and comparing and I don't think the effort is worth it.

>@yusinto Btw, the output looks slightly messy:

alloy:
> @alloy the complete query map file is always regenerated and never updated that's why it appears under 'updated' in the output. Is this what you mean by messy? It's difficult to update query maps because it involved searching and comparing and I don't think the effort is worth it.

Hmm, I see what you mean, that’s unfortunate.

However, then I would kinda expect it under a `Created:` section. The reason it feels messy is because it doesn’t follow the other output in style and thus stands out. E.g. it also uses a full path rather than just the basename.

It’s definitely not a blocker imo, I’ll leave that up to the Relay team. I’m just a bit of a clean/consistent UI aficionado ;)

alloy:
@yusinto I wanted to let you know that we will start using this PR in production at Artsy with the next version of our iOS app: https://github.com/artsy/emission/pull/999

One piece of feedback I wanted to give is that I ended up doing the client-server interaction a little different. Namely, I went with `documentID` instead of the `queryId` that’s outlined in the docs in this PR, because `documentID` is the parameter that was settled on in the open PR to express-graphql: https://github.com/graphql/express-graphql/pull/109/files#diff-04c6e90faac2675aa89e2176d2eec7d8R61. So I think it would make sense to reflect that name in this PR too.

soneymathew:
@yusinto have you had some time to look into this? Any update? Eagerly waiting for this. :)

yusinto:
Hi Im off in South America for the next two weeks. Ill get to this when I
return.

dwwoelfel:
A lot of good work went into this, but I think it's unlikely to be merged as-is. It adds too many assumptions about how persisting queries should work. For example, Facebook wouldn't be able to use a queryMap file because they have far too many persisted queries. Even more importantly, they need to support every query that has ever been persisted since some people never update their apps.

Persisting queries is already supported in Relay Modern in a pluggable way, but it requires a lot of manual effort and basically writing your own RelayCompilerBin. I think a better approach here would be to leverage WriterConfig/ReaderConfig and allow users to easily use their own writerConfigs and readerConfigs when invoking relay. Then you could provide a custom writerConfig that does its work from the outside instead of adding new assumptions to core files like `writeRelayGeneratedFile` and `RelayFileWriter`.

alloy:
> For example, Facebook wouldn't be able to use a queryMap file because they have far too many persisted queries.
> […]
> Persisting queries is already supported in Relay Modern in a pluggable way, but it requires a lot of manual effort and basically writing your own RelayCompilerBin. 

Correct me if I’m wrong, but doesn’t Facebook already use its own RelayCompilerBin? That renders the former point moot imo. The point of this PR is mostly to offer a simple solution that will work for many out-of-the-box or just starting. Users that have more demands can still do so, like they do today.

MatthewHerbst:
Agreed. Relay needs to start shipping features with default options that work for _most_ users. It doesn't need to support every use case out of the box, and advanced users should always be able to replace modular functionality (such as persisted queries) with their own solutions. Without sensible defaults that are easy to use (as in don't require a lot of complicated setup and config) more and more users, including myself, are going to feel the draw of switching to Apollo simply due to lack of feature parity.

docs/Modern-PersistedQueries.md
```diff
@@ -0,0 +1,164 @@
...
+### Simple server example
+Once your server has access to the query map, you can perform the mapping. The solution varies depending on the server and
+database technologies you use, so we'll just cover the most common and basic example here.
+
+If you use `express-graphql` and have access to the query map file, you can import the `complete.queryMap.json` file directly and
+perform the matching using the `matchQueryMiddleware` from [relay-compiler-plus](https://github.com/yusinto/relay-compiler-plus).
+
+```js
+import Express from 'express';
+import expressGraphl from 'express-graphql';
```
soneymathew: nit: typo! correct to expressGraph"Q"L :)

yusinto: Fixed. Thank you.

johntran:
@kassens @jstejada Free for another review :).

alloy:
@johntran @yusinto Btw, this request for change is still outstanding https://github.com/facebook/relay/pull/2354#issuecomment-379756685.

If/when you do that, could you also maybe squash all the changes into a single commit? That’s going to happen anyways when FB merges it and in the meantime it makes rebasing/merging it into fork branches a lot easier 🙏

johntran:
I've created a PR with @alloy's suggestions to yusinto's Relay repo. This merges the latest master (Relay 1.7.0.rc-1) and renamed all instances of queryId with documentId (only in files where the persisted query PR added).
https://github.com/yusinto/relay/pull/3

I am not confident in my rebasing skills, but I definitely want to learn. I would want to rebase all the changes in the persisted query branch and squash them; how would I do that? Any guides you can show me?

I just finished a product launch, so I'm able to dedicate some time every week towards updating this PR.

alloy:
@johntran I’ll follow-up on https://github.com/yusinto/relay/pull/3

johntran:
@jstejada @alunyov @kassens I think we are ready for another review

Thanks @soneymathew for the detailed rebase & squash instructions :).

soneymathew:
@yusinto done! Please review.
https://github.com/facebook/relay/compare/master...yusinto:persisted-queries
If time permits It will be nice if you could amend the commit message to be more concise.

Change summary
- git rebase to master with squashed commits
- removed references to BATCHED_REQUEST
- Added copyright headers to some files
- ensured that all tests/typecheck/prettier pass

soneymathew:
@yusinto done!
All the commits are squashed to one, resolved the merge conflicts and added FB headers to some files that missed it.
https://github.com/facebook/relay/compare/master...yusinto:persisted-queries
I had to remove the references to BATCHED_REQUEST as master doesn't have it as well

mrtnzlml:
Hello, can you please share with us what is blocking this PR? Thank you. :)

josephsavona:
cc @alunyov Thoughts on this? Seems like it would be ideal to avoid creating query "map" files adjacent to every generated file only to later accumulate that data into a single map at the top level - and even then, only write that file if the user opts into creating it.

alloy:
@josephsavona What are you thoughts on @soneymathew's last comment?

josephsavona:
@alloy I’ve imported this but am changing the implementation a bit to emit only a single aggregated query file with all persisted queries. It should be easy to later extract the logic if folks prefer that, but I want to get *something* landed to at least unblock folks and give a point to build from.

alloy:
Sounds great, thanks @josephsavona 👍

PR #2357, opened 2018-03-04
fix typo grapqhl => graphql in quick start guide

j-nolan:
`graphql` is spelled `grapqhl` in some places of the documentation.

Not a huge deal, but easy to fix so here is a PR.


jstejada:
thanks!!

PR #2359, opened 2018-03-06
Allow compatible variable types with `@relay(mask: true)` 

### LINKED ISSUE #2358: Equivalence check is too strict when comparing arguments in unmasked fragments
This was introduced with Relay 1.5.0

For this schema:
```graphql
schema {
  query: Query
}

type Query {
  foo(return: ID): String
  bar(return: ID!): String
}
```

and these graphql tagged templates
```js
const query = graphql`
    query entryQuery (
        $myReturn: ID!
    ) {
        ...A @relay(mask: false)
        ...B @relay(mask: false)
    }
`

const fragA = graphql`
    fragment A on Query {
        foo(return: $myReturn)
    }
`

const fragB = graphql`
    fragment B on Query {
        bar(return: $myReturn)
    }
`
```

`relay-compiler` will produce the below error, even though they are both compatible with the type defined on the query.

```
RelayMaskTransform: Cannot unmask fragment spread `A` because argument `myReturn` has been declared in `B` and they are not the same.
```

This does not occur if at least one of the fragments are used without `@relay(mask: true)`.

The error is thrown when [calling `isEquivalentType `](https://github.com/facebook/relay/blob/master/packages/relay-compiler/transforms/RelayMaskTransform.js#L152) which fails the [check that ensures that both types have the same nullability](https://github.com/facebook/relay/blob/master/packages/relay-compiler/graphql-compiler/core/isEquivalentType.js#L30-L33).

Can this check be removed? Is there a more accurate check that can be done in its place?

robrichard:
Fixes #2358 

See https://github.com/facebook/relay/issues/2358 for details and an example.

I think this is safe to remove this. If the variables are actually not compatible it will be caught by the GraphQL validation step.

### DIFF
--- a/packages/relay-compiler/transforms/RelayMaskTransform.js
+++ b/packages/relay-compiler/transforms/RelayMaskTransform.js
@@ -61,29 +61,30 @@ function visitFragment(fragment: Fragment, state: State): Fragment {
   if (state.hoistedArgDefs.size === 0) {
     return result;
   }
-  const existingArgDefs = new Map();
+  const combinedArgDefs = new Map();
   result.argumentDefinitions.forEach(argDef => {
-    existingArgDefs.set(argDef.name, argDef);
+    combinedArgDefs.set(argDef.name, argDef);
   });
-  const combinedArgDefs = result.argumentDefinitions.slice(); // Copy array
+
   state.hoistedArgDefs.forEach((hoistedArgDef, argName) => {
-    const existingArgDef = existingArgDefs.get(argName);
+    const existingArgDef = combinedArgDefs.get(argName);
+    let newArgDef = hoistedArgDef.argDef;
     if (existingArgDef) {
+      newArgDef = findCompatibleArgDef(existingArgDef, hoistedArgDef.argDef);
       invariant(
-        areSameArgumentDefinitions(existingArgDef, hoistedArgDef.argDef),
+        !!newArgDef,
         'RelayMaskTransform: Cannot unmask fragment spread `%s` because ' +
           'argument `%s` has been declared in `%s` and they are not the same.',
         hoistedArgDef.source,
         argName,
         fragment.name,
       );
-      return;
     }
-    combinedArgDefs.push(hoistedArgDef.argDef);
+    combinedArgDefs.set(argName, newArgDef);
   });
   return {
     ...result,
-    argumentDefinitions: combinedArgDefs,
+    argumentDefinitions: [...combinedArgDefs.values()],
   };
 }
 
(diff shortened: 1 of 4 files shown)


jstejada:
Interesting, thanks @robrichard. cc @leebyron who originally added this check

robrichard:
@leebyron do you have a moment to take a look?

packages/relay-compiler/transforms/RelayMaskTransform.js
```diff
@@ -149,7 +148,6 @@ function areSameArgumentDefinitions(
   return (
     argDef1.kind === argDef2.kind &&
     argDef1.name === argDef2.name &&
-    isEquivalentType(argDef1.type, argDef2.type) &&
```
jstejada: hey @robrichard I think that instead of entirely removing this check, would it make sense to instead check that the input type is a suptype of the expected argument? i.e. in your example in #2358 that $myReturn is a subtype of `return:`.

Does that make sense?

robrichard:
@jstejada this is ready for another look. I think the easiest way is to use the subtype of either of the two variables if any of them are `NonNull`. I added a test case for this too.

jstejada:
@josephsavona @kassens this looks reasonable to me but want to double check with you before importing to phabricator

robrichard:
@josephsavona thanks for reviewing. I pushed an update that I believe will account for all the scenarios listed.

packages/relay-compiler/transforms/RelayMaskTransform.js
```diff
@@ -59,29 +61,30 @@ function visitFragment(fragment: Fragment, state: State): Fragment {
...
+
   state.hoistedArgDefs.forEach((hoistedArgDef, argName) => {
-    const existingArgDef = existingArgDefs.get(argName);
+    const existingArgDef = combinedArgDefs.get(argName);
+    let newArgDef = hoistedArgDef.argDef;
     if (existingArgDef) {
+      newArgDef = findCompatibleArgDef(existingArgDef, hoistedArgDef.argDef);
       invariant(
-        areSameArgumentDefinitions(existingArgDef, hoistedArgDef.argDef),
+        !!newArgDef,
```
josephsavona: nit: use a null check here (`newArgDef != null`)

packages/relay-compiler/transforms/RelayMaskTransform.js
```diff
@@ -141,17 +145,39 @@ function isUnmaskedSpread(spread: FragmentSpread): boolean {
...
+    isEquivalentType(currentType, newType) &&
     // Only LocalArgumentDefinition defines defaultValue
-    (argDef1: any).defaultValue === (argDef2: any).defaultValue
-  );
+    (currentArgDef: any).defaultValue === (newArgDef: any).defaultValue
+  ) {
+    return compatibleArgDef;
+  }
+
+  return null;
```
josephsavona: I think this could be simplified quite a bit, to something like:

```javascript
const {isTypeSubTypeOf} = require('graphql');
...
function findCompatibleArgDef(prevArgDef, nextArgDef, schema): ?ArgumentDefinition {
  if (prevArgDef.kind !== nextArgDef.kind || prevArgDef.name !== nextArgDef.name || (prevArgDef: any).defaultValue !== (nextArgDef: any).defaultValue) {
    return null;
  } else if (isTypeSubTypeOf(schema, nextArgDef.type, prevArgDef.type) {
    return prevArgDef;
  } else if (isTypeSubTypeOf(schema, prevArgDef.type, nextArgDef.type) {
    return nextArgDef;
  } else {
    return null;
  }
}

```

robrichard:
@josephsavona I was not aware of `isTypeSubTypeOf`. The code looks much better now. Ready for another review.

josephsavona:
@robrichard Looks good, i'm going to import and add a few more tests

PR #2361, opened 2018-03-06
[RelayCompilerBin] Don’t include server schema in client extensions.

### LINKED ISSUE #2346: [relay-compiler] After upgrade to 1.5.0, it throws ERROR: Type "Address" already exists in the schema. It cannot also be defined in this type definition.
In our schema.graphql, we define a type Address for common address 
```graphql
type Address {
  city: String
  latitude: Float
  longitude: Float
  state: String
  staticMap: Image
  street: String
  zip: String
}
```
It works when we are using 1.4.1 without issues. But today when we upgrade to 1.5.0 (along with relay-runtime/react-relay) we no longer able to compile the static queries.

react-compiler keeps throwing ERROR:
Type "Address" already exists in the schema. It cannot also be defined in this type definition.

After we downgrade to 1.4.1, the compiler works again.

alloy:
Fixes #2346

cc @edvinerikson 

Add the schema path to the exclusion paths for client schema extensions, which is needed when the server schema resides _inside_ the source root.

### DIFF
--- a/packages/relay-compiler/bin/RelayCompilerBin.js
+++ b/packages/relay-compiler/bin/RelayCompilerBin.js
@@ -132,6 +132,13 @@ Ensure that one such file exists in ${srcDir} or its parents.
   const useWatchman = options.watchman && (await WatchmanClient.isAvailable());
 
   const schema = getSchema(schemaPath);
+
+  const graphqlSearchOptions = {
+    extensions: ['graphql'],
+    include: options.include,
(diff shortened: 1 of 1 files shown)


alloy:
Could well be, although I liked that one didn’t need to pass more options. Was there a reason to not do that immediately at the time?

edvinerikson:
No real reason, I asked the same question in the initial PR but got no response on it.

jstejada:
Thanks for adding this! Going to test that it works internally

PR #2366, opened 2018-03-13
Update documentation on optimistic update

### LINKED ISSUE #2365: optimisticResponse has some undocumented requirements
I tested the optimisticResponse feature using Github's GraphQL API, and found some undocumented behaviors that might be confusing to users. The mutation code is below:
```javascript
import { graphql, commitMutation } from 'react-relay'

const mutation = graphql`
  mutation AddStarMutation($input: AddStarInput!) {
    addStar(input: $input) {
      starrable {
        ... on Repository {
          viewerHasStarred
        }
      }
    }
  }
`
function commit(environment, id) {
  return commitMutation(environment, {
    mutation,
    variables: {
      input: {
        starrableId: id,
      },
    },
    optimisticResponse: {
      addStar: {
        starrable: {
          id: id, // even though the mutation doesn't ask for this field, we still need to provide id for optimisitic update to work
          viewerHasStarred: true,
          __typename: 'can be any string', // ought to be 'Repository'
        },
      },
    },
  })
}
export default { commit }
```  
Here `starrable` is of type `Starrable`, which is an interface implemented by both `Repository` and `Gist` type. The start quide states that for optimistic update to  work:

>In the simplest case above, we just need to pass an optimisticResponse option, which should refer to an object having the same shape as the mutation response payload. 

But here, the requirements seems to be:

1. When the type of a field is an interface or an union, even if you use inline fragment to query the concrete type, you still need to provide a __typename for that field. Weirdly, the __typename need not to be accurate. Any string will do.

2. As you can see in the above example, the shape of optimisticResponse and the mutation need not to be same. What's important is that optimisticResponse must have an id, whether the mutation has id or not.

Are these bugs or technical limitations? If someone can clarify on this or point me to the related source code, I'd love to help with either code or documentation.

irrigator:
Closes #2365.

### DIFF
--- a/docs/Introduction-QuickStartGuide.md
+++ b/docs/Introduction-QuickStartGuide.md
@@ -566,7 +566,13 @@ function commit(
 export default {commit};
 ```
 
-In the simplest case above, we just need to pass an `optimisticResponse` option, which should refer to an object having the same shape as the mutation response payload. Please note that the actual query and response payload may not have the exact same shape as your code, because sometimes Relay will add extra fields for you. For example, Relay will add an `id` field if it exists on the type for cache purpose. Relay will also add an `__typename` field if the type is an union or an interface. You can inspect the network request or response to see the exact shape.
+In the simplest case above, we just need to pass an `optimisticResponse` option, which should refer to an object having the same shape as the mutation response payload. Please note that the actual query and response payload may not have the exact same shape as the selection in your code, because sometimes Relay will add extra fields for you. For example:
+
+* Relay will add an `id` field if it exists on the type for caching purpose.
+
+* Relay will also add an `__typename` field if the type is an union or an interface.
+
+You can inspect the network request or response to see the exact shape.
 
 When we pass this option, Relay will know to immediately update our local data with the optimistic response, and then update it with the actual server response or roll it back if an error occurs.
 
--- a/docs/Introduction-QuickStartGuide.md
+++ b/docs/Introduction-QuickStartGuide.md
@@ -568,7 +568,7 @@ export default {commit};
 
 In the simplest case above, we just need to pass an `optimisticResponse` option, which should refer to an object having the same shape as the mutation response payload. When we pass this option, Relay will know to immediately update our local data with the optimistic response, and then update it with the actual server response or roll it back if an error occurs.
 
-Please note that the actual query and response payload may not have the exact same shape as the selection in your code, because sometimes Relay will select extra fields for you. For example:
+Please note that the actual query and response payload may not have the exact same shape as the selection in your code, because sometimes Relay will select extra fields for you, and you need to add these fields to your optimistic response. For example:
 
 * Relay will select an `id` field if it exists on the type for caching purpose.
 
--- a/docs/Introduction-QuickStartGuide.md
+++ b/docs/Introduction-QuickStartGuide.md
@@ -566,7 +566,9 @@ function commit(
 export default {commit};
 ```
 
-In the simplest case above, we just need to pass an `optimisticResponse` option, which should refer to an object having the same shape as the mutation response payload. When we pass this option, Relay will know to immediately update our local data with the optimistic response, and then update it with the actual server response or roll it back if an error occurs.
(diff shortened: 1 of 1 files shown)


docs/Introduction-QuickStartGuide.md
```diff
@@ -566,7 +566,9 @@ function commit(
 export default {commit};
 ```
 
-In the simplest case above, we just need to pass an `optimisticResponse` option, which should refer to an object having the same shape as the mutation response payload. When we pass this option, Relay will know to immediately update our local data with the optimistic response, and then update it with the actual server response or roll it back if an error occurs.
+In the simplest case above, we just need to pass an `optimisticResponse` option, which should refer to an object having the same shape as the mutation response payload. Please note that the actual query and response payload may not have the exact same shape as your code, because sometimes Relay will add extra fields for you. For example, Relay will add an `id` field if it exists on the type for cache purpose. Relay will also add an `__typename` field if the type is an union or an interface. You can inspect the network request or response to see the exact shape.
```
alloy: Maybe replace “as your code” with “as the selection in your code” and “for cache purpose” with “for caching purposes”.

I also wonder if it’s possible to use markdown lists and if it would make sense to make this an unordered list?

irrigator: Done

docs/Introduction-QuickStartGuide.md
```diff
@@ -566,7 +566,15 @@ function commit(
...
-In the simplest case above, we just need to pass an `optimisticResponse` option, which should refer to an object having the same shape as the mutation response payload. When we pass this option, Relay will know to immediately update our local data with the optimistic response, and then update it with the actual server response or roll it back if an error occurs.
+In the simplest case above, we just need to pass an `optimisticResponse` option, which should refer to an object having the same shape as the mutation response payload. Please note that the actual query and response payload may not have the exact same shape as the selection in your code, because sometimes Relay will add extra fields for you. For example:
+
+* Relay will add an `id` field if it exists on the type for caching purpose.
+
+* Relay will also add an `__typename` field if the type is an union or an interface.
+
+You can inspect the network request or response to see the exact shape.
+
+When we pass this option, Relay will know to immediately update our local data with the optimistic response, and then update it with the actual server response or roll it back if an error occurs.
```
alloy: Thanks for the updates!

Re-reading it now, I think that it would make more sense to move this paragraph back to where it was before so that the explanation of what this option does is completed _first_ and make the notes, about extra fields that Relay may select, a new paragraph.

irrigator: That does make a better reading experience. Updated.

docs/Introduction-QuickStartGuide.md
```diff
@@ -568,6 +568,14 @@ export default {commit};
 
 In the simplest case above, we just need to pass an `optimisticResponse` option, which should refer to an object having the same shape as the mutation response payload. When we pass this option, Relay will know to immediately update our local data with the optimistic response, and then update it with the actual server response or roll it back if an error occurs.
 
+Please note that the actual query and response payload may not have the exact same shape as the selection in your code, because sometimes Relay will select extra fields for you. For example:
```
alloy: Do you think it would make sense to explicitly state that this means people need to add these fields to their optimistic response?

docs/Introduction-QuickStartGuide.md
```diff
@@ -568,6 +568,14 @@ export default {commit};
 
 In the simplest case above, we just need to pass an `optimisticResponse` option, which should refer to an object having the same shape as the mutation response payload. When we pass this option, Relay will know to immediately update our local data with the optimistic response, and then update it with the actual server response or roll it back if an error occurs.
 
+Please note that the actual query and response payload may not have the exact same shape as the selection in your code, because sometimes Relay will select extra fields for you, and you need to add these fields to your optimistic response. For example:
```
jstejada: could you specify "when" Relay selects extra fields for you? i.e. during the compilation step.

I wonder if it's clearer to say, "Relay will add extra fields to your query", instead of "selecting extra fields"

jstejada:
Thanks for adding this! Just left a minor comment

docs/Introduction-QuickStartGuide.md
irrigator: I was using "add", but seeing @alloy keeps mentioning "select" makes me think that's a more accurate word 😆.

Will do.

alloy: Yeah a field entry in a GraphQL document is called a ‘selection’ in the GraphQL nomenclature.

kassens:
Awesome, thanks for the addition!

PR #2367, opened 2018-03-13
Use commonjs imports in relay-runtime

robrichard:
Doing the same that was done in https://github.com/facebook/relay/commit/97f24f8ebe80ebfd849db8e88f8449bfc9f9f5eb for react-relay in relay-runtime.

This is part of the first step in https://github.com/facebook/relay/issues/1590#issuecomment-372484579.

I left the `@providesModule` in the files in `relay-runtime/util/*.js` and a few others that used directly in the other packages. I can address that in the next PR by exporting them in `RelayRuntime.js`. I didn't want this one to get bigger than it already is.

### DIFF
--- a/packages/relay-runtime/RelayRuntime.js
+++ b/packages/relay-runtime/RelayRuntime.js
@@ -11,30 +11,30 @@
 
 'use strict';
 
-const RelayConcreteNode = require('RelayConcreteNode');
-const RelayConnectionHandler = require('RelayConnectionHandler');
-const RelayConnectionInterface = require('RelayConnectionInterface');
-const RelayCore = require('RelayCore');
-const RelayDeclarativeMutationConfig = require('RelayDeclarativeMutationConfig');
-const RelayInMemoryRecordSource = require('RelayInMemoryRecordSource');
-const RelayMarkSweepStore = require('RelayMarkSweepStore');
-const RelayModernEnvironment = require('RelayModernEnvironment');
-const RelayModernGraphQLTag = require('RelayModernGraphQLTag');
-const RelayNetwork = require('RelayNetwork');
-const RelayObservable = require('RelayObservable');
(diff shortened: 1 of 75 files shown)


alloy:
Nice 👌 

I think it would make sense to add a lint rule, such as [this one](https://github.com/facebook/react-native/pull/13014) by @janicduplessis, to this repo. There’s probably a bunch of changes needed of all open PRs and it could be hard to remember for FB employees to not use Haste imports as it will just work for them.

kassens:
Sorry, this is blocked by deep reliance on the `@providesModule` names to import flow types and private modules all over the place. I hope I can make some progress on this at some point, but it's not the highest priority :(

When we have this, I think the open source will be cleaner, but it's a mess to change this.

robrichard:
@kassens is there any point in leaving the `@providesModule` but still updating the `requires` to relative imports?

kassens:
In one of our environments we don't support a mix of both. The `@providesModule` version is the one "native" to most FB code and commonjs somewhat of a hack. I'll see if I can make slow progress on this as it would make a bunch of things easier like proper flow type exports.

alloy:
Very naive thought, but would it be feasible to have a script for that one environment that changes commonjs imports to haste ones and moves all the files back into a flat structure?

robrichard:
@kassens would it be possible to update relay to use ES6 modules while still maintaining compatibility with the FB haste system?

If it's possible, I'd like to open PR to do that to get closer to shipping optimized flat bundles (https://github.com/facebook/relay/issues/1590#issuecomment-372484579).

kassens:
Hey, I think it doesn't make sense to have this as a PR as the hardest part is figuring out the internal cleanup which probably needs to happen somewhat incrementally to keep the changes sane. (A bunch of internal references need to be updates when a module is moved).

Issue #2351, opened 2018-02-27
How to handle the "%future added value" in flow types generated enums


HichamELBSI:
Hi everyone,

How to use a flow type generated enum with the for a known enum (see the use case below)

```javascript
// ./src/events/__generated__/my_generated.graphql.js
...
export type EVENT_TYPES = ('activity' | 'formation' | '%future added value');
export type Event = {
     type: EVENT_TYPES
}
...
```

```javascript
// ./src/components/icon.js
type IconProps = {
    type: 'menu' | 'arrow' | 'activity' | 'formation',
};

const Icon: React.ComponentType<IconProps> = ...

export default Icon;
```

```javascript
// ./src/events/index.js

import type {Event} from './__generated__/my_generated.graphql.js';

type EventPageProps = {
     myEvent: Event
}

const EventPage = ({myEvent}: EventProps) => (
     <div>
               /* Flow error : 
                         Cannot create Icon element because string literal 
                         %future added value [1] is incompatible with string enum [2] in property type.     
              */
               <Icon type={myEvent.type} /> 
     </div>
)
```

leebyron:
These are generated as a reminder that GraphQL services often expand in capabilities and may return new enum values. To be future-proof, clients should account for this possibility and do something reasonable to avoid a broken product. Flow types help us catch cases where that hasn't yet occurred.

AugustinLF:
Hmmm, but couldn't you say that about any relay generated type? That we rely on the schema to be correct? If tomorrow the `type` property of `Event` becomes nullable, our code will fail where it shouldn't, right?

leebyron:
That is also true! I think this area of GraphQL is the least well defined at the moment however historically we've seen changes of what type a field returns (including becoming nullable) being considered as server breaking changes and thus avoided, while we've seen additions to enums not being considered server breaking changes and occurring (we now have a tool in graphql.js that detects this as a "dangerous" change). Another related example is a Union type adding a new possible type which can be returned - this hasn't historically been considered breaking but is dangerous for the same reason.

To bring the example closer to your code - if in the future you introduce a new kind of event and ship it, then your client may crash because it got `undefined` where it expected an icon. Relay's generated flow types are trying to help you out by pointing out this possible future scenario.

Right now this is always how this works, but I think if you really thought this wasn't useful then we would definitely entertain a PR which allows for configuration to exclude this "future added value" entry

AugustinLF:
[...]

PR #2368, opened 2018-03-14
Makes flow generated '%future added value' for enums optional

AugustinLF:
As suggested by @leebyron in #2351, this adds an option to the `relay-compiler`  so it doesn't add `%future added value` to each GraphQL generated enum. My reasoning is that you might have a strict control over your GraphQL service, so you'd have to handle any future value by rerunning the compiler, with the new version of the schema.

I'm not totally convinced by the name `strictFlowTypes`, but I don't really have much better now. My thought between this name is that the option would be a generic option that configures the generated flow types, and would not be specifically to enums.

Regarding test cases, I'm not sure of how I should tackle that, since the only tests are snapshots, and there are no tests with different configurations of the `RelayFlowGenerator`. I could however try to make that happen if you'd like!

### DIFF
--- a/packages/relay-compiler/bin/RelayCompilerBin.js
+++ b/packages/relay-compiler/bin/RelayCompilerBin.js
@@ -94,6 +94,7 @@ async function run(options: {
   watch?: ?boolean,
   validate: boolean,
   quiet: boolean,
+  strictFlowTypes: boolean,
 }) {
   const schemaPath = path.resolve(process.cwd(), options.schema);
   if (!fs.existsSync(schemaPath)) {
@@ -163,7 +164,7 @@ Ensure that one such file exists in ${srcDir} or its parents.
   };
   const writerConfigs = {
     js: {
-      getWriter: getRelayFileWriter(srcDir),
+      getWriter: getRelayFileWriter(srcDir, options.strictFlowTypes),
       isGeneratedFile: (filePath: string) =>
         filePath.endsWith('.js') && filePath.includes('__generated__'),
       parser: 'js',
@@ -194,7 +195,7 @@ Ensure that one such file exists in ${srcDir} or its parents.
   }
 }
 
-function getRelayFileWriter(baseDir: string) {
+function getRelayFileWriter(baseDir: string, strictFlowTypes: boolean) {
   return ({
     onlyValidate,
     schema,
@@ -218,6 +219,7 @@ function getRelayFileWriter(baseDir: string) {
         inputFieldWhiteListForFlow: [],
         schemaExtensions,
         useHaste: false,
+        strictFlowTypes,
       },
       onlyValidate,
       schema,
@@ -332,6 +334,12 @@ const argv = yargs
         'Looks for pending changes and exits with non-zero code instead of ' +
         'writing to disk',
       type: 'boolean',
+      default: false,
+    },
+    strictFlowTypes: {
+      describe:
+        'Generate flow types that more closely match the schema. This should be avoided ' +
+        'if the GraphQL server can change enum values without updating its schema',
       default: false,
     },
   })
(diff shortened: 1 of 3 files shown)


alloy:
With an eye on [this PR](https://github.com/facebook/relay/pull/2293) having traction at the moment, it would be great to not introduce new language containing ‘Flow’ as this option would apply to all language plugins. What about just `strictTypes`?

Btw, there’s also a non-snapshot test for the Flow generator in that PR: https://github.com/facebook/relay/pull/2293/files#diff-0863bb70c9b489414458b93fb3d2be18R51

leebyron:
I would suggest a more specific naming to the intended purpose. Notably this is *less strict* type safety since it removes future-proof protection from the type checker, so I’d like to avoid any naming containing the word “strict” to avoid confusion.

leebyron:
How about `—noFutureProofEnums`?

packages/relay-compiler/bin/RelayCompilerBin.js
```diff
@@ -334,6 +336,12 @@ const argv = yargs
       type: 'boolean',
       default: false,
     },
+    strictFlowTypes: {
+      describe:
+        'Generate flow types that more closely match the schema. This should be avoided ' +
+        'if the GraphQL server can change enum values without updating its schema',
```
leebyron: This is a bit confusing. Additional enum values is definitionally a schema update. Perhaps rephrase to talk about the case where enums add new values in the future

AugustinLF: What do you think of "Generate flow types that more closely match the schema. This means that you will have to update your application code whenever the GraphQL server adds new enum values. Not doing so could break your front-end."? Moreover, I'd like to explain a bit more the idea behind this, perhaps give an example, but I don't think that the `yargs` parameters description is the best case for that. I didn't see proper docs for the relay compiler, would a comment in this file would be ok for you?

alloy: It would also be great to convey that this relates to the `%future added value` entry, as people may otherwise not immediately understand the relation and this being a great time to explain a bit more on that values purpose.

How about something like this?

> This option controls whether or not a catch-all entry is added to enum type definitions for values that may be added in the future.

alloy: The last sentence is a bit confusing, this also applies to people doing server-side rendering.

AugustinLF: Your client? Your relay application? I'd like to emphasise that what I'm not speaking of the server, but the "client".

alloy: Maybe something like this?

> This option controls whether or not a catch-all entry is added to enum type definitions for values that may be added in the future. Enabling this means you will have to update your application whenever the GraphQL server schema adds new enum values to prevent it from breaking.

AugustinLF:
I think I addressed all the requested changes! I factored the `RelayFlowGenerator` tests following what was done in @alloy's PR.

leebyron:
Thanks for making these updates!

nirrek:
FWIW our team would find this flag very useful.

We often leverage flow to get exhaustive switch statements when working with enums. For example:

```js
type Mode = 'a' | 'b';
switch (mode: Mode) {
  case 'a': processA(); break;
  case 'b': processB(); break;
  default: (mode: empty);
}
```

Typically when we are doing these exhaustiveness checks it's because we know that there is no default fallback that makes sense in that context. We need to explicitly handle all the cases. Having flow catch when our client code is no longer in sync with our schema is very helpful for us coordinating changes.

Without this PR it's possible we would start writing highly questionable stuff like the following to get those same guarantees (either that or maintain a fork of relay-compiler).

```js
switch (mode: Mode) {
  case 'a': processA(); break;
  case 'b': processB(); break;
  case '%future added value': break;
  default: (mode: empty);
}
```

Thanks for getting the ball rolling on this one @AugustinLF 👍

PR #2373, opened 2018-03-16
Fix CI for node < 8

kassens:
Thanks @alloy for noticing that https://github.com/facebook/relay/commit/0d24a725c7ad5e2414e755c5ab47850fb85b824c broke CI.  The reason is that node < 8 does not support trailing commas for function calls which prettier adds in our configuration. This should fix it.


alloy:
Thanks! 👍

PR #2374, opened 2018-03-17
Track fragments on connections that yield no edges

wbyoung:
Like connections that return `null`, tracking the fragment is necessary in order to properly generate future queries that affect the connection since traversal of the edges is not generating details on split queries and tracking those individually.

Here's [a bit of code that demonstrates what problem this is solving](https://gist.github.com/wbyoung/f4ef16828ea1f4b0071bcf3eb27335e5) that runs against a simple backend w/ the idea of a `List` model that has an `owner` among other fields. Hopefully the idea of the schema that it's running against and the code itself is pretty self explanatory. This is intended to demonstrate the issue w/ a minimal amount of code — in reality this situation came up when changing between different components/queries being rendered as the user navigates between pages in an app.


yuzhi:
Thank you for your pull request. However, it's been a while since Relay Classic has be deprecated. We are not planning on making any further changes for it. Our experiences showed that diff queries were inefficient and can delay requests by several hundred ms and make the query signficantly longer. Please consider upgrading to Relay Modern.

PR #2382, opened 2018-03-22
Fixed comment

dminkovsky:


kassens:
Thanks, I think we used to dedupe on cursers in classic.

PR #2386, opened 2018-03-28
(RelayModernEnvironment) Verify can normalizePayload before `do`

### LINKED ISSUE #2385: [Modern] QueryRenderer fetch operation throws uncaught error
It looks like in `1.5.0` (https://github.com/facebook/relay/commit/f76f12a56b2ed7486b8d4a0ce5d8efdf206c2b9a), `relay-runtime` introduced a bug that manifests as an uncaught exception when handling a Query response that contains GraphQL errors.

```
RelayObservable.js:644 Uncaught Error: No data returned for operation `MyQuery`, got error(s):
HAHA!
```

The root cause looks to be that `normalizePayload` throws errors, and moving that call from the observable `map` call (which handles errors), to a sink (which doesn't handle errors) prevents relay from catching and handling those errors when thrown.

ajhyndman:
Fixes #2385

I suspect there's a much nicer solution to this, but I thought it was worth starting somewhere.

I'm also not sure how to add test assertions to highlight the behaviour, but this does demonstrate a working fix.

### DIFF
--- a/packages/relay-runtime/store/RelayModernEnvironment.js
+++ b/packages/relay-runtime/store/RelayModernEnvironment.js
@@ -193,6 +193,10 @@ class RelayModernEnvironment implements Environment {
     let optimisticResponse;
     return this._network
       .execute(operation.node, operation.variables, cacheConfig || {})
+      .map(executePayload => {
+        normalizePayload(executePayload);
+        return executePayload;
+      })
       .do({
         next: executePayload => {
           const responsePayload = normalizePayload(executePayload);


packages/relay-runtime/store/RelayModernEnvironment.js
```diff
@@ -192,6 +192,10 @@ class RelayModernEnvironment implements Environment {
     let optimisticResponse;
     return this._network
       .execute(operation.node, operation.variables, cacheConfig || {})
+      .map(executePayload => {
+        normalizePayload(executePayload);
```
sibelius: won't this cause 2 normalizePayload on `executePayload`?

ajhyndman: Yep, that's true.

The first one is being used as validation and is not actually returned, so in the case that we need the value normalized, we have to run it again.  There would certainly be more drastic refactors possible, if anyone wants to suggest something.

I was just aiming for a minimal, low-risk change that definitely fixed the error.

josephsavona:
Thanks for the PR. We're open to changes but are cautious about changes to core that can affect correctness, performance, etc - having a duplicate normalization here isn't tenable from a perf perspective. Thankfully, in the meantime we've significantly refactored the query execution & normalization flow since this was posted, and are catching errors thrown by normalization.

PR #2397, opened 2018-04-06
Allow connection records with no filter calls to invoke getRangeBehavior

bttf:
If a connection record has no filter calls, we are never able to invoke `getRangeBehavior` when handling `RANGE_ADD` from a mutation

I have a use case where I've set `rangeBehavior` to a function that returns `'prepend'` to all calls, regardless of arguments. This in conjunction with the fact that we fetch connections just for connection fields (`total`) in some places, without fetching edges.

It would be nice if all connection records could call getRangeBehavior regardless

screen shot 2018-04-05 at 6 02 11 pm


sibelius:
can you migrate to relay modern?

relay classic is not under development anymore

PR #2401, opened 2018-04-06
[docs] remove clientMutationId and Relay Mutation spec

kassens:
RelayModern has no requirements on mutations beyond the general mutation specs. Specifically, the additional requirements that are no longer required are:

- `clientMutationId` (even RelayClassic only required this because of FB internal reasons)
- `$input` objects, with RelayModern the input to mutations can come either as one big object or multiple mutation arguments.


robrichard:
@kassens there is one bug that some people have been using `clientMutationId` to work around. I have the details in my PR: https://github.com/facebook/relay/pull/2349

kassens:
Now that #2349 finally landed, I think we're ready for this too!

julioxavierr:
@kassens 

I saw this in breaking changes for v8:
```
Mutations are committed to unique root ids #2349. In mutation updaters, fields in the mutation will no longer be available in the store root returned from store.getRoot. Usages like store.getRoot().getLinkedRecord('create_comment_mutation') will break.
```

Will usages like `store.getLinkedRecord('create_comment_mutation')` also break?

PR #2416, opened 2018-04-18
support scalar handles in data loader

mattkrick:
Currently, the ~~data loader~~ `DataChecker` does not support scalar handles. 

## Reproduction
1. Set `dataFrom: 'STORE_THEN_NETWORK'` in the queryFetcher
2. Run a query like the following:

```js
query RideBikes {
  foo {
    bar @__clientField(handle: "handleBars")
  }
}
```


mattkrick:
rebased from master to work on DataChecker

mattkrick:
closed in favor of #2791

PR #2417, opened 2018-04-19
 GraphQLIRVisitor: visit ObjectValue and ObjectFieldValue

pieh:
This fixes variables used in InputObject not being detected in `StripUnusedVariablesTransform.js` and therefore being stripped from resulting query.

Before this:
```gql
query TestQuery3(
  $usedInInputObject: String!
) {
  items(filter: { date: $usedInInputObject }) {
    date
  }
}
```

Would be compiled to (missing query argument declaration)
```gql
query TestQuery3 {
  items(filter: {date: $usedInInputObject}) {
    date
  }
}
````

I'm not familiar with codegen part, so please advise on how to deal with how my change affected it (see RelayCodeGenerator-test.js.snap change).

### DIFF
--- a/packages/graphql-compiler/core/GraphQLIRVisitor.js
+++ b/packages/graphql-compiler/core/GraphQLIRVisitor.js
@@ -41,6 +41,8 @@ const NodeKeys = {
   LinkedField: ['args', 'directives', 'selections'],
   Literal: [],
   LocalArgumentDefinition: [],
+  ObjectValue: [`fields`],
+  ObjectFieldValue: [`value`],
   Request: ['root'],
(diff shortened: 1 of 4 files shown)


pieh:
Hm failed Travis CI build *seems* unrelated to my changes https://travis-ci.org/facebook/relay/jobs/368562847 (failed before any tests even started)

josephsavona:
Relay doesn't support input objects that contain variables - the example would error in codegen if not earlier. That said, we should update IRVisitor if we haven't already.

josephsavona:
We've removed support for *parsing* variables except as top-level input values - they can no  longer  be nested within input objects/lists. This was previously validated in codegen, so it doesn't change the effective behavior.

PR #2424, opened 2018-04-24
Fix CI for node 6

kassens:
This file is executed without transforms so we can't have trailing commas in function calls until we drop node 6.


orta:
😀 it's always this file - hah

kassens:
@orta yes, I hope the `@noformat` will make it less likely to break again. Luckily this is always really easy to fix too.

PR #2435, opened 2018-05-04
+ Strategies section

English3000:


docs/Modern-Debugging.md
```diff
@@ -3,7 +3,24 @@ id: relay-debugging
...
+**A few questions to ask yourself:**
+* _Is my [compilation](https://facebook.github.io/relay/docs/en/installation-and-setup.html#set-up-relay-compiler) up-to-date?_
+* _Is my query valid?_ You can test this on your GraphiQL endpoint.
+
+**If so:**
+1. Put your entire query at the top-level (your `QueryRenderer`) and comment out all fragment containers. Pass the data through your component hierarchy down to the components that were using fragments.
+2. If your app isn't rendering properly, use `console.log(props);` for every component to see the props you are actually passing.
+3. Once your app is rendering, uncomment the deepest fragment container and add that fragment back to your top-level query. Everything should still render. If not, use `console.log(props);` again.
+4. Continue uncommenting fragments and confirming that your app renders until your top-level query is as desired.
+5. If this still fails, come back to your code later and try debugging with a fresh mind. Additionally, you can [post an issue](https://github.com/facebook/relay/issues/new) and someone should get back to you within 1-2 days.
```
jstejada: thanks so much for adding this!! my only nit here is that I'm not sure that it will always be the case that someone will reply within 1 or 2 days

jstejada: otherwise, looks great!

English3000:
Hey Juan, I modified the language.

English3000:
@jstejada Does everything look good?

orta:
IMO this is good, we should get it in 👍

PR #2446, opened 2018-05-13
Remove unnecessary arguments to relay compiler

VinayakBagaria:
Extra `--` arguments on relay-compiler. Even the command line says to execute `yarn run relay --watch`. Note no error comes while using the extra characters.


kassens:
Good catch!

VinayakBagaria:
Thanks!

PR #2449, opened 2018-05-14
chore: clone last 5 commits

DanielRuf:


.circleci/config.yml
```diff
@@ -39,3 +39,6 @@ workflows:
           filters:
             branches:
               only: master
+
+git:
+  depth: 5
```
kassens: Could you update the PR message to include why this should be added? It's not clear to me what this does or what it solves/improves.

DanielRuf: By default Travis CI clones the last 50 commits, 5 are sufficient and faster.
Do we really need a comment here?

See https://docs.travis-ci.com/user/customizing-the-build#Git-Clone-Depth

kassens: Thanks for adding the link here!

Issue #2463, opened 2018-06-08
Compiler: __typename alias doesn't work in the "other" branch of the interface type


mrtnzlml:
Hi! When I try to use `__typename` on the interface type:

```graphql
fragment TripOverview on Booking {
  allTypes {
    __typename
    ... on BookingOneWay {
      ...OneWayTimeline
    }

    # ...
  }
}
```

Compiler generates this type:

```
export type TripOverview = {|
  +allTypes: ?({|
    +__typename: "BookingOneWay",
    +$fragmentRefs: OneWayTimeline$ref,
  |} | {|
    +__typename: "BookingReturn",
    +$fragmentRefs: ReturnTimeline$ref,
  |} | {|
    +__typename: "BookingMulticity",
    +$fragmentRefs: MulticityTimeline$ref,
  |} | {|
    // This will never be '%other', but we need some
    // value in case none of the concrete values match.
    +__typename: "%other"
  |}),
  +$refType: TripOverview$ref,
|};
```

However, when I try to alias the `__typename`:

```graphql
fragment TripOverview on Booking {
  allTypes {
    type: __typename
    ... on BookingOneWay {
      ...OneWayTimeline
    }

    # ...
  }
}
```

It generates this:

```
export type TripOverview = {|
  +allTypes: ?({|
    +type: "BookingOneWay",
    +$fragmentRefs: OneWayTimeline$ref,
  |} | {|
    +type: "BookingReturn",
    +$fragmentRefs: ReturnTimeline$ref,
  |} | {|
    +type: "BookingMulticity",
    +$fragmentRefs: MulticityTimeline$ref,
  |} | {|
    // This will never be '%other', but we need some
    // value in case none of the concrete values match.
    +__typename: "%other"
  |}),
  +$refType: TripOverview$ref,
|};
```

But I would expect this instead:

```diff
  |} | {|
    // This will never be '%other', but we need some
    // value in case none of the concrete values match.
-   +__typename: "%other"
+   +type: "%other"
  |}),
```

It challenging to use Flow in this case because when checking the `type` it complains about missing "type" field in the last branch. The only way how to use it sanely is to keep the `__typename` as is. This way it's the same in every branch and it's simple to use it...

PR #2465, opened 2018-06-09
Use aliases for __typename generated fields in the union

mrtnzlml:
Closes: https://github.com/facebook/relay/issues/2463


mrtnzlml:
@kassens Hi, is there anything I can improve or do differently? :)

kassens:
Haven't really considered aliasing __typename, but makes sense. Thank you 👍

PR #2492, opened 2018-07-19
feat(entria): add entria as user

jgcmarins:


oskojess:
Nice.

kassens:
Sorry for the delay here!

PR #2496, opened 2018-07-30
add edge order specifications to Connections documentation

### LINKED ISSUE #2466: Add clarifications/missing specifications about order of edges to the relay cursor connections specification
Recently while working on improving `graphql-sequelize`'s implementation of connections, I discovered that, as far as I understand, someone misinterpreted the specification: their implementation returns `edges` in the opposite order when the `last` argument is used.  This also means the `startCursor` and `endCursor` are swapped, and their reverse pagination test fetches the previous page by requesting `last: 3, before: pageInfo.endCursor`.  But as far as I understand `before` should always take the `startCursor` of the current `pageInfo`.

I've submitted a PR to fix their implementation but I'd also like to help prevent this kind of misinterpretation for anyone in the future.

Adding the following clarifications to the Relay Cursor Connections Specification would precisely specify the ordering of `edges`:
* The ordering of `edges` should be the same when using `first/after` as when using `last/before`, all other arguments being equal
* To fetch the next page one should use `after: pageInfo.endCursor`, not `after: pageInfo.startCursor`
* To fetch the previous page one should use `before: pageInfo.startCursor`, not `before: pageInfo.endCursor`
* When `before: cursor` is used, the edge closest to `cursor` should come **last** in the result `edges`.
* When `after: cursor` is used, the edge closest to `cursor` should come **first** in the result `edges`.

**EDIT**: merrywhether pointed out that `startCursor` and `endCursor` are not standard right now, so I've updated my PR to refer to the edge cursors instead.

jedwards1211:
Closes #2466

### DIFF
--- a/website/spec/Connections.md
+++ b/website/spec/Connections.md
@@ -251,6 +251,8 @@ The server should use those two arguments to modify the edges returned by
 the connection, returning edges after the `after` cursor, and returning at
 most `first` edges.
 
+You should generally pass the previous page's `pageInfo.endCursor` for `after`.
+
 ## Backward pagination arguments
 
 To enable backward pagination, two arguments are required.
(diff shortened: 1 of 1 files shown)


sibelius:
maybe you should link to relay connection handler as well https://github.com/facebook/relay/blob/master/packages/relay-runtime/handlers/connection/RelayConnectionHandler.js#L42

jedwards1211:
@sibelius It doesn't seem quite right to me for a specification to refer to an *implementation* of the specification...a specification should be able to stand on its own, don't you think?

jedwards1211:
@kassens did you author this spec or just copy it from elsewhere?
I want to get some FB eyes on this to comment or approve it.

ngbrown:
Why are these changes not live on the website? (https://facebook.github.io/relay/graphql/connections.htm) 

It seems only the `graphql/` directory of the website isn't getting updated during publishes, because the most recent build on the `gh-pages` branch is from a version that includes these changes (https://github.com/facebook/relay/commit/be9e7dfc9b45125ba0afd0e27c06639f9ad7936e).

jstejada:
@ngbrown should be live now

jedwards1211:
Thank you guys!

ngbrown:
@jstejada https://facebook.github.io/relay/graphql/connections.htm still doesn't have any mention of `startCursor` or `endCursor`.  The history for the `gh-pages` branch shows no recent activity in the /graphql directory. https://github.com/facebook/relay/tree/gh-pages/graphql

PR #2500, opened 2018-08-02
[WIP] Migration to the New Context API

cyan33:
Newly added module:

* `classic/tools/ReactRelayContext.js` (maybe should be put under `modern`?)
* `classic/tools/useContext.js`

The migration follows the pattern below:

* `getChildContext` => `<ReactRelayContext.Provider value={{...}}>`
* `this.context` => `const context = useContext(ReactRelayContext)`
* Removed all the `contextTypes` and `childContextTypes`

Blockers:

* Flow typing the context shape (previously we used `prop-types`)
* A few broken tests
* Wait for React to release the version that includes the new Context API 
* Should we expose the RelayContext as a public API (See below comment of the reason why we need it)? Will Suspense mitigate this issue when it's required in SSR?
* Improper usage of relay's context in the production code inside Facebook.


robrichard:
@cyan33 do you think that `ReactRelayContext` could be exported publicly in ReactRelayPublic.js? Then we would not need to maintain this external package: https://github.com/relay-tools/relay-context-provider

cc @kassens @jstejada

cyan33:
I don't know exactly why users would need this in their product code. But I discussed @kassens and think the context should not be exposed to users. (I might be wrong though).

Also the bad part about the legacy context API is that you could get access to the content of the context in your component through `this.context.relay` once you add a `contextTypes` and neither React or Relay cannot forbid this.

robrichard:
We needed to use it for server side rendering. We execute a fetch manually with the exported `fetchQuery` function and then render the child fragment containers using `ReactDOM.renderToString`. The context needs to be set for the fragment containers to work. The only current official way to do that currently is with a `<QueryRenderer>`, which if used on the server would require multiple render passes while waiting for the fetched data.

It sounds like React Suspense will address this case, but until then I think this is the only viable solution.

cyan33:
Added it to the blockers in the description. Thanks @robrichard.

sibelius:
can we expose a `connect` function? much like redux or formik one https://github.com/jaredpalmer/formik#connect?

TrySound:
@sibelius I think there are a few "connect" functions.
- createFragmentContainer
- createPaginationContainer
- createRefetchContainer

or what do you mean?

sibelius:
this would be helpful for SSR, like they use this https://github.com/relay-tools/relay-context-provider right now

robrichard:
@sibelius I think a `connect` function is different than what `relay-context-provider` does. `relay-context-provider` is the `Provider` half of the new context API (which would be useful to expose publicly for SSR I think). 

A `connect` function is HOC that would use the `Consumer` half of the new context API. I think you can get the same functionality now by doing `createFragmentContainer(MyComponent, {})`.

cyan33:
@jstejada Hey, I'm not sure if the broken tests are addressed by the recent fix (if there has been any). But feel free to take over and do what is necessary. Tell me if you need any help from me.

jstejada:
Landed by https://github.com/facebook/relay/commit/03b452443f34d89f88ca1412108c0d1d9946eb6c

PR #2501, opened 2018-08-04
Allow literals in @arguments

robrichard:
Sometimes it would be nice to use a fragments that has `@argumentDefinitions` defined with a known static value. This could be done by passing a literal to `@arguments` as in this example:

```graphql
query TestQuery($id: ID!) {
  node(id: $id) {
    id
    ...Profile @arguments(pictureSize: [128])
  }
}

fragment Profile on User @argumentDefinitions(
  pictureSize: {type: "[Int]", nonNull: true}
) {
  id
  name
  profilePicture(size: $pictureSize) {
    uri
  }
}
```

However, this will cause `relay-compiler` to throw an error `All @arguments() args must be variables.` One solution is to define a new global variable on the query, but this has the downside of polluting the query with more globals. Alternatively, the following workaround could also be used:

```graphql
query TestQuery($id: ID!) {
  node(id: $id) {
    id
    ...ProfileWrapper
  }
}

fragment ProfileWrapper on User @argumentDefinitions(
  pictureSize: {type: "[Int]", defaultValue: [128]}
) {
  ...Profile @arguments(pictureSize: $pictureSize)
}


fragment Profile on User @argumentDefinitions(
  pictureSize: {type: "[Int]"}
) {
  id
  name
  profilePicture(size: $pictureSize) {
    uri
  }
}
```

It would reduce a lot of boilerplate to allow literals directly in `@arguments`. This PR removed the invariant that prevents literals and adds new test cases to verify behavior.

### DIFF
--- a/packages/graphql-compiler/core/GraphQLParser.js
+++ b/packages/graphql-compiler/core/GraphQLParser.js
@@ -546,19 +546,15 @@ class GraphQLParser {
     if (argumentDirectives.length) {
       args = (argumentDirectives[0].arguments || []).map(arg => {
         const argValue = arg.value;
-        invariant(
-          argValue.kind === 'Variable',
-          'GraphQLParser: All @arguments() args must be variables, got %s. ' +
-            'Source: %s.',
-          argValue.kind,
-          this._getErrorContext(),
-        );
 
         return {
           kind: 'Argument',
           metadata: null,
           name: getName(arg),
-          value: this._transformVariable(argValue),
+          value:
+            argValue.kind === 'Variable'
+              ? this._transformVariable(argValue)
+              : this._transformValue(arg.value, null),
           type: null, // TODO: can't get type until referenced fragment is defined
         };
       });
--- a/packages/relay-compiler/transforms/__tests__/__snapshots__/RelayApplyFragmentArgumentTransform-test.js.snap
+++ b/packages/relay-compiler/transforms/__tests__/__snapshots__/RelayApplyFragmentArgumentTransform-test.js.snap
@@ -91,6 +91,69 @@ fragment Profile_4CNNX6 on User {
 
 `;
 
+exports[`RelayApplyFragmentArgumentTransform matches expected output: literal-argument.graphql 1`] = `
+~~~~~~~~~~ INPUT ~~~~~~~~~~
+query TestQuery(
+  $id: ID!
+) {
+  node(id: $id) {
+    id
+    ...Profile @arguments(pictureSize: [128])
+    friends(first: 10) {
+      edges {
+        node {
+          ...Profile @arguments(pictureSize: [256])
+        }
+      }
+    }
+  }
+}
+
+fragment Profile on User @argumentDefinitions(
+  pictureSize: {type: "[Int]", nonNull: true}
+) {
+  id
+  name
+  profilePicture(size: $pictureSize) {
+    uri
+  }
+}
+
+~~~~~~~~~~ OUTPUT ~~~~~~~~~~
+query TestQuery(
+  $id: ID!
+) {
+  node(id: $id) {
+    id
+    ...Profile_17y1Ti
+    friends(first: 10) {
+      edges {
+        node {
+          ...Profile_1tVpuC
+        }
+      }
+    }
+  }
+}
+
+fragment Profile_17y1Ti on User {
+  id
+  name
+  profilePicture(size: [128]) {
+    uri
+  }
+}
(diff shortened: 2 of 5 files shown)


packages/relay-compiler/transforms/__tests__/__snapshots__/RelayApplyFragmentArgumentTransform-test.js.snap
```diff
@@ -91,6 +91,69 @@ fragment Profile_4CNNX6 on User {
 
 `;
 
+exports[`RelayApplyFragmentArgumentTransform matches expected output: literal-argument.graphql 1`] = `
+~~~~~~~~~~ INPUT ~~~~~~~~~~
+query TestQuery(
+  $id: ID!
+) {
+  node(id: $id) {
```
soneymathew: Should we add a test case for `pictureSize` coming in as a global query variable as well?
```
query TestQuery(
  $id: ID!, pictureSize: Int!
) 
```

packages/graphql-compiler/core/GraphQLParser.js
```diff
@@ -546,19 +546,15 @@ class GraphQLParser {
...
 
         return {
           kind: 'Argument',
           metadata: null,
           name: getName(arg),
-          value: this._transformVariable(argValue),
+          value:
+            argValue.kind === 'Variable'
+              ? this._transformVariable(argValue)
+              : this._transformValue(arg.value, null),
```
soneymathew: Sorry if this is a noob question.
Will this change work well with `FlattenTransform` and `SkipRedundantNodeTransform` in case the literal values are resolved to be the the same?

robrichard: I added another test where the values are the same and they do get merged.

packages/relay-compiler/transforms/__tests__/__snapshots__/RelayApplyFragmentArgumentTransform-test.js.snap
robrichard: The new test also has a global variable

alloy:
I agree that scalar-only support would be useful to us as well. Thus far I’ve only really felt the need to use literals when toggling a boolean switch to conditionally include a fragment.

robrichard:
I'm going to close this and leave it up to someone else to make an attempt at a scalar-only version. The workaround of using `@argumentDefinitions` is not too cumbersome so I will stick with that approach.

taion:
@jstejada

Could you clarify your comment, actually – what's the significance of getting the type in question here? It looks like `_transformValue` only uses the type to do some validation that e.g. types are appropriately input types, &c., and in general there's not so much validation.

More broadly, as-is, I don't get compile-time errors if I pass a variable of the wrong type into `@arguments`, so I'd argue that there would be no inconsistency around the use of scalars without additional type validation or checking.

josephsavona:
We'd rather avoid adding a hack just for specific literals. The right way to handle this is to delay parsing (untyped) GraphQL ASTs into (typed) IR until all the asts are available. That would allow us to look up (and on-demand type-check) fragments during IR creation. Whenever you see a fragment spread you load and check the fragment, then type check the arguments. We'd like to to this, the main blocker is that we're working on a more incremental version of the compiler (mainly the outer layer than handles incrementally recompiling), after which we'll have less code to refactor.

taion:
I'm not sure I see the hack here, though. In effect, this change is equivalent to removing the check, and replacing `this._transformVariable(argValue)` with `this._transformValue(argValue)`. It seems to "just work" naturally.

To take this PR and simplify the code a bit, the change just looks like: https://github.com/facebook/relay/compare/master...taion:arguments?expand=1#diff-261c85de44b20666fe4c13b299a67b4c.

josephsavona:
The problem is calling `transformValue()` without a type can yield effectively invalid IR, so we'd have to scope it by doing something hacky like `transformValue(argValue, GraphQLScalarType)` in order to make sure that objects and lists would throw. But then the error message wouldn't be super clear...etc.

josephsavona:
Implemented in e8433b9582bc09d78753e379a4723ca5965e6f66. Arbitrary literals - including lists, input objects, enums, etc - are supported. The only catch is that the fragment being spread must explicitly define the argument with `@argumentDefinitions`. This is necessary so that Relay can determine the type to use when parsing the literal without a deep check to see how the argument may be used. This will go out in the next release.

cc @robrichard @taion

Issue #2157, opened 2017-10-18
[Modern] RANGE_ADD Mutation with configs


valorize:
Hi, The docu here: https://facebook.github.io/relay/docs/mutations.html#range-add
suggests to use configs to configure the store updating mechanism. 

Unfortunately it is unclear to me what the parentId: 'shipId' references to and where the connectionKey is coming from.

Let's suppose I have the following query: 
```
graphql`query AppQuery($rows: Int!) {
        allChatMessages(first: $rows) {
            totalCount
            edges {
                node {
                    message
                    username
                }
             }
        }
    }`
```

I suppose my configs would look like:
```
const configs = [{
    type: 'RANGE_ADD',
    parentID: 'id',
    connectionInfo: [{
      key: 'allChatMessages',
      rangeBehavior: 'append',
    }],
    edgeName: 'chatMessageEdge',
  }];
```

with a mutation like
```
    mutation createChatMessageMutation($input: CreateChatMessageInput!) {
        createChatMessage(input: $input) {
            chatMessage {
                username
                message
                updated
            }
            chatMessageEdge {
                node {
                    id
                }
            }
        }
    }
```
is this correct?

zth:
I think you need to annotate your connection and refer to that in the config. Like this:

```
graphql`query AppQuery($rows: Int!) {
        allChatMessages(first: $rows) @connection(key: "AppQuery_allChatMessages") {
            totalCount
            edges {
                node {
                    message
                    username
                }
             }
        }
    }`
```

...and then refer to `AppQuery_allChatMessages` as the connection key in your config.

However, I have no clue about the `parentID` when it has no parent. Looking at the source it looks like it is required: https://github.com/facebook/relay/blob/6d87a701de69ce1e271895ff689ddbda2b39b4c2/packages/relay-runtime/mutations/setRelayModernMutationConfigs.js#L114-L122 

Maybe someone from the Relay team can clarify how this works with connections not nested under a parent but on the top level of the query...?

valorize:
My query field has an id field which I know pass. But my MessagesList component does not automatically update after the graphql response :(

valorize:
With a lot of searching in code I found out that, altough my graphql store returns "query" as parent-id, my relay store used 'client:root' as parent-id. 
Optimistic Update only worked with a updater function. You can see the different usages in Delete and Create Mutations at: https://github.com/valorize/sm-chat/tree/master/src/mutations

nickhudkins:
[...]

PR #2504, opened 2018-08-10
Export root id on public api

julioxavierr:
This is useful to update a connection that haves `Query` as it's parent. Like [here](https://github.com/facebook/relay/issues/2157).

Hence, I don't know if this is the best way to do it.


kassens:
We already export `ROOT_ID` from `relay-runtime`: https://github.com/facebook/relay/blob/7d43533d12d96eddf3b9e0d8f06e9579543405b7/packages/relay-runtime/index.js#L197

We currently don't have good guidelines what we export via the react bindings and what from the "low level" API.

- We could re-export everything a typical product should need to react-relay and leave relay-runtime for frameworks (maybe a flux binding or Vue etc).
- We could export everything only via one module, so `react-relay` would only export React binding specific things and everything else, such as mutations would be imported via runtime?

Any thoughts here?

PR #2507, opened 2018-08-14
Send last fetched vars to refetchVariables() in RefetchContainer

### LINKED ISSUE #2506: RefetchContainer refetchVariables do not get updated
Hello Relay team,

I'm trying to build something similar to https://facebook.github.io/relay/docs/en/refetch-container.html#loading-more-data but using page and pageSize on my connection instead. I managed to get everything working only for the first page. The page variable would not increase after the first page and it seems to be a bug. Here's my createRefetchContainer:

```
export default createRefetchContainer(
  OrganizationsTable,
  graphql`
    fragment OrganizationsTable on Query
      @argumentDefinitions(page: { type: "Int", defaultValue: 1 }, pageSize: { type: "Int", defaultValue: 3 }) {
      organizations(page: $page, pageSize: $pageSize) {
        edges {
          node {
            ... on Organization {
              id
              name
            }
          }
        }
      }
    }
  `,
  graphql`
    query OrganizationsTable_RefetchQuery(
      $page: Int
      $pageSize: Int
    ) {
      ...OrganizationsTable
        @arguments(
          page: $page
          pageSize: $pageSize
        )
    }
  `,
);
```

and here's my `_fetch`:

```
_refetch = () => {
    this.props.relay.refetch(
      // TODO: relay is broken?
      ({ page, pageSize }) => {
        return { page: page + 1, pageSize };
      }
    );
  }
```
When `refetch` is first called I'm getting `{ page: 1, pageSize: 3 }` and returned `{ page: 2, pageSize: 3 }` so everything is great. I see the next page fetched. When I refetch again, I see that i'm getting the `{ page: 1. ... }` again instead of `{ page: 2, ... }`.

Further digging I see that right before my `refetchVariables` callback is called, at [ReaactRelayRefetchConatiner.js:L334](https://github.com/facebook/relay/blob/master/packages/react-relay/modern/ReactRelayRefetchContainer.js#L334), I see the following probing in console:
image

So, my question is, is this a bug? I'm expecting to receive `this.state.localVariables` in this case, but Relay is calling `this._getFragmentVariables()` which are outdated. Or should [the line setting localVariables](https://github.com/facebook/relay/blob/master/packages/react-relay/modern/ReactRelayRefetchContainer.js#L363) be updated? I see a TODO there. Thanks for your time, guys!

kenips:
We currently advertises that refetchVariables, when it is a function,
receives the last queried variables, but it does not. This commit fixes it.

Fixes #2506

### DIFF
--- a/packages/react-relay/modern/ReactRelayRefetchContainer.js
+++ b/packages/react-relay/modern/ReactRelayRefetchContainer.js
@@ -333,7 +333,7 @@ function createContainerWithFragments<
       );
       let fetchVariables =
         typeof refetchVariables === 'function'
-          ? refetchVariables(this._getFragmentVariables())
+          ? refetchVariables(this.state.localVariables || this._getFragmentVariables())
           : refetchVariables;
       fetchVariables = {...rootVariables, ...fetchVariables};
       const fragmentVariables = renderVariables
--- a/packages/react-relay/modern/__tests__/ReactRelayRefetchContainer-test.js
+++ b/packages/react-relay/modern/__tests__/ReactRelayRefetchContainer-test.js
@@ -752,6 +752,49 @@ describe('ReactRelayRefetchContainer', () => {
       expect(relayContext.variables).toEqual(refetchVariables);
     });
 
+    it('passes last fetched variables to refetchVariables()', () => {
+      expect.assertions(1);
+
+      const refetchVariables = {
+        cond: false,
+        id: '4',
+      };
+      refetch(refetchVariables, null, jest.fn());
+
+      environment.mock.resolve(UserQuery, {
+        data: {
+          node: {
+            id: '4',
+            __typename: 'User',
+            name: 'Zuck',
+          },
+        },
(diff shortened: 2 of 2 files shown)


jstejada:
hey @kenips thanks for the contribution and apologies for the delay in getting to this. 

Unfortunately I don't think we're going to accept this API. Righ tnow the "refetching" behavior is not well specified, and we've had a lot of these types of fixes that end up subtly breaking the behavior of some other part of the code.

We're revisiting the refetch api along with the pagination api to make them well specified and a lot simpler to use, and hopefully we'll address these issues then.

kenips:
Thanks for the update!

ntelkedzhiev:
@jstejada Will you revisit this any time soon? How else are we supposed to access the current fragment variables? Not that relay uses the the Context API, the only option is to update them locally (either on every refetch or on every render if we are using the new Context). But in that case, what's the purpose of having fragmentVariables???

PR #2519, opened 2018-08-21
Add Data Masking talk to docs

robrichard:
Add the talk I did at GraphQL Europe about Data Masking to the docs


kassens:
Awesome, we pruned a bunch of old content here a while ago. Thanks for adding recent content!

PR #2522, opened 2018-08-28
remove babel-cli dependency

kassens:
I don't see any references to `babel-cli` nor `babel-node` in the codebase, I don't think we need this anymore.


alloy:
Nice dep cleanups 👌

PR #2526, opened 2018-08-30
Fix remove_range examples not being strings

danielholmes:
Pretty sure this was just an oversight in the original doc and was meant to include string quotes


alloy:
@jstejada 👍

kassens:
This was fixed as part of a different PR rewriting parts of the mutation docs. Thanks for catching this though!

PR #2536, opened 2018-09-19
[relay] Upgrade build config to use Babel 7 and Webpack 4

jstejada:
**Test Plan:**

- `yarn`
- `yarn build`
- `yarn test`
- Sizes:
  - old_dist: 256B
  - new_dist: 256B
- `diff -Ewbur old_dist new_dist`: https://gist.github.com/jstejada/d52d5d9a311371b723ec5a78edc3fe8f
- Added the following changes to the code and verified that the new syntax could be compiled and built without errors:

```
const obj = {};
const foo = obj?.foo ?? 10;

const ReactRelayContext = React.createContext<RelayContext | null>(
  null,
);

```


nodkz:
Any plans about publishing a new package version?

PR #2549, opened 2018-10-18
Update version names

yachaka:
Versions above 1.4.1 were missing the starting `v`, which was causing the links pointing to release notes to be false.

Link of the page where this can be seen: https://facebook.github.io/relay/en/versions.html
Any links to release notes above version 1.4.1 leads to a 404.


jstejada:
Thanks @yachaka! Would you mind verifying that the links to the documentation still work correctly?

PR #2571, opened 2018-11-12
RFC: Custom Opaque Types for ID fields

sheki:
Before I go too deep into a bad path. I would like to get feedback on the approach for generating Opaque types for ID fields. 

In this PR if the user provides config for using a custom opaque types, all ID fields will be replaced with the provided type. 

I think it is better developer experience if relay-came with a global ID field as the fallback instead of a string if the user does not provide this opaque type. 

Tests pending.


jstejada:
hey @sheki, apologies if I'm missing something. What's the motivation for this change?

PR #2573, opened 2018-11-14
fix: use correct file in "main" of relay-compiler

Haroenv:
I noticed this when trying out yarn PnP, which is more strict about the module resolution algorithm than node is (which is technically a bug in Yarn, cc @arcanis), but anyway: let's fix the file here!

Node will fall back to `index.js`, while Yarn tried to use the non-existing `RelayCompilerPublic`

### DIFF
--- a/packages/relay-compiler/package.json
+++ b/packages/relay-compiler/package.json
@@ -10,7 +10,7 @@
   "homepage": "https://facebook.github.io/relay/",
   "bugs": "https://github.com/facebook/relay/issues",
   "repository": "facebook/relay",
-  "main": "lib/RelayCompilerPublic",
+  "main": "lib/RelayCompilerPublic.js",
   "bin": {
     "relay-compiler": "bin/relay-compiler"
   },
--- a/packages/relay-compiler/package.json
+++ b/packages/relay-compiler/package.json
@@ -10,7 +10,7 @@
   "homepage": "https://facebook.github.io/relay/",
   "bugs": "https://github.com/facebook/relay/issues",
   "repository": "facebook/relay",
-  "main": "RelayCompilerPublic",
+  "main": "lib/RelayCompilerPublic",
   "bin": {
     "relay-compiler": "bin/relay-compiler"
   },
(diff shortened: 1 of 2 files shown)


sibelius:
missing main here as well https://github.com/facebook/relay/blob/master/packages/react-relay/package.json

Haroenv:
@sibelius, that's likely not a problem (didn't test, I don't have it in my dependencies now), but I did see `graphql-compiler` also has the file without `lib`. Maël told me it could be related to Haste, so I'll leave it up to the contributors of this project to enlighten me :)

sibelius:
I think missing main could cause problems in another scenarios, like jest

arcanis:
No, a missing main will simply fallback to `index.js`, it should be fine. The problem was that this PR had a more fringe case: there is a `main`, but it points to a non-existing file.

Haroenv:
looks like this breaks the tests, will wait until someone who understands haste more than me comments :)

yuzhi:
@alunyov can you take a look at this one and figure out the next steps?

packages/relay-compiler/package.json
```diff
@@ -10,7 +10,7 @@
   "homepage": "https://facebook.github.io/relay/",
   "bugs": "https://github.com/facebook/relay/issues",
   "repository": "facebook/relay",
-  "main": "RelayCompilerPublic",
+  "main": "lib/RelayCompilerPublic",
```
alunyov: Hey @Haroenv, thank you for looking into this. It looks like `lib/RelayCompilerPublic` does not exist. This should probably be just `./RelayCompilerPublic`...

Can you squash your changes into single commit?
Thank you!

Haroenv: This is exactly what the PR is for. While the file `/RelayCompilerPublic` exists in the source, once published it doesn't, and the main file should be index.js instead of the non-existent file (then) `RelayCompilerPublic`. I'm not sure which tooling depends on the main field, but the other packages had the `lib/xxx` also listed when I checked last time. If possible, the main key can also be removed before publishing? This would also solve the issue I think.

I can squash once we decide on which way to go @alunyov

jstejada: ahh I see what the problem is. @alunyov @kassens do we do CommonJS imports for the relay-compiler package? is it safe to just change the main entry point to point to `lib/*`?

kassens:
We've fixed this with restructuring the gulpfile.

Haroenv:
Thanks!

PR #2575, opened 2018-11-14
[Docs] Clarify behavior of query param when null

zandaleph:
cacheConfig is displayed with a ? to make it obvious it is optional (it then follows with the word Optional, which helps.)  I missed that query was optional without the same signaling.


zandaleph:
ugh, this is one of those "optional parameters" versus "nullable parameters" things.  Lemme see if I can rework this to be more correct and also more obvious...

jstejada:
thanks @zandaleph, let me know when it's ready for review

PR #2610, opened 2019-01-24
New function will be created with each render

KhogaEslam:
[docs] New function will be created with each render of the module, so render function passed inside render of QueryRenderer needs to be inside the class itself, so that it is created once and used any time after that.

Somebody will say, why not doing the same for query function ?
the answer is this depends on object pooling, and relay shares created queries objects.


TrySound:
Or class can be converted to function component :)

josephsavona:
Per @TrySound functional components for the parent (using QueryRender) and the render prop make a lot of sense here.

KhogaEslam:
Ok, what about now ?
Any additional suggestions ?

PR #2618, opened 2019-01-28
relay-compiler false positive detecting circular references

robrichard:
The following error is thrown in some cases:

```
Error: Found a circular reference from fragment 'X'
```

This can happen when a fragment that uses `@argumentDefinitions` is referenced multiple times and the fragment has no selection due to the default arguments.

A workaround is to add a selection field (like `__typename`) that will always be present in the fragment.

I haven't investigated a fix for this, but this PR contains test cases demonstrating the issue and workaround.


josephsavona:
Thanks for reporting this and for the test case!

josephsavona:
Fixed in ba013f8

PR #2619, opened 2019-01-30
[docs] Some updates around fetch policies.

alloy:
It wasn't clear to me how to update the website to pull in the latest docs. What command should I run to do so?


docs/Modern-QueryRenderer.md
```diff
@@ -14,6 +14,7 @@ However, a `QueryRenderer` will not start loading its data until it is mounted,
 * `environment`: The [Relay Environment](./relay-environment.html)
 * `query`: The `graphql` tagged query. **Note:** To enable [compatibility mode](./relay-compat.html), `relay-compiler` enforces the query to be named as `<FileName>Query`. Optional, if not provided, an empty `props` object is passed to the `render` callback.
 * `cacheConfig?`: Optional object containing a set of cache configuration options, i.e. `force: true` requires the fetch to be issued regardless of the state of any configured response cache.
+* `dataFrom?`: Optional prop to indicate if data already present in the store should be used to render immediately and updated from the network afterwards using the `NETWORK_ONLY` key. Using the `STORE_THEN_NETWORK` key, which is the default behavior, ignores data already present in the store and waits for the network results to come back.
```
robrichard: Are the keys backwards here?

alloy:
```suggestion
* `dataFrom?`: Optional prop to indicate if data already present in the store should be used to render immediately and updated from the network afterwards using the `STORE_THEN_NETWORK` key. Using the `NETWORK_ONLY ` key, which is the default behavior, ignores data already present in the store and waits for the network results to come back.
```

alloy: Good catch, thanks!

docs/Modern-QueryRenderer.md
```diff
@@ -14,6 +14,7 @@ However, a `QueryRenderer` will not start loading its data until it is mounted,
 * `environment`: The [Relay Environment](./relay-environment.html)
 * `query`: The `graphql` tagged query. **Note:** To enable [compatibility mode](./relay-compat.html), `relay-compiler` enforces the query to be named as `<FileName>Query`. Optional, if not provided, an empty `props` object is passed to the `render` callback.
 * `cacheConfig?`: Optional object containing a set of cache configuration options, i.e. `force: true` requires the fetch to be issued regardless of the state of any configured response cache.
+* `dataFrom?`: Optional prop to indicate if data already present in the store should be used to render immediately and updated from the network afterwards using the `STORE_THEN_NETWORK` key. Using the `NETWORK_ONLY ` key, which is the default behavior, ignores data already present in the store and waits for the network results to come back.
```
jstejada: hmm, not sure if we should actually document this prop cc @kassens @josephsavona 

If we do, we should also mention that this will likely not work if the query gets garbage collected, so you'd have to retain any query you'd like to use from cache manually

alloy: I'd add that this prop is currently being used by people and so it might actually be helpful to add docs (or a comment) that state it is still considered experimental.

josephsavona: I don’t have strong opinions. The new suspense query renderer that @jstejada built has a slightly different (better) api around using cache and/or server, so it might be confusing to document once only to change again soon?

PR #2621, opened 2019-01-30
[split-modern] remove classic import from RelayModernGraphQLTag

kassens:
Code in modern shouldn't import from Classic in preparation to delete it. Turns out the type definition here wasn't actually needed to type check.


mrtnzlml:
@kassens Do you have any plans _when_ to delete Classic?

kassens:
@mrtnzlml hopefully within the next weeks. If you depend on it, you can obviously keep using the version published on npm (we haven't made real changes to classic in the last 2-3 years anyways).

Removing compat/classic frees us to make more improvements to modern (maybe soon just Relay?) without considering interoperability working on the legacy core.

PR #2624, opened 2019-01-31
[docs] Add doc on type emission and usage.

### LINKED ISSUE #2516: Add docs page about flow type usage
We're generating flow types with the modern files and even if not officially exported yet, there are extracted types available via https://github.com/flow-typed/flow-typed.

We should document how the generated types are meant to be used.

alloy:
Closes #2516 

This also updates docusaurus to add support for [language-specific code tabs](https://docusaurus.io/docs/en/doc-markdown#language-specific-code-tabs). I don't think any other docs need to include examples in other languages, but I felt that for this page it makes sense.

### DIFF
--- a/docs/Modern-TypeEmission.md
+++ b/docs/Modern-TypeEmission.md
@@ -1,0 +1,325 @@
+---
+id: type-emission
+title: Type Emission
+---
+
+As part of its normal work, `relay-compiler` will emit type information for your language of choice that helps you write type-safe application code. These types are included in the artifacts that `relay-compiler` generates to describe your operations and fragments.
+
+Regardless of your choice of language, all language plugins will emit roughly the same sort of type-information, but be sure to read the documentation for other [language plugins](#language-plugins) to learn about their specifics.
+
+### Operation input data
+
+The shape of the variables object used for query, mutation, or subscription operations.
+
+In this example the emitted type-information would require the variables object to contain a `page` key with a non-null string.
+
+<!--DOCUSAURUS_CODE_TABS-->
+<!--Flow-->
+```javascript
+/**
+ * export type ExampleQueryVariables = {|
+ *   +artistID: string
+ * |}
+ */
+import type { ExampleQueryVariables } from "__generated__/ExampleQuery.graphql"
+
+const variables: ExampleQueryVariables = {
+  artistID: 'banksy',
+}
+
+<QueryRenderer
+  query={graphql`
+    query ExampleQuery($artistID: ID!) {
+      artist(id: $artistID) {
+        name
+      }
+    }
+  `}
+  variables={variables}
+/>
+```
+<!--TypeScript-->
+```javascript
+/**
+ * export type ExampleQueryVariables = {
+ *   readonly artistID: string
+ * }
+ * export type ExampleQuery = {
+ *   readonly variables: ExampleQueryVariables
+ * }
+ */
+import { ExampleQuery } from "__generated__/ExampleQuery.graphql"
+
+<QueryRenderer<ExampleQuery>
+  query={graphql`
+    query ExampleQuery($artistID: ID!) {
+      artist(id: $artistID) {
+        name
+      }
+    }
+  `}
+  variables={{
+    artistID: 'banksy',
+  }}
+/>
+```
+<!--END_DOCUSAURUS_CODE_TABS-->
+
+### Operation/Fragment selection-set data
+
+The shape of the data selected in a operation or fragment, following the [data-masking] rules. That is, excluding any data selected by fragment spreads, unless the `@relay(mask: false)` directive is used.
+
+In this example the emitted type-information describes the response data available to the operation’s render function.
+
+<!--DOCUSAURUS_CODE_TABS-->
+<!--Flow-->
+```javascript
+/**
+ * export type ExampleQueryResponse = {|
+ *   +artist: ?{|
+ *     +name: string
+ *   |}
+ * |}
+ */
+import type { ExampleQueryResponse } from "__generated__/ExampleQuery.graphql"
+
+<QueryRenderer
+  query={graphql`
+    query ExampleQuery {
+      artist(id: "banksy") {
+        name
+      }
+    }
+  `}
+  render={({ props }: { props?: ExampleQueryResponse }) => {
+    if (props) {
+      return props.artist && <div>{props.artist.name} is great!</div>
+    }
+    return <div>Loading</div>
+  }}
+/>
+```
+<!--TypeScript-->
+```javascript
+/**
+ * export type ExampleQueryResponse = {
+ *   readonly artist?: {
+ *     readonly name: string
+ *   }
+ * }
+ * export type ExampleQuery = {
+ *   readonly response: ExampleQueryResponse
+ * }
+ */
+import { ExampleQuery } from "__generated__/ExampleQuery.graphql"
+
+<QueryRenderer<ExampleQuery>
+  query={graphql`
+    query ExampleQuery {
+      artist(id: "banksy") {
+        name
+      }
+    }
+  `}
+  render={({ props }) => {
+    if (props) {
+      return props.artist && <div>{props.artist.name} is great!</div>
+    }
+    return <div>Loading</div>
+  }}
+/>
+```
+<!--END_DOCUSAURUS_CODE_TABS-->
+
+Similarly, in this example the emitted type-information describes the prop data that the container expects to receive.
+
+<!--DOCUSAURUS_CODE_TABS-->
+<!--Flow-->
+```javascript
+/**
+ * export type ExampleFragment_artist = {|
+ *   +name: string
+ * |}
+ */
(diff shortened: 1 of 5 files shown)


docs/Modern-TypeEmission.md
```diff
@@ -0,0 +1,325 @@
...
+title: Type Emission
+---
+
+As part of its normal work, `relay-compiler` will emit type information for your language of choice that helps you write type-safe application code. These types are included in the artifacts that `relay-compiler` generates to describe your operations and fragments.
+
+Regardless of your choice of language, all language plugins will emit roughly the same sort of type-information, but be sure to read the documentation for other [language plugins](#language-plugins) to learn about their specifics.
+
+### Operation input data
+
+The shape of the variables object used for query, mutation, or subscription operations.
```
ashfurrow: I think this should be `variable objects` instead of `variables object`.

docs/Modern-TypeEmission.md
```diff
@@ -0,0 +1,325 @@
...
+  variables={{
+    artistID: 'banksy',
+  }}
+/>
+```
+<!--END_DOCUSAURUS_CODE_TABS-->
+
+### Operation/Fragment selection-set data
+
+The shape of the data selected in a operation or fragment, following the [data-masking] rules. That is, excluding any data selected by fragment spreads, unless the `@relay(mask: false)` directive is used.
```
ashfurrow: > The shape of the data selected in a operation or fragment, following the [data-masking] rules.

This isn't a complete sentence, I'm not quite sure what it's saying.

docs/Modern-TypeEmission.md
```diff
@@ -0,0 +1,325 @@
...
+      biography
+    }
+  `
+)
+```
+<!--END_DOCUSAURUS_CODE_TABS-->
+
+### Fragment references
+
+The opaque identifier described in [data-masking] that a child container expects to receive from its parent, which represents the child container’s fragment spread inside the parent’s fragment.
```
ashfurrow: I guess, related to the above comment about incomplete sentences, is this a definition for fragment references? 

(Note: the technical term for incomplete sentences is "sentence fragment", so this is actually very on-brand 👍)

docs/Modern-TypeEmission.md
```diff
@@ -0,0 +1,325 @@
...
+
+_Please read [this important caveat](#single-artifact-directory) about actually enabling type-safe fragment reference checking._
+
+Consider a component that composes the above fragment container example. In this example, the emitted type-information of the child container receives a unique opaque identifier type, called a fragment reference, which the type-information emitted for the parent’s fragment references in the location where the child’s fragment is spread. Thus ensuring that the child’s fragment is spread into the parent’s fragment _and_ the correct fragment reference is passed to the child container at runtime.
+
```
ashfurrow: This section is the least clear to me – I think some additional notes on why and how you might use this would help. Maybe even a sentence like:

> Fragment references are how Relay connects different fragment containers together based on which fragments they themselves spread.

Or something like that. It's a tricky concept to try to describe 😅

docs/Modern-TypeEmission.md
```diff
@@ -0,0 +1,325 @@
...
+  }}
+/>
+```
+<!--END_DOCUSAURUS_CODE_TABS-->
+
+## Single artifact directory
+
+An important caveat to note is that by default strict fragment reference type-information will _not_ be emitted, instead they will be typed as `any` and would allow you to pass in any data to the child container.
+
+To enable this feature, you will have to tell the compiler to store all the artifacts in a single directory, like so:
```
ashfurrow: > To enable ~this feature~ strict fragment reference type emission

docs/Modern-TypeEmission.md
```diff
@@ -0,0 +1,325 @@
...
+
+```json
+{
+  "plugins": [
+    ["relay", { "artifactDirectory": "./src/__generated__" }]
+  ]
+}
+```
+
+It is recommended to alias this directory in your module resolution configuration such that you don’t need to specify relative paths in your source files. This is what is also done in the above examples, where artifacts are imported from a `__generated__` alias, rather than relative paths like `../../../../__generated__`.
```
ashfurrow: Nice 👍

docs/Modern-TypeEmission.md
```diff
@@ -0,0 +1,325 @@
...
+
+### Background information
+
+The reason is that `relay-compiler` and its artifact emission is stateless. Meaning that it does not keep track of locations of original source files and where the compiler previously saved the accompanying artifact on disk. Thus, if the compiler were to emit artifacts that try to import fragment reference types from _other_ artifacts, the compiler would:
```
ashfurrow: > The reason is that

It's unclear what the reason is for.

docs/Modern-TypeEmission.md
```diff
@@ -0,0 +1,325 @@
...
+
+* first need to know where on disk that other artifact exists;
+* and update imports when the other artifact changes location on disk.
+
+Facebook uses a module system called [Haste], in which all source files are considered in a flat namespace. This means that an import declaration does not need to specify the path to another module and thus there is no need for the compiler to ever consider the above issues. I.e. an import only needs to specify the basename of the module filename and Haste takes care of actually finding the right module at import time. Outside of Facebook, however, usage of the Haste module system is non-existent nor encouraged, thus the decision to not import fragment reference types but instead type them as `any`.
+
+At its simplest, we can consider Haste as a single directory that contains all module files, thus all module imports always being safe to import using relative sibling paths. This is what is achieved by the single artifact directory feature. Rather than co-locating artifacts with their source files, all artifacts are stored in a single directory, allowing the compiler to emit imports of fragment reference types.
```
ashfurrow: A+, this is super-clear.

docs/Modern-TypeEmission.md
```diff
@@ -0,0 +1,325 @@
...
+
+## Language plugins
+
+* Flow: This is the default and builtin language plugin. You can explicitly enable it like so:
+
+  ```shell
+  $ relay-compiler --language javascript […]
+  ```
+
+* [TypeScript](https://github.com/relay-tools/relay-compiler-language-typescript): This is a language plugin for the TypeScript language maintained by the community. Install and enable it like so:
```
ashfurrow: I would make `a language plugin for TypeScript` the link to https://github.com/relay-tools/relay-compiler-language-typescript instead of `TypeScript`.

docs/Modern-TypeEmission.md
orta: Yeah, this is like the best description I've seen of haste anywhere (and I [tried to describe it](http://artsy.github.io/blog/2016/11/14/JS-Glossary/#haste-map) once)

website/package.json
```diff
@@ -15,7 +15,7 @@
     "crowdin-download": "crowdin-cli --config ../crowdin.yaml download -b master"
   },
   "dependencies": {
-    "docusaurus": "1.6.2",
+    "docusaurus": "1.7.2",
```
jstejada: let's bump the docusaurus version on a separate commit

website/i18n/en.json
```diff
@@ -388,6 +394,27 @@
       },
       "version-v1.7.0-videos": {
         "title": "Videos"
+      },
```
jstejada: hmm, why were these changes generated?

alloy: I believe they were an artifact of running `yarn build` and they _seemed_ like they should have been committed previously and so I thought I’d include them. I’m fine with removing them.

website/package.json
alloy: Sounds good. I tend to forget that these commits get squashed.

docs/Modern-TypeEmission.md
```diff
@@ -0,0 +1,325 @@
...
+
+export const ExampleFragment = createFragmentContainer(
+  (props: { artist: ExampleFragment_artist }) => (
+    <div>About the artist: {props.artist.biography}</div>
+  ),
+  graphql`
+    fragment ExampleFragment_artist on Artist {
+      biography
+    }
+  `
```
josephsavona: let's change this to 

```
  {
    artist: graphql`...`,
  }
```

The fact that this syntax works is an unfortunate accident and we're going to move toward dropping support (first marking as deprecated, of course). We've received a lot of feedback that confirms our intuition that having multiple ways of defining fragments is confusing to users, and it also makes it harder to do things like have Flow understand the content of the tag (something we're working on :-) )

docs/Modern-TypeEmission.md
```diff
@@ -0,0 +1,325 @@
...
```
josephsavona: same here and anywhere else this pattern is used

docs/Modern-TypeEmission.md
orta: oh nice, I've never seen this format - definitely reads better.

alloy: Yeah, I always have that indecisive tension, so good to hear 👍

PR #2636, opened 2019-02-05
[split-modern] remove Classic and Compat

kassens:
The day has come, modern and compat code no longer references RelayClassic which means we can remove it from OSS.


kassens:
This nuked about 50% of `*.js` files in this repo. ♻️

PR #2637, opened 2019-02-05
remove __typename as exclusion criterion in copyFields

mattkrick:
normalizing 2 payloads of the same union type could result in the same `storageKey`. I don't think this is a real-world issue today (aside from the 1 changed test), but it is a blocker for custom publish queues #2482 

NOTE:

v2.0.0 RelayModernRecord::update gives a warning that is a false positive for conflicting types.
"RelayModernRecord: Invalid record update, expected both versions of record...."
Using the example above, 2 subscription payloads of a union type could have the same `storageKey` but different types. Maybe we should timestamp subscription payloads as they come in?

### DIFF
--- a/packages/relay-runtime/store/RelayModernRecord.js
+++ b/packages/relay-runtime/store/RelayModernRecord.js
@@ -84,17 +84,20 @@ function clone(record: Record): Record {
 /**
  * @public
  *
- * Copies all fields from `source` to `sink`, excluding `__id` and `__typename`.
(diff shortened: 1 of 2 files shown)


mattkrick:
@devknoll any word on this getting merged to master?

devknoll:
I think @josephsavona took this over.

mattkrick:
closing in favor of #2791, please see newer PR

PR #2638, opened 2019-02-05
Fix issue with Relay serialization

### LINKED ISSUE #2633: Invariant error when printing Enums
Given the schema:

```js
var graphql = require('graphql')

var TestEnumType = new graphql.GraphQLEnumType({
  name: 'TestEnum',
  values: {
    FOO: {
      value: 1
    },
    BAR: {
      value: 2
    }
  }
})

var schema = new graphql.GraphQLSchema({
  query: new graphql.GraphQLObjectType({
    name: 'Query',
    fields: {
      print: {
        type: graphql.GraphQLString,
        args: {
          test: {
            type:  TestEnumType
          }
        }
     }
   }
 })
})
```

Having literal enum in the query results in error

```js
relayCompiler.Printer.print(
  relayCompiler.Parser.parse(
    schema, 
    'query Foo { print(test: FOO) }'
  )[0]
)
```

```
Invariant Violation: GraphQLIRPrinter: Expected value of type TestEnum to be a string, got `1`.
    at invariant (/home/freiksenet/Work/gatsby/gatsby/node_modules/fbjs/lib/invariant.js:40:15)
    at printLiteral (/home/freiksenet/Work/gatsby/gatsby/node_modules/relay-compiler/lib/GraphQLIRPrinter.js:255:105)
    at printValue (/home/freiksenet/Work/gatsby/gatsby/node_modules/relay-compiler/lib/GraphQLIRPrinter.js:243:12)
    at /home/freiksenet/Work/gatsby/gatsby/node_modules/relay-compiler/lib/GraphQLIRPrinter.js:212:24
    at Array.forEach (<anonymous>)
    at printArguments (/home/freiksenet/Work/gatsby/gatsby/node_modules/relay-compiler/lib/GraphQLIRPrinter.js:211:8)
    at printField (/home/freiksenet/Work/gatsby/gatsby/node_modules/relay-compiler/lib/GraphQLIRPrinter.js:56:81)
    at printSelection (/home/freiksenet/Work/gatsby/gatsby/node_modules/relay-compiler/lib/GraphQLIRPrinter.js:74:11)

```

freiksenet:
Fixes #2633.

I had to add dependency to actually add enums with proper test values, because monolithic SDL-first is so limited.

### DIFF
--- a/packages/relay-compiler/core/GraphQLIRPrinter.js
+++ b/packages/relay-compiler/core/GraphQLIRPrinter.js
@@ -16,6 +16,7 @@ const {DEFAULT_HANDLE_KEY} = require('../util/DefaultHandleKey');
 const {
   GraphQLEnumType,
   GraphQLInputObjectType,
+  GraphQLScalarType,
   GraphQLList,
   GraphQLNonNull,
 } = require('graphql');
@@ -312,6 +313,10 @@ function printLiteral(value: mixed, type: ?GraphQLInputType): string {
       value,
     );
     return result;
+  }
+  if (type instanceof GraphQLScalarType) {
+    const result = type.serialize(value);
+    return JSON.stringify(result);
   }
   if (Array.isArray(value)) {
     invariant(
--- a/packages/relay-compiler/core/GraphQLIRPrinter.js
+++ b/packages/relay-compiler/core/GraphQLIRPrinter.js
@@ -314,7 +314,7 @@ function printLiteral(value: mixed, type: ?GraphQLInputType): string {
     );
     return result;
   }
-  if (type instanceof GraphQLScalarType) {
+  if (type instanceof GraphQLScalarType && value != null) {
     const result = type.serialize(value);
     return JSON.stringify(result);
   }
--- a/packages/relay-compiler/core/GraphQLIRPrinter.js
+++ b/packages/relay-compiler/core/GraphQLIRPrinter.js
@@ -304,13 +304,14 @@ function printLiteral(value: mixed, type: ?GraphQLInputType): string {
     type = type.ofType;
   }
   if (type instanceof GraphQLEnumType) {
+    const result = type.serialize(value);
     invariant(
-      typeof value === 'string',
-      'GraphQLIRPrinter: Expected value of type %s to be a string, got `%s`.',
+      typeof result === 'string',
+      'GraphQLIRPrinter: Expected value of type %s to be a valid enum value, got `%s`.',
       type.name,
       value,
     );
-    return value;
+    return result;
   }
   if (Array.isArray(value)) {
     invariant(
--- a/packages/relay-compiler/core/__tests__/__snapshots__/GraphQLIRVisitor-test.js.snap
+++ b/packages/relay-compiler/core/__tests__/__snapshots__/GraphQLIRVisitor-test.js.snap
@@ -88,7 +88,7 @@ fragment UserFragment_mutated on User @argumentDefinitions(
     ... on Mutated @skip(if: $cond_mutated) {
       name_mutated
     }
-    thumbnail: profilePicture_mutated(size_mutated: "32_mutated") {
+    thumbnail: profilePicture_mutated(size_mutated: 42) {
       height_mutated
       width_mutated
       src: uri_mutated
--- a/packages/relay-compiler/core/__tests__/__snapshots__/GraphQLIRVisitor-test.js.snap
+++ b/packages/relay-compiler/core/__tests__/__snapshots__/GraphQLIRVisitor-test.js.snap
@@ -79,10 +79,10 @@ fragment UserFragment_mutated on User @argumentDefinitions(
   ... on Mutated @include(if: $cond_mutated) {
     id_mutated
     __typename_mutated
-    checkins_mutated(environments_mutated: [WEB_mutated]) {
+    checkins_mutated(environments_mutated: [MOBILE]) {
       __typename_mutated
     }
-    friends_mutated(after_mutated: $after_mutated, first_mutated: $first_mutated, traits_mutated: [HELPFUL_mutated]) {
+    friends_mutated(after_mutated: $after_mutated, first_mutated: $first_mutated, traits_mutated: [DERISIVE]) {
       count_mutated
     }
     ... on Mutated @skip(if: $cond_mutated) {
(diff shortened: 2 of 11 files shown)


freiksenet:
For some reason it's not picking up a dependency.

nodkz:
@freiksenet i'll drop a line here when directives sdl support will be added to `graphql-compose`

ETA today or tomorrow

nodkz:
Done `graphql-compose@5.8.1` fixes parsimg and schema construction from SDL with directives in `schemaComposer.addTypeDefs(sdl: string)` method

packages/relay-test-utils/RelayTestSchema.js
```diff
@@ -15,7 +15,37 @@ const RelayTestSchemaPath = require('./RelayTestSchemaPath');
...
 
 const {buildASTSchema, parse} = require('graphql');
+const {SchemaComposer} = require('graphql-compose');
 
-module.exports = buildASTSchema(
-  parse(fs.readFileSync(RelayTestSchemaPath, 'utf8'), {assumeValid: true}),
-);
+function buildSchema() {
+  // Compose upstream is going to add AST directives soon, making this simpler
+  const composer = new SchemaComposer();
```
josephsavona: graphql-js supports extending schemas, can we just use that?

packages/relay-compiler/core/__tests__/__snapshots__/GraphQLIRVisitor-test.js.snap
```diff
@@ -79,10 +79,10 @@ fragment UserFragment_mutated on User @argumentDefinitions(
   ... on Mutated @include(if: $cond_mutated) {
     id_mutated
     __typename_mutated
-    checkins_mutated(environments_mutated: [WEB_mutated]) {
+    checkins_mutated(environments_mutated: [MOBILE]) {
       __typename_mutated
     }
-    friends_mutated(after_mutated: $after_mutated, first_mutated: $first_mutated, traits_mutated: [HELPFUL_mutated]) {
+    friends_mutated(after_mutated: $after_mutated, first_mutated: $first_mutated, traits_mutated: [DERISIVE]) {
```
josephsavona: use positive language? :-)

packages/relay-compiler/core/GraphQLIRPrinter.js
```diff
@@ -304,13 +304,14 @@ function printLiteral(value: mixed, type: ?GraphQLInputType): string {
...
     invariant(
-      typeof value === 'string',
-      'GraphQLIRPrinter: Expected value of type %s to be a string, got `%s`.',
+      typeof result === 'string',
+      'GraphQLIRPrinter: Expected value of type %s to be a valid enum value, got `%s`.',
       type.name,
       value,
     );
-    return value;
+    return result;
```
josephsavona: I can't remember if GraphQLInputObject and GraphQLList also support `serialize()`, if they do can we just use the approach here for all types (ie replace the body of `printLiteral` w the body of the enum case)? Even if lists/objects don't work, we could do this for all enums and scalars.

packages/relay-test-utils/package.json
```diff
@@ -13,6 +13,7 @@
   "dependencies": {
     "@babel/runtime": "^7.0.0",
     "fbjs": "^1.0.0",
+    "graphql-compose": "^5.8.0",
```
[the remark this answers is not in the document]
josephsavona: Same question, we should use the built-in mechanism to extend schemas.

freiksenet:
Which mechanism? extendSchema only supports extending through AST.

packages/relay-test-utils/package.json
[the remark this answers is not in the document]
nodkz: It's my "lodash" for GraphQL type system 😈

If not today then it will be used tomorrow. Just give me some time to make this lib popular and right solution for such kind of problems.

packages/relay-test-utils/RelayTestSchema.js
freiksenet: Could you point me to that? extendSchema only support extending via AST.

packages/relay-compiler/core/__tests__/__snapshots__/GraphQLIRVisitor-test.js.snap
freiksenet: I can't, serialization for enums only returns valid values and HEPLFUL_mutated isn't a valid enum value for traits.

packages/relay-compiler/core/GraphQLIRPrinter.js
freiksenet: Only scalars and enums have serialize.

packages/relay-test-utils/package.json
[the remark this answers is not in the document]
freiksenet: Which mechanism? extendSchema only supports extending through AST.

freiksenet:
It's pretty critical for us to get this merged and released, because 2.0.0 is breaking for us (Gatsby). If you can't merge additional libraries for test, then I think this should be merged without new tests.

KyleAMathews:
Appreciate y'alls help getting this merged!

josephsavona:
> extendSchema only supports extending through AST.

Right - what I'm wondering is if we can start with a very minimal schema defined in code, and then extend it with an AST schema derived from the testschema.graphql file. I'm concerned about adding third-party dependencies to something as critical as our core test suite (which could affect both test performance and reliability). 

If that doesn't work, let's split this into two steps as you suggested: first we can land the printer changes to use `serialize()`, then explore a follow-up PR to add regression tests that unions w custom values work.

freiksenet:
> Right - what I'm wondering is if we can start with a very minimal schema defined in code, and then extend it with an AST schema derived from the testschema.graphql file. I'm concerned about adding third-party dependencies to something as critical as our core test suite (which could affect both test performance and reliability).

Right, I didn't think about doing it in other direction. I'll try to do it tomorrow.

freiksenet:
OK, I had to do double extend for this to work, because there seems to be a bug in AST Builder that sometimes makes it not use the defined type map function, probably when it's input types or maybe argument types.

freiksenet:
Any update on this?

freiksenet:
Disregard the closure, I misclicked when commenting.

freiksenet:
Pushed additional test for another regression we found that is fixed by this.

josephsavona:
I manually updated the imported version to account for the latest changes - note that I had to make some changes for backwards compatibility purposes for internal code, so updates on this PR won't apply cleanly to the imported version. Do you foresee any more changes?

josephsavona:
Thanks for letting us know about this and sending a fix!

PR #2639, opened 2019-02-05
Properly handle null cursor in PaginationContainer

danielberkompas:
I made this change to fix a bug I encountered. If I called `refetchConnection` with `cursor: null`, subsequent `loadMore()` calls would always load the same page of results from the server. 

This spreading of `this._refetchVariables` is apparently intended to make sure that `fetchVariables` has the expected shape, in case `connectionConfig.getVariables` doesn't return all the needed keys.

However, this breaks pagination if `refetchConnection` is called with `{cursor: null}`. This sets `this._refetchVariables` to contain `{cursor: null}`, and then if you call `loadMore()`, the cursor is always overridden by `null`, causing the same page of results to be loaded again and again. It's easily fixed by simply reversing the order of the spreading.

### DIFF
--- a/packages/react-relay/modern/ReactRelayPaginationContainer.js
+++ b/packages/react-relay/modern/ReactRelayPaginationContainer.js
@@ -708,8 +708,8 @@ function createContainerWithFragments<
         componentName,
       );
       fetchVariables = {
-        ...fetchVariables,
         ...this._refetchVariables,
+        ...fetchVariables,
       };
 
       const cacheConfig: ?CacheConfig = options


alloy:
Thanks! Could you add a test case that shows this change fixes the issue?

danielberkompas:
Hi @jstejada,

Thanks for the detailed reply. I am avoiding passing in the cursor explicitly now, and everything is working properly. I thought it'd be worth proposing the change anyway, but I can see why this is a risky change for you guys to merge. 

Hopefully this PR will help anyone else who passed a cursor into the `refetchConnection` function to at least know not to do that.

PR #2641, opened 2019-02-07
Move persisted id generation to RelayCompilerMain

janicduplessis:
### Summary

The current hash that we use for persisted queries (`sourceHash`) does not include the full query text and only the query node.

For example the following query and fragments:

App A:
```graphql
query CardsScreenQuery {
  viewer {
    ...CardsScreen_viewer
  }
}

fragment CardsScreen_viewer on Viewer {
  name
}
```

App B:
```graphql
query CardsScreenQuery {
  viewer {
    ...CardsScreen_viewer
  }
}

fragment CardsScreen_viewer on Viewer {
  logo
}
```

Would generate the same id because the current hash includes only the query part and not the fragment. This means it only hashes:

```graphql
query CardsScreenQuery {
  viewer {
    ...CardsScreen_viewer
  }
}
```

### Fix

To fix this I added a hash for the whole query text and pass that as the id to `persistQuery`. This makes sure the previous two queries in the example generate a different id.

Note that the ids used to work prior to upgrading to version 2.0 and using `--persisted-output`. Before I used a custom relay-compiler-bin that generated the queries map in `persistQuery`. 

### Test plan

Tested that this fix works in my apps where I discovered the issue. Basically I have 2 apps with some modules with the same name / query but different fragments. When generating the persisted queries for each app and merging them together I added a check to make sure there where not 2 different queries with the same id. The check fails before this change and passes after.

### DIFF
--- a/packages/relay-compiler/bin/RelayCompilerMain.js
+++ b/packages/relay-compiler/bin/RelayCompilerMain.js
@@ -22,6 +22,7 @@ const RelayFileWriter = require('../codegen/RelayFileWriter');
 const RelayIRTransforms = require('../core/RelayIRTransforms');
 const RelayLanguagePluginJavaScript = require('../language/javascript/RelayLanguagePluginJavaScript');
 
+const crypto = require('crypto');
 const fs = require('fs');
 const path = require('path');
 
@@ -304,7 +305,10 @@ function getRelayFileWriter(
     let queryMap;
     if (persistedQueryPath != null) {
       queryMap = new Map();
-      persistQuery = (text: string, id: string) => {
+      persistQuery = (text: string) => {
+        const hasher = crypto.createHash('md5');
+        hasher.update(text);
+        const id = hasher.digest('hex');
         queryMap.set(id, text);
         return Promise.resolve(id);
       };
--- a/packages/relay-compiler/codegen/RelayFileWriter.js
+++ b/packages/relay-compiler/codegen/RelayFileWriter.js
@@ -60,7 +60,7 @@ export type WriterConfig = {
   optionalInputFieldsForFlow: Array<string>,
   outputDir?: ?string,
   generatedDirectories?: Array<string>,
-  persistQuery?: ?(text: string, id: string) => Promise<string>,
+  persistQuery?: ?(text: string) => Promise<string>,
   platform?: string,
   schemaExtensions: Array<string>,
   noFutureProofEnums: boolean,
--- a/packages/relay-compiler/codegen/writeRelayGeneratedFile.js
+++ b/packages/relay-compiler/codegen/writeRelayGeneratedFile.js
@@ -49,7 +49,7 @@ async function writeRelayGeneratedFile(
   _generatedNode: GeneratedNode,
   formatModule: FormatModule,
   typeText: string,
-  _persistQuery: ?(text: string, id: string) => Promise<string>,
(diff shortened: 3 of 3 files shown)


alloy:
Can you please add some test coverage?

janicduplessis:
@alloy @jstejada Any pointers on how to test this? I had a look and there isn't any test coverage for 
 persisted queries, `writeRelayGeneratedFile` or `RelayFileWriter`. I'm not familiar with this code so I don't know how I should get the parameters these modules need.

alloy:
@janicduplessis How about moving it into a separate function so you can test it in isolation?

packages/relay-compiler/codegen/__tests__/persistGeneratedNode-test.js
```diff
@@ -0,0 +1,105 @@
...
+    `;
+    const initialGeneratedNode = createGeneratedNode(query)[0];
+    const generatedNode = await persistGeneratedNode(
+      initialGeneratedNode,
+      (text, id) => id,
+    );
+    expect(generatedNode).toBe(initialGeneratedNode);
+  });
+
+  it('handles similar queries with different fragments', async () => {
```
janicduplessis: This test would catch the regression this fixes

janicduplessis:
@alloy @jstejada Moved some logic for persisted queries to a separate file and added some tests.

janicduplessis:
@jstejada This should be good to ship

kassens:
Sorry to come to this so late!

We should certainly take the full text into account when creating the persisted ids in the default environment.

The one change that would make this cleaner is to move the hash algorithm out of the "core" parts of the codegen. I think we should remove the second (previously broken) parameter to `persistQuery` and make it just `persistQuery(text: string): Promise<string>`.

To have the default out-of-the-box experience, we can update the implementation of `persistQuery` in `packages/relay-compiler/bin/RelayCompilerMain.js` to do the hashing.

janicduplessis:
@kassens Makes sense, this is actually a lot cleaner. I tested that it still generates the proper ids. This will probably require some minor internal changes at fb.

kassens:
We don't use the second argument, we actually persist all queries in our database.

You can see this if you look at some facebook.com traffic and inspect the requests to /api/graphql ;) Commenting/liking on posts for example goes through RelayModern.

janicduplessis:
@kassens I see, when using the OSS relay compiler there is no way to customize `persistQuery` without forking `RelayCompilerMain` so it will always use the id generated by the default `persistQuery` in the generated files. In my case I use this id as the database id for my persisted queries. This is why this caused issues when the id was not unique because it did not hash the full text.

mrtnzlml:
> We don't use the second argument, we actually persist all queries in our database.

@kassens How does it work with the persisted queries invalidation? When do you remove old queries from the database? I imagine something like when the query ID is not used for a long time then it's time to delete it? Or what is the logic you use?

_(Sorry for abusing this issue for my question. Is there a better way where to ask "how does it work in FB" kind of questions? 😊)_

kassens:
@mrtnzlml We don't clean up old queries and don't allow breaking changes to the schema (but you can for example start returning `null` for some fields if a feature is removed).

This is important because Facebook doesn't deprecate mobile clients and force upgrade people (it might be very difficult if you only have 2G mobile internet access). So for example a random Facebook Android installation from 3 years ago still sends its persisted queries and should work!

Issue #2617, opened 2019-01-28
Watchman required with 2.0.0 when not watching files.


sheki:
When running a non watch command like
` relay-compiler --schema ./webpack/assets/javascripts/relay/schema.json --extensions js jsx --src ./webpack` in relay 2.0.0, the lack of `watchman` on the system fails the build.

This was not the case with relay 1.7.x. Our CI builds now need watchman. Watchman does not come with prebuilt packages and its not something I want to add to our build.

Is there a chance that this is a regression with 2.0.0

```
 relay-compiler --schema ./webpack/assets/javascripts/relay/schema.json --extensions js jsx --src ./webpack
Watchman:  Watchman was not found in PATH.  See https://facebook.github.io/watchman/docs/install.html for installation instructions
HINT: pass --watch to keep watching for changes.

Writing js
Unchanged: 476 files
Watchman:  Watchman was not found in PATH.  See https://facebook.github.io/watchman/docs/install.html for installation instructions
events.js:183
      throw er; // Unhandled 'error' event
      ^

Error: Watchman was not found in PATH.  See https://facebook.github.io/watchman/docs/install.html for installation instructions
    at _errnoException (util.js:1022:11)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:190:19)
    at onErrorNT (internal/child_process.js:372:16)
    at _combinedTickCallback (internal/process/next_tick.js:138:11)
    at process._tickCallback (internal/process/next_tick.js:180:9)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

sibelius:
you can disable watchman, check https://github.com/facebook/relay/blob/708f842c30e3cbb709a6b516002529b7de11dfe3/packages/relay-compiler/bin/RelayCompilerMain.js#L135

`relay-compiler --watchman false`

alloy:
I don't think it should just be disabled by default, but it _should_ only be enabled when `--watch` is enabled imo.

m4rcelofs:
> 
> 
> I don't think it should just be disabled by default, but it _should_ only be enabled when `--watch` is enabled imo. @sheki Care to create a PR for that?

The [Quick Start Guide](https://github.com/facebook/relay/issues/2617) indicates that watchman is an optional dependency, so throwing an error when just following the tutorial-provided command seems... weird, indeed.

jstejada:
cc @alunyov

alunyov:
[...]

PR #2643, opened 2019-02-10
What if Watchman CLI is missing...

alunyov:
Here is an updated version of the `GraphQLWatchmanClient.isAvailable`. (as a fix for https://github.com/facebook/relay/issues/2617)

Now, it's also checking if the `watchman` CLI is available. 

Test plan (Tested with `relay-examples:relay-publish-test`  (https://github.com/relayjs/relay-examples/pull/89)

First of all, I've removed watchman from my machine.
Then,
  - Build the compiler
  - Install updated dependencies relay-examples (yarn)
  - Run the compiler (yarn build)

Compiled without errors.


packages/relay-compiler/core/GraphQLWatchmanClient.js
```diff
@@ -24,6 +25,14 @@ class GraphQLWatchmanClient {
...
   static isAvailable(): Promise<boolean> {
     return new Promise(resolve => {
+      try {
+        const checkWatchman = execSync('command -v watchman').toString().trim();
+        if (checkWatchman === '') {
+          return resolve(false);
+        }
+      } catch {
+        return resolve(false);
+      }
```
alloy: TIL about `command` ! ✨ 

How about using the exit status, though? (And made it async too, although it probably matters very little overall.)

```js
const exec = require("util").promisify(require("child_process").exec)
undefined
> exec("command -v does-not-exist").then(() => console.log("Exists!")).catch(() => console.log("Does not exist!"))
Promise { … }
> Does not exist!
> exec("command -v ls").then(() => console.log("Exists!")).catch(() => console.log("Does not exist!"))
Promise { … }
> Exists!
```

alloy:
Can you clarify if using watchman without `--watch` is faster at yielding files than using `fast-glob` instead? (E.g. does watchman keep a live cached version of the results?)

The amount of faffing with ‘watch’ related options in the code is starting to confuse me a bit, so it would be nice if we could simplify that without a loss in performance.

packages/relay-compiler/core/GraphQLWatchmanClient.js
alunyov: I just realized that this won't work on windows. I will add something else.

alunyov:
I'm not sure about the performance differences between watchman and fast-glob. But I think at this point, it's fine to keep it as it is.

We've been working recently on the new relay compiler. It's ready, but still has a lot of dependencies on internal tools. And we still are using both versions - but I think once we complete the migration, we will replace an open source version too.

Issue #2647, opened 2019-02-18
Expand PluginInterface support


dispalt:
I just wrote a plugin for [scalajs](http://www.scala-js.org/) (very rough but works), using the plugin interface and two things would be super nice to support.

- Extra schema, like in the relayCompilerBin you add directives, it would be nice for a plugin to be able to add custom directives.  
- The getFileFilter function looks for the string graphql, we should be able to override that.

Otherwise, it went well.

sibelius:
related to https://github.com/facebook/relay/issues/2518

alloy:
Awesome! Just for my own interest, is the source of the plugin available?

> * Extra schema, like in the relayCompilerBin you add directives, it would be nice for a plugin to be able to add custom directives.

Could you elaborate on the motivation for this?

> * The getFileFilter function looks for the string graphql, we should be able to override that.

Huh, yeah I didn’t even notice that when I added plugin support. Makes sense. What would you like to be able to override it with?

dispalt:
https://github.com/dispalt/scala-relay/tree/next/node-compiler

The code is kinda terrible, but it works in production =)

> Awesome! Just for my own interest, is the source of the plugin available?
> 
> > * Extra schema, like in the relayCompilerBin you add directives, it would be nice for a plugin to be able to add custom directives.
> 
> Could you elaborate on the motivation for this?

So I used some generator specific options, scalajs is much more rigid type-wise than flow or ts, so I'd use it as a way to pass options to the specific fragment to effect the generated code.  Specifically there are a couple was to do nullable in scalajs, and one way creates more baggage but is more correct, and I wanted to be able to choose on a per fragment basis whether it was on.

> 
> > * The getFileFilter function looks for the string graphql, we should be able to override that.
> 
> Huh, yeah I didn’t even notice that when I added plugin support. Makes sense. What would you like to be able to override it with?

well I used to use an annotation `@gql("""fragment ... """)def bar` but it just seems like something that shouldn't below there, or at least be override-able.

dispalt:
I guess if I were to prioritize the adding schema in a plugin would be my top choice.

@alloy does this make sense?

alloy:
Re directives, makes sense 👍 I can see similar directives being useful to control generation for other languages. E.g. in some cases the application code will know for a fact that a user is signed in and a field will never return `null`.

I think both these should be fairly easy to add, so probably not necessary to prioritise one over the other. Were you looking to create a PR for this yourself?

[...]

PR #2652, opened 2019-02-20
Add support for three additional optional attributes for the language PluginInterface

dispalt:
- getFileFilter, optionally pass in some filter function as a first pass filter.
- processCompilerTransforms, hook into the compilerTransforms.
- extraSchema, allow a plugin to add some extra schema.

### DIFF
--- a/packages/relay-compiler/bin/RelayCompilerMain.js
+++ b/packages/relay-compiler/bin/RelayCompilerMain.js
@@ -188,9 +188,9 @@ Ensure that one such file exists in ${srcDir} or its parents.
 
   const useWatchman = options.watchman && (await WatchmanClient.isAvailable());
 
-  const schema = getSchema(schemaPath);
-
   const languagePlugin = getLanguagePlugin(options.language);
+
+  const schema = getSchema(schemaPath, languagePlugin);
 
   const inputExtensions = options.extensions || languagePlugin.inputExtensions;
   const outputExtension = languagePlugin.outputExtension;
@@ -200,6 +200,7 @@ Ensure that one such file exists in ${srcDir} or its parents.
 
   const sourceModuleParser = RelaySourceModuleParser(
     languagePlugin.findGraphQLTags,
+    languagePlugin.getFileFilter,
   );
 
   const providedArtifactDirectory = options.artifactDirectory;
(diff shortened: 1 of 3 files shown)


dispalt:
Refs #2647

sibelius:
can you add some tests?

dispalt:
Sure wanted to make sure it's something you guys would accept before doing the work.

packages/relay-compiler/bin/RelayCompilerMain.js
```diff
@@ -350,8 +357,10 @@ function getRelayFileWriter(
   };
 }
 
-function getSchema(schemaPath: string): GraphQLSchema {
+function getSchema(schemaPath: string, languagePlugin: PluginInterface): GraphQLSchema {
```
alloy: This parameter can be narrowed down to just the extra schema parameter.

packages/relay-compiler/bin/RelayCompilerMain.js
```diff
@@ -188,10 +188,10 @@ Ensure that one such file exists in ${srcDir} or its parents.
 
   const useWatchman = options.watchman && (await WatchmanClient.isAvailable());
 
-  const schema = getSchema(schemaPath);
-
   const languagePlugin = getLanguagePlugin(options.language);
 
+  const schema = getSchema(schemaPath, languagePlugin);
```
alloy: Just pass in the schema additions.

packages/relay-compiler/language/RelayLanguagePluginInterface.js
```diff
@@ -34,6 +49,9 @@ export type PluginInterface = {
   findGraphQLTags: GraphQLTagFinder,
   formatModule: FormatModule,
   typeGenerator: TypeGenerator,
+  processCompilerTransforms?: CompilerTransforms => CompilerTransforms,
+  extraSchema?: string,
```
alloy: Let’s rename this to `schemaAdditions` and add some doc with an example of adding language specific directives.

packages/relay-compiler/language/RelayLanguagePluginInterface.js
```diff
@@ -34,6 +49,9 @@ export type PluginInterface = {
...
+  getFileFilter?: string => FileFilter,
```
alloy: Can you name the parameter so readers understand it’s a file path.

packages/relay-compiler/core/RelaySourceModuleParser.js
```diff
@@ -88,13 +88,15 @@ module.exports = (tagFinder: GraphQLTagFinder) => {
...
 
-  function getFileFilter(baseDir: string): FileFilter {
+  function getFileFilterDefault(baseDir: string): FileFilter {
     return (file: File) => {
       const text = fs.readFileSync(path.join(baseDir, file.relPath), 'utf8');
       return text.indexOf('graphql') >= 0;
     };
   }
 
+  const getFileFilter = fileFilter ? fileFilter : getFileFilterDefault;
```
alloy: What do you think of changing this to always read the file and yield the contents to the plugin, rather than having the plugin read the file?

I also feel like not making this optional but having each plugin implement their version would make the API clearer, although it would technically be a breaking change. Not sure in how much the Relay team would care about that, especially as there’s really only the TS plugin that’s currently in existence (afaik). @jstejada what say you?

packages/relay-compiler/language/RelayLanguagePluginInterface.js
alloy: Or perhaps `schemaExtensions`, as per the `extendSchema` API 🤔

sibelius:
I think this should go to relay.config.js as well

PR #2665, opened 2019-02-26
improve the relay store doc by using the consistent imported name

lisongx:
This part of the doc is very confusing, as you can see we import the pubic api in [L458](https://github.com/facebook/relay/pull/2665/files#diff-af6b98bdc4cbe6dedaed354e06c40bf4R458): `import {ConnectionHandler} from 'relay-runtime'` , but using `RelayConnectionHandler` in the rest of this page.

This patch change all of them to the imported one `ConnectionHandler`.


kassens:
Thank you!

PR #2681, opened 2019-03-07
Sort selections in a stable sort order

### LINKED ISSUE #2675: [help wanted] jest tests not passing on node 11
When enabling travis for node 11 or running node 11 locally, some compiler tests fail with keys of the generated code being in different order.

I assume there's some edge case bug in `Map` in the V8 version used by node 11. Would be good to debug and potentially file an upstream bug with a reduced use case.

If someone wants to investigate this, please say so in the comments here so we don't duplicate the effort!

eliperkins:
This fixes an issue that was raised in Node 11, where `sort` may call to items in a different order than other versions of Node.

Fixes #2675


packages/relay-test-utils/__tests__/__snapshots__/RelayMockPayloadGenerator-test.js.snap
```diff
@@ -331,14 +331,19 @@ fragment TestFragment on User @argumentDefinitions(condition: {type: "Boolean!"}
     "profile_picture": {
       "uri": "<mock-value-for-field-\\"uri\\">"
     },
-    "__typename": "User",
     "actor": {
       "username": "<mock-value-for-field-\\"username\\">",
-      "__typename": "Comment",
-      "body": {
-        "text": "<mock-value-for-field-\\"text\\">"
+      "__typename": "User",
```
eliperkins: This looks like an odd change... 🤔

packages/relay-compiler/transforms/SkipRedundantNodesTransform.js
```diff
@@ -207,13 +207,19 @@ function transformNode<T: Node>(
...
+      const isScalarOrLinkedField =
+        next.kind === 'ScalarField' || next.kind === 'LinkedField';
+      if (isScalarOrLinkedField) {
+        return [[...scalarsAndLinkedFields, next], rest];
+      }
+      return [scalarsAndLinkedFields, [...rest, next]];
+    },
+    [[], []],
+  );
+  return [...scalarsAndLinkedFields, ...rest];
```
eliperkins: Should these be sorted before merging them back together? Or is the order of the existence good enough?

packages/relay-test-utils/__tests__/__snapshots__/RelayMockPayloadGenerator-test.js.snap
```diff
@@ -745,12 +750,16 @@ fragment TestFragment on Viewer {
 {
   "variables": {},
   "payload": {
-    "__typename": "Page",
```
eliperkins: This too. 🤔 🤔 🤔

packages/relay-compiler/transforms/SkipRedundantNodesTransform.js
josephsavona: Repeatedly creating new arrays like this is incredibly inefficient. Instead, let’s create just two new arrays and partition the elements into them. There’s a helper function partitionArray() (or similar) in RelayParser that we can extract into a module and reuse it for this.

josephsavona: I don’t think we need to sort, just partition

packages/relay-test-utils/__tests__/__snapshots__/RelayMockPayloadGenerator-test.js.snap
josephsavona: I think what’s happening is that we’re able to skip more redundant nodes as a result of this change (exactly what I’d expect). I need to look at this more and try it locally to confirm.

packages/relay-compiler/transforms/SkipRedundantNodesTransform.js
eliperkins: Done in 516ae8c

kassens:
This looks good to me. Now testing internally. The only update I made was adding the file headers and keeping the old sort behind a [`Rollout`](https://github.com/facebook/relay/blob/master/packages/relay-compiler/util/Rollout.js) flag since this introduces so many changes we have to land them in multiple waves internally 😂

PR #2682, opened 2019-03-07
Fixed links

LorenzNickel:
Both links can not be accessed any longer, for one I found an alternative

### DIFF
--- a/packages/relay-compiler/core/GraphQLSchemaUtils.js
+++ b/packages/relay-compiler/core/GraphQLSchemaUtils.js
@@ -82,8 +82,6 @@ function canHaveSelections(type: GraphQLType): boolean {
  * Implements duck typing that checks whether a type has an id field of the ID
  * type. This is approximating what we can hopefully do with the __id proposal
  * a bit more cleanly.
- *
- * https://github.com/graphql/graphql-future/blob/master/01%20-%20__id.md
  */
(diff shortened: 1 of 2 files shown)


LorenzNickel:
Okay, can somebody help me to pass this check? Seems like the new url in my snapshot does not match something else, how can I fix this?

packages/relay-compiler/core/GraphQLSchemaUtils.js
```diff
@@ -82,8 +82,6 @@ function canHaveSelections(type: GraphQLType): boolean {
  * Implements duck typing that checks whether a type has an id field of the ID
  * type. This is approximating what we can hopefully do with the __id proposal
  * a bit more cleanly.
- *
- * https://github.com/graphql/graphql-future/blob/master/01%20-%20__id.md
```
kassens: This link still exist. As long as we have to comment above, it makes sense for me to keep?

LorenzNickel: I don't really understand what you mean. If I open this link (or just [the repo of it](https://github.com/graphql/graphql-future/)) I get a 404-Github page. Is this repo private?
image

kassens: Oh, I guess it's been made private as the last change was 3 years ago.

PR #2688, opened 2019-03-15
Cleanup: remove single iterall dependency

kassens:
We only test on arrays here, so we don't really need an external dep for this.

Test Plan: CI


josephsavona:
Yay one less dependency

PR #2690, opened 2019-03-15
Export flow types to npm

kassens:
I still need to figure out how to test this, but I think this might work to allow npm consumers to use the generated flow types.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -33,6 +33,7 @@ const gulpUtil = require('gulp-util');
 const header = require('gulp-header');
 const once = require('gulp-once');
 const path = require('path');
+const rename = require('gulp-rename');
 const webpack = require('webpack');
 const webpackStream = require('webpack-stream');
 
@@ -252,6 +253,20 @@ const modules = gulp.parallel(
   ),
 );
 
+const flowDefs = gulp.parallel(
+  ...builds.map(
+    build =>
+      function modulesTask() {
+        return gulp
+          .src(['**/*.js', '!**/__tests__/**/*.js', '!**/__mocks__/**/*.js'], {
+            cwd: PACKAGES + '/' + build.package,
+          })
+          .pipe(rename({extname: '.js.flow'}))
+          .pipe(gulp.dest(path.join(DIST, build.package)));
+      },
+  ),
+);
+
 const copyFilesTasks = [];
 builds.forEach(build => {
   copyFilesTasks.push(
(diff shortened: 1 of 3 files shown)


zth:
I've been testing this in a few different projects, and from what I can see, this generally works very well :+1: 

To get it working it's needed to replicate most of the `.flowconfig` of Relay, which I think is fine, but one thing I noticed was that I needed to switch module system in Flow to `haste`. I wonder what type of implications this has for projects not using `haste` with Flow already, but I can't seem to find that much information about what it really does when used in Flow.

Is a lot of the Relay codebase still deeply tied to `haste`? I saw that most of the `haste`-stuff seemed to be helpers from `fbjs`, which I _think_ could be swapped for regular imports fairly easy, right? Would you accept a PR for that?

So glad this is so close to happening, it'll be really nice to be able to use the actual definitions!

mrtnzlml:
> I've been testing this in a few different projects...

@zth How did you test it? I would like to test it as well but it's not released yet. We are currently shipping [our own](https://github.com/kiwicom/relay/blob/0fccd51eb0c244468a9e4df1d85efcc779e1784c/src/createFragmentContainer.js) (slightly modified) types to share it with all our projects as a workaround.

zth:
@mrtnzlml I cloned master, applied the PR, built it, and copied the built packages to the various projects. However it might be smarter to checkout a specific version tag like 3.0 and then apply the PR manually, because master contains a breaking change that forces you to use the object syntax for fragment definitions (not simply a graphql-tag), and our code bases are full of definitions with just tags and not the object syntax.

Another note is that you'll need to copy most of the settings from Relays flowconfig (including setting the module system to haste), so it's quite an invasive change. 

Yeah we also have some hand rolled/copied types we use right now. The absolute best thing would be if Flow could export libdefs from codebases automatically, that way we wouldn't need to change so much config (or stay on the same Flow version as Relay even), but from what I've heard that's going to be a while before it lands in Flow.

zth:
In waiting for this to land (any update btw @kassens ?) I've published a package that attempts to add these types on the fly without you needing to resort to using a fork of Relay with this PR applied. Basically, you install https://www.npmjs.com/package/relay-flow-types-installer and then add `relay-flow-types-installer` to your `postinstall` script. 

The package contains all the types generated via this PR for Relay versions `2.0.0` -> `5.0.0`, and will copy the proper version of the Flow typed source files (`.js.flow`) into each Relay package directory in `node_modules` on install. 

Please the package's readme for more information.

kassens:
I'm landing this after the 9.0 release to give us some time to test this in the latest release.

PR #2694, opened 2019-03-17
Remove nested backticks from pagination container example

martinbooth:
Pagination Container example is invalid javascript because it has backticks in the graphql string which are not escaped


kassens:
Thank you @martinbooth!

PR #2695, opened 2019-03-23
Adds the sketch file for the homepage re-design

orta:
While the design is WIP and home isn't implemented, this is just to raise visibility for that file. I'll be amend commiting my changes over time.


orta:
I switched to figma, and added a link in #2696

PR #2700, opened 2019-03-31
[docs] Add docs about language plugin that emits uncommented Flow types

zth:
Some editors/IDEs do not understand Flow properly unless it's in plain source. This PR adds information about a community maintained language plugin for the compiler that emits the Flow types as plain source instead of inside comments.


jstejada:
thanks!

PR #2707, opened 2019-04-11
Community Resources: add link to Relay examples repository

mrtnzlml:
Hello guys! 👋 

We had internal Relay workshop some time ago and we decided to record some parts and share it with the community along with short documentation as well as with working code. I believe this could serve as a good source of information even for other people working with Relay.

Thanks for considering it.

### DIFF
--- a/docs/Community-LearningResources.md
+++ b/docs/Community-LearningResources.md
@@ -6,6 +6,9 @@ title: Community Learning Resources
 ## Relay Modern Learning Blog Posts Thread on Twitter
 https://twitter.com/sseraphini/status/1078595758801203202
 
+## Relay examples - common patterns explained via code and videos
+https://github.com/kiwicom/relay-example
+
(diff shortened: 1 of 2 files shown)


sibelius:
this is awesome

my only feedback is about the color of videos

image

it is the same color of apollo docs

mrtnzlml:
@sibelius That's just a coincidence. Definitely not an intention. The guy who created these videos used our own company color (unfortunately the old one - should be this https://orbit.kiwi/visual-style/color/).

Also, I checked their docs and seems like it's quite different? https://www.apollographql.com/docs/ Either way, they were using our color... :D

mrtnzlml:
Hi, is here anything I can do to make this PR better (mergeable)? :)

jstejada:
Thanks for adding this, apologies we didn't get to it sooner!

PR #2709, opened 2019-04-27
Clarify use of onCompleted vs onError

milosa:
This took me a while to figure out, so I think it's a good idea to add this information to the documentation.

[This article](https://itnext.io/the-definitive-guide-to-handling-graphql-errors-e0c58b52b5e1) writes:

> As of March 2018, neither Apollo-Client (including subscriptions-transport-ws) nor Relay Modern is perfect at handling errors. 
> Relay’s mutation API comes close with its onCompleted(result, errors) callback, but this is sorely missed for queries and 
> subscriptions. 

and

>  If GraphQL gives you a result with data, even if that result contains errors, it is not an error.

### DIFF
--- a/docs/Modern-Mutations.md
+++ b/docs/Modern-Mutations.md
@@ -37,7 +37,10 @@ commitMutation(
 * `config`:
   * `mutation`: The `graphql` tagged mutation query.
   * `variables`: Object containing the variables needed for the mutation. For example, if the mutation defines an `$input` variable, this object should contain an `input` key, whose shape must match the shape of the data expected by the mutation as defined by the GraphQL schema.
-  * `onCompleted`: Callback function executed when the request is completed and the in-memory Relay store is updated with the `updater` function. Takes a `response` object, which is the "raw" server response, and `errors`, an array containing any errors from the server. .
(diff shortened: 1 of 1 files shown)


docs/Modern-Mutations.md
```diff
@@ -37,7 +37,10 @@ commitMutation(
 * `config`:
   * `mutation`: The `graphql` tagged mutation query.
   * `variables`: Object containing the variables needed for the mutation. For example, if the mutation defines an `$input` variable, this object should contain an `input` key, whose shape must match the shape of the data expected by the mutation as defined by the GraphQL schema.
-  * `onCompleted`: Callback function executed when the request is completed and the in-memory Relay store is updated with the `updater` function. Takes a `response` object, which is the "raw" server response, and `errors`, an array containing any errors from the server. .
+  * `onCompleted`: Callback function executed when the request is completed and the in-memory Relay store is updated with the `updater` function. Takes a `response` object, which is the "raw" server response, and `errors`, an array containing any errors from the server. 
+**Please note:** 
+    * This callback (and not `onError`) will be executed whenever the response contains `data`, even if `data` is `null`.
```
taion: This partly depends on your network layer. If the network layer throws an exception, then you hit `onError`.

jstejada: @milosa mind adding that clarification?

PR #2711, opened 2019-04-29
Upgrade Docusaurus to 1.9.0

yangshun:
Docusaurus just released v1.9.0 and I thought I would help Relay upgrade their version and at the same time use some of the newer features:

- Clean HTML (no more `.html` in the path)
- Show last updated time and author
- Collapsed sidenav (helpful especially when Relay has such a long list of side navigation items)
- Scroll to top button

Screen Shot 2019-04-29 at 4 25 05 PM
Screen Shot 2019-04-29 at 4 21 18 PM

### DIFF
--- a/website/i18n/en.json
+++ b/website/i18n/en.json
@@ -86,310 +86,310 @@
       "videos": {
         "title": "Videos"
       },
-      "version-1.5.0-installation-and-setup": {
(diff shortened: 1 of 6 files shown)


jstejada:
thanks!

PR #2715, opened 2019-05-03
Allow Declarative Mutation Configs to support root field aliases

levibuzolic:
Currently declarative mutation configs silently fail when using aliased root fields, this is a simple PR to add support for field aliases. 

I haven't had a chance to write a test case for this yet, I'd love some help there if anybody has any interest in this.


levibuzolic:
Anybody got a change to look at this?

sibelius:
can you add a test case for this?

levibuzolic:
Appears to be working in Relay 5.0.0

PR #2719, opened 2019-05-06
Fix require statements with single artifact directory on windows+metro

martinbooth:
File paths on windows have backslashes, but this code expects forward
slashes (see relativeReference variable).
Backslashes are ok in node, but metro bundler does not support them and
consequently the paths generated do not resolve when used with it
Lets use forward slashes on windows and unix as they work everywhere

### DIFF
--- a/packages/babel-plugin-relay/createModernNode.js
+++ b/packages/babel-plugin-relay/createModernNode.js
@@ -125,12 +125,12 @@ function getRelativeImportPath(
   const relative = path.relative(
     path.dirname(filename),
     path.resolve(artifactDirectory),
-  );
+  ).replace(/\\/g, '/');
 
   const relativeReference =
     relative.length === 0 || !relative.startsWith('.') ? './' : '';
 
-  return relativeReference + path.join(relative, fileToRequire);
+  return relativeReference + path.posix.join(relative, fileToRequire);
 }
 
 module.exports = createModernNode;


packages/babel-plugin-relay/createModernNode.js
```diff
@@ -125,12 +125,12 @@ function getRelativeImportPath(
   const relative = path.relative(
     path.dirname(filename),
     path.resolve(artifactDirectory),
-  );
+  ).replace(/\\/g, '/');
```
anikethsaha: @martinbooth  In windows, it gives two backslashes so I think  the regex will be
`replace(/\\\\/g, '/')`

anikethsaha:
@martinbooth 
> metro bundler does not support them

Are you sure about this?

sibelius:
can we have a test for this?

anikethsaha:
@martinbooth  can you please comment on the output of `console.log(relative)`.

LOLdevelopr:
The current code will work fine even with backslashes and forwards. Not required to make it all to `osx` convertable. 

```bash
In Windows

> path.resolve(""/hh/kl");
'H:\\hh\\kl'

> path.posix.resolve("/hh/kl");
'/hh/kl'

```

They are not same.

packages/babel-plugin-relay/createModernNode.js
martinbooth: Maybe you're looking at the output of evaluating that expression in node? That escapes the output: i.e.
```
> path.relative(path.dirname("C:\\dir1\\dir2\\dir3"), path.resolve("C:\\"))
'..\\..'
```
If you print it to the console unescaped you'll see that there is just one backslash:
```
> console.log(path.relative(path.dirname("C:\\dir1\\dir2\\dir3"), path.resolve("C:\\")))
..\..
```

packages/babel-plugin-relay/createModernNode.js
```diff
@@ -125,12 +125,12 @@ function getRelativeImportPath(
...
 
   const relativeReference =
     relative.length === 0 || !relative.startsWith('.') ? './' : '';
```
martinbooth: This line does not use the operating system's delimiter, so we are mixing forward slashes and backslashes on windows

martinbooth:
> @martinbooth
> 
> > metro bundler does not support them
> 
> Are you sure about this?

Here is the output of starting metro bundler without this fix applied:
```
$ react-native start --reset-cache
┌──────────────────────────────────────────────────────────────────────────────┐
│                                                                              │
│  Running Metro Bundler on port 8081.                                         │
│                                                                              │
│  Keep Metro running while developing on any JS projects. Feel free to        │
│  close this tab and run your own Metro instance if you prefer.               │
│                                                                              │
│  https://github.com/facebook/react-native                                    │
│                                                                              │
└──────────────────────────────────────────────────────────────────────────────┘

Looking for JS files in
   C:\source\demo

warning: the transform cache was reset.
::1 - - [08/May/2019:19:48:30 +0000] "GET /launch-js-devtools HTTP/1.1" 200 - "-" "-"
Loading dependency graph, done.
error: bundling failed: Error: Unable to resolve module `..\__generated__\nativeIndexQuery.graphql` from `C:\source\demo\src\native\index.tsx`: The module `..\__generated__\nativeIndexQuery.graphql` could not be found from `C:\source\demo\src\native\index.tsx`. Indeed, none of these files exist:
  * `C:\source\__generated__\nativeIndexQuery.graphql(.native||.windows.js|.native.js|.js|.windows.json|.native.json|.json|.windows.ts|.native.ts|.ts|.windows.tsx|.native.tsx|.tsx)`
  * `C:\source\__generated__\nativeIndexQuery.graphql\index(.native||.windows.js|.native.js|.js|.windows.json|.native.json|.json|.windows.ts|.native.ts|.ts|.windows.tsx|.native.tsx|.tsx)`
    at ModuleResolver.resolveDependency (C:\source\demo\node_modules\metro\src\node-haste\DependencyGraph\ModuleResolution.js:120:15)
    at ResolutionRequest.resolveDependency (C:\source\demo\node_modules\metro\src\node-haste\DependencyGraph\ResolutionRequest.js:49:18)
    at DependencyGraph.resolveDependency (C:\source\demo\node_modules\metro\src\node-haste\DependencyGraph.js:218:16)
    at Object.resolve (C:\source\demo\node_modules\metro\src\lib\transformHelpers.js:141:30)
    at dependencies.map.result (C:\source\demo\node_modules\metro\src\DeltaBundler\traverseDependencies.js:373:31)
    at Array.map (<anonymous>)
    at resolveDependencies (C:\source\demo\node_modules\metro\src\DeltaBundler\traverseDependencies.js:369:18)
    at C:\source\demo\node_modules\metro\src\DeltaBundler\traverseDependencies.js:188:33
    at Generator.next (<anonymous>)
    at step (C:\source\demo\node_modules\metro\src\DeltaBundler\traverseDependencies.js:298:30)
::1 - - [08/May/2019:19:49:18 +0000] "GET /src/native/index.delta?platform=windows&dev=true&hot=false&deltaBundleId=9d9ba1a079a07dae HTTP/1.1" 500 - "http://localhost:8081/debugger-ui/" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/73.0.3683.103 Safari/537.36"
 DELTA  [windows, dev] src\native/index.js ░░░░░░░░░░░░░░░░ 0.0% (0/2), failed.
```
You can see it is trying to resolve: `..\__generated__\nativeIndexQuery.graphql` relative to the path: `C:\source\demo\src\native\index.tsx` and producing a path: `C:\source\__generated__\nativeIndexQuery.graphql`
You could likely repro this without relay at all, just using an import statement (or require) in a react native app that uses backslashes instead of forward slashes.

I imagine this could alternatively be fixed in metro (https://github.com/facebook/metro/issues/51 may even be the same issue) but on the other hand, I've never seen backslashes used in require calls or import statements written by hand, so I don't see why generated ones should be any different

packages/babel-plugin-relay/createModernNode.js
anikethsaha: Yeah..my mistake...yes this regex is fine.. 👍

PR #2724, opened 2019-05-08
Fix a typo in an error thrown by RelayParser

VinceOPS:
Hi there :open_hands: 

This is a super minor PR, but I found a typo while using gatsby, which relies on the RelayParser for its graphql queries... So I thought it could be nice to fix it!

Cheers


djD-REK:
> 
> 
> Hi there open_hands
> 
> This is a super minor PR, but I found a typo while using gatsby, which relies on the RelayParser for its graphql queries... So I thought it could be nice to fix it!
> 
> Cheers

Good catch! I was hunting down this typo myself, thanks for fixing it!

PR #2733, opened 2019-05-13
fix(QueryRenderer cache)

morrys:
dispose the pending request before retrieving the selectionReferences in order to creating a new instance of ReactRelayQueryFetcher.

### DIFF
--- a/packages/react-relay/__tests__/ReactRelayQueryRenderer-test.js
+++ b/packages/react-relay/__tests__/ReactRelayQueryRenderer-test.js
@@ -1602,6 +1602,68 @@ describe('ReactRelayQueryRenderer', () => {
 
       expect(render.mock.calls).toHaveLength(1);
     });
+    it('check GC with single pending request', () => {
+
+      const fetch = async () => response;
+      store = new Store(new RecordSource());
+      environment = new Environment({
+        network: Network.create(fetch),
+        store,
+      });
+
+      const renderer = ReactTestRenderer.create(
+          <ReactRelayQueryRenderer
+            environment={environment}
+            query={TestQuery}
+            render={render}
+            variables={variables}
+          />,
+      );
+
+      expect(store._roots.size).toEqual(1);
+      renderer.unmount();
+      expect(store._roots.size).toEqual(0);
+    });
+    it('check GC with multiple pending requests', () => {
+
+      const fetch = async () => response;
+      store = new Store(new RecordSource());
+      environment = new Environment({
+        network: Network.create(fetch),
+        store,
+      });
+
+      const renderer = ReactTestRenderer.create(
+        <PropsSetter>
+          <ReactRelayQueryRenderer
+            environment={environment}
+            query={TestQuery}
+            render={render}
+            variables={variables}
+          />
+        </PropsSetter>,
+      );
+
+      renderer.getInstance().setProps({
+        cacheConfig,
+        environment,
+        query: TestQuery,
+        render,
+        variables: { id: '5' },
+      });
+
+      renderer.getInstance().setProps({
+        cacheConfig,
+        environment,
+        query: TestQuery,
+        render,
+        variables: { id: '6' },
+      });
+
+      expect(store._roots.size).toEqual(3);
+      renderer.unmount();
+      expect(store._roots.size).toEqual(0);
+    });
   });
 
   describe('When retry', () => {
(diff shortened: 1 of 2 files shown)


morrys:
Hi,
in order to better analyze the fix I added some details of why I made that change.
The disposeRequest() invokes the dispose of the pending request which adds new functions to the variable selectionReferences (request.unsubscribe()).
That's why it's necessary to call the disposeRequest before retrieving the selectionReferences in order to create a new instance of ReactRelayQueryFetcher.

I hope this helps

jstejada:
Hey @morrys, before we can land this, could you add some regression tests that illustrate the broken behavior and how this change fixes it?

morrys:
hi jstejada, you're absolutely right. This week I hope to be able to realize the test case that shows the error and that it doesn't happen in the modified version.

packages/react-relay/__tests__/ReactRelayQueryRenderer-test.js
```diff
@@ -1602,6 +1602,68 @@ describe('ReactRelayQueryRenderer', () => {
 
       expect(render.mock.calls).toHaveLength(1);
     });
+    it('check GC with single pending request', () => {
```
jstejada: what exactly are we testing here? that GC runs after the component is unmounted?

packages/react-relay/__tests__/ReactRelayQueryRenderer-test.js
```diff
@@ -1602,6 +1602,68 @@ describe('ReactRelayQueryRenderer', () => {
...
+            render={render}
+            variables={variables}
+          />,
+      );
+
+      expect(store._roots.size).toEqual(1);
+      renderer.unmount();
+      expect(store._roots.size).toEqual(0);
+    });
+    it('check GC with multiple pending requests', () => {
```
jstejada: same question here

morrys:
`Why exactly do we want to create a new ReactRelayQueryFetcher.`
I think the main motivation is security to have a clean fetcher query (eg. OnDataChange).

`I think the original idea in the code was precisely to access the selection references before calling dispose, so we had access to them before they were cleared`

I think the mistake is just that. The method invoked is disposeRequest (and did not dispose).
the "disposeRequest" does not delete the selection references (
is the dispose that erases them).
```
disposeRequest () {
    this._error = null;
    this._snapshot = null;

    //order is important, dispose of pendingFetch before selectionReferences
    if (this._pendingRequest) {
      this._pendingRequest.dispose ();
    }
    if (this._rootSubscription) {
      this._rootSubscription.dispose ();
      this._rootSubscription = null;
    }
}
```
as we can see from the comment `order is important, dispose of pendingFetch before selectionReferences, this is important because the dispose  of pendingRequest adds a new reference in selection references.


`Otherwise prevSelectionReferences would always be empty right?`
No, with this change we make sure that requests that are still pending are disposed and taken to the next fetcher query.

 `And what we want to do is carry them over to the next query fetcher, right?`
Exactly, in this way when the query renderer is unmounted, all the requests made will be correctly disposed.

packages/react-relay/__tests__/ReactRelayQueryRenderer-test.js
morrys: When the queryrenderer is unmounted it performs the "dispose" of the fetcher query and then all the references are disposed. The "dispose" of the references foresee two behaviors:
* deletion of the root from the store
* schedule the GC

The GC provides for the cancellation of all entries from the record source associated with the roots no longer present.
So we're testing the actual deletion of all the roots created during the queryrenderer life cycle.

packages/react-relay/__tests__/ReactRelayQueryRenderer-test.js
morrys: same answer here but in a context where there are more references selected to be dispose. (this test shows how the previous version of react-relay did not delete all the referenced roots in the query renderer lifecycle)

jstejada:
@morrys, the new tests are failing internally, since they're accessing a private property and that's disallowed in our internal infra (by a babel transform).

I can try to get the tests to work, but we were thinking it might not be worth trying to merge this pr since:

- It's not entirely clear to us why this change is necessary, or what exactly this is fixing. Is there some product behavior that is incorrect right now? Do the new tests fail without this change? (apologies if we've missed this)
- Even though there's new tests, it's a risky change and hard to guarantee that this won't break some product behavior internally (which has happened in the past when changing this class).
- We're planning to change this entire implementation really soon, in favor of our updated infra inside relay-experimental, so I think we'd prefer doing that instead of landing this small change which we don't fully understand the effects of.

PR #2734, opened 2019-05-15
Fix grammar in the spec

cgriego:


cgriego:
Anything else I need to do?

josephsavona:
Looks good, thanks! We’ll import soon.

kassens:
Sorry for taking so long, importing now!

Issue #2738, opened 2019-05-17
Feature: Standardize the type of fetch policy


morrys:
ReactRelayQueryRenderer uses this as a type of fetchPolicy:

```
const NETWORK_ONLY = 'NETWORK_ONLY';
const STORE_THEN_NETWORK = 'STORE_THEN_NETWORK';
const DataFromEnum = {
  NETWORK_ONLY,
  STORE_THEN_NETWORK,
};
type DataFrom = $Keys<typeof DataFromEnum>;
```
while

ReactRelayRefetchContainer uses this as a type of fetchPolicy (in ReactRelayTypes.js):

```
export type RefetchOptions = {|
  +force?: boolean,
  +fetchPolicy?: 'store-or-network' | 'network-only',
|};
```

It would be more convenient to also have DataFrom in ReactRelayTypes and change RefetchOptions like this:

```
export type RefetchOptions = {|
  +force?: boolean,
  +fetchPolicy?: DataFrom,
|};
```

PR #2739, opened 2019-05-17
Feature: centralizing the logic of the "lookup in the store".

morrys:
Simplification of the logics present in the relative components.

Perfect when combined with this issue #2738


alloy:
This should probably come with some test coverage.

morrys:
I have seen that the ReactRelayQueryRenderer already has tests with the dataFrom property set to "STORE_THEN_NETWORK". Do you say to add tests for the ReactRelayRefetchContainer with the fetchPolicy "store-or-network"?

alloy:
I see. Yeah then perhaps not, although I guess some extra coverage isn’t a problem either.

PR #2741, opened 2019-05-22
Fix section header in Pagination Container docs

etcinit:
The `hasMore` header had an extra `()` appended at the end, unlike other functions in the same section.


kassens:
Thank you for the polish!

Issue #2518, opened 2018-08-21
[RFC] Allow for more complex configuration of relay-compiler / babel plugin


alloy:
While I’m not a fan of lots of configuration –and thus think that where possible we should have generally applicable defaults, such as with persisted queries– I’m starting to feel like relay-compiler + the babel plugin could do with a more advanced way of configuration, in the form of an evaluated JS config module:

* Being able to specify custom scalar mappings (#2162)
* Specifying custom persisted query logic (
https://github.com/facebook/relay/pull/2354#issuecomment-36927167, 0https://github.com/facebook/relay/pull/2354#issuecomment-395483449, and #1846)
* Currently the path to the single artefact directory needs to be passed to _both_ the CLI and the babel plugin. It would be great if this could be specified in one place instead. (https://github.com/relay-tools/relay-compiler-language-typescript#configuration)

```js
export default {
  schema: "./data/schema.json",
  language: "typescript",
  artifactDirectory: "./src/__relay_artifacts__",
  customScalars: { URL: "String" },
  persistedQueries: (…) => { … },
}
```

alloy:
cc @jstejada @kassens

jstejada:
cc @alunyov

sibelius:
consuming a user babel config in js is pretty easy

you can just import like a normal `.js` file

bencallaway:
@alloy Is the "language" option already supported? If not how would that work? Should it compile on the fly?

alloy:
@bencallaway It is. See https://github.com/relay-tools/relay-compiler-language-typescript#relay-compiler-language-typescript

bencallaway:
@alloy I have been using the `relay-compiler-language-typesecript` plugin, and very thankful for your work there btw. I was curious about the `babel-plugin-relay` config itself. As of `1.7.0-rc.1` it appears "language" is not a plugin option (yet? <- Partly my previous question). However, after looking back at the code I now see there's a `buildCommand` option that I had failed to notice before (https://github.com/facebook/relay/blob/v1.7.0-rc.1/packages/babel-plugin-relay/BabelPluginRelay.js). It would seem I can pass in my own `relay-compiler --language typescript` and the compilation will occur on the fly. (I am currently doing this with a separate `--watch` process while developing.

bencallaway:
Ok, retracting some of that last comment. It appears that `buildCommand` does not handle compilation on the fly, it's just used for error reporting. I'll need to spend some more time looking through the plugin's source.

[...]

PR #2745, opened 2019-05-27
Expose custom scalars configuration from CLI.

### LINKED ISSUE #2162: relay-compiler: config of customScalars for accurate flow types
I have a custom scalar on my backend called `DataUrl` which is just a string which has been validated to be a `data:` url.

In https://github.com/facebook/relay/blob/master/packages/relay-compiler/bin/RelayCompilerBin.js#L177, `customScalars` is passed as an empty object, but could be configuration-based.

It is used by https://github.com/facebook/relay/blob/master/packages/relay-compiler/core/RelayFlowTypeTransformers.js#L74. There appears to be special support for the nonstandard `Url` scalar type already.

alloy:
Closes #2162.

In preparation of #2518, as requested by @josephsavona we should have CLI versions of each possible argument, hereby CLI support for custom scalars.

I ended up going with the dot-notation syntax of yargs, which seems a little foreign (to me), but it comes builtin with yargs. Let me know what you think.

E.g.

```
$ relay-compiler --customScalars.URL=String --customScalars.Date=String
```

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -260,6 +260,7 @@ const modules = gulp.parallel(
               '!**/__tests__/**',
               '!**/__flowtests__/**',
               '!**/__mocks__/**',
+              '!**/node_modules/**',
             ],
             {
               cwd: path.join(PACKAGES, build.package),
--- a/packages/relay-compiler/bin/RelayCompilerBin.js
+++ b/packages/relay-compiler/bin/RelayCompilerBin.js
@@ -100,6 +100,14 @@ const argv = yargs
       type: 'string',
       default: null,
     },
+    customScalars: {
+      describe:
+        'Mappings from custom scalars in your schema to built-in GraphQL ' +
+        'types, for type emission purposes. (Uses yargs dot-notation, e.g. ' +
+        '--customScalars.URL=String)',
+      type: 'object',
+      default: {},
+    },
   })
   .help().argv;
 
--- a/packages/relay-compiler/bin/RelayCompilerMain.js
+++ b/packages/relay-compiler/bin/RelayCompilerMain.js
@@ -49,6 +49,23 @@ import type {
   PluginInterface,
 } from '../language/RelayLanguagePluginInterface';
 
+type Options = {|
+  schema: string,
+  src: string,
+  extensions: Array<string>,
+  include: Array<string>,
+  exclude: Array<string>,
+  verbose: boolean,
+  watchman: boolean,
+  watch?: ?boolean,
+  validate: boolean,
+  quiet: boolean,
+  persistOutput: ?string,
+  noFutureProofEnums: boolean,
+  language: string,
+  artifactDirectory: ?string,
+|};
+
 function buildWatchExpression(options: {
   extensions: Array<string>,
   include: Array<string>,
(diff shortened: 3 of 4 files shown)


packages/relay-compiler/bin/RelayCompilerBin.js
```diff
@@ -100,6 +100,14 @@ const argv = yargs
       type: 'string',
       default: null,
     },
+    customScalars: {
```
alloy: Re case, see https://github.com/facebook/relay/issues/2518#issuecomment-496176977

packages/relay-compiler/bin/RelayCompilerMain.js
```diff
@@ -174,86 +176,107 @@ Root files can include:
...
+    ? await codegenRunner.watchAll()
+    : await codegenRunner.compileAll();
+
+  if (result === 'ERROR') {
+    process.exit(100);
+  }
+  if (options.validate && result !== 'NO_CHANGES') {
+    process.exit(101);
+  }
+}
```
alloy: Most of these changes are to refactor configuration of the codegen runner out into a standalone function that can be tested.

gulpfile.js
```diff
@@ -260,6 +260,7 @@ const modules = gulp.parallel(
               '!**/__tests__/**',
               '!**/__flowtests__/**',
               '!**/__mocks__/**',
+              '!**/node_modules/**',
```
alloy: I lost a bunch of time faffing around trying to find the best way to do continuous development on the package while integrated into one of our apps, which led me to at some point running `yarn install` in the `packages/relay-compiler` dir. After that building failed in confusing ways, so this addition is really just to guard other devs from the same silly mistake.

alloy:
Note that I had to do some synthetic tests, as none of our schema’s custom scalars really maps to a single builtin GraphQL scalar type, but doing so did update the [typescript] artifacts in our app as expected.

For instance, we have a `FormattedNumber` scalar, which holds either a string or a number, but afaict the existing custom scalar support does not allow something like `{ "FormattedNumber": "String | Int" }`. Is that correct?

TrySound:
Awesome! I want it so much. Currently to make date/time scalars type safe I have to copy/paste RelayCompilerBin and RelayCompilerMain into my project.

packages/relay-compiler/bin/RelayCompilerBin.js
```diff
@@ -100,6 +100,14 @@ const argv = yargs
...
+      describe:
+        'Mappings from custom scalars in your schema to built-in GraphQL ' +
+        'types, for type emission purposes. (Uses yargs dot-notation, e.g. ' +
+        '--customScalars.URL=String)',
+      type: 'object',
```
jstejada: @alloy I'm getting flow errors for this type. Is it actually supported? I can't find it in the docs: http://yargs.js.org/docs/#api-optionkey-opt

jstejada: going to merge this in anyway and we can fix in a follow up

alloy: It doesn’t exist [yet](https://github.com/yargs/yargs/issues/890), in the sense that it checks the input type, but setting it here does make it show up in the help banner and so I think it’s helpful to include already.

I did not notice any flow errors at the time, but also the relay-config PR applies custom typings to these options, so perhaps that overrides any current failures here.

jstejada: got it! thanks! I'll just leave it with a $FlowFixMe for now then

PR #2746, opened 2019-05-28
Add relay-config package for centralised configuration

### LINKED ISSUE #2518: [RFC] Allow for more complex configuration of relay-compiler / babel plugin

alloy:
Closes #2518 

_Note: Includes #2745, so that should probably be merged first and then this can be rebased._

I tried multiple times to come up with a doc describing the integration a few times, but then yesterday I had a whole day and figured I’d just write _a_ implementation to show what it could look like. Please feel free to poke holes as much as you need.

----

What this change does:

* Add a new `relay-config` package

* Usage of the package is optional. If installed it will be picked-up, otherwise relay will function as it does today.

* RelayConfig relies on [cosmiconfig](https://github.com/davidtheclark/cosmiconfig) to do its bidding. It’s configured to load from:

  - a `relay` key in `package.json`

    ```json
    {
      "relay": {
        "src": "./src"
      }
    }
    ```

  - a `relay.config.json` file

    ```json
    {
      "src": "./src"
    }
    ```

  - or a `relay.config.js` file

    ```js
    module.exports = {
      src: "./src",
    }
    ```

* It accepts all the same configuration as the CLI does. (As a bonus, I typed the configuration object and the Yargs options object.)

* Additionally, when using the `relay.config.js` file, a configuration entry like the language plugin also accepts an actual function:

  ```js
  const typescript = require("relay-compiler-language-typescript")

  module.exports = {
    language: typescript,
  }
  ```

  In the future, other entries such as `persistedQueries` and `customScalars` could also be configured as such and allow for project specific setup.

* It is used by all pieces of Relay that take configuration–i.e. `relay-compiler` and `babel-plugin-relay`. For instance, previously one would have needed to specify `artifactDirectory` on the CLI to `relay-compiler` and the same setting in `.babelrc` for `babel-plugin-relay`. This can now be reduced to a single centralised setting.

  Additional external tooling can also start relying on this. For instance, the `vscode-apollo` extension, with [this change](https://github.com/apollographql/apollo-tooling/pull/1288), will allow for custom GraphQL validation rules to surface in the UI. Leveraging the centralised config allows one to surface `relay-compiler` specific validations without needing to take in the same configuration in a different place yet again.

* Finally, this also adds some previously missing test coverage around language plugin loading.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -260,6 +260,7 @@ const modules = gulp.parallel(
               '!**/__tests__/**',
               '!**/__flowtests__/**',
               '!**/__mocks__/**',
+              '!**/node_modules/**',
             ],
             {
               cwd: path.join(PACKAGES, build.package),
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -54,6 +54,7 @@ const babelOptions = require('./scripts/getBabelOptions')({
     'babel-plugin-macros': 'babel-plugin-macros',
     chalk: 'chalk',
     child_process: 'child_process',
+    cosmiconfig: 'cosmiconfig',
     crypto: 'crypto',
     'fast-glob': 'fast-glob',
     'fb-watchman': 'fb-watchman',
@@ -238,6 +239,21 @@ const builds = [
         entry: 'index.js',
         output: 'relay-test-utils',
         libraryName: 'RelayTestUtils',
+        target: 'node',
+        noMinify: true, // Note: uglify can't yet handle modern JS
+      },
+    ],
+  },
+  {
+    package: 'relay-config',
+    exports: {
+      index: 'index.js',
+    },
+    bundles: [
+      {
+        entry: 'index.js',
+        output: 'relay-config',
+        libraryName: 'RelayConfig',
         target: 'node',
         noMinify: true, // Note: uglify can't yet handle modern JS
       },
--- a/packages/babel-plugin-relay/BabelPluginRelay.js
+++ b/packages/babel-plugin-relay/BabelPluginRelay.js
@@ -17,6 +17,12 @@ const getValidGraphQLTag = require('./getValidGraphQLTag');
 const getValidRelayQLTag = require('./getValidRelayQLTag');
 const invariant = require('./invariant');
 
+let RelayConfig;
+try {
+  // eslint-disable-next-line no-eval
+  RelayConfig = eval('require')('relay-config');
+} catch (_) {}
+
 import type {Validator} from './RelayQLTransformer';
 
 export type RelayPluginOptions = {
@@ -38,7 +44,7 @@ export type RelayPluginOptions = {
   snakeCase?: boolean,
   substituteVariables?: boolean,
   validator?: Validator<any>,
-  // Directory as specified by outputDir when running relay-compiler
+  // Directory as specified by artifactDirectory when running relay-compiler
   artifactDirectory?: string,
 };
 
@@ -115,7 +121,8 @@ module.exports = function BabelPluginRelay(context: {types: $FlowFixMe}): any {
   return {
     visitor: {
       Program(path, state) {
-        path.traverse(visitor, state);
+        const config = RelayConfig && RelayConfig.loadConfig();
+        path.traverse(visitor, {...state, opts: {...config, ...state.opts}});
       },
     },
   };
--- a/packages/relay-compiler/bin/RelayCompilerBin.js
+++ b/packages/relay-compiler/bin/RelayCompilerBin.js
@@ -100,6 +100,14 @@ const argv = yargs
       type: 'string',
       default: null,
     },
+    customScalars: {
+      describe:
+        'Mappings from custom scalars in your schema to built-in GraphQL ' +
+        'types, for type emission purposes. (Uses yargs dot-notation, e.g. ' +
+        '--customScalars.URL=String)',
+      type: 'object',
+      default: {},
+    },
   })
   .help().argv;
 
--- a/packages/relay-compiler/bin/RelayCompilerBin.js
+++ b/packages/relay-compiler/bin/RelayCompilerBin.js
@@ -15,101 +15,156 @@ require('@babel/polyfill');
 const {main} = require('./RelayCompilerMain');
 const yargs = require('yargs');
 
+import type {Config} from './RelayCompilerMain';
+
+type OptionBase<T> = {|
+  describe: string,
+  default?: T,
+|};
+
+/**
+ * Maps from a value type to a corresponding yargs option object type.
+ *
+ * For instance, a required array string type maps to:
+ *
+ *     { describe: string, default?: Array<string>, type: 'string', array: true, demandOption: true }
+ *
+ * Whereas an optional single string type maps to:
+ *
+ *     { describe: string, default?: string, type: 'string', array: false, demandOption: false }
+ */
+type OptionTypeMap<T> = $Call<
+  | ((
+      Array<string>,
+    ) => {|...OptionBase<Array<string>>, array: true, type: 'string'|})
+  | ((
+      ?string,
+    ) => {|
+      ...OptionBase<string>,
+      array: false,
+      demandOption: false,
+      type: 'string',
+    |})
+  | (string => {|
+      ...OptionBase<string>,
+      array: false,
+      demandOption: true,
+      type: 'string',
+    |})
+  | ((?boolean) => {|...OptionBase<boolean>, type: 'boolean'|})
+  | (({}) => {|...OptionBase<{}>, type: 'object'|}),
+  T,
+>;
+
+/**
+ * Generates an object type with all keys from Config and values mapped to yargs option object types.
+ */
+type Options = $ObjMap<Config, <T>(T) => OptionTypeMap<T>>;
+
+const options: Options = {
+  schema: {
+    describe: 'Path to schema.graphql or schema.json',
+    demandOption: true,
+    type: 'string',
+    array: false,
+  },
+  src: {
+    describe: 'Root directory of application code',
+    demandOption: true,
+    type: 'string',
+    array: false,
+  },
+  include: {
+    describe: 'Directories to include under src',
+    type: 'string',
+    array: true,
+    default: ['**'],
+  },
+  exclude: {
+    describe: 'Directories to ignore under src',
+    type: 'string',
+    array: true,
+    default: ['**/node_modules/**', '**/__mocks__/**', '**/__generated__/**'],
+  },
+  extensions: {
+    array: true,
+    describe:
+      'File extensions to compile (defaults to extensions provided by the ' +
+      'language plugin)',
+    type: 'string',
+  },
+  verbose: {
+    describe: 'More verbose logging',
+    type: 'boolean',
+  },
+  quiet: {
+    describe: 'No output to stdout',
+    type: 'boolean',
+  },
+  watchman: {
+    describe: 'Use watchman when not in watch mode',
+    type: 'boolean',
+    default: true,
+  },
+  watch: {
+    describe: 'If specified, watches files and regenerates on changes',
+    type: 'boolean',
+  },
+  validate: {
+    describe:
+      'Looks for pending changes and exits with non-zero code instead of ' +
+      'writing to disk',
+    type: 'boolean',
+    default: false,
+  },
+  persistOutput: {
+    describe:
+      'A path to a .json file where persisted query metadata should be saved',
+    demandOption: false,
+    type: 'string',
+    array: false,
+  },
+  noFutureProofEnums: {
+    describe:
+      'This option controls whether or not a catch-all entry is added to enum type definitions ' +
+      'for values that may be added in the future. Enabling this means you will have to update ' +
+      'your application whenever the GraphQL server schema adds new enum values to prevent it ' +
+      'from breaking.',
+    type: 'boolean',
+    default: false,
+  },
+  language: {
+    describe:
+      'The name of the language plugin used for input files and artifacts',
+    demandOption: false,
+    type: 'string',
+    array: false,
+    default: 'javascript',
+  },
+  artifactDirectory: {
+    describe:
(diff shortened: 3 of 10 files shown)


alloy:
For our apps, this:

```bash
$ relay-compiler \
  --src ./src \
  --schema data/schema.graphql \
  --language typescript \
  --artifactDirectory ./src/__generated__ \
  --persist-output ./data/complete.queryMap.json \
  --exclude '**/node_modules/**,**/__mocks__/**,**/__generated__/**'
```

```json
{
  "plugins": [
    ["relay", { "artifactDirectory": "./src/__generated__" }]
  ]
}
```

…is replaced with:

```js
module.exports = {
  src: "./src",
  schema: "./data/schema.graphql",
  language: "typescript",
  artifactDirectory: "./src/__generated__",
  persistOutput: "./data/complete.queryMap.json",
  exclude: ["**/node_modules/**", "**/__mocks__/**", "**/__generated__/**"],
}
```

```json
{
  "plugins": ["relay"]
}
```

packages/relay-compiler/bin/RelayCompilerMain.js
```diff
@@ -97,74 +116,67 @@ type LanguagePlugin = PluginInitializer | {default: PluginInitializer};
...
-      ? pluginPath
-      : `relay-compiler-language-${language}`;
-    try {
-      // eslint-disable-next-line no-eval
-      let languagePlugin: LanguagePlugin = eval('require')(requirePath);
-      if (languagePlugin.default) {
-        languagePlugin = languagePlugin.default;
+    let languagePlugin: LanguagePlugin;
+    if (typeof language === 'string') {
+      const pluginPath = path.resolve(process.cwd(), language);
```
alloy: This module loading from a path could probably be deprecated with the ability to just pass a function using a `relay.config.js` file.

zth:
This is so great @alloy! Centralized configuration, yummies!

gulpfile.js
```diff
@@ -243,6 +244,21 @@ const builds = [
...
+    bundles: [
+      {
+        entry: 'index.js',
+        output: 'relay-config',
+        libraryName: 'RelayConfig',
+        target: 'node',
+        noMinify: true, // Note: uglify can't yet handle modern JS
+      },
+    ],
+  },
```
alloy: (Cargo-culted this from the test utils package.)

packages/relay-compiler/bin/__tests__/RelayCompilerMain-test.js
```diff
@@ -10,44 +10,144 @@
...
+        expect(RelayFileWriter.writeAll).toHaveBeenCalledWith(
+          expect.objectContaining({
+            config: expect.objectContaining({customScalars: {URL: 'String'}}),
+          }),
+        );
+      });
+    });
+  });
+
+  describe('concerning language plugins', () => {
```
josephsavona: `describe('getLanguagePlugin()', ...)` since that's what this is testing?

packages/relay-compiler/bin/__tests__/RelayCompilerMain-test.js
```diff
@@ -10,44 +10,144 @@
...
-      new Error('--persist-output path does not exist: /nonexisting/output'),
+      new Error(
+        `--persistOutput path does not exist: ${path.resolve(
+          'non-existent/output',
+        )}`,
+      ),
     );
   });
+
+  describe('concerning the codegen runner', () => {
```
josephsavona: `describe('getCodegenRunner', ...)`

packages/relay-compiler/bin/__tests__/RelayCompilerMain-test.js
```diff
@@ -10,44 +10,144 @@
...
+    const options = {
+      schema: testSchemaPath,
+      language: 'javascript',
+      include: [],
+      exclude: [],
+      src: '.',
+    };
+
+    describe('and its writer configurations', () => {
```
josephsavona: seems like an unnecessary level of nesting

packages/babel-plugin-relay/BabelPluginRelay.js
```diff
@@ -17,6 +17,12 @@ const getValidGraphQLTag = require('./getValidGraphQLTag');
 const getValidRelayQLTag = require('./getValidRelayQLTag');
 const invariant = require('./invariant');
 
+let RelayConfig;
+try {
+  // eslint-disable-next-line no-eval
+  RelayConfig = eval('require')('relay-config');
```
josephsavona: i know that we were already using it, but why the `eval('require')('foo')` pattern? is this because the bundler can't find it or something?

packages/babel-plugin-relay/BabelPluginRelay.js
```diff
@@ -17,6 +17,12 @@ const getValidGraphQLTag = require('./getValidGraphQLTag');
...
+} catch (_) {}
```
josephsavona: hmm - why suppress this error?

packages/relay-compiler/bin/RelayCompilerBin.js
```diff
@@ -15,96 +15,175 @@ require('@babel/polyfill');
...
-        'A specific directory to output all artifacts to. When enabling this ' +
-        'the babel plugin needs `artifactDirectory` set as well.',
-      type: 'string',
-      default: null,
-    },
-  })
+  .options(options)
+  // Apply externally loaded config through the yargs API so that we can leverage yargs' defaults and have them show up
+  // in the help banner.
+  .config(config)
```
josephsavona: so this is just populating defaults,  and command line args will override them?

packages/relay-compiler/bin/RelayCompilerMain.js
josephsavona: yeah, let's do that. we can deprecate this in a follow-up and  delete in a later release

packages/relay-compiler/bin/RelayCompilerMain.js
```diff
@@ -174,86 +186,105 @@ Root files can include:
...
   }
-  if (options.verbose && options.quiet) {
+  if (config.verbose && config.quiet) {
     throw new Error("I can't be quiet and verbose at the same time");
   }
 
-  const reporter = new ConsoleReporter({
-    verbose: options.verbose,
-    quiet: options.quiet,
+  const watchman = config.watchman && (await WatchmanClient.isAvailable());
```
josephsavona: hmm the previous name made it clear that this wasn't identical to `config.watchman` -  the logic has actually changed, see comment below

packages/relay-compiler/bin/RelayCompilerMain.js
```diff
@@ -174,86 +186,105 @@ Root files can include:
...
-
   const parserConfigs = {
     [sourceParserName]: {
-      baseDir: srcDir,
+      baseDir: config.src,
       getFileFilter: sourceModuleParser.getFileFilter,
       getParser: sourceModuleParser.getParser,
       getSchema: () => schema,
-      watchmanExpression: useWatchman
+      watchmanExpression: config.watchman
```
josephsavona: revert? we should only pass a watchman expression  if watchman is available, not just if config.watchman is  set.

packages/relay-compiler/bin/RelayCompilerMain.js
```diff
@@ -174,86 +186,105 @@ Root files can include:
...
         ? buildWatchExpression(sourceSearchOptions)
         : null,
-      filepaths: useWatchman
+      filepaths: config.watchman
```
josephsavona: same here

packages/relay-compiler/bin/RelayCompilerMain.js
```diff
@@ -174,86 +186,105 @@ Root files can include:
...
-        : getFilepathsFromGlob(srcDir, sourceSearchOptions),
+        : getFilepathsFromGlob(config.src, sourceSearchOptions),
     },
     graphql: {
-      baseDir: srcDir,
+      baseDir: config.src,
       getParser: DotGraphQLParser.getParser,
       getSchema: () => schema,
-      watchmanExpression: useWatchman
+      watchmanExpression: config.watchman
```
josephsavona: same here and below

packages/babel-plugin-relay/BabelPluginRelay.js
sibelius: I think this is to avoid webpack messing `require` calls

josephsavona:
One more thought: should we *only* support `relay.config.js`, or at least heavily encourage it in the docs? Several useful features like persisted queries only work with a function, and it seems good to encourage people to use a config format that they won't  have to change later when they realize they want some feature.

packages/relay-compiler/bin/__tests__/RelayCompilerMain-test.js
alloy: Ah my bad, I should have followed the patterns in other test files 👍 (I personally most of the times don’t use just symbols as descriptions.)

packages/babel-plugin-relay/BabelPluginRelay.js
alloy: That’s half of it, yeah.

1. For the single webpack bundle, to avoid webpack from failing to build (as it can’t find user project files at build time), we can’t refer to the `require` symbol. Normally with web pack you could use [`__webpack_require__`](https://webpack.js.org/api/module-variables#__webpack_require__-webpack-specific) instead…
1.  However, as we also build commonjs modules, we can’t use a webpack specific API and have to use plain `require`.

This seemed to be a pattern people use to avoid that situation.

packages/babel-plugin-relay/BabelPluginRelay.js
alloy: Because we still want the plugin to be able to load its configuration from `.babelrc` if the user doesn’t use a single configuration file.

alloy:
I’m up for that 👍 I’ll follow up with a doc PR first.

packages/relay-compiler/bin/RelayCompilerBin.js
alloy: The order of overrides is:

1. Defaults defined in `options` above
1. `config` from `relay-config`
1. CLI args

packages/relay-compiler/bin/RelayCompilerMain.js
```diff
@@ -97,163 +116,183 @@ type LanguagePlugin = PluginInitializer | {default: PluginInitializer};
...
+    throw new Error("I can't be quiet and verbose at the same time");
   }
-  const srcDir = path.resolve(process.cwd(), options.src);
-  if (!fs.existsSync(srcDir)) {
-    throw new Error(`--src path does not exist: ${srcDir}`);
+
+  config = getPathBasedConfig(config);
+  config = await getWatchConfig(config);
+
+  const codegenRunner = module.exports.getCodegenRunner(config);
```
alloy: Other than adding optional function parameters to these functions to allow for DI, this appears to be _a_ suggestion to being able to mock functions that reside in the same module as the caller https://github.com/facebook/jest/issues/936#issuecomment-214939935.

I have no real preference for one or the other, let me know if you do.

alloy:
Btw here’s a PoC of how having this unified config interface will allow other tooling to hook in better with one’s relay setup–in this case we can have (and share with the community) a generic vscode-apollo config: https://github.com/artsy/emission/blob/b7f138d2b4311d0eed5923b39219e124a6a96a65/apollo.config.js

_(I would have loved to include validation rules about eg operation/fragment names having to start with the name of the module they reside in, but I just noticed those aren’t proper validation rules, but [operate directly on the parsed GraphQL tags](https://github.com/facebook/relay/blob/4de953d1322ea1049e555ced864e471a67f43535/packages/relay-compiler/core/RelayFindGraphQLTags.js) 😞.)_

alloy:
One more thought on my end, if we were to only allow configuration through `relay.config.js` at some point, does that meant that perhaps we can move some of the configuration validation and expansion of eg globs could move into `relay-config` as well? That would allow other tooling relying on the configuration to get the same quality as `relay-compiler` and `babel-plugin-relay` would and alleviate e.g. [this](https://github.com/artsy/emission/blob/b7f138d2b4311d0eed5923b39219e124a6a96a65/apollo.config.js#L26).

alloy:
As noted [here](https://github.com/facebook/relay/commit/6ba5890127f5b3ad124691075335525060ad10bb#r33772388), the CI failures exist on `master` as well.

enisdenjo:
Maybe I am doing it wrong, but the config file is not picked up for me...

### What I did:

- Installed `relay-config` alongside all other Relay deps (all at `v5.0.0`)
- Created `relay.config.js` at the root of my project (same level as `package.json`)
_Having the config in `package.json` changes nothing..._
- `relay.config.js` has the same contents as @alloy's [comment](https://github.com/facebook/relay/pull/2746#issuecomment-496647836)
- Added a `relay` script in `package.json` which just does: `relay-compiler`
- Running the script results in: `Error: --schema path does not exist: [...]`

Any ideas? Thanks! :)

enisdenjo:
Never mind guys, I was searching for `schema.graphql` while I had `schema.json`... Silly me 😬

PR #2749, opened 2019-06-03
Improved DX for working on Relay as non-FB dev

alloy:
* Only clean build on install when publishing, not just any `yarn install`.
* Add a `yarn build:watch` task which leaves the `dist` dir in tact, which makes `yarn link` work and allows for easy integration into the dev’s project.
* Use incremental builds for `yarn build` and `yarn build:watch`.


alloy:
The one thing I’m unsure about is the change to not always build on `yarn install`. Perhaps it works best when using JIT compilation in one’s test setup 🤔

package.json
```diff
@@ -7,10 +7,13 @@
   "bugs": "https://github.com/facebook/relay/issues",
   "repository": "facebook/relay",
   "scripts": {
-    "build": "gulp",
+    "build": "gulp dist",
```
alloy: This will now perform incremental builds, so running `yarn install` after a `git pull` will be faster.

package.json
```diff
@@ -7,10 +7,13 @@
...
+    "build:clean": "gulp cleanbuild",
+    "build:watch": "gulp watch",
     "jest": "NODE_ENV=test jest \"$@\"",
     "lint": "eslint --max-warnings 0 .",
-    "prepublish": "node node_modules/fbjs-scripts/node/check-dev-engines.js package.json && npm run build",
+    "postinstall": "npm run build",
```
alloy: This triggers the build after `yarn install`.

package.json
```diff
@@ -7,10 +7,13 @@
...
+    "prepublishOnly": "node node_modules/fbjs-scripts/node/check-dev-engines.js package.json && npm run build:clean",
```
alloy: This only runs when actually invoking `npm publish` and will perform a clean before the build.

PR #2750, opened 2019-06-04
Fix flow on GitHub

kassens:
These fixmes also apply to OSS. The `emptyFunction` type isn't
compatible with flow 0.99, so this ignores this simple function that we
probably just shouldn't use in the first place.

### DIFF
--- a/packages/relay-compiler/bin/RelayCompilerMain.js
+++ b/packages/relay-compiler/bin/RelayCompilerMain.js
@@ -109,15 +109,11 @@ function getLanguagePlugin(language: string): PluginInterface {
       // eslint-disable-next-line no-eval
       let languagePlugin: LanguagePlugin = eval('require')(requirePath);
       if (languagePlugin.default) {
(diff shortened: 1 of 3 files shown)


packages/relay-compiler/bin/RelayCompilerMain.js
```diff
@@ -109,15 +109,11 @@ function getLanguagePlugin(language: string): PluginInterface {
...
-         * error found when Flow v0.99 was deployed. To see the error, delete
-         * this comment and run Flow. */
+        // $FlowFixMe - Flow no longer considers statics of functions as any
         languagePlugin = languagePlugin.default;
       }
       if (typeof languagePlugin === 'function') {
-        /* $     /* $FlowFixMe(>=0.99.0 site=react_native_fb) This comment suppresses an
-         * error found when Flow v0.99 was deployed. To see the error, delete
-         * this comment and run Flow. */
+        // $FlowFixMe
```
alloy: Thanks 👍

Issue #2762, opened 2019-06-12
Enable mocking value as null when using RelayMockPayloadGenerator


sibelius:
It is defaulting back to default scalar value when setting a field as null

```jsx
test('should return null if mockResolver has null on scalar field', () => {
    testGeneratedData(
      `
      query TestQuery @relay_test_operation {
        node(id: "my-id") {
          ... on User {
            id
            name
          }
        }
      }
    `,
      {
        User() {
          return {
            id: 'my-user-id',
            name: null,
          };
        },
      },
    );
  });
```

Mock Output
```
{
  "data": {
    "node": {
      "__typename": "User",
      "id": "my-user-id",
      "name": "<mock-value-for-field-\\"name\\">"
    }
  }
}
```

As I'm defining user { name }  as `null`, it should be `null` instead of default mock value for `String()`

alunyov:
@sibelius thank for working on this!

PR #2763, opened 2019-06-12
test(mock): add failing test for mocking null values, see #2762

sibelius:


alunyov:
Good catch @sibelius! Thank you for this fix

Issue #2759, opened 2019-06-06
relay-compiler has implicit dependency on lodash>=4 and core-js<3


wyattanderson:
Hi there! I've been working on upgrading our codebase to Relay 4.0.0 and ran into a few roadblocks. It seems that `relay-compiler` has an implicit (i.e. not specified in `package.json` and therefore not resolved by `npm install` when installing `relay-compiler`) dependency on `lodash>=4`, and that if your application explicitly pins `lodash<4` (for historical reasons, let's not dive into that), `relay-compiler` can't run at all because it can't import `lodash/clone`:

```
ERROR:
Error writing modules:
Error: Cannot find module 'lodash/clone'
    at Function.Module._resolveFilename (module.js:538:15)
    at Function.Module._load (module.js:468:25)
    at Module.require (module.js:587:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/Users/wyatt/git/relay-lodash-reproducer/node_modules/relay-compiler/bin/relay-compiler:24273:18)
    at __webpack_require__ (/Users/wyatt/git/relay-lodash-reproducer/node_modules/relay-compiler/bin/relay-compiler:30:30)
    at _clone (/Users/wyatt/git/relay-lodash-reproducer/node_modules/relay-compiler/bin/relay-compiler:24226:39)
    at keys.forEach.key (/Users/wyatt/git/relay-lodash-reproducer/node_modules/relay-compiler/bin/relay-compiler:24257:38)
    at Array.forEach (<anonymous>)
    at builder (/Users/wyatt/git/relay-lodash-reproducer/node_modules/relay-compiler/bin/relay-compiler:24253:8)
```

I've built a minimal reproducer at https://github.com/wyattanderson/relay-lodash-reproducer which you can trigger by cloning, running `npm install`, and then running `node_modules/.bin/relay-compiler  --extensions jsx --schema schema.graphql  --src src/`, which will crash with the above error.

And, there's a bonus! On the `core-js` branch in that repository, I've pinned `core-js@3.1.3` (the latest version), and the same steps reproduce the issue in #2701:

```
Error: Cannot find module 'core-js/es6'
    at Function.Module._resolveFilename (module.js:538:15)
    at Function.Module._load (module.js:468:25)
    at Module.require (module.js:587:17)
    at require (internal/module.js:11:18)
[...]

PR #2764, opened 2019-06-12
Don't include @babel packages in generated bundles

wyattanderson:
See background in #2759.

Before this change, the `relay-compiler` binary pulls in `lodash`:

```
dist/relay-compiler/bin/relay-compiler
23734:module.exports = require("lodash/clone");
26487:module.exports = require("lodash/uniq");
27035:module.exports = require("lodash/isPlainObject");
27041:module.exports = require("lodash/isRegExp");
28413:module.exports = require("lodash/isInteger");
28419:module.exports = require("lodash/repeat");
```

And, only these files in `dist/` reference `lodash`:

```
$ rg lodash dist/ -l
dist/relay-compiler/relay-compiler.js
dist/relay-compiler/relay-compiler.min.js
dist/relay-compiler/bin/relay-compiler
```

After this change, no files in `dist/` contain references to `lodash`. This seems consistent with what's contained in Relay packages before they were upgraded to Babel 7 (where the `@` was added) in #2536.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -118,7 +118,10 @@ const PRODUCTION_HEADER =
 
 const buildDist = function(filename, opts, isProduction) {
   const webpackOpts = {
-    externals: [/^[-/a-zA-Z0-9]+$/],
+    externals: [
+      /^[-/a-zA-Z0-9]+$/,
+      /^@babel\/.+$/,
+    ],
     target: opts.target,
     node: {
       fs: 'empty',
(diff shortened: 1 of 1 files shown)


kassens:
Thank you! I had trouble figuring out what was wrong with our config to not inline these requires!

wyattanderson:
Hey @kassens @josephsavona, are there any plans for a Relay release soon? Even a minor version bump including this fix would be great. It's been nearly 2 months since 5.0.0 was released.

PR #2766, opened 2019-06-13
Remove Haste configuration from .flowconfig

AlicanC:
Removed Haste stuff from .flowconfig and refactored the project accordingly so when Flow types are shipped on NPM (#2690) users won't have to configure Haste on their projects.


zth:
This is great!

AlicanC:
I thought I did it, but didn't realize the builds were broken. (I guess it was cached on my local.)

Trying to fix that right now.

AlicanC:
@alunyov I was working on this, but your latest commit 9dc19e3 broke the whole thing for me. (`relay-test-utils-internal` can not be found.)

alunyov:
@AlicanC sorry for that. I'm working on the fix now

jstejada:
cc @kassens

kassens:
The problem here is that internally, we use the source code as is without going through npm. Internally, we have the `warning` module etc, but no `fbjs/lib/warning`, so we need to require `warning` directly.

I should think of a workaround, some of the modules we could probably just get rid of or inline..

PR #2768, opened 2019-06-14
Include script to redirect facebook.github.io/relay to relay.dev

yangshun:
I've added `js/redirect.js` to `facebook/facebook.github.io` repo and modified Relay website to load a `/js/redirect.js` file. Loading this JS file on:

- When someone goes to `facebook.github.io/relay` they will be redirected to `relay.dev`
- When someone goes to `relay.dev` will see a `js/redirect.js` 404 loading error because there's no real `/js/redirect.js` there. I could add a stub file in relay/website if you don't want to see the 404 in the console. Lemme know if you want that.

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -127,6 +127,7 @@ const siteConfig = {
   enableUpdateTime: true,
   enableUpdateBy: true,
   docsSideNavCollapsible: true,
+  scripts: ['/js/redirect.js'],
 };
 
 module.exports = siteConfig;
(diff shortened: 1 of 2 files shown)


kassens:
I think you forgot to `git add js/redirect.js`. I don't see that file.

yangshun:
@kassens I explained on Workplace:

I added the actual script in https://facebook.github.io/js/redirect.js. The script doesn't need to be in the relay website repo.

When a user goes to https://facebook.github.io/relay/, the page will try to load https://facebook.github.io/js/redirect.js (a file outside the Relay website!) which then redirects you properly. I have tested it on the few URLs which gave you errors and the redirection worked fine.

I have since added a stub file to relay/website so that you don't see 404 when you load the page on relay.dev

PR #2779, opened 2019-06-19
fix(mock): fix RelayMockPayloadGenerator for defaultValues as array o…

### LINKED ISSUE #2778: can't generate mock for array of enums using array as a default value
- It should be possible to generate a defaultValue for array of enum

```jsx
test('generate mock for enum in arrays', () => {
    testGeneratedData(
      `
        query TestQuery @relay_test_operation {
          node(id: "my-id") {
            ... on User {
              id
              traits
            }
          }
        }
      `,
      {
        User(context) {
          return {
            traits: ['CHEERFUL', 'DERISIVE'],
          };
        },
      },
    );
  });
```

It is throwing the following error:

```
 Invariant Violation: RelayMockPayloadGenerator: Invalid value "CHEERFUL,DERISIVE" provided for enum field "node.traits" via MockResolver.Expected one of the following values: "CHEERFUL", "DERISIVE", "HELPFUL", "SNARKY".

      400 |           .includes(String(value).toUpperCase())
      401 |       ) {
    > 402 |         invariant(
          |                   ^
      403 |           false,
      404 |           'RelayMockPayloadGenerator: Invalid value "%s" provided for enum ' +
      405 |             'field "%s" via MockResolver.' +
```

sibelius:
…n enum fields, fix #2778

### DIFF
--- a/packages/relay-test-utils/RelayMockPayloadGenerator.js
+++ b/packages/relay-test-utils/RelayMockPayloadGenerator.js
@@ -389,8 +389,7 @@ class RelayMockPayloadGenerator {
     // But missing case should be acceptable, we will just use
     // a correct spelling from enumValues
     const correctSpellingValues = valueToValidate.map(v => {
-      const correctSpellingEnumIndex = enumValues
-        .map(s => s.toUpperCase())
+      const correctSpellingEnumIndex = enumValuesNormalized
         .indexOf(String(v).toUpperCase());
 
       return enumValues[correctSpellingEnumIndex];
--- a/packages/relay-test-utils/RelayMockPayloadGenerator.js
+++ b/packages/relay-test-utils/RelayMockPayloadGenerator.js
@@ -391,32 +391,60 @@ class RelayMockPayloadGenerator {
     ) {
       value = defaultValues[applicationName];
 
-      // Let's validate the correctness of the provided enum value
-      // We will throw if value provided by mock resolvers is invalid
-      if (
-        enumValues !== null &&
-        !enumValues
-          .map(s => s.toUpperCase())
-          .includes(String(value).toUpperCase())
-      ) {
-        invariant(
-          false,
-          'RelayMockPayloadGenerator: Invalid value "%s" provided for enum ' +
-            'field "%s" via MockResolver.' +
-            'Expected one of the following values: %s.',
-          String(value),
-          `${path.join('.')}.${applicationName}`,
-          enumValues.map(v => `"${v}"`).join(', '),
-        );
+      if (Array.isArray(value)) {
+        const enumValuesNormalized = enumValues.map(s => s.toUpperCase());
+
+        const correctValues = value.filter(v => enumValuesNormalized.includes(v));
+
+        if (correctValues.length !== value.length) {
+          invariant(
+            false,
+            'RelayMockPayloadGenerator: Invalid value "%s" provided for enum ' +
+              'field "%s" via MockResolver.' +
+              'Expected one of the following values: %s.',
+            value,
+            `${path.join('.')}.${applicationName}`,
+            enumValues.map(v => `"${v}"`).join(', '),
+          );
+        }
+      } else {
+        // Let's validate the correctness of the provided enum value
+        // We will throw if value provided by mock resolvers is invalid
+        if (
+          enumValues !== null &&
+          !enumValues
+            .map(s => s.toUpperCase())
+            .includes(String(value).toUpperCase())
+        ) {
+          invariant(
+            false,
+            'RelayMockPayloadGenerator: Invalid value "%s" provided for enum ' +
+              'field "%s" via MockResolver.' +
+              'Expected one of the following values: %s.',
+            String(value),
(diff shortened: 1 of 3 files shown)


sibelius:
We could move this enum logic to other functions to be easy to reason about it

alunyov:
@sibelius you're the best!

sibelius:
fixed an edge case when enumValues as null

packages/relay-test-utils/RelayMockPayloadGenerator.js
```diff
@@ -391,32 +391,60 @@ class RelayMockPayloadGenerator {
...
-        invariant(
-          false,
-          'RelayMockPayloadGenerator: Invalid value "%s" provided for enum ' +
-            'field "%s" via MockResolver.' +
-            'Expected one of the following values: %s.',
-          String(value),
-          `${path.join('.')}.${applicationName}`,
-          enumValues.map(v => `"${v}"`).join(', '),
-        );
+      if (Array.isArray(value) && enumValues !== null) {
```
alunyov: @sibelius let's have another iteration on this, and actually, move enumValues generation to a separate method. So, in the `mockScalar` we would call something like `value = this._getCorrectDefaulEnum(value, enumValues);`?

sibelius: it makes sense

I'm gonna modify this

packages/relay-test-utils/RelayMockPayloadGenerator.js
```diff
@@ -353,6 +353,77 @@ class RelayMockPayloadGenerator {
     return mockData;
   }
 
+  /**
+   * Generate default enum value
+   * @private
+   */
+  _getCorrectDefaultEnum(
+    enumValues: $ReadOnlyArray<string> | null,
```
alunyov: I would make this `enumValues: $ReadOnlyArray<string>`

packages/relay-test-utils/RelayMockPayloadGenerator.js
```diff
@@ -391,33 +462,7 @@ class RelayMockPayloadGenerator {
...
-
-      // But missing case should be acceptable, we will just use
-      // a correct spelling from enumValues
-      if (enumValues !== null && value !== undefined) {
-        const correctSpellingEnumIndex = enumValues
-          .map(s => s.toUpperCase())
-          .indexOf(String(value).toUpperCase());
-        value = enumValues[correctSpellingEnumIndex];
-      }
+      value = this._getCorrectDefaultEnum(enumValues, value, path, applicationName);
```
alunyov: and here we would call it `if (enumValues != null) { ...` which will provide more context on why we're calling `_getCorrectDefaultEnum(...)`

packages/relay-test-utils/RelayMockPayloadGenerator.js
```diff
@@ -353,6 +353,77 @@ class RelayMockPayloadGenerator {
...
+        );
+      }
+    }
+
+    // But missing case should be acceptable, we will just use
+    // a correct spelling from enumValues
+    if (value !== undefined) {
+      if (Array.isArray(value)) {
+        return value.map(v => {
+          const correctSpellingEnumIndex = enumValues
```
alunyov: We can use `enumValuesNormalized` here, right?

packages/relay-test-utils/RelayMockPayloadGenerator.js
```diff
@@ -353,6 +353,77 @@ class RelayMockPayloadGenerator {
...
+          .map(s => s.toUpperCase())
+          .indexOf(String(v).toUpperCase());
+
+          return enumValues[correctSpellingEnumIndex];
+        });
+      } else {
+        const correctSpellingEnumIndex = enumValues
```
alunyov: and here (enumValuesNormalized)

packages/relay-test-utils/RelayMockPayloadGenerator.js
```diff
@@ -353,6 +353,77 @@ class RelayMockPayloadGenerator {
...
+   * Generate default enum value
+   * @private
+   */
+  _getCorrectDefaultEnum(
+    enumValues: $ReadOnlyArray<string> | null,
+    value: mixed | Array<mixed>,
+    path: $ReadOnlyArray<string>,
+    applicationName: string,
+  ) {
+    if (enumValues == null) {
```
alunyov: should we return early if `value === undefined`?

packages/relay-test-utils/RelayMockPayloadGenerator.js
```diff
@@ -353,6 +353,77 @@ class RelayMockPayloadGenerator {
...
+      return value;
+    }
+
+    const enumValuesNormalized = enumValues.map(s => s.toUpperCase());
+
+    if (Array.isArray(value)) {
```
alunyov: I think we can do a little trick here:

```
const valueToValidate = Array.isArray(value) ? value : [value];
```
and the just leave the first part of this if/else statement - where we are validating an array.

PR #2780, opened 2019-06-22
Fix links in relay-compiler readme

zpao:


zpao:
There are still broken links in that doc since 7b730e0e13383c76fe47751966a79ef06f16b412. I don't know where `graphql-compiler` is now but it's still published so I guess that's where I'd want to go to read up?

josephsavona:
@zpao thanks for the PR! We moved graphql-compiler contents back into relay-compiler and are no longer updating the latter package.

PR #2752, opened 2019-06-04
IDE friendly errors

alloy:
These changes make it so file/line locations in errors can be made linkable by IDEs. This is the format used by jest, flow, typescript, and many others.

### Before

relay-ide-friendly-errors-before

### After

relay-ide-friendly-errors-after

PR #2784, opened 2019-06-26
[RelayCompilerError] Use graphql-js exported utils

alloy:
⚠️ 
* This includes #2752, so that one should ideally be merged first. The real change for this PR is in 73671f7.
* ~~The graphql-js changes are in master but not yet in a release, so this PR needs to be updated to point to a release once that’s out.~~

Follow-up from https://github.com/facebook/relay/pull/2752#issuecomment-502046563, https://github.com/graphql/graphql-js/pull/1985#issuecomment-502401686, and as [discussed in Slack](https://graphql.slack.com/archives/C044QJ51J/p1559654699006600) with @josephsavona and @IvanGoncharov.

This change removes graphql-js code that was vendored and uses the newly exported graphql-js utilities instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "fbjs": "^1.0.0",
     "fbjs-scripts": "^1.1.0",
     "flow-bin": "^0.102.0",
-    "graphql": "^14.2.0",
+    "graphql": "^14.4.0",
     "gulp": "4.0.0",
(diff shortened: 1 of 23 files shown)


kassens:
Since we're 100% focused on the compiler rewrite and this required some internal changes. I'll close this for now. Better IDE integration is one of our new goals :)

PR #2791, opened 2019-07-11
Support external publish queues

mattkrick:
The following changes make it possible to support a 3rd party PublishQueue, such as https://yarnpkg.com/en/package/relay-linear-publish-queue

Quick description of the problem & solution here: https://dev.to/mattkrick/distributed-state-101-why-i-forked-facebook-s-relay-1p7d

Please let me know what else is needed to move this along!

### DIFF
--- a/packages/relay-runtime/store/DataChecker.js
+++ b/packages/relay-runtime/store/DataChecker.js
@@ -286,6 +286,7 @@ class DataChecker {
           this._traverseSelections(selection.selections, dataID);
           break;
         case SCALAR_HANDLE:
+          break;
         case FRAGMENT_SPREAD:
           invariant(
             false,
--- a/packages/relay-runtime/store/RelayModernQueryExecutor.js
+++ b/packages/relay-runtime/store/RelayModernQueryExecutor.js
@@ -328,10 +328,6 @@ class Executor {
   }
 
   _processResponse(response: GraphQLResponseWithData): void {
-    if (this._optimisticUpdate !== null) {
-      this._publishQueue.revertUpdate(this._optimisticUpdate);
-      this._optimisticUpdate = null;
-    }
     const payload = normalizeResponse(
       response,
       this._operation.root,
@@ -342,7 +338,8 @@ class Executor {
     this._incrementalResults.clear();
     this._source.clear();
     this._processPayloadFollowups(payload);
-    this._publishQueue.commitPayload(this._operation, payload, this._updater);
+    this._publishQueue.commitPayload(this._operation, payload, this._updater, this._optimisticUpdate);
+    this._optimisticUpdate = null;
     const updatedOwners = this._publishQueue.run();
     this._updateOperationTracker(updatedOwners);
 
--- a/packages/relay-runtime/store/RelayModernRecord.js
+++ b/packages/relay-runtime/store/RelayModernRecord.js
@@ -85,17 +85,20 @@ function clone(record: Record): Record {
 /**
  * @public
  *
- * Copies all fields from `source` to `sink`, excluding `__id` and `__typename`.
+ * Copies all fields from `source` to `sink`, excluding `__id`.
  *
  * NOTE: This function does not treat `id` specially. To preserve the id,
  * manually reset it after calling this function. Also note that values are
  * copied by reference and not value; callers should ensure that values are
  * copied on write.
+ *
+ * `__typename` is not excluded because two normalized payloads could
+ * have the same `storageKey`, but different `__typename` values
  */
 function copyFields(source: Record, sink: Record): void {
   for (const key in source) {
     if (source.hasOwnProperty(key)) {
-      if (key !== ID_KEY && key !== TYPENAME_KEY) {
+      if (key !== ID_KEY) {
         sink[key] = source[key];
       }
     }
--- a/packages/relay-runtime/store/RelayPublishQueue.js
+++ b/packages/relay-runtime/store/RelayPublishQueue.js
@@ -149,7 +149,11 @@ class RelayPublishQueue implements PublishQueue {
     operation: OperationDescriptor,
     {fieldPayloads, source}: RelayResponsePayload,
     updater?: ?SelectorStoreUpdater,
+    optimisticUpdate?: OptimisticUpdate | null
   ): void {
+    if (optimisticUpdate) {
+      this.revertUpdate(optimisticUpdate)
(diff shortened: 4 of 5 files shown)


josephsavona:
Thanks again for the PR and for linking to the post with more context. We've made some recent changes that touch on the PublishQueue/Store implementations and interface, so I'll briefly describe those as well as how Relay already supports the linear-update pattern you're trying to achieve - no forking required. 

> Instead of processing updates in the order they occurred, Relay processes server updates, then local updates, then optimistic updates. 

First a few updates on changes since this PR:
* PublishQueue now supports multiple optimistic payloads, which can occur if an optimistic payload uses advanced features such as data-driven dependencies. This makes it less obvious how to  swap optimistic payloads for final payloads as done in this PR.
* The previous server-update then local-update was confusing, so we've changed the logic to be just "final" updates in order, followed by optimistic updates in order.
* We've also changed the mechanism for reverting/rebasing optimistic updates, which shifts some of the costs wrt to building abstractions on top of the built-in support for final/optimistic updates. 

>  That means even though an optimistic update occurred first, Relay applies it after the local update. That's the crux of the bug.

> Let's use that logic in a simple component like a volume slider that goes from 1 to 10. Say the volume is 3 and I optimistically add 1 to it. Then, I locally set the volume to 10. What's the result? If you guessed 10, you've correctly applied Lamport's relation. If you guessed 11, then you've got a broken app and a bright future at Facebook (Kidding. I'm totally kidding. 😉).

As I've pointed out in previous discussions, your tone here isn't especially productive and doesn't make me feel super enthused about answering or helping. I realize it's on your blog, but I think it's appropriate to comment on given that you linked to it from the PR. 

Regardless: there are two ways to look at this: one is to say that Relay doesn't work the way you expect and therefore it's a bug. Another is to take the time to understand what features Relay provides, and then consider how to achieve the desired affect using those features.

Specifically: Relay is designed to apply server updates in order, then optimistic updates in order. That supports a common pattern of quickly applying (some of) the effects of mutations locally in order to improve perceived responsiveness, while the actual mutation completes in the background. So for an application that wants to maintain linear ordering - to issue optimistic update A, then issue *some* update B, and always have B to be applied after A - it needs to main an external queue of operations. For that use-case, "final" (non-optimistic) updates that occur after optimistic updates have to be first applied as optimistic, then later converted to final updates once previous updates have all committed.

So for the volume example, the app would have to do the following sequence:

1. Optimistically increment the volume from 3 ->  4 via applyUpdate
2. Final update to set volume to  10 *via applyUpdate*  - this  can't actually be final until  all previous  optimistic  updates have committed
3. When the optimistic update from 1 is reverted or commits, then revert the update from 2 and commit it as final w commitUpdate.

I hope this clarifies that the update ordering is a feature, not a bug, and that it's possible to implement linear ordering without forking and using the primitives provided. Given this, we'll go ahead and close the PR, though we still need to investigate the copyFields change from #2637 separately.

mattkrick:
Thanks for the reply, couple questions:

> Relay is designed to apply server updates in order, then optimistic updates in order

Can we agree that the order of server updates is non-deterministic due to latency? If event A sets the volume to `2` and event B sets the volume to `5`, then it's possible that event B returns first, right? If that happens, then Relay's default logic would have the volume end up being `2`, even if optimistic updates existed to hint at the order. It's OK if we disagree, but I'd like to respectfully point that this does not follow lamport's happened-before relation. In practice, I think any user who clicks `2` first then `5` 2nd would like to result to converge on `5`.
 
> it needs to main an external queue of operations

please correct me if i'm not understanding this correctly, as I'm often wrong. You're suggesting to replace a single linear publish queue with an external queue that has custom logic which decides whether to apply, commit, or revert events & then calls those methods on the primary publish queue?

No disrespect intended, I appreciate the response & simply want to get the math straight.

josephsavona:
> Can we agree that the order of server updates is non-deterministic due to latency?

Yes. As I noted in an earlier issue/PR about this topic, Relay defers to the application's network layer for ordering of server responses. For example, an application could choose to serialize all requests to the server at the network layer in order to guarantee linear ordering, or possibly choose to only  serialize mutations.

> please correct me if i'm not understanding this correctly, as I'm often wrong. You're suggesting to replace a single linear publish queue with an external queue that has custom logic which decides whether to apply, commit, or revert events & then calls those methods on the primary publish queue?

Yes. Perhaps the naming of the internal module  (PublishQueue) is what's confusing here. That  internal module is really about batching a set of operations together with a single call to notify affected subscribers. The external queue I described above would operate on a longer-time scale, coordinating the execution of multiple requests over time.

mattkrick:
good good, I think i can work with that. If using an external queue, will it still be passed in as a custom publish queue? To guarantee linearizability, the only thing I need is to match the optimistic update to the final update(s), is there a way to achieve that?

josephsavona:
An external queue is something external to Relay, you'd likely then want to use it to execute all of your operations instead of directly running them against your Environment instance. The queue would have to keep track of in-flight operations and manage the process of applying each one as optimistic or final, potentially upgrading from optimistic->final as in step 3 of the above example.

mattkrick:
i see. so to be clear, in order to avoiding forking relay with 5 lines of code, i should:
- write a custom network layer including an event queue to guarantee mutations are received in order
- modify my graphql server to forward the mutation operation id back down to the custom network layer
- write a custom external queue which does everything the publish queue does, except publish
- write a custom environment which implements that custom external queue

Please don't interpret this as being disrespectful, I simply want to understand that this is what you meant when you said: 
> Relay already supports the linear-update pattern you're trying to achieve - no forking required.

If i misunderstood your suggestion in any way, please let me know!

josephsavona:
Totally fine to clarify. I made multiple recommendations above about things you might want to do depending on exactly what you were trying to achieve. Not all of those are required and some overlap. You *could* choose to implement serialization both at the network layer and at a layer that wraps the environment, but that would be duplication. 

Concretely if you want to linearize mutations I would:

* Implement a mutation queue that handles the linearization, optimistic->final upgrading.
* Use that to execute your mutations.

This is pretty straightforward to implement: not 5 lines, but that comparison isn't valid anyway (see my earlier comments about recent changes to PublishQueue/Store).

Felix-N:
hi josephsavona,

i just ran into the same issue. Honestly I'm totally fine how relay handles responses, but it would've been quite helpful to have the behavior very explicitly named in the documentation.

PR #2796, opened 2019-07-13
Remove promise polyfill

TrySound:
relay-runtime depends on fbjs/lib/resolveImmediate which depends on
promise package.

In modern apps promises are not always necessary to polyfill. On the
other side users may already have globally defined polyfill.

`promise` package with its `asap` dependency adds 10kb (minified) to our
bundle.

I suggest to reimplement resolveImmediate with globally defined promise.

### DIFF
--- a/packages/relay-runtime/store/RelayModernStore.js
+++ b/packages/relay-runtime/store/RelayModernStore.js
@@ -20,7 +20,7 @@ const deepFreeze = require('../util/deepFreeze');
 const defaultGetDataID = require('./defaultGetDataID');
 const hasOverlappingIDs = require('./hasOverlappingIDs');
 const recycleNodesInto = require('../util/recycleNodesInto');
-const resolveImmediate = require('resolveImmediate');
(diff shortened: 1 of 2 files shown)


kassens:
Would love some more opinions on this? What do other libraries do that need to use `Promise`? Is depending on the global `Promise` object the right solution these days?

sibelius:
it is recommend to let users bring their own polyfill implementations

Even Relay docs recommends this https://relay.dev/docs/en/classic/installation-and-setup#javascript-environment-requirements

morrys:
What about this? https://github.com/facebook/relay/blob/master/scripts/getBabelOptions.js#L32

morrys:
Hi @kassens, 
now I can't prove it but if I'm not mistaken when the environment for jest is created in the babel options [the env parameter is set to test](https://github.com/facebook/relay/blob/master/scripts/jest/preprocessor.js#L19)

can't you use [here](https://github.com/facebook/relay/blob/master/scripts/getBabelOptions.js#L27-L39) that information to load a different map?

PR #2802, opened 2019-07-15
Add persistFunction option to relay compiler

robrichard:
Adds a new option to relay compiler `persistFunction`. This allows customizing how the `id` is generated for persisted queries

### DIFF
--- a/packages/relay-compiler/bin/RelayCompilerMain.js
+++ b/packages/relay-compiler/bin/RelayCompilerMain.js
@@ -64,6 +64,7 @@ export type Config = {|
   persistOutput?: ?string,
   noFutureProofEnums: boolean,
   language: string | PluginInitializer,
+  persistFunction?: ?string | ?((text: string) => Promise<string>),
   artifactDirectory?: ?string,
   customScalars?: ScalarTypeMapping,
 |};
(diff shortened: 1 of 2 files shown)


josephsavona:
The overall idea makes sense, but the idea of relay-config was that this could be specified via the config file and/or CLI flags. Can you make sure this can be specified via relay-config as well?

robrichard:
@josephsavona thanks for taking a look. I tested the following scenarios:

1. Passing a path to a module via the cli (e.g.`--persistFunction=persist.js`)
2. Passing a path to a module via a `relay.config.json` or a `relay.config.js`
3. Passing an inline function via a `relay.config.js`
4. Passing an inline function directly to `relayCompiler` when combined with https://github.com/facebook/relay/pull/2800

Module paths will work with a file that assigns a function to `module.exports` or a default export that was transpiled to commonjs.

packages/relay-compiler/bin/RelayCompilerMain.js
```diff
@@ -157,6 +158,39 @@ function getLanguagePlugin(
...
+function getPersistQueryFunction(
+  config: Config,
+): ?(text: string) => Promise<string> {
+  if (config.persistFunction == null) {
+    return null;
+  } else if (typeof config.persistFunction === 'string') {
+    try {
+      // eslint-disable-next-line no-eval
+      const persistFunction = eval('require')(
+        path.resolve(process.cwd(), String(config.persistFunction)),
```
robrichard: String wrapper here to appease flow. Not sure why it's flagging since it's wrapped in a typeof check.

packages/relay-compiler/bin/RelayCompilerMain.js
```diff
@@ -333,17 +377,16 @@ function getRelayFileWriter(
...
+      const persistImplmentation = persistFunction || defaultPersistFunction;
+      persistQuery = async (text: string) => {
+        const id = await persistImplmentation(text);
         queryMap.set(id, text);
-        return Promise.resolve(id);
+        return id;
       };
     }
-    const results = RelayFileWriter.writeAll({
+    const results = await RelayFileWriter.writeAll({
```
alloy: So making `persistQuery` a promise is automatically making this function return a promise too? Nice.

packages/relay-compiler/bin/RelayCompilerMain.js
alloy: Flow novice here, but perhaps Flow would only consider that when capturing the value in a local variable?

```js
const { persistFunction } = config
if (persistFunction == null) {
  return null;
} else if (typeof persistFunction === 'string') {
  // ...
}
```

josephsavona: yeah, i suspect assigning to a local const would solve this

packages/relay-compiler/bin/RelayCompilerMain.js
josephsavona: Not quite: RelayFileWriter.writeAll was already returning a promise, this is just changing to use async/await instead of directly returning the promise

josephsavona:
Thanks for taking a look, @alloy!

packages/relay-compiler/bin/RelayCompilerMain.js
alloy: Ahh, ok yeah that makes more sense. Thanks 👍

PR #2809, opened 2019-07-31
fix: Check for single-value-as-array earlier

stefanprobst:
`GraphQLIRPrinter`'s `printLiteral` does currently check for single values on List input fields, but does not handle a single object on a List field, because this will be caught by the `if` block before. Solved by moving the block up.

Not sure about your test setup. Please advise if this should be done differently/somewhere else.


kassens:
Looks good! Thanks for adding the test

PR #2810, opened 2019-07-31
Expose way to override isGeneratedFile function via language plugin.

zth:
This PR lets language plugins override `isGeneratedFile`. My use case is emitting ReasonML types, where the generated files need to be called `SomeModule_graphql.re` rather than `SomeModule.graphql.re`, since ReasonML does not accept `.` in the file name other than for the extension.


zth:
Anyone had time to look at this?

zth:
I've updated this PR to also pass the isGeneratedFile function to deleteExtraFiles in order to prevent deleting files marked as generated.

cc @jstejada @kassens anything I can do to help move this forward? Explain exactly why I need it etc?

PR #2811, opened 2019-07-31
Move check for graphql string in source files to language plugin

zth:
This moves the check for the string `graphql` in source files to the JS language plugin, away from `RelaySourceModuleParser`. This has two reasons:

1. Language plugins might not require/use the string 'graphql'. I have that case when emitting ReasonML types - Writing GraphQL there looks like `[%relay.mutation ...]` and `[%relay.query ...]`, so `graphql` is not naturally inside any of the source files I want the compiler to pick up.
2. The requirement was quite "hidden" and it took me a few hours to understand why my files weren't picked up even though they seemed to satisfy all requirements of the compiler + configuration. So, also reducing confusion.

If this lands I'll PR the TypeScript language plugin to add the same type of check into that as well.


jstejada:
cc @alunyov or @kassens might want to take a look at this

zth:
Hey @kassens ,

I think that sounds good! I made some changes - is this more in line of what you had in mind?

zth:
cc @alunyov since you're on a merging spree ;) Would you mind having a look at this one too? 

It's the last piece of the puzzle to make the compiler to support what I need for making my Reason integration work without needing to maintain a fork.

mrtnzlml:
> If this lands I'll PR the TypeScript language plugin to add the same type of check into that as well.

Hi @zth, I just wanted to remind you of this. Do you have time to update the TS language plugin? :)

zth:
@mrtnzlml I actually believe this PR was changed enough from its original state that the TS plugin won't need PRing. Or have you encountered a bug because of this?

mrtnzlml:
I thought so, but I was mistaken - it was a different issue. Sorry for the useless ping. 😬

zth:
Haha no worries, just glad it works :D

PR #2818, opened 2019-08-07
Remove USERS.md

jstejada:
USERS.md no longer power the users page in the Relay website, so we should remove it to avoid confusion. Instead, users should be added to the siteConfig.js file.

Anyone on this list not currently on siteConfig.js should submit another pull request adding themselves to siteConfig.js if they're still using Relay.


orta:
:wave:

kassens:
@jstejada I think we can land this?

Issue #2457, opened 2018-05-27
PaginationContainer inconsistent with connections spec


bellkev:
I recently added the PageInfo and Edge Type interfaces to my GraphQL schema in support of a relay PaginationContainer and noticed an inconsistency with the [connection spec](https://facebook.github.io/relay/graphql/connections.htm#sec-undefined.PageInfo). Relay requires both `startCursor` and `endCursor` fields in PageInfo, which are not mentioned in the spec. It also appears not to use the `cursor` field on Edge Types which is required by the spec (not needing this field actually simplifies things in my case, as my backing datastore has no native concept of a per-item cursor/token).

Should the spec be updated to reflect the type of connection Relay expects?

merrywhether:
It would be great to see this happen. It's hard enough convincing API maintainers to add this to their GraphQL schema when it's not in the official GraphQL spec, but even more so when it's not even in the official Relay spec.

jedwards1211:
Wow, I'm amazed in all the time I spent reworking graphql-sequelize's connections implementation I never noticed this

jedwards1211:
In specs for this we should make sure to require that `startCursor` be `edges[0].cursor` and `endCursor` be `edges[edges.length - 1].cursor`.

jedwards1211:
@merrywhether you mean you have trouble convincing API maintainers to use connections instead of lists?

jedwards1211:
@bellkev considering that the first line of the spec is as follows, this is tragically funny:
> Relay’s support for pagination relies on the GraphQL server exposing connections in a standardized way.

merrywhether:
@jedwards1211 Most of my company's projects use apollo-client, and our GraphQL API has mostly been tuned to work with that. When I asked about startCursor/endCursor to easily support pagination in Relay, they weren't super enthusiastic about it, and even initially pointed me at the Relay Connections spec and said that it wasn't required there, which did not help my point.

jedwards1211:
Heh I see, that sounds frustrating.
Just a thought...as long as the core APIs support pagination of some kind (if only offset) It's probably possible to create wrapper queries that consume the core API and convert the data to a connection for relay.

ngbrown:
I noticed that only the `graphql/` directory of the website isn't getting updated during publish.

[...]

PR #2819, opened 2019-08-07
Run update-spec in website package to update connections spec

merrywhether:
As referenced in #2457, the connections spec markdown file is up-to-date but the published artifact related to it is not. This PR just runs `yarn update-spec` in the website package in order to get the spec site back in sync with the current version of the spec.

This was the simplest solution to the immediate problem, but is not long term. I could see a more robust solution that removes `website/static/graphql` from the git repo entirely and moves `yarn update-spec` (maybe renamed to `yarn generate-spec` or something) into the circleCI config so that the html artifacts are generated during the publish and thus cannot get out of sync, but I did not want to overstep (though could easily wire that up as a follow-up if desired).


kassens:
Thanks, good catch!

PR #2832, opened 2019-08-23
Remove fbjs usage: emptyFunction, forEachObject, removeFromArray

kassens:
We just use a small set of dependencies from fbjs that are just as clean when we don't depend on them.


TrySound:
Would be good to drop promise polyfill too. Users usually have own if it's required. What do you about this?
https://github.com/facebook/relay/pull/2796

kassens:
@TrySound thanks, I was wondering what to do about `Promise`, thanks for opening the PR. Let's have the discussion there!

PR #2838, opened 2019-08-29
fix: use reporter to report changes

jquense:
At the moment there is no way to intersect these logs.

### DIFF
--- a/packages/relay-compiler/codegen/CodegenDirectory.js
+++ b/packages/relay-compiler/codegen/CodegenDirectory.js
@@ -182,6 +182,7 @@ class CodegenDirectory {
     Profiler.run('CodegenDirectory.printChanges', () => {
       const output = CodegenDirectory.formatChanges(changes, options);
 
+      // eslint-disable-next-line no-console
       console.log(output);
     });
   }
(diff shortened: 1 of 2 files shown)


packages/relay-compiler/codegen/CodegenDirectory.js
```diff
@@ -148,32 +148,41 @@ class CodegenDirectory {
...
-        printFiles('Created', changes.created);
-        printFiles('Updated', changes.updated);
-        printFiles('Deleted', changes.deleted);
-        output.push(`Unchanged: ${changes.unchanged.length} files`);
-      }
-      // eslint-disable-next-line no-console
-      console.log(output.join('\n'));
+      const output = CodegenDirectory.formatChanges(changes, options);
+
+      console.log(output);
```
jquense: IDK if ya'll want this here still but i tried not to break any existing usages...also the instance method  version doesn't appear to be used anywhere

jquense:
```suggestion
      // eslint-disable-next-line no-console
      console.log(output);
```

PR #2840, opened 2019-09-03
remove @babel/polyfill from relay-compiler

robrichard:
Is babel polyfill still needed? I can't think of anything relay-compiler would be using that's not supported natively in Node 8 and higher.


kassens:
I think your right. Our babel config is now pretty minimal in terms of transforms.

PR #2781, opened 2019-06-24
Optionally output eager es modules

TrySound:
In this PR I added `eagerESModules` option to `relay-runtime` and `babel-plugin-relay`. I have two problem which can be solved with this change

1. Rollup handles only es modules. To handle commonjs there is a plugin
which converts it into esm. Though it always skips module if
import/export statements are found. As a result I get runtime error:
"require is not defined".

To workaround this I wrote custom plugin. Though it would be cool to
have proper solution out of the box.

```js
{
  name: 'relay-generated',
  transform(code) {
    // convert __generated__ requires into imports
    // remove after relay-compiler will be able to emit esm
    if (code.includes('__generated__')) {
      let i = -1;
      const paths = [];
      const processed = code.replace(
        /require\((.+__generated__.+)\)/g,
        (req, pathString) => {
          i += 1;
          paths.push(pathString);
          return `__generated__${i}`;
        },
      );
      const imports = paths.map(
        (p, i) => `import __generated__${i} from ${p};\n`,
      );
      return {
        code: imports.join('') + processed,
      };
    }
  },
},
```

2. Another problem is flow bug (https://github.com/facebook/flow/issues/7444) which treats all not existent types as any.
This leads to many type unsafe places.

```js
import type {NotExistentType} from './__generated__/MyQuery.graphql';

type Props = {|
  my: NotExistentType
|}
```

PR #2841, opened 2019-09-05
Add an option to generate esmodule imports/exports in babel-plugin-relay & relay-compiler

vincentriemer:
In order to be compatible with bundlers like rollup, which without extra configuration does not support commonjs modules, so this PR adds an `esmodules` option to both `babel-plugin-relay` and `relay-compiler` so that it can generate es modules instead of commonjs. The new config option in both packages will default to `esmodules: false` so this shouldn't (hopefully) be a breaking change.

### DIFF
--- a/packages/babel-plugin-relay/BabelPluginRelay.js
+++ b/packages/babel-plugin-relay/BabelPluginRelay.js
@@ -32,6 +32,9 @@ export type RelayPluginOptions = {
 
   // Directory as specified by artifactDirectory when running relay-compiler
   artifactDirectory?: string,
+
+  // Generate output with esmodule imports instead of commonjs requires
+  esmodules?: boolean,
 };
 
 export type BabelState = {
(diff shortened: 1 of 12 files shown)


alloy:
Rather than adding another configuration option, I wonder if this would be possible to do as a language plugin. For instance, our TypeScript plugin emits ES modules, rather than CommonJS (which is also why the Babel plugin is currently already able to read ES module files).

I would imagine the plugin would basically just re-export the Flow plugin, but amend the generated module. This also alleviates the need for a “hacky” approach to exist in the canonical code-base.

vincentriemer:
That makes sense to me, but I'd still need to keep the esmodule option on the babel plugin (thankfully not as hacky). I'll remove the relay-compiler commit.

TrySound:
@vincentriemer Did you see my PR?
https://github.com/facebook/relay/pull/2781

alloy:
@TrySound I did not, sorry, was out of the loop for a bit.

@jstejada Any thoughts?

vincentriemer:
@TrySound Sorry I didn't! Your PR looks cleaner so I'll close this one, thanks!

Issue #2845, opened 2019-09-09
Some invalid Internal import paths in relay-experimental 


babangsund:
Once the package has been built and packed, these paths are invalid. 😅 

#### in `RelayEnvironmentProvider`:
https://github.com/facebook/relay/blob/cf5566eb0fbee1f75f821d27ad738c23788cea6c/packages/relay-experimental/RelayEnvironmentProvider.js#L15

It exists as a public module in `react-relay`, so it can be replaced with the following:
```javascript
 const {ReactRelayContext} = require('react-relay'); 
```

#### in `useRelayEnvironment`:
https://github.com/facebook/relay/blob/cf5566eb0fbee1f75f821d27ad738c23788cea6c/packages/relay-experimental/useRelayEnvironment.js#L14

Same here - it exists as a public module in `react-relay`:
```javascript
 const {ReactRelayContext} = require('react-relay'); 
```

#### in `useRefetchableFragmentNode`:

https://github.com/facebook/relay/blob/cf5566eb0fbee1f75f821d27ad738c23788cea6c/packages/relay-experimental/useRefetchableFragmentNode.js#L591

Already exists in the public API:
```javascript
const {ID_KEY} = require('relay-runtime');
```

---

https://github.com/facebook/relay/blob/cf5566eb0fbee1f75f821d27ad738c23788cea6c/packages/relay-experimental/useRefetchableFragmentNode.js#L533

https://github.com/facebook/relay/blob/cf5566eb0fbee1f75f821d27ad738c23788cea6c/packages/relay-experimental/useRefetchableFragmentNode.js#L560

Although a decision for the Relay team, you could simply add `RelayModernRecord` to the public `relay-runtime` API - i.e:

**Export**
```javascript
// relay-runtime/index.js

const RelayModernRecord = require('./store/RelayModernRecord');
...
...
module.exports = {
  Record: RelayModernRecord,
};
```

**Import**
```javascript
// useRefetchableFragmentNode.js
const {Record: RelayModernRecord} = require('relay-runtime');
```

babangsund:
Merged

PR #2847, opened 2019-09-09
Fix broken relay-experimental imports; Add Record: RelayModernRecord to relay-runtime API

babangsund:
Suggested fix for https://github.com/facebook/relay/issues/2845

### DIFF
--- a/packages/relay-experimental/useRelayEnvironment.js
+++ b/packages/relay-experimental/useRelayEnvironment.js
@@ -11,7 +11,7 @@
 
 'use strict';
 
-const ReactRelayContext = require('react-relay/ReactRelayContext');
+const {ReactRelayContext} = require('react-relay');
 
 const invariant = require('invariant');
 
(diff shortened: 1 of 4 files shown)


packages/relay-experimental/useRelayEnvironment.js
```diff
@@ -11,7 +11,7 @@
 
 'use strict';
 
-const ReactRelayContext = require('react-relay/ReactRelayContext');
+const {ReactRelayContext} = require('react-relay');
```
jstejada: the problem here is that we don't actually want to bring in all of 'react-relay' in our web bundle, let me see if I can get this to work

babangsund: Hadn't considered this aspect. 🤔

jstejada: I think we still want to be compatible with the apis in `react-relay`, so you can use both apis together and migrate incrementally to hooks if you want (without having to convert everything to hooks at once), which means that we need to use the same context; but for cases where we're exclusively using relay-experimental we don't to bring in all of react-relay.

I think I might have a solution for this though

babangsund: Okay. Looking forward to being blown away. 😉

PR #2848, opened 2019-09-10
Add relay-config documentation to v5

babangsund:
The relay-config documentation in "Installation and setup" is missing from the v5 documentation.

cc @sibelius


kassens:
Thanks!

PR #2849, opened 2019-09-13
[easy] clean up title of QueryRenderer docs

kassens:
None of the other React pages have the `< />` wrapper and it just doesn't look clean in the side bar.


babangsund:
@kassens Wouldn't it make sense to also remove it from the other versions? 🤔

kassens:
@babangsund updating the old versions is a bit more pain.

PR #2850, opened 2019-09-13
Add docs for @inline

robrichard:


babangsund:
I like how you snuck the `pinned` in there. Bold! 😉

babangsund:
> I like how you snuck the `pinned` in there. Bold! 😉

EDIT: 

To which I'd add that someone like @entria probably deserves being higher on the pinned list, considering their contributions and evangelism in the Relay community. 👍 

I could be ignorant here - there may be other deciding factors for being "pinned" worthy that I'm unaware of. 🙂 (In which case, please disregard my comment 😅)

docs/Modern-GraphQLInRelay.md
```diff
@@ -89,11 +89,68 @@ fragment TodoApp_app on App {
 }
 ```
 
+### `@inline`
+
+By default Relay will only expose the data for fields explicitly requested by a [component's fragment](./fragment-container.html#createfragmentcontainer), which is known as [data masking](./thinking-in-relay#data-masking). Fragment data is unmasked for use in React components by `createFragmentContainer`. However, you may want to use fragment data in non-React functions that are called from React.
+
+Non-React functions can also take advantage of data masking. A fragment can be defined with the `@inline` directive and stored in a local variable. The non-React function can than "unmask" the data using the `readInlineData` function.
```
babangsund: can **then** "unmask"

docs/Modern-GraphQLInRelay.md
```diff
@@ -89,11 +89,68 @@ fragment TodoApp_app on App {
...
+```javascript
+// React Component
+function MyComponent({item}) {
+  function handleClick() {
+    processItemData(item);
+  }
+
+  return (
+    <button onClick={handleClick}>Process {item.title}</button>
+  )
```
babangsund: Missing semi 😂

website/siteConfig.js
```diff
@@ -54,7 +54,7 @@ const users = [
     caption: '1stdibs',
     image: '/img/logos/1stdibs.png',
     infoLink: 'https://www.1stdibs.com/',
-    pinned: false,
+    pinned: true,
```
kassens: Sneaky!

docs/Modern-GraphQLInRelay.md
```diff
@@ -89,11 +89,68 @@ fragment TodoApp_app on App {
...
+  sendToThirdPartyApi({
+    title: item.title,
+    price: item.price,
+    creatorName: item.creator.name
+  });
+}
+```
+
+```javascript
+// React Component
```
jstejada: we should probably show this component's fragment too

robrichard: This component's fragment is a few lines down inside `createFragmentContainer`, did you have a different way of displaying it in mind?

babangsund:
🎉

PR #2855, opened 2019-09-16
LocalQueryRenderer docs

babangsund:
LocalStateManagement docs
- Add a note on LocalQueryRenderer in v6/next
- Add missing React fragment
- Sync across v4/5/6

Todo: ./local-query-renderer


babangsund:
Still trying to decide if LocalQueryRenderer warrants a separate section in the API reference,
or if it should just exist as sub-section in the QueryRenderer docs.

Leaning towards the first option.

PR #2860, opened 2019-09-19
Update Modern-NetworkLayer.md

juffalow:
In one of my old projects, the structure of `relay-runtime/lib` is:
* *file.js*
* *file.js*
* RelayQueryResponseCache.js
* *file.js*

But with the new version (`6.0.0`) the `relay-runtime/lib` is changed to:
* folder
* folder
* network
  * RelayQueryResponseCache.js
  * *file.js*
* index.js

So the import doesn't work like the old one anymore.

### DIFF
--- a/docs/Modern-NetworkLayer.md
+++ b/docs/Modern-NetworkLayer.md
@@ -66,7 +66,8 @@ import {
   RecordSource,
   Store,
 } from 'relay-runtime';
-import RelayQueryResponseCache from 'relay-runtime/lib/RelayQueryResponseCache';
+import { QueryResponseCache as RelayQueryResponseCache } from 'relay-runtime/lib';
+// or import RelayQueryResponseCache from 'relay-runtime/lib/network/RelayQueryResponseCache';
 
 const oneMinute = 60 * 1000;
 const cache = new RelayQueryResponseCache({ size: 250, ttl: oneMinute });


docs/Modern-NetworkLayer.md
```diff
@@ -66,7 +66,8 @@ import {
   RecordSource,
   Store,
 } from 'relay-runtime';
-import RelayQueryResponseCache from 'relay-runtime/lib/RelayQueryResponseCache';
+import { QueryResponseCache as RelayQueryResponseCache } from 'relay-runtime/lib';
```
babangsund: `import { QueryResponseCache as RelayQueryResponseCache } from 'relay-runtime';`

docs/Modern-NetworkLayer.md
```diff
@@ -66,7 +66,8 @@ import {
...
+// or import RelayQueryResponseCache from 'relay-runtime/lib/network/RelayQueryResponseCache';
```
mrtnzlml: I wouldn't recommend this direct access in docs since it's not needed anymore.

docs/Modern-NetworkLayer.md
babangsund: In fact, I think it makes sense to change all occurrences. i.e. no need for the `as` alias.

babangsund:
You should update v6 docs as well... :-)

sibelius:
I don't think we update old docs

babangsund:
> I don't think we update old docs

I don't see any reason not to keep "old" docs in good shape? :|

kassens:
Yeah, updating old versions of the docs is a bit to involved. We just need to update the version more frequently to release the fixes.

Thanks for the fix, I'll update it to remove the `as`. We probably going to reduce the amount of prefixed modules slowly. It's kind of repetitive to have `Relay...` everything.

PR #2846, opened 2019-09-09
Curry printRequireModuleDependency with extension; add relative ./ path

babangsund:
Suggested fix for https://github.com/facebook/relay/issues/2713

PR #2861, opened 2019-09-19
add missing .

babangsund:
Oops. https://github.com/facebook/relay/pull/2846#issuecomment-533221917

### DIFF
--- a/packages/relay-compiler/codegen/RelayFileWriter.js
+++ b/packages/relay-compiler/codegen/RelayFileWriter.js
@@ -22,7 +22,7 @@ const graphql = require('graphql');
 const invariant = require('invariant');
 const nullthrows = require('nullthrows');
 const path = require('path');
-const writeRelayGeneratedFile = require('./writeRelayGeneratedFile');
+const {writeRelayGeneratedFile} = require('./writeRelayGeneratedFile');
 
 const {
   getReaderSourceDefinitionName,
--- a/packages/relay-compiler/codegen/RelayFileWriter.js
+++ b/packages/relay-compiler/codegen/RelayFileWriter.js
@@ -22,7 +22,7 @@ const graphql = require('graphql');
 const invariant = require('invariant');
 const nullthrows = require('nullthrows');
 const path = require('path');
-const {writeRelayGeneratedFile} = require('./writeRelayGeneratedFile');
+const writeRelayGeneratedFile = require('./writeRelayGeneratedFile');
 
 const {
   getReaderSourceDefinitionName,
--- a/packages/relay-compiler/codegen/__tests__/writeRelayGeneratedFile-test.js
+++ b/packages/relay-compiler/codegen/__tests__/writeRelayGeneratedFile-test.js
@@ -1,0 +1,27 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @emails oncall+relay
+ * @flow strict-local
+ * @format
+ */
+
+'use strict';
+
+const {
+  createPrintRequireModuleDependency,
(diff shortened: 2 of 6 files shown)


mrtnzlml:
OK, you did it faster, thanks. I'd personally add the test though since it's a fix of a fix. 😅

sibelius:
can you add a test?

babangsund:
Probably should. :-)

babangsund:
> Probably should. - although uncertain exactly what you want to test?

Well, here's a tiny test for `createPrintRequireModuleDependency` 🤷‍♂

packages/relay-compiler/codegen/RelayFileWriter.js
```diff
@@ -24,7 +24,7 @@ const graphql = require('graphql');
 const invariant = require('invariant');
 const nullthrows = require('nullthrows');
 const path = require('path');
-const writeRelayGeneratedFile = require('./writeRelayGeneratedFile');
+const {writeRelayGeneratedFile} = require('./writeRelayGeneratedFile');
```
sibelius: why this?

babangsund: So I could export the function we want to test 🤔

sorenhoyer:
After this is merged, would it be possible to make a patch version and do an npm publish, so we don't have to build relay-compiler from master?

mrtnzlml:
@sorenhoyer I think it still doesn't work anyway. We have to wait a bit more. I tried multiple times in the past but seems like refetch container still requires the last argument (which makes this unusable - at least for OSS).

sibelius:
@mrtnzlml refetch on typescript is broken because of this https://github.com/facebook/relay/issues/2874

mrtnzlml:
@sibelius Mm, so it won't work for custom compilers using ESM as well, right? (I use ESM generated output with Flow)

sibelius:
after this https://github.com/facebook/relay/pull/2875 it will

packages/relay-compiler/codegen/RelayFileWriter.js
```diff
@@ -23,7 +23,7 @@ const graphql = require('graphql');
...
```
alunyov: Yeah, this, unfortunately, won't work with our internal FB compiler setup. 

So we should probably keep default export from this `writeRelayGeneratedFile`.

packages/relay-compiler/codegen/__tests__/writeRelayGeneratedFile-test.js
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @emails oncall+relay
+ * @flow strict-local
+ * @format
+ */
+
+'use strict';
+
+const {
+  createPrintRequireModuleDependency,
```
alunyov: We should probably make `createPrintRequireModuleDependency` a standalone module, and import it to `writeRelayGeneratedFile`.

packages/relay-compiler/codegen/writeRelayGeneratedFile.js
```diff
@@ -169,4 +169,7 @@ function extractHash(text: ?string): ?string {
   return match && match[1];
 }
 
-module.exports = writeRelayGeneratedFile;
+module.exports = {
+  createPrintRequireModuleDependency,
+  writeRelayGeneratedFile,
+};
```
alunyov: Let's keep the old version of the export

packages/relay-compiler/codegen/__tests__/writeRelayGeneratedFile-test.js
babangsund: This can fortunately be done in a jiffy. :-)

babangsund:
> Thanks for adding the tests!

My pleasure.

I've moved `createPrintRequireModuleDependency` into a standalone module, renamed the test, and reverted the import/exports.

PR #2862, opened 2019-09-19
[docs] Only search in current version

BorisTB:
Prevent showing search results across all doc versions and show only relevant links

Screenshot 2019-09-19 at 22 23 36

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -154,6 +154,9 @@ const siteConfig = {
   algolia: {
     apiKey: '3d7d5825d50ea36bca0e6ad06c926f06',
     indexName: 'relay',
+    algoliaOptions: {
+      facetFilters: ['version:VERSION']
+    }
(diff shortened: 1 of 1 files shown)


kassens:
I don't really know how this works, have you copied this from somewhere else?

Something else with the algolia config is wrong in that it doesn't seem to index the current version at all (we renamed it to `QueryRenderer` without the `< />` wrapper.

mrtnzlml:
@kassens The behavior is described here: https://docusaurus.io/docs/en/search#extra-search-options

> You can also specify extra search options used by Algolia by using an algoliaOptions field in algolia. This may be useful if you want to provide different search results for the different versions or languages of your docs. Any occurrences of "VERSION" or "LANGUAGE" will be replaced by the version or language of the current page, respectively...

kassens:
Thanks for the link!

PR #2865, opened 2019-09-20
Don't assert variables in context

robrichard:
Doesn't seem like variables are used any more in relay context, so no need to assert for them?


kassens:
Nice

PR #2866, opened 2019-09-23
Allow language plugin to specify how modules should be named.

zth:
This allows the language plugin to specify how modules should be named.

Context: In Reason, `.` is not allowed in file names except for with the file extension. This means that the language plugin I'm working on for Reason needs a way to have modules/generated files be named `OperationName_graphql.re` instead of `OperationName.graphql.js` in order for it to work.

Please let me know what you think!


zth:
@alunyov @kassens hey! Sorry for pinging you, but is there anything we can do to move this PR along? A similar one has popped up in https://github.com/facebook/relay/pull/3015 , so there's definitively some demand for a feature like this in the language plugin interface :)

Thanks!

alloy:
cc @jstejada

zth:
@alloy awesome! Rebasing right away :sweat_smile:

zth:
@alunyov https://github.com/facebook/relay/pull/2810 this one is related to this too, if you have a chance to take a look at that as well while we're at it?

PR #2881, opened 2019-10-08
[Docs] Update fetchQuery to include cacheConfig documentation

damassi:
Noticed that `fetchQuery` is missing docs around `cacheConfig`, which is the fourth slot [per the source](https://github.com/facebook/relay/blob/master/packages/relay-runtime/query/fetchQuery.js#L32).


docs/Modern-fetchQuery.md
```diff
@@ -32,6 +32,7 @@ fetchQuery(environment, query, variables)
 * `environment`: The [Relay Environment](./relay-environment.html)
 * `query`: The `graphql` tagged query. **Note:** `relay-compiler` enforces the query to be named as `<FileName>Query`.
 * `variables`: Object containing set of variables to pass to the GraphQL query, i.e. a mapping from variable name to value.
+* `cacheConfig?`: Optional object containing a set of cache configuration options, i.e. `force: true` requires the fetch to be issued regardless of the state of any configured response cache.
```
alloy: Are there currently any other options? It might be a little confusing to led the reader to believe the set of options may have more than 1 member.

Issue #2566, opened 2018-11-06
Bug: Cache in SSR


HsuTing:
I found [requestCache](https://github.com/facebook/relay/blob/master/packages/react-relay/modern/ReactRelayQueryRenderer.js#L53) caused a bug.
In server side, [componentDidUpdate](https://github.com/facebook/relay/blob/master/packages/react-relay/modern/ReactRelayQueryRenderer.js#L245) never be called. As a result, `requestCache` never be cleaned.

I used [react-relay-network-modern-ssr](https://github.com/relay-tools/react-relay-network-modern-ssr) in my project and this was ok when I loaded the website for the first time. This queried data correctly. However, after I reload the website, this did not request again to get the newest data.

Because `nodejs` never import each module again, `requestCache` will never be cleaned in `server`.

jstejada:
hey @HsuTing , which version are you using?

HsuTing:
@jstejada 
`react-relay@^1.7.0`
Do you need the other version of the packages?

HsuTing:
Same problem

sibelius:
@HsuTing can you try to come up with a failing test ?

robrichard:
Could this request cache also potentially leak data to unauthorized users? If the same query is issued with different environments with different authorization levels, the user with less authorization could potentially get the cached response from the user with higher authorization?

HsuTing:
@sibelius 
[here](https://github.com/HsuTing/relay-bug-example) is a simple example.
If you need the more information, let me know 😸 .

crash7:
Hi all! I can confirm the issue too with a different implementation of Relay SSR than the used in @HsuTing example.

The real problem is that the requestCache[key] is only cleared in `componentDidMount`, which never happens on SSR: 
https://github.com/HsuTing/relay-bug-example/blob/master/src/utils/createEnvironment/server.js

Because of that, the "request" is always in flight: https://github.com/facebook/relay/blob/master/packages/react-relay/modern/ReactRelayQueryRenderer.js#L365

Would be nice to have a way to `force` that fetch someway, maybe with an extra prop in the QueryRender component that I can set when `process.browser` is `false`.

m64253:
I think having "state" like this outside of the component instance is always going to an issue server side?

Maybe this can be moved to the network? 

Would love to help to resolve this but I'm a bit uncertain on what the actual problem being solved by `requestCache` is, simultaneous request duplication?

edvinerikson:
Just speculating.. 

[...]

PR #2883, opened 2019-10-14
Disable global request cache when QueryRenderer runs on a server

psaia:
This is an immediate fix to the issue discussed here: https://github.com/facebook/relay/issues/2566

### DIFF
--- a/packages/relay-runtime/query/fetchQueryInternal.js
+++ b/packages/relay-runtime/query/fetchQueryInternal.js
@@ -32,7 +32,7 @@ type RequestCacheEntry = {|
   +subscription: Subscription,
 |};
 
-const requestCachesByEnvironment = new Map();
+const requestCachesByEnvironment = new WeakMap();
 
 /**
  * Fetches the given query and variables on the provided environment,
--- a/packages/react-relay/ReactRelayQueryRenderer.js
+++ b/packages/react-relay/ReactRelayQueryRenderer.js
@@ -52,7 +52,7 @@ export type RenderProps<T> = {|
  * constructor. If a request is already in flight from a previous call to the
  * constructor, just reuse the query fetcher and wait for the response.
  */
-const requestCache = {};
+let requestCache = {};
 
 export type Props = {|
   cacheConfig?: ?CacheConfig,
@@ -103,6 +103,13 @@ class ReactRelayQueryRenderer extends React.Component<Props, State> {
 
     let queryFetcher;
     let requestCacheKey;
+
+    // If there's not window, we're on the server and do not want a global
+    // request cache to persist.
+    if (typeof window !== 'object') {
+      requestCache = {};
+    }
+
     if (props.query) {
       const {query} = props;
 
--- a/packages/react-relay/ReactRelayQueryRenderer.js
+++ b/packages/react-relay/ReactRelayQueryRenderer.js
@@ -114,7 +114,7 @@ class ReactRelayQueryRenderer extends React.Component<Props, State> {
     let requestCacheKey;
 
     // When server-side rendering the global request cache should not persist.
-    if (!isBrowser) {
+    if (!isBrowser()) {
       requestCache = {};
     }
 
--- a/packages/react-relay/ReactRelayQueryRenderer.js
+++ b/packages/react-relay/ReactRelayQueryRenderer.js
@@ -54,6 +54,15 @@ export type RenderProps<T> = {|
  */
 let requestCache = {};
 
+/**
+ * isBrowser will be true in *real* browser environments. Unfortunately,
+ * the best way to detect this in all environments is to use `new Function`.
+ * As a result, eslint needs to be disabled on this to allow for eval.
+ */
+const isBrowser = new Function( // eslint-disable-line
+  'try {return this===window;}catch(e){ return false;}',
+);
+
 export type Props = {|
   cacheConfig?: ?CacheConfig,
   fetchPolicy?: FetchPolicy,
@@ -104,9 +113,8 @@ class ReactRelayQueryRenderer extends React.Component<Props, State> {
     let queryFetcher;
     let requestCacheKey;
 
-    // If there's not window, we're on the server and do not want a global
-    // request cache to persist.
-    if (typeof window !== 'object') {
+    // When server-side rendering the global request cache should not persist.
+    if (!isBrowser) {
       requestCache = {};
     }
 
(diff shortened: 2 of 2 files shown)


snrbrnjna:
I can confirm, that this PR fixes the problems with relay when QueryRenderer runs on the server. But the fix is incomplete: in some environments or circumstances, there is a `window` object in the server. A 100% reliable version to check if the runtime is nodejs or a browser is this:

`const isBrowser=new Function("try {return this===window;}catch(e){ return false;}");`
For details on this, see https://stackoverflow.com/a/31090240

So please @psaia modify your PR to check if it is run not in the browser with that approach. Apart from that, I hope this will make it soon into an offical release. I assume until then SSR with relay isn't working out for anyone out there.

snrbrnjna:
@facebook-github-bot why are the Checks for this PR failing?

babangsund:
You can look at the test results. For example: https://travis-ci.org/facebook/relay/jobs/597863063#L2389

psaia:
@snrbrnjna Thanks for that. I've pushed an update.

psaia:
I finally had some time to dig into these failing tests. They're passing now. Some notes:

- I used the `new Function` trick you referenced @snrbrnjna 
- The request cache will be disabled when running on a server but **not** when jest runs so the tests simulate the browser as much as possible. This is what caused the tests to fail.

psaia:
@babangsund Any chance you could help get this merged? It would greatly speed up our builds (since we wouldn't need to use my fork).

marc-se:
Will this be merged on Christmas? 🎄🎅🏼
I wish for nothing more.

mattkrick:
@psaia do you call a fetch function on the server?
In my use case there were 2 retainers holding onto the queryFetcher. The 2nd fix was to replace `Map` with `WeakMap` here:
https://github.com/facebook/relay/blob/master/packages/relay-runtime/query/fetchQueryInternal.js#L37

curious how you guys are doing it!

psaia:
@mattkrick Not directly, no. On the server `<QueryRenderer />` from `react-relay` is abstracting the calls (just as it does on the client) - which obviously wraps `fetchQuery`. There isn't much server-specific code. 

However, your fix looks interesting. I'm curious why a "2nd fix" was necessary? What was the side effect you were seeing that the downstream Map/cache was causing? After debugging, it looked as though the issues discussed in #2566 were specifically caused by this `ReactRelayQueryRenderer` component. Though, if there's an SSR issue with `fetchQuery` all higher level components would be effected as well.

Unrelated, but I'm curious if Facebook plans on deprecated this `ReactRelayQueryRenderer` entirely in favor of hook components which is why there's a lack of urgency here.

We have a single wrapper which abstracts the query renderer at the moment since I figure this API could change at any moment:

```typescript
export default function withData<T>(
  ComposedComponent,
  query: T,
  variables: any = {}
) {
  const relayContext = useContext(ReactRelayContext);
  return (
    <QueryRenderer<T>
      fetchPolicy={"store-and-network"}
      environment={relayContext.environment}
      query={query}
      variabley={variables}
      render={({ error, props }) => {
        if (error) {
          return <div>{error.message}</div>;
        } else if (props) {
          return (
            <ComposedComponent
              {...props}
              environment={relayContext.environment}
            />
          );
        } else {
          return <div>Loading</div>;
        }
      }}
    />
  );
}
```

mattkrick:
@psaia the only answer I have for "why the 2nd fix" is the heap dump told me so 🙈 
Screen Shot 2019-12-13 at 2 58 46 AM

Maybe we SSR differently? Mine is a 3-step process:
```js
const environment = new ServerEnvironment()
const element = <Foo environment={environment}/>
ReactDOMServer.renderToStaticMarkup(element) // 1) attempt to render, relay inits fetches
await environment.load() // 2) wait for fetches to turn into results
return ReactDOMServer.renderToStaticMarkup(element) // 3) render again, now with results
```

My environment for SSR is basically this:
```js
class ServerEnvironment extends Environment {
  async load() {
    if (!this.isFetched) {
      this.isFetched = true
      this.result = await this.request
    }
  }

  fetch = (request, variables) => {
    if (!this.isFetched) {
      this.request = callGraphQLServer(request, variables)
    } else {
      return this.result
    }
  }
}
```

psaia:
@mattkrick Mine is similar except for using [react-relay-network-modern](https://github.com/relay-tools/react-relay-network-modern-ssr/blob/master/src/server.js). Here's a [consolidated gist](https://gist.github.com/psaia/4c3800e58a66a3e787442d58d03ed436).

mattkrick:
@psaia looks pretty similar! not readily apparent how you avoided that internal fetch function, but glad it didn't hit ya! if you'd like to add the `WeakMap` addition to your PR i wouldn't be opposed, otherwise i'm happy to submit one later

psaia:
@mattkrick I'd like to fully understand the double fetch issue. Did this happen on the client or server or both? 

Currently there is a double fetch by design (at least it works well for me): one on the server, another post-hydration on the client.

mattkrick:
@psaia 
in my case i'm not rehydrating, i'm just using it to create some HTML for an email.
You can reproduce this on the client, too though. Provided you create a new environment when someone logs out & logs back in, each new environment will get stored in `requestCachesByEnvironment`. Today, you can mitigate that problem by just clearing the store on logout & reusing the environment container.

here's the internal flow (as i understand it)
you pass in a fetch function to the network layer.
when relay determines it needs to fetch, it checks to see if that fetch is a dupe (e.g. it's already in flight). to do this, it looks up all requests for that particular environment in a map called `requestCachesByEnvironment`.

IMO that requestCache should probably be kept on the environment itself, but that PR would be more than 1 line 😄

psaia:
@mattkrick Thanks for explaining. That makes sense. I could see that causing a memory leak as well.

I was able to reproduce and test so I went ahead and committed.

psaia:
@mattkrick Thanks for explaining. That makes sense. I could see that causing a serious memory leak.

I was able to reproduce and test so I went ahead and committed.

ccemeraldeyes:
When will this be merged? It's passing tests and a desired feature.

packages/relay-runtime/query/fetchQueryInternal.js
```diff
@@ -32,7 +32,7 @@ type RequestCacheEntry = {|
   +subscription: Subscription,
 |};
 
-const requestCachesByEnvironment = new Map();
+const requestCachesByEnvironment = new WeakMap();
```
maraisr: This should be wrapped in a `WEAKMAP_SUPPORTED` if block. 

cc  @ccemeraldeyes @psaia

maraisr:
I don't think this should be merged - the reuse of an environment, across SSR requests, is a bad idea, philosophically. I believe the _memory leak_ issue with the retainers come about from using `Maps` to that hold onto Environments. If we can garbage collect those environments, the store naturally get's evicted. Rather than now wrapping a `isBrowser` block.

Please see my PR @ #3014 which has solved this SSR problem for us, as we've been running a fork with that implemented.

cc @ccemeraldeyes @psaia

PR #2889, opened 2019-10-17
Publish master releases from CI to NPM

kassens:
Adds a GitHub action that runs only on pushes to the master branch on facebook/relay. This step pushes a build to the master tag on npm which can be installed with `yarn add relay-runtime@master` or similar.
The master releases have a version like `0.0.0-master-62d0ae50` which includes the Git commit hash for easy reference.

Test Plan:
So far, I did a few test releases to https://www.npmjs.com/package/release-test-react-relay?activeTab=versions, but to test the final code, we'll have to push this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,11 +4,11 @@ on: [push, pull_request]
 
 jobs:
   build:
-    name: Build and Tests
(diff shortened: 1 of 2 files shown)


kassens:
The only thing missing now is hopefully the NPM_TOKEN secret which we should get added soon!

PR #2890, opened 2019-10-17
Fix a bug in the master release condition

kassens:
The `ref` for pushes to master is `refs/heads/master`.


sorenhoyer:
First master version published. Really excited about this! Gj. :-)

PR #2919, opened 2019-10-31
Fixed typos on relay experimental package

rsmelo92:
While studying these packages found some typos.... I think it's very minor, but the less broken windows the better


kassens:
Thank you!

PR #2924, opened 2019-11-03
Allow non-nullable Viewer field in ViewerQueryGenerator

janicduplessis:
This adds support for non-nullable `Viewer` field by using `getNullableType` before using the type of the viewer field.

Also make sure that when generating the refetch operation we use the actual type of the field (`viewerField.type` instead of `viewerType`) to ensure nullability matches the one in the schema.


janicduplessis:
Is there an example of a test using a different schema? I wanted to add one but I don’t want to change the type of the viewer field for all tests.

kassens:
Creating tests with a different schema isn't super easy. I'll see if I can add something when I import this.

PR #2934, opened 2019-11-08
Upgrade yargs to 14.2.0

sugarshin:
`mem` package vulnerability found.

> In nodejs-mem before version 4.0.0 there is a memory leak due to old results not being removed from the cache despite reaching maxAge. Exploitation of this can lead to exhaustion of memory and subsequent denial of service.

ref: https://github.com/sindresorhus/mem/commit/da4e4398cb27b602de3bd55f746efa9b4a31702b

`relay-compiler` depends on that.

```
relay-compiler@6.0.0
  └─┬ yargs@9.0.1
    └─┬ os-locale@2.1.0
      └── mem@1.1.0
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "signedsource": "^1.0.0",
     "webpack": "^4.30.0",
     "webpack-stream": "^5.1.1",
-    "yargs": "^9.0.0"
+    "yargs": "^14.2.0"
   },
   "private": true,
   "devEngines": {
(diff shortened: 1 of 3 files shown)


kassens:
Thank you, I tested the compiler and it seems like the there were no API changes affecting us!

kassens:
I had to end up also updating React Native and Metro as we want to avoid multiple versions.

sugarshin:
@kassens Thank you! Would you please publish a new version to npm

PR #2935, opened 2019-11-08
Allow language plugin to add schema extensions.

zth:
This allows the language plugin to add schema extensions to the compiler. It's allowed to add its own transforms as of now, but there's no way of adding schema extensions in addition to the transforms as of now. This diff fixes that.

### DIFF
--- a/packages/relay-compiler/language/RelayLanguagePluginInterface.js
+++ b/packages/relay-compiler/language/RelayLanguagePluginInterface.js
@@ -36,6 +36,7 @@ export type PluginInterface = {
   findGraphQLTags: GraphQLTagFinder,
   formatModule: FormatModule,
   typeGenerator: TypeGenerator,
+  schemaExtensions?: $ReadOnlyArray<string>,
 };
 
 /**
--- a/packages/relay-compiler/bin/RelayCompilerMain.js
+++ b/packages/relay-compiler/bin/RelayCompilerMain.js
@@ -32,7 +32,7 @@ const {
   fragmentTransforms,
   printTransforms,
   queryTransforms,
-  schemaExtensions,
+  schemaExtensions: relaySchemaExtensions,
 } = RelayIRTransforms;
 
 import type {ScalarTypeMapping} from '../language/javascript/RelayFlowTypeTransformers';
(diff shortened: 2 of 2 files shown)


alunyov:
Hey @zth thank you for working on this!

Can you provide a little bit more context on why do we need this, what is your use-case?
Are there any additional unit-tests can be added to better illustrate this use-case.

Thank you!

zth:
Hello @alunyov ,

Thanks for having a look! I'll look into appropriate places to add unit tests, but let me illustrate the use case I have:

I'm working on a way to use Relay with ReasonML, and I want to add a custom directive called `@reasonRecord`, which when used on an object or fragment changes the type emitted.

I'm able to do what I need through making a custom transform that picks up and transforms the directive. That transform can then be added via the language plugin (as [here](https://github.com/facebook/relay/blob/master/packages/relay-compiler/language/javascript/RelayFlowGenerator.js#L1105) in the official Flow plugin), but there's currently no way to add the actual schema extension I need (`directive @reasonRecord on OBJECT | FRAGMENT_DEFINITION | FIELD`). `SCHEMA_EXTENSION` in that same file does nothing from what I can tell, and when tracking down how it works internally in the compiler, I've found no way to inject my own schema extensions without maintaining my own `RelayCompilerMain.js`, which isn't a great solution I think.

Are you following what I'm looking for here? Maybe there's a better way to achieve this?

packages/relay-compiler/language/RelayLanguagePluginInterface.js
```diff
@@ -36,6 +36,7 @@ export type PluginInterface = {
   findGraphQLTags: GraphQLTagFinder,
   formatModule: FormatModule,
   typeGenerator: TypeGenerator,
+  schemaExtensions?: $ReadOnlyArray<string>,
```
alunyov: I'm not sure this should be in the PluginInterface.
Can we move it to the compiler configs, instead?

We have this `type Config {...}` in the RelayCompilerMain.js - maybe we could add another option there, next to the `customScalars?: ScalarTypeMapping`.

And then you can provide those directive extensions via `relay-config`, right?

zth: Thanks for the review!

Hmm, I'd actually like to argue that it belongs in `PluginInterface` or something exposed by the language plugin. This because it (or at least the particular thing I'm trying to do) is deeply integrated with the actual language plugin (and only makes sense in the language plugin context). Adding it to `relay-config` would mean that every user would have to configure it themselves in addition to using the language plugin, or am I missing something?

alunyov: Oh, I guess you're right! I didn't think about this particular use-case.

Issue #2804, opened 2019-07-18
Feature: shared object between mutation and network


morrys:
I wanted to propose the addition of an optional parameter in commitMutation shared with the network interface, in order to allow add extra information during the execution of the mutation.

This parameter will be very useful in offline contexts, but I think it is also useful for other use cases.

sibelius:
Can’t you use updater?

morrys:
No, because that function is not used by the network.

https://github.com/facebook/relay/blob/master/packages/relay-runtime/store/RelayModernEnvironment.js#L327

perhaps an example makes it clearer:

as is
```
const source = this._network.execute (
         operation.node.params,
         operation.variables,
         {force: true},
         uploadables,
       );
```

to be

```
const source = this._network.execute (
         operation.node.params,
         operation.variables,
         {force: true},
         uploadables,
         extra
       );
```

josephsavona:
This is what what `CacheConfig`  (third arg) is intended for.  The `metadata` can be any object w whatever info  you need.

morrys:
Great news, my problem is solved :) maybe it could be useful to add this property in commitMutation too:

```
commitMutation(environment, {
    mutation,
    variables: {
      input,
    },
    extra
})
```
and to modify RelayModernEnvironment to use it as you suggested:

https://github.com/facebook/relay/blob/master/packages/relay-runtime/store/RelayModernEnvironment.js#L330

`{force: true, metadata: extra }`

what do you think?

josephsavona:
Yeah, totally reasonable to add CacheConfig as an optional arg to those methods. For executeMutation it’s a bit awkward though bc force:false doesn’t make sense.

morrys:
Perhaps it is sufficient to create a subtype with a different name :)

PR #2937, opened 2019-11-09
added metadata option in commitMutation & requestSubscription

morrys:
this PR allows you to set the CacheConfig metadata parameter directly from commitMutation.

issue: https://github.com/facebook/relay/issues/2804


mrtnzlml:
Hola @morrys, this would be very helpful for me as well! :) Do you think it makes sense to add it to `requestSubscription` as well? 🤔

morrys:
hi @mrtnzlml,
being a simple change that does not involve any breaking change it makes sense to adapt the requestSubscription too :)

During the week I check and update the PR.

morrys:
hi @mrtnzlml,
I updated the PR :)

morrys:
@jstejada @josephsavona any news about this PR?

morrys:
Friendly ping @jstejada @josephsavona

jstejada:
@morrys if this metadata is intended to be passed directly to the cache config, do you think it'd make sense to rename it to `networkMetadata` or similar?

morrys:
@jstejada with this PR it is used for this purpose but in my opinion with a more generic name it leaves open the possibility of using it for other purposes.

PR #2950, opened 2019-11-25
Add environment prop

EgorShum:
Seems like QueryRenderer has lost an "environment" prop


zth:
I think QueryRenderer now consumes the environment from context instead of a prop, right?

sibelius:
not yet https://github.com/facebook/relay/blob/master/packages/react-relay/ReactRelayQueryRenderer.js#L123

kassens:
Sorry for overlooking this  a while :(

`QueryRenderer` API hasn't changed and is indeed missing the environment prop. When switching to `useQuery` in the future, the env will be provided by the environment provider.

Issue #2829, opened 2019-08-19
feature: Change the retain function in RelayModernStore


morrys:
I wanted to propose to modify the [retain function in RelayModernStore](https://github.com/facebook/relay/blob/master/packages/relay-runtime/store/RelayModernStore.js#L140) so that it can receive the following information:
* the CacheConfig parameter of the query
* a boolean that indicates if the method is called by [lookupInStore](https://github.com/facebook/relay/blob/master/packages/react-relay/ReactRelayQueryFetcher.js#L296) or during the execution of the query in the [network](https://github.com/facebook/relay/blob/master/packages/react-relay/ReactRelayQueryFetcher.js#L103)

**as is**
```ts
retain (selector: NormalizationSelector): Disposable
```

**to be**
```ts
retain (selector: NormalizationSelector, retainConfig: {network: boolean, cacheConfig ?: CacheConfig}): Disposable
```

This change allowed me to extend the [store](https://github.com/morrys/wora/blob/master/packages/relay-store/src/Store.ts#L62) and implement the TTL management of the queries (we can expect to open an issue to bring this management natively in relay).

Thanks,
Lorenzo

josephsavona:
Hmm, we're actively exploring new features to allow finer grained control of cache invalidation, and  also looking at changes to the Store API to support a new approach to loading connections. We  considered query-level TTL but aren't convinced that's the right long-term API, so I think we need to understand the use-case  better - probably the first step  is documenting our thinking around cache invalidation so that you and others can give feedback.

morrys:
Hi @josephsavona, I have noticed many changes in the management of the Store and I would like to discuss some of its aspects before all the new experimental features become official.

*  A change that I really appreciate is the new management of roots in the retain function:
It is no longer managed through an index but using [the new property](https://github.com/facebook/relay/blob/master/packages/relay-runtime/store/RelayModernStore.js#L216) which identifies the request, very similar to the implementation in the [store](
https://github.com/morrys/wora/blob/master/packages/relay-store/src/Store.ts#L49) I use to manage the persistence of the Relay store.

* TTL vs gcReleaseBufferSize
This theme is certainly the most complex to deal with and for the moment extending the retain function is the simplest solution.

* A change that would **force me to have to fork the Store** is the evolution of the object saved in the roots and its strong use in the Store.
[...]

PR #2959, opened 2019-12-01
feat data store abstraction & configuration

morrys:
This PR allows you to manage the data that guarantees the consistency of the store in a similar way to how the source records are managed.

https://github.com/facebook/relay/issues/2829

its example of how this configuration can be used is given by the library is [wora/cache-persist](https://morrys.github.io/wora/docs/cache-persist.html) used in [react-relay-offline](https://github.com/morrys/react-relay-offline) to manage persistence and offline capabilities.

In short, `wora/cache-persist` is an object that allows you to manage interfacing with storage (localStorage, sessionStorage, indexedDB, AsyncStorage & any custom storage) in "synchronous" mode.
This result is achieved by combining:
 * the initial **restore** of data persisted in an inmemory object (this is asynchronous but a reconciliation is made between the initial state and the persisted state)
 * after restore, **data is read and written synchronously in the inmemory object** (optimizes performance)
 * each mutation of the object in memory is inserted in a **queue** managed asynchronously (the storage update interval is configurable, throttle default 500ms)
 * at the end of the interval a **merge** of the mutations is made in order to avoid repeated writing
 * the writes are grouped and the **multiSet** function of the storage is invoked (better performance if this function is natively implemented as for asyncstorage)
 * removals are grouped and the storage **multiRemove** function is invoked (better performance if this function is natively implemented as for asyncstorage)

### DIFF
--- a/packages/relay-runtime/store/RelayMapStoreSource.js
+++ b/packages/relay-runtime/store/RelayMapStoreSource.js
@@ -1,0 +1,58 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict-local
+ * @format
+ */
+
+// flowlint ambiguous-object-type:error
+
+'use strict';
+
+import type {DataID} from '../util/RelayRuntimeTypes';
+import type {StoreSource} from './RelayStoreTypes';
+
+/**
+ * An implementation of the `StoreSource` interface (defined in
+ * `RelayStoreTypes`) that holds all records in memory (JS Map).
+ */
+class RelayMapStoreSource<I, T> implements StoreSource<I, T> {
+  _stores: Map<I, ?T>;
+
+  constructor() {
+    this._stores = new Map();
+  }
+
+  clear(): void {
+    this._stores = new Map();
+  }
+
+  delete(id: I): void {
+    this._stores.set(id, null);
+  }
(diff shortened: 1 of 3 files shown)


josephsavona:
Hmm, I see what you're trying to achieve here but I think we need to discuss the approach more. Store performance is sensitive and we need to be careful about introducing indirections such as StoreSource here. For example, `publish()` is a point at which you can observe modifications to the store and persist updates accordingly.

morrys:
Hi @josephsavona,
I agree with you on discussing it. So I wait to know where it is better to continue the discussion before making further comments ok?

Let me know,
thanks

josephsavona:
Feel free to discuss here, i just wonder if a more real-time conversation could help too.

morrys:
* **regarding the performances:** I agree that we must be careful about the performances but what I have marked as `StoreSource (roots & connectionEvents)` are not used frequently with respect to how much the `MutableRecordSource` is used. The proposed change replaces a Map with a wrapper over the Map (we could consider managing the data inside the StoreSource with a plain object instead of the Map)
* **on using the publish() function to monitor and update the persistence:** in my opinion this is not enough for two reasons:
   * the StoreSource are modified during retain and GC
   * within the updateTargetFromSource function there are many important logics to guarantee the     consistency of the store

As for how this is used by **wora/cache-persist** I am also available to provide more details on why it is an efficient and correct way for me to manage Relay persistence (and not only Relay ).

maraisr:
What is the latest on this?

PR #2966, opened 2019-12-11
Git source control

edvinerikson:
Hello! 
At my company we commit Relay artifacts to source control, however its currently a manual process and we are looking for ways to automate this. This change adds support for git source-control and it also automatically enables it if the compiler is being executed inside a git repo.

However the automatic activation is likely something we need to adjust, let me know what you think.

### DIFF
--- a/packages/relay-compiler/bin/RelayCompilerMain.js
+++ b/packages/relay-compiler/bin/RelayCompilerMain.js
@@ -42,6 +42,7 @@ const {
   schemaExtensions: relaySchemaExtensions,
 } = RelayIRTransforms;
 
+import type {SourceControl} from '../codegen/SourceControl';
 import type {ScalarTypeMapping} from '../language/javascript/RelayFlowTypeTransformers';
 import type {WriteFilesOptions} from '../codegen/CodegenRunner';
 import type {
--- a/packages/relay-compiler/bin/RelayCompilerMain.js
+++ b/packages/relay-compiler/bin/RelayCompilerMain.js
@@ -27,6 +27,11 @@ const invariant = require('invariant');
 const path = require('path');
 
 const {buildClientSchema, Source, printSchema} = require('graphql');
+const {spawnSync} = require('child_process');
+const {
+  SourceControlGit,
+  SourceControlMercurial,
+} = require('../codegen/SourceControl');
 
 const {
   commonTransforms,
@@ -342,13 +347,21 @@ function getCodegenRunner(config: Config): CodegenRunner {
       baseParsers: ['graphql'],
     },
   };
+
+  let sourceControl: SourceControl | null = null;
+
+  if (isGitSourceControl()) {
+    sourceControl = SourceControlGit;
+  } else if (isMercurialSourceControl()) {
+    sourceControl = SourceControlMercurial;
+  }
+
   const codegenRunner = new CodegenRunner({
     reporter,
     parserConfigs,
     writerConfigs,
     onlyValidate: config.validate,
-    // TODO: allow passing in a flag or detect?
-    sourceControl: null,
+    sourceControl,
   });
   return codegenRunner;
 }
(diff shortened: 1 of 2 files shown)


maraisr:
I highly doubt this would be something that the compiler would do, but may I suggest something like a post-relay-compiler script that simply `git add`'s all the `__generated__` folders.

eg: `git add __generated__`  in a package.json

```
{
  "scripts": {
	"relay": "relay-compiler",
	"postrelay": "git add __generated__"
  }
}
```

at my company we also add all generated artifacts to source control, only we have a husky commit hook that runs `relay-compiler --validate`, with some extra hooks that with a sensible `.gitignore` we simply just run `git add .`

edvinerikson:
> I highly doubt this would be something that the compiler would do, but may I suggest something like a post-relay-compiler script that simply git add's all the __generated__ folders.

Not sure I follow you here, are you saying that the compiler should not add/remove files? Judging from what's already in the codegen runner, it seems that it does support this, but lacks the git integration which is what I suggest that we add. 

To me it sounds reasonable to make this feature available to OSS too. We can of course do it with scripts and workarounds outside of the Relay compiler but I rather bake it in directly to avoid corner cases. Given that the feature already exists I am sure there is a good reason for that. 😊

maraisr:
I guess we're each entitled to an opinion - but afaik Facebook don't use git internally, so there wouldn't be an appetite to add it. And quite frankly, it isn't he compiler's job to integrate with git.

kassens:
Thanks for integrating this in OSS. At FB we also commit generated files as a way to skip build steps (Relay is just one of many potential build steps).

I'm curious to hear what others think? I suppose if you didn't want to commit the generated files, you'd have the generated files in `.gitignore`?

It seems like `git` errors if we'd try to `git add` the file and it's an ignored file:
```
$ git add generated/b.js                                                                                                                                                                                                            The following paths are ignored by one of your .gitignore files:
generated/b.js
Use -f if you really want to add them.
$ echo $?                                                                                                                                                                                                                         1
```

Which I think would translate into a compiler failure. Maybe the right behavior would be to somehow check if the files are visible to `git` first?

tbergq:
Would this mean that the default behaviour after I run relay compiler is that it does `git add` on all changed files?

taion:
We use gitignore for our generated files.

kassens:
@tbergq Yes, the current Git implementation would do that.

kassens:
hg doesn't have the "staging area" of git where you first `git add` stuff before `git commit` to create a commit.

In hg if a file was just created or deleted without using the `hg rm` command, a simple `hg commit -m "changes"` doesn't pick up deletions or new files, but all edits. This made it easy for people to forget adding the generated files to their commits.

Since we're not `git add`ing changes to the staging area for *changes* to existing files the compiler produces,  I think `git add --intent-to-add NEW_OR_REMOVED_FILES` would be the better command. This makes it so a simple `git ci -a -m "my changes"` would add the files, but they'd still show up if you did `git diff`.

Thoughts from folks who use git more actively?

edvinerikson:
> I'm curious to hear what others think? I suppose if you didn't want to commit the generated files, you'd have the generated files in .gitignore?

Yeah, it sounds reasonable that the files would be git ignored in this case but I am sure some people don't. I guess the automatic activation should be considered a breaking change if we keep it this way.

> It seems like git errors if we'd try to git add the file and it's an ignored file:

Fixed this by running the `git check-ignore` command for the paths.

> Since we're not git adding changes to the staging area for changes to existing files the compiler produces, I think git add --intent-to-add ADDED_FILES would be the better command. This makes it so a simple git ci -a -m "my changes" would add the files, but they'd still show up if you did git diff.

Yeah I think it makes sense.  By doing it this way I don't think we need to do anything with removed files right? 
So perhaps something like this would work?
```js
await execFile('git', ['add', '--intent-to-add', ...added]);
```


Since the current SourceControl interface is a bit limited, what do you think about extending the SourceControl interface to accept changed files too and exposing some way of having our own SourceControl implementation supplied when running the compiler? 

While I think the git implementation that you suggest is a good default, I think it can be beneficial to let developers supply their own implementation. Lets say we want to automatically commit the changes, that might not be something Relay would want to do by default since its a quite unexplored area, but it could unlock some innovation in OSS.

alunyov:
Closing for now. Compiler is being migrated to rust, we may need to revisit this PR.

PR #2969, opened 2019-12-19
Docs: update commitMutation.onCompleted docs

mrtnzlml:
Response object in the onCompleted callback is not "raw" server response but updated response from the store. This confuses many users. You can try it easily: write an updater to modify the incoming records from the server and you will get the updated values in onCompleted callback (rather than the raw server response).


kassens:
Thanks!

PR #2980, opened 2020-01-01
Docs: Remove link to nonexistent RelayValidator

jaroslav-kubicek:
Link doesn't work, `RelayValidator` was removed here: https://github.com/facebook/relay/commit/15e8d2201fdad47596e1acfb3a5daa3fb7511a8e


kassens:
Thanks for the update! We found that validation can often happen most efficiently in different stages of the transform pipeline and validation steps are now just "transforms" that don't change anything and might just throw in some cases.

Issue #2984, opened 2020-01-05
Optimizing the RequestIdentifier


morrys:
Now the RequestIdentifier is created by concatenating the id or text of RequestParameters with the serialization of the variables.

https://github.com/facebook/relay/blob/master/packages/relay-runtime/util/getRequestIdentifier.js#L27-L39
```
function getRequestIdentifier(
  parameters: RequestParameters,
  variables: Variables,
): RequestIdentifier {
  const requestID = parameters.id != null ? parameters.id : parameters.text;
  invariant(
    requestID != null,
    'getRequestIdentifier: Expected request `%s` to have either a ' +
      'valid `id` or `text` property',
    parameters.name,
  );
  return requestID + JSON.stringify(stableCopy(variables));
}
```

When there is no id, the RequestIdentifier is a very long string (all the text of the query).

Wouldn't it be better to use the query hash value?

Let me know and I'll proceed with creating the PR.

kassens:
Good catch, I've noticed this before but didn't get around to fixing this.

I think it'd make sense to add a `hash` field in addition to `text` to the query parameters object and use here (if we don't have an `id`).

kassens:
Closed by #2985

PR #2985, opened 2020-01-06
2984 optimizing requestid

morrys:
https://github.com/facebook/relay/issues/2984

Modified the compilation of the artifacts in order to generate a unique identifier (md5 text) for the 'Request' nodes.

modified writeRelayGeneratedFile:

eliminate the switch, the node is always of type Request because inside 'if (generatedNode.kind === RelayConcreteNode.REQUEST) {'

https://github.com/facebook/relay/blob/master/packages/relay-compiler/codegen/writeRelayGeneratedFile.js#L98


sibelius:
this will make persisting the store to cause less memory

more info here https://github.com/morrys/react-relay-offline/issues/34#issuecomment-588248965

morrys:
@jstejada @kassens any news about this PR?
before resolving conflicts I would like to know if you are interested in its integration.
Thanks

morrys:
Friendly ping @kassens @jstejada

jstejada:
@morrys this PR is already imported and we're looking at how to roll it out internally

PR #2999, opened 2020-01-14
Relay Compiler: setup external config conditionally

mrtnzlml:
Relay Config package doesn't exist in every project which results in `yargs.config(undefined)` call. Unfortunately, yargs adds a new `--config` option for these cases. However, this option is not specified by Relay Compiler so it confuses people. I am not sure if it's a bug or feature though because technically it's fine as well. I'd argue that it's better to hide it and expose it manually (when needed) because now it's sometimes available and sometimes not.

### DIFF
--- a/packages/relay-compiler/bin/RelayCompilerBin.js
+++ b/packages/relay-compiler/bin/RelayCompilerBin.js
@@ -12,7 +12,7 @@
 
 'use strict';
 
-const yargs = require('yargs');
+const _yargs = require('yargs');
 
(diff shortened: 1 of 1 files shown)


mrtnzlml:
Yargs docs: http://yargs.js.org/docs/#api-configobject

mrtnzlml:
Cc @alloy - was this intentional or not? :)

alloy:
Nope, definitely not intentional. Good catch :+1:

PR #3010, opened 2020-01-31
Docs: add basic info about 'LocalQueryRenderer' component

mrtnzlml:


mrtnzlml:
Rebased. 🎸

docs/Modern-LocalStateManagement.md
```diff
@@ -94,6 +94,8 @@ const Example = (props) => {
 }
 ```
 
+It is also possible to use `LocalQueryRenderer` component which works the same way like standard Query Renderer component except it doesn't send the request to the server. It's useful when you need to access _only_ local fields. Please note: Local Query Renderer returns only `props` in the `render` method - no errors (since there is no network request going on).
```
josephsavona, line 97: small grammar nit but otherwise looks great: "which works the same way as standard" (s/like/as)

mrtnzlml: @josephsavona Thanks! I need to work on my English. 😊

mrtnzlml:
cc @kassens :)

Issue #3013, opened 2020-02-06
Environment's held on longer than they need too


maraisr:
When using useFragment instead of createFragmentContainer, in an SSR context the resourceMap holds onto the environment longer than it needs too, which causes the application to look like its got a memory leak, as the store is contained inside the environment. As touched on https://github.com/facebook/relay/issues/2854#issuecomment-582276269

I will prefix that with being, that in an SSR setting, you'd ideally want to blow away the store on every request, so you're not leaking store data between requests. So this only came about, when you create a new Environment, for the request. And having the resourceMap Map hold onto now discarded environments.

https://github.com/facebook/relay/blob/02d9e7113774ef4852f90b91308adab09e8729fa/packages/relay-experimental/FragmentResource.js#L440-L451

My suggestion is to move away from a Map, and into a WeakMap, so that once the environment gets gc'd, it cleans up the fragment resourceCache.

```diff
- const dataResources: Map<IEnvironment, FragmentResource> = new Map();
+ const dataResources: WeakMap<IEnvironment, FragmentResource> = new WeakMap(); 
```

PR #3014, opened 2020-02-06
feat: WeakMap when environments are referenced

maraisr:
With this PR I aim to address the situation where environments are removed, but never garbage collected as references are held onto in this Map.

A simple case could be:

```js
const MyApp = () => {

	const [environment, setEnvironment] = useState(new Environment({ network, store }));

	useEffect(() => {
		setTimeout(() => {
			setEnvironment(new Environment({ network, store }));
		}, 1e3);
	}, []);

	return (<RelayEnvironmentProvider environment={environment}>
		<MyComponentTreeWith_A_useFragment/>
	</RelayEnvironmentProvider>);
};
```

as you'll find from that example, the `dataResources` map will be forever growing, with nothing clearing.

fixes: #3013

### DIFF
--- a/packages/relay-experimental/FragmentResource.js
+++ b/packages/relay-experimental/FragmentResource.js
@@ -437,7 +437,7 @@ function createFragmentResource(environment: IEnvironment): FragmentResource {
   return new FragmentResourceImpl(environment);
 }
 
-const dataResources: Map<IEnvironment, FragmentResource> = new Map();
+const dataResources: WeakMap<IEnvironment, FragmentResource> = new WeakMap();
 function getFragmentResourceForEnvironment(
   environment: IEnvironment,
 ): FragmentResourceImpl {
--- a/packages/relay-experimental/FragmentResource.js
+++ b/packages/relay-experimental/FragmentResource.js
@@ -42,6 +42,12 @@ type FragmentResourceCache = Cache<
   Error | Promise<mixed> | SingularOrPluralSnapshot,
 >;
 
+const WEAKMAP_SUPPORTED = typeof WeakMap === 'function';
+interface IMap<K, V> {
+  get(key: K): V | void;
+  set(key: K, value: V): IMap<K, V>;
+}
+
 type SingularOrPluralSnapshot = Snapshot | $ReadOnlyArray<Snapshot>;
 opaque type FragmentResult: {data: mixed, ...} = {|
   cacheKey: string,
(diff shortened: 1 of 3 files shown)


packages/relay-experimental/FragmentResource.js
```diff
@@ -437,7 +437,7 @@ function createFragmentResource(environment: IEnvironment): FragmentResource {
   return new FragmentResourceImpl(environment);
 }
 
-const dataResources: Map<IEnvironment, FragmentResource> = new Map();
+const dataResources: WeakMap<IEnvironment, FragmentResource> = new WeakMap();
```
alunyov: We need to add here a similar check: 

``` const WEAKMAP_SUPPORTED = typeof WeakMap === 'function';```

File: packages/relay-experimental/QueryResource.js [hunk 2]
```diff
@@ -527,7 +527,7 @@ function createQueryResource(environment: IEnvironment): QueryResource {
   return new QueryResourceImpl(environment);
 }
 
-const dataResources: Map<IEnvironment, QueryResource> = new Map();
+const dataResources: WeakMap<IEnvironment, QueryResource> = new WeakMap();
```
alunyov: same here

packages/relay-experimental/FragmentResource.js
maraisr: Wanted too - just wasn't sure how to articulate that to flow?

alunyov: Oh, yeah... I forgot about flow. We probably need to change it to something  ```const dataResources: IMap< IEnvironment, FragmentResource>``` where both WeakMap and Map are implementing this interface IMap.

maraisr:
Thanks for that @alunyov, i have made those IMap changes... I'm not the best Flow writer, so please do tell me if there was a better way.

alunyov:
Thanks for the update!

> Also; should we try and implement something that mimic's the WeakMap 

IMO, we may just keep it as it is right now. They're just a few instances where we are using WeakMap, so - not a huge overhead. And also pulling in the polyfill will most likely conflict with our internal setup.

maraisr:
Oh sorry, I may have no articulated myself correctly. My bad. I'm not saying we include the polyfill within this repo. Im merely saying, that bundlers, runtimers etc.. That they include the polyfill, if they choose. Much like how react isnt including a polyfill for Map, but rather saying that if IE9 is your target, then you'll need to include a polyfill for Map.

What I'm really trying to say is, let this library articulate that we want the behavior of a WeakMap by using it - and simply say that for environments where that isn't directly available, that you have to use polyfill's, by whatever means your setup is like.. So Facebook might use rome, babel, whatever.

PR #3015, opened 2020-02-09
Make .graphql extension be a part of the language plugin

edvinerikson:
This makes it possible to use the language plugin interface in java.

Background:

Relay will generate the file "MyFragment.graphql.java" and a class "MyFragment.graphql". This is however not valid java code, dots "." is not allow in class names.

It's possible to change the class name programatically to solve that problem, but then we will have a public class that do not match the name of the file, which as a result is going to create a java compilation error.

### DIFF
--- a/packages/relay-compiler/bin/RelayCompilerMain.js
+++ b/packages/relay-compiler/bin/RelayCompilerMain.js
@@ -336,7 +336,7 @@ function getCodegenRunner(config: Config): CodegenRunner {
         persistQueryFunction,
       ),
       isGeneratedFile: (filePath: string) =>
-        filePath.endsWith('.graphql.' + outputExtension) &&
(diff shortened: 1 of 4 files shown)


zth:
@edvinerikson On a related note, I have an old PR that accomplishes the same thing but in a slightly different way: https://github.com/facebook/relay/pull/2866

edvinerikson:
That's great! As long as we get traction on one of them I am happy. 🙂

alloy:
I think it makes sense to be as specific as possible about the filename, so I’d vote for @zth’s version.

alunyov:
@edvinerikson  thank you for this PR. 

We just landed the #2866, so I'm going to close this.

Thank you!

edvinerikson:
That's great! Thanks 🙂

PR #3027, opened 2020-02-15
Optional updater parameter for commitPayload

HermantNET:
Allows us optionally pass an `updater` to `commitPayload`.

The app I'm currently working on relies on numerous updaters, I need to be able to re-use them while making normalising data from sources outside of relay easily.

Example code:
```
import createNoteUpdater from "Relay/Updaters/createNote"

// ...
const payload = { create_note: { id: 123, text: "hello" } }

modernEnvironment.commitPayload(
  createOperationDescriptor(CreateNoteMutation, {}),
  payload,
  createNoteUpdater
);
```

I've noticed it referenced in a type here as part of `PublishQueue`:
https://github.com/facebook/relay/blob/8c51286848313ef6b3dc09abaa81aa9b44bd17de/packages/relay-runtime/store/RelayStoreTypes.js#L829-L833

while the type definition under `IEnvironment` doesn't offer the same flexibility:
https://github.com/facebook/relay/blob/8c51286848313ef6b3dc09abaa81aa9b44bd17de/packages/relay-runtime/store/RelayStoreTypes.js#L508-L511

### DIFF
--- a/packages/relay-runtime/store/RelayModernEnvironment.js
+++ b/packages/relay-runtime/store/RelayModernEnvironment.js
@@ -240,7 +240,11 @@ class RelayModernEnvironment implements IEnvironment {
     );
   }
 
-  commitPayload(operation: OperationDescriptor, payload: PayloadData, updater: ?StoreUpdater = null): void {
+  commitPayload(
+    operation: OperationDescriptor,
+    payload: PayloadData,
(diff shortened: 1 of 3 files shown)


HermantNET:
> This is a reasonable change, but please add a unit test so that we avoid regressions.

Thank you! I'll add a unit test when I have some time

PR #3042, opened 2020-03-07
[relay-compiler] Add persisted query text in dev mode

### LINKED ISSUE #2799: Bring back dev only query text behind option
See discussion on 377ca939b5f5b46d57e11d4a1dfa7c4aecf5666b

janicduplessis:
Fixes #2799

- Add `includePersistedQueryTextInDev` option to enable this feature.

- Add `nodeDevOnlyProperties` api to `RelayLanguagePluginInterface` to allow modifying the node object in development. This is similar to `devOnlyAssignments` that was removed but to make it more flexible for other languages we pass the object to merge instead of a string of JavaScript.

## Test plan

- Validate that no changes are made to generated files when the option is disabled.
- Validate that the query text is added properly under the `__DEV__` flag when using peristed queries and the option enabled.


packages/relay-compiler/language/javascript/formatGeneratedModule.js
```diff
@@ -22,13 +24,22 @@ const formatGeneratedModule: FormatModule = ({
...
   const docTextComment =
     docText != null ? '\n/*\n' + docText.trim() + '\n*/\n' : '';
   const hashText = hash != null ? `\n * ${hash}` : '';
+  const devOnlyAssignments = deepMergeAssignments(
+    '(node/*: any*/)',
+    nodeDevOnlyProperties,
+  );
+  const devOnlyAssignmentsText =
+    devOnlyAssignments.length > 0
+      ? `\nif (__DEV__) {\n  ${devOnlyAssignments}\n}`
```
janicduplessis: Not sure if we should use `process.env.NODE_ENV === 'production'` or this

janicduplessis:
@kassens Any chance you (or someone else) could have a look at this?

janicduplessis:
@kassens Is this PR still useful? Looks like a lot of focus is on the new rust compiler, but I think this is still a valuable feature as it makes working with persisted queries a lot easier in DEV, to avoid having to teach the server about WIP queries. If the concern is feature parity with the new compiler I can have a look at implementing this in the new compiler also.

PR #3043, opened 2020-03-10
fix(paginationVariables): UNSTABLE_extraVariables warning

chuckGonzalez:
Right now when i call `loadNext` from `usePaginationFragment`, i get the next warning:

`index.js:1437 Warning: Relay: UNSTABLE_extraVariables provided by caller should not contain count variable first. This variable is automatically determined by Relay.`

That is because getPaginationVariables.js evaluate baseVariables after extraVariables  is merged with parentVariables and fragmentVariables in useLoadMoreFunction.js.

Fix:

1. Pass extraVariables to getPaginationVariables and evaluate if extraVariables has count or cursor instead of baseVariables

### DIFF
--- a/packages/relay-experimental/__tests__/getPaginationVariables-test.js
+++ b/packages/relay-experimental/__tests__/getPaginationVariables-test.js
@@ -30,6 +30,7 @@ describe('getPaginationVariables', () => {
           10,
           'cursor-1',
           {order_by: 'LAST_NAME'},
+          {},
           {forward: null, backward: null, path: []},
         ),
       ).toThrowError(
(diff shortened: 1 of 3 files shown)


sibelius:
fixed here https://github.com/facebook/relay/commit/84565d41eb33af75a2478cf22ad1d6d5b790e950

sorenhoyer:
@sibelius are you sure about that? We're still getting this very "noisy" warning in our console after using usePaginationFragment in latest experimental version

sibelius:
not sure if that commit was released it

PR #3070, opened 2020-05-05
feat(added cacheConfig in RequestDescriptor)

morrys:
This change allows you to associate cache configurations directly in the request object.

In this way, once the operation has been created, it is possible to access all the information necessary to create a request in the network.

In addition, this modification allows you to add parameters to the CacheConfig object and easily use this information within Relay.

An example of a use case is to allow the configuration of a different TTL (https://github.com/facebook/relay/commit/bb888069cd238b10516bd1ddc528b8f84bdf408f) in the queries. So just add the TTL parameter inside CacheConfig and use that information inside the retain function without having to change its interface.


morrys:
hi @kassens,
any news or doubt about this PR? 👍

morrys:
friendly ping @jstejada @josephsavona

josephsavona:
Overall this makes a lot of sense to me. I'm still skeptical of per-query TTL, but overall this could be a clean way to allow alternate store implementations to experiment with such features. My main concern is what to do about the RequestIdentifier, which is intended to identify and de-duplicate requests for the same information. There are various places in Relay where we de-duplicate queries based on this identifier, and we need to decide whether the CacheConfig should be part of that or not. I am inclined to agree w the implementation here and say no - the CacheConfig isn't part of the request id. But we need to think through this and make sure that doesn't cause problems for the intended use-case here.

Thoughts?

morrys:
> I am inclined to agree w the implementation here and say no

I am of the same opinion, because the request is always the same and leaving the identifier the same I have the possibility to recognize that the query is already retained in the store or it is still pending in the network

morrys:
@josephsavona I just resolved the conflicts, any ETA on this?
thanks

morrys:
friendly ping

> @josephsavona I just resolved the conflicts, any ETA on this?
> thanks

josephsavona:
cc @jstejada and @rbalicki2

morrys:
Hi @rbalicki2, any update or ETA on this PR? thanks 👍

PR #3074, opened 2020-05-07
[Do not land] Remove fast-glob

cpojer:
See https://www.internalfb.com/intern/diff/D21428538


alloy:
Everybody on a diet! 🙌

cpojer:
Let's try without flat map.

@alloy do you know how to best verify this code doesn't break anything? We don't use this code path at fb 🤷‍♂️

alloy:
It can probably be tested in https://github.com/artsy/reaction, let me see if I can quickly spin that up.

packages/relay-compiler/bin/RelayCompilerMain.js
```diff
@@ -23,6 +23,7 @@ const WatchmanClient = require('../core/GraphQLWatchmanClient');
 
 const crypto = require('crypto');
 const fs = require('fs');
+const glob = require('glob');
```
alloy, line 26: Is there a reason to always require this rather than inline as before?

cpojer: I assume it doesn't matter but I don't know.

cpojer:
Thank you so much for giving it a try <3

I'll land the internal version of this change after somebody approves it.

packages/relay-compiler/bin/RelayCompilerMain.js
alloy: I guess it was inlined because this code-path isn’t hit when you use watchman?

mrmlnc:
@cpojer, Sorry for the mention.

I'm the maintainer of the fast-glob package and I want to make this package better. Unfortunately, I don't have access to the facebook intranet.

If you don't mind, tell me please, the decision to replace `fast-glob with `glob` is related to bugs, performance reason or any other reasons that I can fix?

JFYI: The `fast-glob@2.x.x` is outdated version and the `fast-glob@3.x.x` has many improvements.

alloy:
@mrmlnc I’m not at FB and don’t know for sure, but I believe it’s related to reducing the size of the overall dependency tree (especially in a react/native app), hence my comment about “diet”.

cpojer:
@mrmlnc @alloy is correct. There is nothing wrong with your package, in fact I actually like it quite a bit. I have been reducing the size of Facebook's React Native node_modules folder internally at Facebook. We use a monorepo, so every product developer has to install dependencies for every single project: Relay, Metro, Jest, Babel, eslint, React Native, jscodeshift and dozens more.

Previously the size of our node_modules folder was 930 MiB and we pulled in thousands of packages, many which were outdated or unnecessary. This lead to a long install time until people could get started, both on CI and locally. I have since been focusing on reducing the size of this folder through a number of (funky) techniques while simultaneously updating everything to the latest major versions. Our node_modules folder is now ~143 MiB and is much faster to install.

One of the techniques was to eliminate dependencies that are big (500k+), used only in one or two places (for example only in Relay) and are already covered by functionality of packages that are shared across a wider range of tools (like "glob"). In this case the impact is very low as people who run into problems with glob in open source can switch to watchman, which for large projects is much more scalable. While part of the source code, we do not use this code path at Facebook either.

To summarize, I'm removing large dependencies that we don't use (much) to make it faster for product developers to get started with React Native. If you are looking for things to improve I recommend reducing the size of the "fast-glob" packages and the dependencies it brings in. Why does it have to be 500k? Can it be 100k? Can it be 10?

sibelius:
@cpojer great work, how are you tracking big packages inside node_modules?

mrmlnc:
@cpojer, Thank you for the quick reply!

In the `fast-glob@3.x.x` we reduce the package size from `2.47MB` to `0.42MB` and «require time» from `534ms` to `78ms` ([Release Notes](https://github.com/mrmlnc/fast-glob/releases/tag/3.0.0), [Package Phobia](https://packagephobia.now.sh/result?p=fast-glob)).

Thank you again for the detailed answer.

cpojer:
@mrmlnc Nice work! I'll definitely consider using `fast-glob` in the future if the need arises :)

packages/relay-compiler/bin/RelayCompilerMain.js
Daniel15: Not sure if we do it for open-source builds, but internally at FB we automatically inline `require`s so the behaviour would be identical in that case.

alloy: iirc inlining all requires was disabled, as it led to worse behaviour of relay-runtime in a SSR context. /cc @robrichard

PR #3089, opened 2020-06-02
Add Black Lives Matter banner

Daniel15:
Similar to the Docusaurus and React sites.

image

React's version, for reference:

image

It seems harder to show it above the menubar with Docusaurus, so I placed it underneath instead. This is consistent with what the Docusaurus site itself is doing.

### DIFF
--- a/website/pages/en/index.js
+++ b/website/pages/en/index.js
@@ -33,17 +33,25 @@ Button.defaultProps = {
 class HomeSplash extends React.Component {
   render() {
     return (
-      <div className="homeContainer">
-        <div className="homeSplashFade">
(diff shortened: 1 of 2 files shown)


josephsavona:
Thank you for sending this PR. We're in the process of importing and landing this change, so it should (barring hiccups) roll out to the website this evening.

We're actively discussing how our team can go beyond simply expressing support and _take action_ to support progress on racial justice. In the meantime, we're adding this banner to help raise awareness of the systemic issues.

Daniel15:
@josephsavona It seems like the styling isn't working correctly on the live site :/ https://relay.dev/

PR #3090, opened 2020-06-03
Add flow types for relay-experimental to NPM

dlvhdr:
This generates the .js.flow files for relay-experimental.
This will allow the users to have flow definitions.

* I linked the local library and saw flow types for `useLazyLoadQuery` as an example.
* Ran yarn test

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -236,6 +236,21 @@ const builds = [
       },
     ],
   },
+  {
+    package: 'relay-experimental',
+    exports: {
+      index: 'index.js',
+      ReactRelayContext: 'ReactRelayContext.js',
+    },
+    bundles: [
+      {
+        entry: 'index.js',
+        output: 'relay-experimental',
+        libraryName: 'RelayExperimental',
+        libraryTarget: 'umd',
+      },
+    ],
+  },
 ];
 
 const modules = gulp.parallel(


kassens:
I think we don't export any flow types at all at the moment. Fixing this could be a pretty nice win for anyone using Relay and Flow.

This issue is actually not specific for relay-experimental and I believe we just need to generally figure out how to export flow types in our build system (that nobody understands super well).


I think the approach would be to update the [gulpfile](https://github.com/facebook/relay/blob/master/gulpfile.js) to:
- *Just like it does today:* webpack/babel compile files as it does right now, outputting files like `dist/react-relay/lib/RelayContext.js` that are fully transformed
- *Additionally:* copy the original `*.js` files into dist with a `*.js.flow` extension that flow would pick up automatically (it prefers .js.flow over the .js files).

To test the gulp changes, I typically:
- build the original
- move the `dist` into `dist-original`
- make gulp changes, build again (might help to `rm -r dist` as caching seems to be too aggressive sometimes)
- look at the diff in `dist` and `dist-original`

To test, this actually works: Probably something about `npm link` or copying manually into a `node_modules` directory of a flow-enabled project to see it work. I don't have a great approach here.

Finally, `relay-experimental` is not an npm package we control, the hooks version is under `react-relay/hooks` released with the `experimental` tag that is published from a branch in this repo. Once figured out for the non-hooks version, this shouldn't be too hard to make the required changes in the gulpfile in the branch.

dlvhdr:
Resolved in latest version.

PR #3100, opened 2020-06-15
Support ScalarHandle in DataChecker

leoasis:
If we have a fragment that makes use of a handler in a scalar field, `DataChecker` will throw an error because the `SCALAR_HANDLE` kind of selection is not supported. Talking with the team we agreed this was a case that should be supported.

I did this by replicating the logic for LINKED_HANDLE, only that in this case we don't need to copy inner selections since this is a scalar field. I created a `cloneRelayScalarHandleSourceField` module similar to its linked handle counterpart. Also while at it, I added some tests to support this functionality and that of LINKED_HANDLE, since it was previously not covered.

### **Wrong assumptions based on @kassens's [comment below,](https://github.com/facebook/relay/pull/3100#issuecomment-647920623) leaving this for posterity:**

~~When figuring out the logic around a handle when checking if the data is complete, given that a handle field lives alongside the original one (which contains the data that came back from the response), it turned out that it would be enough to just skip this selection, as the logic for the original field would be enough to test that the data is complete.~~

~~I went a bit further with this reasoning and did the same with a `LINKED_HANDLE`. The logic here was even creating a clone of the original field and assigning the sub selections, traversing that in the same way we do for linked fields. But if the original linked field is always present next to a handler, then this is already being done when traversing the original field, so there should be no need to do this.~~

~~Probably I'm missing something in the assumptions for the latter case, but figured I would just send a PR and explaining the rationale to start the discussion. However, worth noting that all tests still pass even when skipping a LINKED_HANDLE field. So if this is not the right thing to do, we should at least add a test that validates this is indeed used.~~

### DIFF
--- a/packages/relay-runtime/store/DataChecker.js
+++ b/packages/relay-runtime/store/DataChecker.js
@@ -21,6 +21,7 @@ const RelayRecordSourceProxy = require('../mutations/RelayRecordSourceProxy');
 const RelayStoreUtils = require('./RelayStoreUtils');
 
 const cloneRelayHandleSourceField = require('./cloneRelayHandleSourceField');
+const cloneRelayScalarHandleSourceField = require('./cloneRelayScalarHandleSourceField');
 const invariant = require('invariant');
 
 const {isClientID} = require('./ClientID');
@@ -342,7 +343,7 @@ class DataChecker {
           }
           break;
         }
-        case LINKED_HANDLE:
+        case LINKED_HANDLE: {
           // Handles have no selections themselves; traverse the original field
           // where the handle was set-up instead.
           const handleField = cloneRelayHandleSourceField(
@@ -356,6 +357,17 @@ class DataChecker {
             this._checkLink(handleField, dataID);
           }
           break;
+        }
+        case SCALAR_HANDLE: {
+          const handleField = cloneRelayScalarHandleSourceField(
+            selection,
+            selections,
+            this._variables,
+          );
+
+          this._checkScalar(handleField, dataID);
+          break;
+        }
         case MODULE_IMPORT:
           this._checkModuleImport(selection, dataID);
           break;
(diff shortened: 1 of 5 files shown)


kassens:
I don't think we have any uses for scalar handles internally, and this totally seems like a bug to me to hit the `invariant` there.

For linked handles, I'm pretty sure we need to traverse the handle field. An example is the @connection handler that collects the aggregated page results as a single concatenated list. When this needs to be checked, all the items in the list need to be checked. Agreed a test for this would be good.

cc @josephsavona

leoasis:
@kassens ah right, I see the code that attempts to do this. Sorry for missing this! Will undo that change and only focus on scalar handlers here.

leoasis:
> I don't think we have any uses for scalar handles internally, and this totally seems like a bug to me to hit the `invariant` there.
> 
> For linked handles, I'm pretty sure we need to traverse the handle field. An example is the @connection handler that collects the aggregated page results as a single concatenated list. When this needs to be checked, all the items in the list need to be checked. Agreed a test for this would be good.
> 
> cc @josephsavona

I'm wondering if scalar handle would require the same validation as a linked handle then, in order to support the case where that handler may return the right data vs undefined. So this woudl require a strategy similar to what we do for a linked handle, which would be to clone the field and call the checkScalar method with that cloned field.

josephsavona:
> I'm wondering if scalar handle would require the same validation as a linked handle then, in order to support the case where that handler may return the right data vs undefined. So this woudl require a strategy similar to what we do for a linked handle, which would be to clone the field and call the checkScalar method with that cloned field.

Yeah, I think we'll need to implement this to be fully correct.

leoasis:
@kassens @josephsavona I've updated this PR to now add support for SCALAR_HANDLE the same way we do for LINKED_HANDLE, and added tests. Also updated the description and title to better describe what this is about.

josephsavona:
@leoasis nice, imported. we'll try to include this in our release today (assuming internal tests pass)

Issue #3076, opened 2020-05-12
`react-relay` uses undeclared dependency `scheduler`


unindented:
`yarn berry` complains about `react-relay` requiring an undeclared dependency `scheduler`:

```
./.yarn/$$virtual/react-relay-virtual-8db22a2c03/0/cache/react-relay-npm-0.0.0-experimental-895a6fe0-3b3bf7f4a4-3.zip/node_modules/react-relay/lib/relay-experimental/useBlockingPaginationFragment.js
Module not found: react-relay tried to access scheduler, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
```

I was able to repro by following [the `create-react-app` tutorial](https://relay.dev/docs/en/experimental/step-by-step), using `yarn berry` instead of `yarn v1`.

jgcmarins:
should we add `scheduler` to dependencies to fix that?

unindented:
Yup, I think that's the right thing to do.

sibelius:
Send a PR please

PR #3108, opened 2020-06-29
chore(react-relay): Adds scheduler as a dep

maraisr:
fixes: #3076

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "react": "16.13.1",
     "react-refresh": "^0.4.0",
     "react-test-renderer": "16.13.1",
+    "scheduler": "^0.19.1",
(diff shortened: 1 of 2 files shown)


kassens:
Thanks @maraisr!

kassens:
We'll include this in the next experimental release, but for now it's harder to merge as we use some experimental branch release of Scheduler internally and we like to avoid using different versions in github and externally.

That said, the next experimental branch releases should be fixed.

maraisr:
I think we should keep this one open Stale bot,

cc @en_js @kassens

maraisr:
fixed in https://github.com/facebook/relay/commit/bf361b32ec0f7d0cee90ff448c9ad4e038e82fdb

PR #3111, opened 2020-07-03
add OSS compiler CLI and test project

kassens:
Not quite ready to merge, but this shows how the compiler CLI could look like in OSS.

Pretty rough (just like the internal one).

Examples to run:
```
cd compiler
cargo run -p relay-compiler test-project/relay.json ..
```

Separate build and run using release:
```
cd compiler
cargo build --release -p relay-compiler
target/release/relay-compiler test-project/relay.json ..
```

### DIFF
--- a/compiler/crates/relay-compiler/src/main.rs
+++ b/compiler/crates/relay-compiler/src/main.rs
@@ -1,0 +1,56 @@
+use env_logger::Env;
+use log::{error, info};
+use relay_compiler::{compiler::Compiler, config::Config};
+use std::{path::PathBuf, sync::Arc};
+use structopt::StructOpt;
+
+#[derive(StructOpt)]
+#[structopt(
+    name = "Relay Compiler",
+    about = "Compiler to produce Relay generated files."
+)]
+struct Opt {
+    /// Compile and watch for changes
+    #[structopt(long, short)]
+    watch: bool,
+
+    /// Path to the compiler config file
+    config: PathBuf,
+
+    /// Root directory of the project
+    root: PathBuf,
+}
+
+#[tokio::main]
+async fn main() {
+    env_logger::from_env(Env::default().default_filter_or("info, error, warn")).init();
+
+    let opt = Opt::from_args();
+
+    let config = match Config::load(opt.root, opt.config) {
+        Ok(config) => config,
+        Err(err) => {
+            error!("{}", err);
+            std::process::exit(1);
+        }
+    };
+
+    let compiler = Compiler::new(config, Arc::new(common::NoopPerfLogger));
+
+    if opt.watch {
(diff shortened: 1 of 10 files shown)


compiler/crates/relay-compiler/src/main.rs
```diff
@@ -0,0 +1,56 @@
...
+    /// Path to the compiler config file
+    config: PathBuf,
+
+    /// Root directory of the project
+    root: PathBuf,
+}
+
+#[tokio::main]
+async fn main() {
+    env_logger::from_env(Env::default().default_filter_or("info, error, warn")).init();
```
maraisr, line 26: I feel like we should pull in a different printer so its _pretty_? And potentially scope the errors also?

compiler/crates/relay-compiler/src/main.rs
```diff
@@ -0,0 +1,56 @@
...
+
+    let opt = Opt::from_args();
```
maraisr, line 28: Do we print some sort of "debug" info here, like [maybe the version that is running](https://github.com/clap-rs/clap/blob/master/README.md#L215-L217) etc...

Could potentially use [Clap](https://github.com/clap-rs/clap)

Granted this is just a "initial start"—but think we can definitely jazz it up!

compiler/crates/relay-compiler/src/main.rs
kassens: Yeah, this is still a pretty rough edge.

I see 2 stages:
- Clean this up for end users to be more to the essentials, probably update some messages and the formatting to remove timestamps.
- I dream of something like the Jest test runner / yarn install to show progress (probably more important internally if it takes a couple seconds even with Rust) and just generally nicer output.

For now this is basically debug output level for development.

compiler/crates/relay-compiler/src/main.rs
kassens: This is using https://crates.io/crates/structopt which is a wrapper around Clap.

This should already generate a `--help` command and I think a `--version` based on the Cargo.toml.

compiler/crates/relay-compiler/src/main.rs
```diff
@@ -0,0 +1,56 @@
...
+struct Opt {
+    /// Compile and watch for changes
+    #[structopt(long, short)]
+    watch: bool,
+
+    /// Path to the compiler config file
+    config: PathBuf,
+
+    /// Root directory of the project
+    root: PathBuf,
```
kassens, line 21: I think the `root` should actually come from the `config.json` file. Maybe with the directory of the config.json the default and otherwise a relative root configured in the config.json. Thoughts?

alloy: Yup, that makes sense to me 👍

PR #3112, opened 2020-07-03
feat(rust-compiler): Build FB as a feature flag at compile-time

maraisr:
With this PR I aim to introduce the ability to build the compiler with either Facebook specific things enabled/disabled.

I don't think there are any other specific things at this stage.

The general approach will be to follow the feature spec, which can be done either in code, or as compiler directives.

Facebook will need to apply an extra configuration in Buck to enable this.

- https://doc.rust-lang.org/cargo/reference/features.html
- https://buck.build/rule/rust_library.html#features

---

Might need to investigate an approach here where unit tests could be ran against both versions. Snapshots will obviously fail. But the issue is the feature is defined in the `relay-compiler` crate, but transformers own the reference? So maybe that feature flag _should_ be reiterated per crate that requires that level of context?

### DIFF
--- a/compiler/crates/graphql-transforms/src/connections/connection_interface.rs
+++ b/compiler/crates/graphql-transforms/src/connections/connection_interface.rs
@@ -22,27 +22,32 @@ pub struct ConnectionInterface {
     pub start_cursor_selection_name: StringKey,
 }
 
-lazy_static! {
-    pub static ref OSS_CONNECTION_INTERFACE: Arc<ConnectionInterface> =
-        Arc::new(ConnectionInterface {
-            cursor_selection_name: "cursor".intern(),
-            edges_selection_name: "edges".intern(),
-            end_cursor_selection_name: "endCursor".intern(),
-            has_next_page_selection_name: "hasNextPage".intern(),
-            has_prev_page_selection_name: "hasPreviousPage".intern(),
-            node_selection_name: "node".intern(),
-            page_info_selection_name: "pageInfo".intern(),
-            start_cursor_selection_name: "startCursor".intern(),
-        });
-    pub static ref FB_CONNECTION_INTERFACE: Arc<ConnectionInterface> =
-        Arc::new(ConnectionInterface {
-            cursor_selection_name: "cursor".intern(),
-            edges_selection_name: "edges".intern(),
-            end_cursor_selection_name: "end_cursor".intern(),
-            has_next_page_selection_name: "has_next_page".intern(),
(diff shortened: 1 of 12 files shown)


compiler/crates/graphql-transforms/src/connections/connection_interface.rs
```diff
@@ -22,27 +22,32 @@ pub struct ConnectionInterface {
...
-            end_cursor_selection_name: "end_cursor".intern(),
-            has_next_page_selection_name: "has_next_page".intern(),
-            has_prev_page_selection_name: "has_previous_page".intern(),
-            node_selection_name: "node".intern(),
-            page_info_selection_name: "page_info".intern(),
-            start_cursor_selection_name: "start_cursor".intern(),
-        });
-}
+#[cfg(feature = "is_FB")]
+    lazy_static! {
```
maraisr, line 26: You cannot apply compiler directives inside the macro. However we could have done this via a `if cfg!()` so let me know if this duplicated lazy_static is okay—or if we simply define both in the same lazy, and wrap in a if.

compiler/crates/relay-compiler/Cargo.toml
```diff
@@ -14,6 +14,10 @@ path = "src/lib.rs"
 name = "relay_compiler_compile_relay_artifacts_test"
 path = "tests/compile_relay_artifacts_test.rs"
 
+[features]
+default = []
```
maraisr, line 18: default is magic, and gets applied—by default. I was thinking of first going `is_OSS` and `is_FB`, but that _could_ be dangerous because what-if, both are applied.

The array simply allows us to nest features, ie maybe have a feature for `fb_connections`, that the `is_FB` turns on. etc..

compiler/crates/relay-compiler/src/build_project/mod.rs
```diff
@@ -34,7 +34,7 @@ pub use generate_artifacts::{
 };
 use generate_extra_artifacts::generate_extra_artifacts;
 use graphql_ir::Program;
-use graphql_transforms::FB_CONNECTION_INTERFACE;
+use graphql_transforms::CONNECTION_INTERFACE;
```
maraisr, line 37: Figured it wasn't really a callsite's concern which connection it should pull in.

kassens:
Since we run the tests also with the the OSS style (for OSS tests), I implemented something similar, but moved the connectionConfig into the config JSON file (1386365510126b337a1409dd2a7ff11dbe895f66).

The config key is optional and defaults to the OSS style.

Thanks for taking a first try at this!

PR #3116, opened 2020-07-04
chore: Removed Gatbsy and pinned AG

maraisr:
Since Nov last year [Gatbsy hasn't used relay](https://github.com/gatsbyjs/gatsby/pull/19665)—so removing them from the site.

Also a cheeky "pin" of AutoGuru to fill that space.


kassens:
Test 09b823699

PR #3118, opened 2020-07-06
Add scalar type name to object output

wakeless:
I've been working on introspecting the output of the Relay object artifacts (using it to inform client side outputs)
and I'd noticed that the Scalar name was not in the output. So I've poked at the compiler and hacked this in.

This PR is by no means finished, but I'm putting it up as a POC and wonder if it was up to scratch if it is something that would be merged in.

TBH, I'm not even sure if the rust compiler is the compiler that's being used, if not I'd obviously have to do the same thing in the shipped compiler.

Before shipping this, it would at the very least require nullability information to be added to the output.

Would love comments/feedback/information about getting this in the codebase.

### DIFF
--- a/compiler/crates/relay-codegen/src/build_ast.rs
+++ b/compiler/crates/relay-codegen/src/build_ast.rs
@@ -503,6 +503,10 @@ impl<'schema, 'builder> CodegenBuilder<'schema, 'builder> {
             ),
             (CODEGEN_CONSTANTS.name, Primitive::String(name)),
             (
+                CODEGEN_CONSTANTS.type_,
+                Primitive::String(self.schema.get_type_name(schema_field.type_.inner())),
+            ),
+            (
                 CODEGEN_CONSTANTS.storage_key,
                 match args {
                     None => Primitive::Null,
--- a/compiler/crates/relay-compiler/tests/compile_relay_artifacts/fixtures/client-fields-of-client-type.expected
+++ b/compiler/crates/relay-compiler/tests/compile_relay_artifacts/fixtures/client-fields-of-client-type.expected
@@ -32,6 +32,7 @@ type Foo {
       "args": null,
       "kind": "ScalarField",
       "name": "id",
+      "type": "ID",
       "storageKey": null
     },
     {
@@ -50,6 +51,7 @@ type Foo {
               "args": null,
               "kind": "ScalarField",
               "name": "client_name",
+              "type": "string",
               "storageKey": null
             },
             {
@@ -71,6 +73,7 @@ type Foo {
                   "args": null,
                   "kind": "ScalarField",
                   "name": "uri",
+                  "type": "string",
                   "storageKey": null
                 }
               ],


josephsavona:
Thanks for your interest! In this case, we intentionally exclude the types of scalar fields in generated artifacts because relay-runtime doesn't need to know scalar types at runtime (it only needs to distinguish between scalars and non-scalars). We generally try to avoid including any information in the artifacts that isn't needed at runtime, in order to keep artifact size smaller. 

Can you provide more context about how you're using this information? Note that we support the `@relay_test_operation` directive on queries which will emit extra data in the `metadata` of a query about the type of each field in the response (by its field path).

wakeless:
Thanks for the swift reply @josephsavona .

I'm building out some auto-configuring lib code that's driven by the relay fragments. This will be something like this pseudocode:

``` 
let fragment = graphql` fragment TableRow on Customer {
  archived
  dateCreated
  name
}

let columnDefs = fragment.selections.map((i) => i.type == "BOOLEAN" ? <BooleanCell> : <TextCell>)

let query = ... // select data from the server

//render data selected with columns generated from the metadata selected from the query selection
```

This is all looking really good thus far, the artifacts that are available are very comprehensive, except for this particular blind spot.

wakeless:
I've looked at `@relay_test_operation` and it's kinda what I need, but the main downside is that it doesn't exist on fragments. It's also not really in the shape of the AST which is what's really appealing to this approach. Though a directive to enable it definitely sounds like something I'd be happy with. I'm totally in understanding of the "keep it lean" by default nature of the defaults.

josephsavona:
I see, thanks for clarifying the use-case. I wonder if a schema introspection query would work here? GraphQL is designed to support querying the schema, which would allow you to fetch type information explicitly.

wakeless:
I'm sure I could get it, with something like that, but the goal was to have it fully driven by Relay, so an engineer can simply define a fragment as they do elsewhere and pass it through. I would definitely have gone down a different path if the absolute majority of the data wasn't already coming out of the relay compiler.

What made me think this was the right place for this is that the types generated from the compiler will have a number, or an int, or a bool in them, ie. they have more type-level context than the relay compiler outputs itself.

It really then wouldn't be a hard task for the flow (or typescript) bindings to just sit on top of the relay AST rather than be a first-class artifact it outputs.

A lot of that is YAGNI I suppose, but it certainly was surprising for me that the AST contained less than the types.

josephsavona:
Thanks for clarifying the use-case, I can definitely see how having the scalar types is helpful in this case. However, generally speaking we think of the generated artifacts as solely for consumption by relay-runtime: the format is subject to change at any time (and we've frequently iterated on it over the years), and for efficiency reasons we want to avoid adding unnecessary data. While we could enable emitting types via a directive, that adds complexity to the compiler. Features tend to be add-only, in the sense that once we add a feature and someone uses it, it's difficult to remove and we then have to support it. Given that, we won't be able to accept this PR.

I'm very curious if the approach of using introspection to get this information could work.

sibelius:
can you solve this using an `union` type?

so you can have different selection per different type?

josephsavona:
@sibelius yeah, that’s another great way to handle this. If the types matter to the client, then you can make them part of the data model. You could have a Form type with a list of Field values, each w type info, a default value, etc.

PR #3121, opened 2020-07-07
Support defer/stream spec proposal

robrichard:
This change will allow Relay to support `@defer`/`@stream` in the form that is currently being proposed to the GraphQL WG:

Draft Spec: https://github.com/graphql/graphql-spec/pull/742
GraphQL-js: https://github.com/graphql/graphql-js/pull/2319

This works by allowing either `initial_count` or `initialCount` as an argument for `@stream`. Whichever one is passed will be forwarded to the GraphQL server.

Additionally, either `extensions.is_final: true` or `hasNext: false` will be used to track final payloads.

### DIFF
--- a/compiler/crates/relay-transforms/src/apply_transforms.rs
+++ b/compiler/crates/relay-transforms/src/apply_transforms.rs
@@ -143,14 +143,21 @@ fn apply_common_transforms(
     )?;
 
     program = log_event.time("transform_connections", || {
-        transform_connections(&program, &project_config.schema_config.connection_interface)
+        transform_connections(&program, &project_config.schema_config)
     });
     program = log_event.time("mask", || mask(&program));
     program = log_event.time("transform_defer_stream", || {
-        transform_defer_stream(&program)
+        transform_defer_stream(
+            &program,
+            &project_config.schema_config.defer_stream_interface,
+        )
     })?;
     program = log_event.time("transform_match", || {
-        transform_match(&program, &project_config.feature_flags)
+        transform_match(
+            &program,
+            &project_config.feature_flags,
+            &project_config.schema_config.defer_stream_interface,
+        )
     })?;
     program = log_event.time("transform_subscriptions", || {
         transform_subscriptions(&program)
@@ -236,13 +243,23 @@ fn apply_reader_transforms(
     )?;
 
     program = log_event.time("inline_data_fragment", || inline_data_fragment(&program))?;
-    program = log_event.time("skip_unreachable_node", || skip_unreachable_node(&program))?;
+    program = log_event.time("skip_unreachable_node", || {
+        skip_unreachable_node(
+            &program,
+            &project_config.schema_config.defer_stream_interface,
+        )
+    })?;
     program = log_event.time("remove_base_fragments", || {
         remove_base_fragments(&program, &base_fragment_names)
     });
 
     log_event.time("flatten", || flatten(&mut program, true, false))?;
-    program = log_event.time("skip_redundant_nodes", || skip_redundant_nodes(&program));
+    program = log_event.time("skip_redundant_nodes", || {
+        skip_redundant_nodes(
+            &program,
+            &project_config.schema_config.defer_stream_interface,
+        )
+    });
     program = log_event.time("generate_data_driven_dependency_metadata", || {
         generate_data_driven_dependency_metadata(&program)
     });
@@ -408,7 +425,12 @@ fn apply_normalization_transforms(
         print_stats("hash_supported_argument", &program);
     }
 
-    program = log_event.time("skip_unreachable_node", || skip_unreachable_node(&program))?;
+    program = log_event.time("skip_unreachable_node", || {
+        skip_unreachable_node(
+            &program,
+            &project_config.schema_config.defer_stream_interface,
+        )
+    })?;
     if let Some(print_stats) = maybe_print_stats {
         print_stats("skip_unreachable_node", &program);
     }
@@ -433,7 +455,12 @@ fn apply_normalization_transforms(
         print_stats("flatten", &program);
     }
 
-    program = log_event.time("skip_redundant_nodes", || skip_redundant_nodes(&program));
+    program = log_event.time("skip_redundant_nodes", || {
+        skip_redundant_nodes(
+            &program,
+            &project_config.schema_config.defer_stream_interface,
+        )
+    });
     if let Some(print_stats) = maybe_print_stats {
         print_stats("skip_redundant_nodes", &program);
     }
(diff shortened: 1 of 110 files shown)


josephsavona:
Awesome, thanks for sending this! Any interest in adding equivalent support to the rust version of the compiler? :-)

robrichard:
@josephsavona I added support to the rust compiler. I did not run it against any actual code though, just worked through the tests.

compiler/crates/graphql-transforms/src/transform_connections.rs
```diff
@@ -125,6 +125,7 @@ impl<'s> ConnectionTransform<'s> {
             for arg in &connection_directive.arguments {
                 if arg.name.item == DEFER_STREAM_CONSTANTS.if_arg
                     || arg.name.item == DEFER_STREAM_CONSTANTS.initial_count_arg
+                    || arg.name.item == DEFER_STREAM_CONSTANTS.initial_count_arg_oss
```
maraisr, line 128: I don't think this is the right approach. It should be either, not both. Have a look at how i'm currently supporting fb/oss for connection things.

https://github.com/facebook/relay/pull/3112/files#diff-86ef1b8ad53dcd24d2a88fabf01a1fc7R40

There is a few issues with it though, so still ironing out the kinks

robrichard: My assumption is the schema you are working with is not likely to support both. In this case if you did pass both `initialCount` and `initial_count` to a `@stream_connection` directive, you will get a schema validation error that one of those args is invalid.

maraisr: No sorry. Its more that _both_ shouldn't be supported at the same time. As the spec _will_ dictate the oss variant should be used. All I was saying, is to instead of _or_ the 2, put it behind a compiler flag, such that OSS people only get the spec version.

josephsavona: Agree here: rather than have the code allow either argument name, this would ideally be a compiler config (similar to what we do for connections, which also have different naming internally and in OSS). Then the code would just check for that single name.

robrichard: @josephsavona is the compiler config for the js compiler in the Relay repo? I only see the OSS names https://github.com/facebook/relay/blob/master/packages/relay-runtime/handlers/connection/ConnectionInterface.js#L55

josephsavona: Yeah we have an internal version of that file too, which uses FB's naming (basically the same but snake case). I think we can do the same thing, just put all the constants in a file similar to ConnectionInterface, and then we can maintain an internal version.

robrichard: @josephsavona I updated the PR to add a config to the JS compiler similar to the ConnectionInterface. '

The `hasNext` => `extensions.is_final` is still a runtime check, since it is not a just a property name change (and the check is in relay-runtime not the compiler).

I also removed the rust change. @maraisr maybe you can help out and add it to your pr https://github.com/facebook/relay/pull/3112?

robrichard:
@josephsavona the spec is not approved. The consensus from the last GraphQL WG meeting was that the next step is to get `@defer` and `@stream` merged to graphql-js behind an experimental flag. The intention is to make it easier for people to try it out and start getting real word feedback from outside of Facebook and 1stdibs before finalizing the spec. Maybe it makes sense to get this merged once the related code is in GraphQL-js (https://github.com/graphql/graphql-js/pull/2319)?

compiler/crates/graphql-transforms/src/transform_connections.rs
kassens: See 1386365510126b337a1409dd2a7ff11dbe895f66 for how we can add this to the config in the Relay Rust compiler.

robrichard:
@josephsavona @kassens official experimental branches of graphql-js with support for `@defer` and `@stream` [are now being published](https://www.npmjs.com/package/graphql/v/15.4.0-experimental-stream-defer.1). The working group is aligned on this approach but is waiting on community feedback before finalizing.

Do you think this can be merged so Relay users can easily start testing out `@defer` and `@stream`?

Still working on/trying to get help on the rust compiler, I can open a separate PR when that is ready.

sibelius:
it would be great for the next Relay release

robrichard:
@josephsavona @kassens I updated this PR to make the same change to the rust compiler. The check failures are also failing on master, locally `cargo build` and `cargo test` are working as expected.

It was a little more involved than the `ConnectionInterface` since the codegen also needs to be passed the configurable `DeferStreamInterface`. I'm still a Rust beginner so let me know if anything could be done in a better way.

sorenhoyer:
`Module not found: - Unknown directive 'defer'.` - I'm so ready to start using this now!

chriserickson:
Curious what the status of this PR is - it looks like it was being held until the spec was approved (which it has been now), but is now a bit stale and has conflicts. Definitely interested in using this and could help contribute.

chriserickson:
> Curious what the status of this PR is - it looks like it was being held until the spec was approved (which it has been now), but is now a bit stale and has conflicts. Definitely interested in using this and could help contribute.

Oh, I see, they are still experimental.

sorenhoyer:
Would be great to have this merged soon, so we can start experimenting with defer and stream in Relay. Seems like Apollo Client has planned support for this in September. :-)

robrichard:
@alunyov @rbalicki2 This branch up to date and working again. It's set up the same was as ConnectionInterface, a new DeferStreamInterface can be passed to the config, replacing the static DEFER_STREAM_CONSTANTS. The changes are somewhat large, but very repetitive, as the new interface needed to be passed to several transforms. I've separated the changes to relay-runtime, compiler code, test code, and test fixtures into separate commits which should make it easier to review.

compiler/crates/relay-transforms/src/apply_transforms.rs
```diff
@@ -122,14 +122,25 @@ fn apply_common_transforms(
     let log_event = perf_logger.create_event("apply_common_transforms");
     log_event.string("project", project_config.name.to_string());
     let mut program = log_event.time("transform_connections", || {
-        transform_connections(&program, &project_config.schema_config.connection_interface)
+        transform_connections(
+            &program,
+            &project_config.schema_config.connection_interface,
```
alunyov, line 127: I think we should consider adding `schema_config` directly to the `program`.

We already have `Arc<Schema>` there. And I think it would be easier if we can just always have them together: `schema` and `schema_config`.

Let me try this internally.

alunyov: Update: I could not come up with a better version of the Schema with Config. We have some some non-relay users of Programs/Schema, so adding a dependency on `relay-config` doesn't make sense for them. And adding something like a trait for `Program` is a bigger change. 

Let's keep it for now, but maybe in this specific line: let's pass a `&project_config.schema_config` to the `transform_connections` instead of individual config options.

robrichard: @alunyov I've updated `transform_connections` to accept schema_config

alunyov:
@robrichard, I have imported this PR internally. It still needs some additional handling to land it internally. Also, @rbalicki2 was working on this integration recently. I think, once Robert is back from the holidays, he may have some feedback.

But overall, this looks great, and I'm pretty optimistic that this will land soon. Thank you for your work!

Tiedye:
Any updates on this?

robrichard:
Continued in https://github.com/facebook/relay/pull/4467

PR #3125, opened 2020-07-14
remove persisted query id invariant from loadQuery

zth:
This removes the persisted query id invariant, allowing use of `loadQuery` without persisted queries enabled.


packages/relay-experimental/loadQuery.js
```diff
@@ -194,13 +194,6 @@ function loadQuery<TQuery: OperationType, TEnvironmentProviderOptions>(
     const request = getRequest(graphQlTaggedNode);
     params = request.params;
```
morrys, line 196: @zth, @josephsavona I think it would be better to manage the moduleId using the id for persisted queries and the cacheID for non-persisted queries instead of removing the check.

const moduleId = params.cacheID ? params.cacheID : params.id;

PR #3135, opened 2020-07-18
Plural support for @deleteRecord directive

zth:
This adds support for using the new `@deleteRecord` directive on a list of `ID` in addition to a single `ID`.

There's a few things around the implementation I'm looking for guidance on before this is ready for a real review.

### DIFF
--- a/packages/relay-compiler/transforms/DeclarativeConnectionMutationTransform.js
+++ b/packages/relay-compiler/transforms/DeclarativeConnectionMutationTransform.js
@@ -66,11 +66,7 @@ function visitScalarField(field: ScalarField): ScalarField {
   }
   const schema = this.getContext().getSchema();
 
-  if (
-    ['ID', 'ID!', '[ID]', '[ID!]', '[ID!]!'].includes(
-      schema.getTypeString(field.type),
-    ) === false
-  ) {
+  if (!schema.isId(schema.getRawType(field.type))) {
     throw createUserError(
       `Invalid use of @${DELETE_RECORD} on field '${
         field.name
--- a/packages/relay-compiler/transforms/DeclarativeConnectionMutationTransform.js
+++ b/packages/relay-compiler/transforms/DeclarativeConnectionMutationTransform.js
@@ -70,7 +70,7 @@ function visitScalarField(field: ScalarField): ScalarField {
     throw createUserError(
       `Invalid use of @${DELETE_RECORD} on field '${
         field.name
-      }'. Expected field type to be a single or list of ID, got ${schema.getTypeString(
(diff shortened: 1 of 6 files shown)


packages/relay-compiler/transforms/DeclarativeConnectionMutationTransform.js
```diff
@@ -65,11 +65,18 @@ function visitScalarField(field: ScalarField): ScalarField {
     return field;
   }
   const schema = this.getContext().getSchema();
-  if (!schema.isId(field.type)) {
+
+  if (
+    !schema.isId(field.type) &&
+    (!schema.isList(field.type) ||
+      (schema.isList(field.type) && !schema.isId(field.type.ofType)))
```
zth, line 72: This results in a Flow error, since I couldn't figure out how to properly refine `field.type` into a list of `ID`s. Any guidance on how to refine that type properly using `schema` is appreciated.

packages/relay-compiler/transforms/DeclarativeConnectionMutationTransform.js
```diff
@@ -65,11 +65,18 @@ function visitScalarField(field: ScalarField): ScalarField {
     return field;
   }
   const schema = this.getContext().getSchema();
-  if (!schema.isId(field.type)) {
+
+  if (
+    ['ID', 'ID!', '[ID]', '[ID!]', '[ID!]!'].includes(
```
zth, line 70: What's your take on this approach? Problematic?

josephsavona: you can use `if (!schema.isId(schema.getRawType(field.type)))` to unwrap NonNull/List wrappers and get at the underlying type.

packages/relay-compiler/transforms/__tests__/__snapshots__/DeclarativeConnectionMutationTransform-test.js.snap
```diff
@@ -91,7 +110,7 @@ mutation CommentDeleteMutation($input: CommentDeleteInput) {
 ~~~~~~~~~~ OUTPUT ~~~~~~~~~~
 THROWN EXCEPTION:
 
-Invalid use of @deleteRecord on field '__typename'. Expected field type ID, got String!.
+Invalid use of @deleteRecord on field '__typename'. Expected field type to be a single or list of ID, got String!.
```
josephsavona, line 113: maybe "Expected field to return an ID or list of ID values, got String!"

zth:
Thanks for the review @josephsavona ! I've fixed the issues you pointed out.

PR #3138, opened 2020-07-18
Run tests for`relay-compiler-neon`and `relay-compiler-wasm` with GH Actions

alunyov:
This PR adds two new steps for our GH Actions - Tests for `neon` and `wasm` extensions of the `relay-compiler`.

Both steps will build an extension and run `jest` tests for it with different versions of `Node` (10,12,14).


josephsavona:
@alunyov just a heads up that when i run the relay-compiler-wasm unit tests locally w `yarn test` i'm getting an error on the second test case (with intentionally invalid syntax) because [this line](https://github.com/facebook/relay/blob/master/compiler/crates/relay-compiler-wasm/src/lib.rs#L22) panics rather than returning an Err value.

alunyov:
Thanks @josephsavona, I'll send PR for this later. I think the test is passing because in the test I'm just doing `expect(...).toThrow()`. And it throws, and prints panic message to console

PR #3148, opened 2020-07-26
Add connections argument to @deleteRecord directive

zth:
This adds support for removing nodes from connections when using the `@deleteRecord` directive.


packages/relay-runtime/handlers/connection/MutationHandlers.js
```diff
@@ -27,9 +27,37 @@ import type {
 const DeleteRecordHandler = {
   update: (store: RecordSourceProxy, payload: HandleFieldPayload) => {
     const record = store.get(payload.dataID);
+
     if (record != null) {
       const id = record.getValue(payload.fieldKey);
+
       if (typeof id === 'string') {
+        const {connections} = payload.handleArgs;
```
zth, line 35: I'm stuck on `connections` not appearing here in `handleArgs`, even though I _think_ I'm doing the transform right for passing the arguments on to here. But yeah, it's not working so obviously I'm doing something wrong.

josephsavona:
cc @kassens @tyao1

n1ru4l:
Hey I already mentioned it in the [Twitter Thread a while back](https://twitter.com/___zth___/status/1291085643662217216) and I wonder whether something like this could become part of relay:

> An interesting approach would be to automatically remove a deleted record from all referenced lists + remove all records that reference the deleted record (in a non-nullable way) up to the next nullable field/list

zth:
@tyao1 sounds good! I'll change this to its own directive.

zth:
Quick update: I'll fix this as soon as https://github.com/facebook/relay/pull/3155 lands, as that touches a lot of the same things.

zth:
Superseded by https://github.com/facebook/relay/pull/3177.

PR #3155, opened 2020-07-31
Add @appendNode and @prependNode directives for updating the store

zth:
This adds two new directives, `@appendNode` and `@prependNode`, for updating the store declaratively. The directives will create an edge for the annotated object and prepend/append that edge to the provided connections. It works with single nodes as well as lists of nodes.

Example:
```
mutation AddComments(
  $input: AddCommentsInput!, 
  $connections: [String!]!, 
  $edgeTypeName: String!
) {
  addComments(input: $input) {
    # It works for lists of nodes as well as for single nodes
    comments @appendNode(connections: $connections, edgeTypeName: $edgeTypeName) {
      id
    }

    # This mutation naturally doesn't make sense, but is just here to illustrate that this works for a single node as well
    comment @appendNode(connections: $connections, edgeTypeName: $edgeTypeName) {
      id
    }
  }
}
```


packages/relay-compiler/transforms/DeclarativeConnectionMutationTransform.js
```diff
@@ -100,55 +115,111 @@ function visitLinkedField(field: LinkedField): LinkedField {
...
+      throw createUserError(
+        `Unsupported use of @${nodeDirective.name} on field '${transformedField.name}', 'edgeTypeName' argument must be provided.`,
+        [targetDirective.loc],
+      );
+    }
+    const rawType = schema.getRawType(transformedField.type);
+    if (
+      schema.isObject(rawType) ||
+      schema.isUnion(rawType) ||
+      schema.isInterface(rawType)
```
zth, line 195: Are these the right constraints?

josephsavona:
cc @kassens @tyao1

packages/relay-compiler/transforms/DeclarativeConnectionMutationTransform.js
tyao1: I think `schema.canHaveSelections` is more accurate.

zth: Great, thanks! I've fixed that now.

zth:
@tyao1 Thanks for having a look. Yes it does. I added a test to show it, please see the latest commit.

anatooly:
ERROR:
Variable '$connections' was defined as type '[String!]!' but used in a location expecting the type '[ID!]!'

tobias-tengler:
Would be great if these directives could be documented!

anatooly:
How do this variant?
```
mutation CommentCreateMutation(
  $connections: [String!]!
  $input: CommentCreateInput
) {
  commentCreate(input: $input) @appendNode(
      connections: $connections
      edgeTypeName: "CommentEdge"
    ) {
      id
      ...Comment_comment
    }
  }
}
```
index.js:1 Warning: MutationHandlers: Expected target node to exist.

giautm:
Hey @anatooly , I found if I use directives on a field with arguments they will not work. Please retry with a field without agrument, I believe there is a bug in implement of those directives

PR #3166, opened 2020-08-18
Remove unused promise-polyfill dependency

kassens:
I think we used to auto-require `Promise` when `Promise` was used in the
code and that would then be rewritten to `require('promise-polyfill')`.
We no longer ship a Promise implementation, so I think this can go now.

Test Plan:
yarn, then grep over `dist/`.


kassens:
oh, this is used for jest tests to run promises with `jest.runAllTimers()`.

PR #3177, opened 2020-09-05
implement deleteEdge store updater directive

zth:
_This supersedes https://github.com/facebook/relay/pull/3148 which I've now closed._

This implements a new declarative for updating the store declaratively after mutations called `@deleteEdge`. As the name hints at, this particular directive allows you to remove a node's edge from the provided connections. Please note that this directive does not _delete the node_, only edge(s) for the node. There's `@deleteRecord` already for deleting a record, which can be combined with this directive.

It's intended to be used like this:

```
mutation DeleteComment($input: DeleteCommentInput!, $connections: [String!]!) {
  deleteComment(input: $input) {
    deletedCommentId @deleteEdge(connections: $connections)
    # This will delete any edge for the node with id `deletedCommentId` from the 
    # connections provided through `$connections`
  }
}
```

It works for single IDs (as demonstrated above) as well as a list of IDs.

### DIFF
--- a/packages/relay-compiler/codegen/NormalizationCodeGenerator.js
+++ b/packages/relay-compiler/codegen/NormalizationCodeGenerator.js
@@ -384,13 +384,7 @@ function generateScalarField(node): Array<NormalizationSelection> {
   const handles =
     (node.handles &&
       node.handles.map(handle => {
-        if (handle.dynamicKey != null) {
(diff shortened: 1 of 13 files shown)


zth:
cc @tyao1

tyao1 (approved):
Awesome! Thanks for implementing this!

Issue #3181, opened 2020-09-11
Typescript generator as something configured in the compiler config.json


maraisr:
Following on from #3180, this here is to track the progress and implementation of the Typescript relay-typegen initiative. 

I think a good place to start looking is https://github.com/facebook/relay/tree/master/compiler/crates/relay-typegen/src

But @kassens  if you don't mind dropping some notes (or even edit this issue) about where to go, what to do and some rough implementation guide. I'd be more than happy to tackle this, this weekend! (I know I've said this in the past, but got stuck pretty quick).

kassens:
This was resolved by ab5c96b, I'll try to add more details to the main issue how to run from OSS!

Issue #3180, opened 2020-09-10
[epic] enable Rust compiler for OSS


kassens:
This issue collects what I think is missing to update the Relay compiler to the Rust version in OSS. If someone want's to give any of this a shot, please comment here!

- [x] Typescript generator as something configured in the compiler config.json
- [x] Non-watchman file source. Currently we convert the directories in `config.json` into a watchman query for watch mode and simple builds. This was designed with non-watchman in mind and it should be possible to build out something that uses a different file-crawler and maybe file watcher crate. This is not strictly required, but I think removing this external required dependency would be nice. (Details in #3193)
- [x] Figure out how we deploy a pre-build binary using npm. The `flow-bin` package might be a good inspiration as that somehow loads the binary for that version on first use. (Details/discussion in #3194)
- [x] Some missing config options in how types are generated (Details/discussion in #3195)
- [x] Refine Typescript generation (Details/discussion in #3196)
- [ ] https://github.com/facebook/relay/issues/3438
- [x] https://github.com/facebook/relay/issues/3437
- [x] Look at assumptions of using hg
- [x] Figure out situation with required variables with default values (see: https://github.com/facebook/relay/pull/3612 )
- [x] Documentation/Blog Post about the release of new version of the compiler
- [x] Deprecate JS version of the compiler
- [x] #3627
- [ ] Fix issues listed here: https://github.com/vincentriemer/new-relay-compiler-issues-repro

kassens:
@MaartenStaa made a huge leap start on this with #3182!

I understand there's a few improvements we can make to the types still by copying more what the JS OSS typescript generator produces: https://github.com/relay-tools/relay-compiler-language-typescript/blob/e1b092f081d4e816ecd17fb0d3b4e62cdf3db917/test/__snapshots__/TypeScriptGenerator-test.ts.snap

I'll file separate issues for the remaining checkboxes.

kassens:
A good intro way to contribute is #3199 to convert the runtime unit tests to use artifacts generated by the Rust compiler. More details in there.

alunyov:
Thanks @stale. I think we still need the issue. I'm looking at converting the rest of the tests. It may take a while, but we'll see.

alunyov:
[...]

PR #3182, opened 2020-09-11
Add support for Typescript in Rust Relay compiler

MaartenStaa:
See #3181 and #3180.

This PR adds a language option to the typegen_config and refactors the relay-typegen crate to allow completely different output based on the configured language, currently typescript and flow (which is the default when no language is configured, or an invalid one is set).

I also made some changes so the watchman query looks for .ts files when using typescript, and the output files also use .ts as their extension.

This probably needs a bit more work before it's completely ready. Off the top of my head:

- [ ] In `relay-compiler/build_project/artifact_content.rs` some output is generated such as `@flow` and a commented out `import type` statement? Those should probably be left out for Typescript.
- [x] Currently the language used for typegen is Typescript if the language in the config file is "typescript", and Flow otherwise. We should probably display an error if an invalid language is set, e.g. "typesrcipt".
- [ ] I was not able to actually test the compiler due to an error returned by Watchman: `failed to parse query: must use ["suffix", "suffixstring"]`, and I don't have a clue why. This error was already returned before I made the changes to look for .ts files when using Typescript.
- [ ] I did not at all look into `relay-lsp` and whether any changes are needed there for Typescript.

Looking forward to your feedback!

cc @kassens @josephsavona @maraisr

### DIFF
--- a/compiler/crates/relay-compiler/src/watchman/query_builder.rs
+++ b/compiler/crates/relay-compiler/src/watchman/query_builder.rs
@@ -29,10 +29,11 @@ pub fn get_watchman_expr(config: &Config) -> Expr {
                 Some(p) if p.enabled => Some(Expr::All(vec![
                     // Ending in *.js(x) or *.ts(x) depending on the project language.
                     Expr::Suffix(match &p.typegen_config.language {
-                        TypegenLanguage::Flow => vec!["js", "jsx"],
-                        TypegenLanguage::Typescript => vec!["ts", "tsx"],
-                    }
-                    .into()),
+                        TypegenLanguage::Flow => vec![PathBuf::from("js"), PathBuf::from("jsx")],
+                        TypegenLanguage::Typescript => {
+                            vec![PathBuf::from("ts"), PathBuf::from("tsx")]
+                        }
+                    }),
                     // In the related source root.
                     Expr::DirName(DirNameTerm {
                         path: path.clone(),
--- a/compiler/crates/relay-compiler/src/watchman/query_builder.rs
+++ b/compiler/crates/relay-compiler/src/watchman/query_builder.rs
@@ -27,12 +27,12 @@ pub fn get_watchman_expr(config: &Config) -> Expr {
             })
             .filter_map(|(path, project)| match project {
                 Some(p) if p.enabled => Some(Expr::All(vec![
-                    // Ending in *.js or *.ts depending on the project lanague.
-                    Expr::Suffix(vec![match &p.typegen_config.language {
-                        TypegenLanguage::Flow => "js",
-                        TypegenLanguage::Typescript => "ts",
+                    // Ending in *.js(x) or *.ts(x) depending on the project language.
+                    Expr::Suffix(match &p.typegen_config.language {
+                        TypegenLanguage::Flow => vec!["js", "jsx"],
+                        TypegenLanguage::Typescript => vec!["ts", "tsx"],
                     }
-                    .into()]),
+                    .into()),
                     // In the related source root.
                     Expr::DirName(DirNameTerm {
                         path: path.clone(),
--- a/compiler/crates/relay-compiler/src/watchman/query_builder.rs
+++ b/compiler/crates/relay-compiler/src/watchman/query_builder.rs
@@ -7,6 +7,7 @@
 
 use crate::compiler_state::SourceSet;
 use crate::config::{Config, SchemaLocation};
+use relay_typegen::TypegenLanguage;
 use std::path::PathBuf;
 use watchman_client::prelude::*;
 
@@ -28,8 +29,8 @@ pub fn get_watchman_expr(config: &Config) -> Expr {
                 Some(p) if p.enabled => Some(Expr::All(vec![
                     // Ending in *.js or *.ts depending on the project lanague.
                     Expr::Suffix(vec![match &p.typegen_config.language {
-                        Some(language) if language == "typescript" => "ts",
-                        _ => "js",
+                        TypegenLanguage::Flow => "js",
+                        TypegenLanguage::Typescript => "ts",
                     }
                     .into()]),
                     // In the related source root.
--- a/compiler/crates/relay-compiler/src/watchman/query_builder.rs
+++ b/compiler/crates/relay-compiler/src/watchman/query_builder.rs
@@ -5,22 +5,43 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+use crate::compiler_state::SourceSet;
 use crate::config::{Config, SchemaLocation};
 use std::path::PathBuf;
 use watchman_client::prelude::*;
 
 pub fn get_watchman_expr(config: &Config) -> Expr {
-    let mut sources_conditions = vec![
-        // ending in *.js
-        Expr::Suffix(vec!["js".into()]),
-        // in one of the source roots
-        expr_any(
-            get_source_roots(&config)
-                .into_iter()
-                .map(|path| Expr::DirName(DirNameTerm { path, depth: None }))
-                .collect(),
-        ),
-    ];
+    let mut sources_conditions = vec![expr_any(
+        config
+            .sources
+            .iter()
+            .flat_map(|(path, name)| match name {
+                SourceSet::SourceSetName(name) => {
+                    std::iter::once((path, config.projects.get(&name))).collect::<Vec<_>>()
+                }
+                SourceSet::SourceSetNames(names) => names
+                    .iter()
+                    .map(|name| (path, config.projects.get(name)))
+                    .collect::<Vec<_>>(),
+            })
+            .filter_map(|(path, project)| match project {
+                Some(p) if p.enabled => Some(Expr::All(vec![
+                    // Ending in *.js or *.ts depending on the project lanague.
+                    Expr::Suffix(vec![match &p.typegen_config.language {
+                        Some(language) if language == "typescript" => "ts",
+                        _ => "js",
+                    }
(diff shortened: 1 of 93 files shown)


josephsavona:
Awesome! I took a quick look through and the high-level approach makes sense, of having typegen create an abstract AST of types and then having Flow/TS-specific writers to convert that AST to code. Thanks for the contribution! 

I'll leave to @tyao1, @kassens, and @rbalicki2 for review

zth:
This is so cool, awesome work @maartenstaa ! 😍

compiler/crates/relay-compiler/src/watchman/query_builder.rs
```diff
@@ -5,22 +5,44 @@
...
+                    .iter()
+                    .map(|name| (path, config.projects.get(name)))
+                    .collect::<Vec<_>>(),
+            })
+            .filter_map(|(path, project)| match project {
+                Some(p) if p.enabled => Some(Expr::All(vec![
+                    // Ending in *.js or *.ts depending on the project lanague.
+                    Expr::Suffix(vec![match &p.typegen_config.language {
+                        TypegenLanguage::Flow => "js",
+                        TypegenLanguage::Typescript => "ts",
```
maraisr, line 33: with react, this has to be `tsx`. So could this be an array perhaps?

MaartenStaa: You're right, I adjusted it! I set it to look at .jsx as well, just in case people use that.

MaartenStaa: @maraisr Fixed the build now as well.

compiler/crates/relay-typegen/src/config.rs
```diff
@@ -9,9 +9,26 @@ use fnv::FnvHashMap;
...
+#[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields, rename_all = "camelCase")]
+pub enum TypegenLanguage {
+    Flow,
+    Typescript,
+}
+
+impl Default for TypegenLanguage {
+    fn default() -> Self {
+        Self::Flow
```
kassens, line 21: I was even thinking of making Typescript the default, but just realized `*.ts` files wouldn't work with vanilla JS anymore. The Flow generated `*.js` files using Flow comment style syntax allows people to just treat them as pure JS files.

Maybe we could even make this optional and skip generating types altogether to make the files a bit smaller.

compiler/crates/relay-compiler/src/watchman/query_builder.rs
```diff
@@ -5,22 +5,45 @@
...
+                    .map(|name| (path, config.projects.get(name)))
+                    .collect::<Vec<_>>(),
+            })
+            .filter_map(|(path, project)| match project {
+                Some(p) if p.enabled => Some(Expr::All(vec![
+                    // Ending in *.js(x) or *.ts(x) depending on the project language.
+                    Expr::Suffix(match &p.typegen_config.language {
+                        TypegenLanguage::Flow => vec![PathBuf::from("js"), PathBuf::from("jsx")],
+                        TypegenLanguage::Typescript => {
+                            vec![PathBuf::from("ts"), PathBuf::from("tsx")]
```
kassens, line 34: For a TS project, I suppose you'd have *all* your own product code in `*.tsx?` files or should this also include `.js` files for hybrid projects?

kassens (commented):
First, I want to repeat everyone here and say this is awesome and pretty much aligns with how I thought it'd be done ideally.

Did you come up with the way the fragment references are modeled in typescript or was this based on the existing (community driven) typescript types? I know this was pretty tricky to get right in Flow.

alloy:
Awesome, very nice to see this @MaartenStaa  👏

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/unmasked-fragment-spreads.expected
```diff
@@ -0,0 +1,87 @@
...
+  readonly uri?: string,
+  readonly width?: number,
+};
+export type RecursiveFragment$data = RecursiveFragment;
+export type RecursiveFragment$key = {
+  readonly $data?: RecursiveFragment$data,
+  readonly $fragmentRefs: RecursiveFragment$ref,
+};
+-------------------------------------------------------------------------------
+import type { PhotoFragment$ref } from "PhotoFragment.graphql";
```
alloy, line 70: It would be better to avoid any need to import these types, as that entirely removes the need to have a single artefact directory (or Haste). See for example [this output](https://github.com/relay-tools/relay-compiler-language-typescript/blob/e1b092f081d4e816ecd17fb0d3b4e62cdf3db917/test/__snapshots__/TypeScriptGenerator-test.ts.snap#L495) of the existing TS plugin.

kassens:
- I'd like to just have TS output be included as I'm not sure of a good way to make it pluggable and maintainable.
- Unfortunately, I (and the Rest of FB Relay engineers) unfortunately have limited experience and time to polish the TS output and figure out how it's supposed to work.
- This PR seems like a great start and make it easier to folks to start contributing changes.

Given the above, I'm somewhat inclined to merge this and have it open up for contributions to tweak the output in follow up PRs. Thoughts?

compiler/crates/relay-compiler/src/watchman/query_builder.rs
maraisr: I think generally its all ts or all js. But can't presume that, so think yeah maybe hybrids are nice?

maraisr:
Yeah @kassens I think that is a great way forward. Also a great way for non-Rust contributors to have a go, as is not as daunting as starting something from scratch. (Mind you @MaartenStaa did just that 😛).

I'm super keen to investigate a cleaner named interface, and proper union types. And think it was outlined in another ticket already, around not merging unions and forcing things to be nullable.

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/conditional.expected
```diff
@@ -0,0 +1,36 @@
...
+  readonly id?: string,
+  readonly $refType: ConditionField$ref,
+};
+export type ConditionField$data = ConditionField;
+export type ConditionField$key = {
+  readonly $data?: ConditionField$data,
+  readonly $fragmentRefs: ConditionField$ref,
+};
+-------------------------------------------------------------------------------
+import { FragmentReference } from "relay-runtime";
```
maraisr, line 25: think though maybe this should `import type {}` no?

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/unmasked-fragment-spreads.expected
maraisr: Yeah... I think this is actually a blocker for a merge. Typescript wouldn't know what that means (please correct me here).

At the very least just prefix with a `./`, 

```diff
-import type { PhotoFragment$ref } from "PhotoFragment.graphql";
+import type { PhotoFragment$ref } from "./PhotoFragment.graphql";
```

alloy: Hmm, yeah that’s a good point. @kassens can you amend that in your import?

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/conditional.expected
alloy: Not a requirement, but definitely something to follow up with.

MaartenStaa: @maraisr correct. It was a conscious decision to use this option since `import type` was a relatively recent addition to Typescript, and I wouldn't want to make this unusable for people on older versions.

MaartenStaa:
> Given the above, I'm somewhat inclined to merge this and have it open up for contributions to tweak the output in follow up PRs. Thoughts?

This sounds like a reasonable way to move forward, though I feel somewhat uneasy about having this merged at this stage since I wasn't able to test this yet. 

>(Mind you @MaartenStaa did just that 😛).

To be fair, I have played around with Rust before in my spare time, so it's not like I dove in with zero experience. 

> I'm super keen to investigate a cleaner named interface, and proper union types. And think it was outlined in another ticket already, around not merging unions and forcing things to be nullable. 

I think you may be referring to my other PR, and how we ended up here in the first place: #3178.

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/conditional.expected
alloy: The fragment reference import _does_ currently use the `type` keyword, though.

alloy:
> This sounds like a reasonable way to move forward, though I feel somewhat uneasy about having this merged at this stage since I wasn't able to test this yet.

I think it would be good to port e.g. the TS example that exists in the TS plugin repo to use the rust compiler and then iterate on any issues and the few discussed here over the next few weeks. Others (e.g. me) can chime in too.

As long as the relay team is ok with not putting out a public release until the dust has settled (@kassens ?).

@MaartenStaa What do you think?

mrjackdavis:
For those of you coming here from google to fix this error:

```sh
Watchman error: Watchman error: The watchman server reported an error: "failed to parse query: must use ["suffix", "suffixstring"]", while executing command
```

I had to upgrade watchman

```sh
brew upgrade watchman
```

And restart watchman

```sh
watchman shutdown-server
```

PR #3187, opened 2020-09-18
A few small fixes to relay-compiler-neon. A follow-up for relay-compiler-wasm with a better error handling. 

alunyov:
In `relay-compiler-neon`, we've changed the unit-test (to make the build `green`).

In `relay-compiler-wasm` added a better (a little) error handling for document parsing. We still do not translate the actual error message to WASM.


alunyov:
Test is already fixed - de38d415f

PR #3202, opened 2020-09-30
change the ID type for directives handling connection IDs

zth:
This changes the type of the `connections: [String]` arguments to the new store updater directives handling connections from `String` to `ID`. This is to make it match better type-wise and semantically to what it actually is (an ID coming from the store), but since both `String` and `ID` normally serialize to `string`, it won't change the behavior of any product code.

It is however a breaking change in the sense that all operations that use the new store directives with `connections` will need to change from `String` to `ID` in the definition itself.

### DIFF
--- a/compiler/crates/graphql-transforms/tests/declarative_connection/fixtures/append-edge-unspported.invalid.expected
+++ b/compiler/crates/graphql-transforms/tests/declarative_connection/fixtures/append-edge-unspported.invalid.expected
@@ -1,7 +1,7 @@
 ==================================== INPUT ====================================
 # expected-to-throw
 mutation CommentCreateMutation(
-  $connections: [String!]!
+  $connections: [ID!]!
   $input: CommentCreateInput
 ) {
   commentCreate(input: $input) {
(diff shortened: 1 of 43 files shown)


maraisr:
Are `@connection(key: "example")` analogous to an `ID`? I thought it was just a string, and not an ID.

zth:
No you're right that's a string. This just changes the instances where you can supply a list of connection IDs (which are obtained by selecting __id on the connection itself). So a key is still a string.

maraisr:
> 
> 
> No you're right that's a string. This just changes the instances where you can supply a list of connection IDs (which are obtained by selecting __id on the connection itself). So a key is still a string.

But isnt that the same key you pass the `$connectionIds` variable? Or do you pass something different to it?

@zth

zth:
@maraisr it's not. In short, the `key` supplied to `@connection` is just one piece of what'll make up the cache id for that `@connection` in the store. In later versions of Relay, you can request the actual cache ID for a connection by selecting `__id` on the connection itself. You can then feed that id to directives as this PR highlights, or you could use it to access the connection directly in the store without having to look up the parent node first and then use `ConnectionHandler`. An example in pseudo-code:

```graphql
fragment SomeFragment on SomeType {
  someConnectionField @connection(key: "SomeFragment_someConnectionField", filters: []) {
    __id # This is the cache id
    edges {
      node {
       ....
    }
  }
}
```

```javascript
const data = useFragment(...);
const connectionId = data.someConnectionField.__id;

....
store.get(connectionId); // You can use the id to get a connection directly, without having to go through ConnectionHandler
```

So, that's the ID you feed to `$connections`. I hope that makes sense.

PR #3211, opened 2020-10-12
chore: Re-export graphql for the hooks

maraisr:
From what I can see the "internally" released, relay experimental releases with a re-export of graphql, can we also articulate this in the OSS version?

https://unpkg.com/browse/react-relay@0.0.0-experimental-183bdd28/lib/hooks.js

Really tripped me up when I got the types aligned in the new DT types (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/48638/files#diff-cf027a09fee4ac3fe3dddd1ca1649cccR1)


maraisr:
I think we should keep this one open Stale bot,

cc @en_js @kassens

Issue #3142, opened 2020-07-22
[modern] calling preloadQuery causes "Object(...) is not a function" error


Choongkyu:
Hello. When running preloadQuery, I'm running into an issue where I get this message:
```
×
TypeError: Object(...) is not a function
App
src/App.tsx:19
  16 |   }
  17 | `;
  18 | function App() {
> 19 |   const rootQuery = preloadQuery(
  20 |     environment,
  21 |     AppQuery,
  22 |     { userId: "2" },
View compiled
▼ 16 stack frames were expanded.
renderWithHooks
node_modules/react-dom/cjs/react-dom.development.js:14803
mountIndeterminateComponent
node_modules/react-dom/cjs/react-dom.development.js:17482
beginWork
node_modules/react-dom/cjs/react-dom.development.js:18596
HTMLUnknownElement.callCallback
node_modules/react-dom/cjs/react-dom.development.js:188
invokeGuardedCallbackDev
node_modules/react-dom/cjs/react-dom.development.js:237
invokeGuardedCallback
node_modules/react-dom/cjs/react-dom.development.js:292
beginWork$1
node_modules/react-dom/cjs/react-dom.development.js:23203
performUnitOfWork
node_modules/react-dom/cjs/react-dom.development.js:22157
workLoopSync
node_modules/react-dom/cjs/react-dom.development.js:22130
performSyncWorkOnRoot
node_modules/react-dom/cjs/react-dom.development.js:21756
scheduleUpdateOnFiber
node_modules/react-dom/cjs/react-dom.development.js:21188
updateContainer
node_modules/react-dom/cjs/react-dom.development.js:24373
(anonymous function)
node_modules/react-dom/cjs/react-dom.development.js:24758
unbatchedUpdates
node_modules/react-dom/cjs/react-dom.development.js:21903
legacyRenderSubtreeIntoContainer
node_modules/react-dom/cjs/react-dom.development.js:24757
render
node_modules/react-dom/cjs/react-dom.development.js:24840
▲ 16 stack frames were expanded.
Module../src/index.tsx
src/index.tsx:9
   6 | import { RelayEnvironmentProvider } from "react-relay/hooks";
   7 | import { environment } from "./RelayEnvironment";
   8 | 
>  9 | ReactDOM.render(
  10 |   <React.StrictMode>
  11 |     <RelayEnvironmentProvider environment={environment}>
  12 |       <React.Suspense fallback="loading...">
View compiled
__webpack_require__
[...]

PR #3218, opened 2020-10-16
Update StepByStep docs to prevent issues

dantxal:
While going through the guide I ended up with the error in the issue #3142 .

I just used the versions suggested by @goldhand in the issue.

With this fixes, any new user will be able to go through the guide without any problem.


sibelius (changes requested):
it does not make sense to mark a specific version on docs

dantxal:
Now that you said it, I agree, can you close this PR for now?

RafaelVidaurre:
@sibelius I think it does make some sense for an experimental branch, since things are unstable people cannot try it out when the latest experimental release is broken

PR #3221, opened 2020-10-21
Add support for input object extensions

tberman:
I would like to be able to add input types on the client. I am building client side mutations using the missingFields handler but can't create the new input object.


sibelius (approved):

josephsavona:
This seems reasonable - this should already work in the new Rust-based compiler too.

PR #3224, opened 2020-10-22
docs: Talk about the log option for the environment

maraisr:
Here I aim to tell the community about the logging option on Environment

### DIFF
--- a/docs/Modern-RelayEnvironment.md
+++ b/docs/Modern-RelayEnvironment.md
@@ -59,7 +59,7 @@ function handlerProvider(handle) {
 
 ## Adding a `log` function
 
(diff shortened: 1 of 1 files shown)


docs/Modern-RelayEnvironment.md
```diff
@@ -56,3 +56,37 @@ function handlerProvider(handle) {
   );
 }
 ```
+
+## Adding a `log` function
+
+You can get access to the inner life cycle stages of Relay, and log those. Super helpful for tracing, and debugging. For an in-depth look at what log events you can access, I urge you to try it out. Log a few things, and have a play. Alternively you can find a [list of possible events](https://github.com/facebook/relay/blob/37cee8039b888755e4ad4219e703be13a9a49ebb/packages/relay-runtime/store/RelayStoreTypes.js#L410) in the source code.
```
zth, line 62: I don't think the rest of the docs are in "I" form, so maybe drop that here too?

sibelius (approved):
this is great

I do miss the old log using console.groupCollapsed()

maraisr:
Yeah! I mean I'd personally like a context concept to like be able to get variables or operation for the .next etc... because the only way to get that is to save it against the transactionID, but you can't WeakMap that. But I'll pr something for that in the future.

tyao1:
Hi @maraisr! Thanks for adding the doc! 
Currently the log option is unstable and it was just changed a bit in https://github.com/facebook/relay/commit/95606c5ed. We will wait until the option is more stable before merging it into the docs.

PR #3239, opened 2020-10-30
[rust] Refine typescript type generation

### LINKED ISSUE #3196: [rust] refine typescript type generation
@MaartenStaa send a huge first chunk of type generation for TypeScript on the new Rust based compiler.

There is an existing OSS project to generate TypeScript types and there are some differences in the output. We should probably check what's different and what makes sense to incorporate.

OSS project: https://github.com/relay-tools/relay-compiler-language-typescript/blob/e1b092f081d4e816ecd17fb0d3b4e62cdf3db917/test/__snapshots__/TypeScriptGenerator-test.ts.snap

MaartenStaa:
Fixes #3196

The generated output should now be basically identical to what is being generated over at relay-tools/relay-compiler-language-typescript. I didn't write a script to compare the fixtures in an automated manner, but compared samples by hand and fixed cases where the output from the Rust compiler differed.

Looking forward to your feedback!


MaartenStaa:
Just realized I should probably ping @kassens @maraisr and @alloy at least.

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/fragment-spread.expected
```diff
@@ -48,137 +48,118 @@ fragment UserFrag2 on User {
...
-  readonly $refType: FragmentSpread$ref,
+  readonly justFrag: {
+    readonly " $fragmentRefs": FragmentRefs<"PictureFragment">
+  } | null,
+  readonly fragAndField: {
+    readonly uri: string | null,
+    readonly " $fragmentRefs": FragmentRefs<"PictureFragment">,
+  } | null,
+  readonly " $fragmentRefs": FragmentRefs<"OtherFragment" | "UserFrag1" | "UserFrag2">,
+  readonly " $refType": FragmentRefs<"FragmentSpread">,
```
alloy, line 84: @kassens Just a FYI: If you were to switch the Flow emission to use types like these (based on strings) as well, instead of opaque types, we could get rid of any additional code related to haste or a specific artifact directory.

alloy (approved):
I just reviewed the emitted TS and it looks stellar to me 👌 Thanks! 🙏

kassens:
Sorry for missing this!

Curious about the import type, that's a new TypeScript feature, so it's there to support versions before and after TS 3.8?

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/fragment-spread.expected
kassens: Oh, interesting. This is using the fragment names and the fact that they're globally unique as their identifier, clever!

What about the property names? Why `" $refType"` as the key (leading space?)

MaartenStaa: So I wasn't involved with the original design behind the Typescript types, but I believe the leading space causes Typescript to omit this property in its autocomplete.

MaartenStaa:
> Sorry for missing this!
> 
> Curious about the import type, that's a new TypeScript feature, so it's there to support versions before and after TS 3.8?

No problem! Yeah, the `import type` syntax was introduced in TypeScript 3.8. So in older versions you'd just use the regular ES `import x from y` syntax to import both values and types. This still works in in later versions as well, but `import type` is semantically more clear. Also, some lint rules would complain about unused values if you don't use `import type`, though I have personally not run into this problem.

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/fragment-spread.expected
alloy: > So I wasn't involved with the original design behind the Typescript types, but I believe the leading space causes Typescript to omit this property in its autocomplete.

Correct 👍

kassens:
Sorry for the delay, I was on vacation last week, but this should be merged in a bit!

kassens:
Thanks for the contribution again! Helps us slowly to get OSS ready!

PR #3256, opened 2020-11-13
feat - Added the TTL configuration for single query

morrys:
With this PR it is possible to configure a TTL different from the queryCacheExpirationTime configured in the Store for each query.

To do this, simply add the new ttl property to the cacheConfig property of the QueryRenderer

### DIFF
--- a/packages/relay-runtime/store/RelayModernStore.js
+++ b/packages/relay-runtime/store/RelayModernStore.js
@@ -202,7 +202,7 @@ class RelayModernStore implements Store {
       operationAvailability,
       operationLastWrittenAt,
       rootEntry?.fetchTime,
-      this._queryCacheExpirationTime,
+      operation.request?.cacheConfig?.ttl ?? this._queryCacheExpirationTime,
     );
   }
 
@@ -225,11 +225,12 @@ class RelayModernStore implements Store {
       rootEntry.refCount--;
 
       if (rootEntry.refCount === 0) {
-        const {_queryCacheExpirationTime} = this;
+        const ttl =
+          operation.request?.cacheConfig?.ttl ?? this._queryCacheExpirationTime;
         const rootEntryIsStale =
           rootEntry.fetchTime != null &&
-          _queryCacheExpirationTime != null &&
-          rootEntry.fetchTime <= Date.now() - _queryCacheExpirationTime;
+          ttl != null &&
+          rootEntry.fetchTime <= Date.now() - ttl;
 
         if (rootEntryIsStale) {
           this._roots.delete(id);
--- a/packages/relay-runtime/store/__tests__/RelayModernEnvironment-QueryCacheExpirationTime-test.js
+++ b/packages/relay-runtime/store/__tests__/RelayModernEnvironment-QueryCacheExpirationTime-test.js
@@ -28,6 +28,7 @@ const {generateAndCompile} = require('relay-test-utils-internal');
 describe('query cache expiration time', () => {
   let environment;
   let operationDescriptor;
+  let operationCustomTTLDescriptor;
   let ParentQuery;
   let source;
   let store;
@@ -59,6 +60,11 @@ describe('query cache expiration time', () => {
       store,
     });
     operationDescriptor = createOperationDescriptor(ParentQuery, {size: 32});
+    operationCustomTTLDescriptor = createOperationDescriptor(
+      ParentQuery,
+      {size: 32},
+      {ttl: QUERY_CACHE_EXPIRATION_TIME + 10},
+    );
   });
 
   afterEach(() => {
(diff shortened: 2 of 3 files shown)


maraisr (approved):

josephsavona:
While we now _can_ support per-query TTL, we need to consider whether we _should_. Will users expect that queries aren't always "cached" until they expire? It may be non-intuitive that queries can be purged from the cache before the TTL is reached, as can happen if the release buffer overflows. Also, what happens if the same query is fetched with two different TTLs? Finally, there's the question of whether users even have enough context to choose a good TTL for most queries. If you do have a query that's so important it needs its own custom TTL, then you probably also want other things too, like ensuring that that query doesn't get bumped from the release buffer (and therefore you might just retain it and manage its TTL manually). My sense is that the release buffer and store-level TTL allows a good degree of control and can help ensure consistency across the app. Thoughts?

morrys:
@josephsavona, here is my thought:

> While we now can support per-query TTL, we need to consider whether we should. 

It's okay to analyze whether _we should_ but also why _we shouldn't_.

> Will users expect that queries aren't always "cached" until they expire? It may be non-intuitive that queries can be purged from the cache before the TTL is reached, as can happen if the release buffer overflows

I do not agree that it is not intuitive because it is the user who configures the TTL, be it the store or a custom one.
As you say, this happens with the release buffer and is the main reason I'm not a fan of this feature.

> Also, what happens if the same query is fetched with two different TTLs?

I would manage it considering the latest TTL and put some warnings in the development phase.

> Finally, there's the question of whether users even have enough context to choose a good TTL for most queries

This is why you do it. Giving the opportunity those who have a good context to do it

> If you do have a query that's so important it needs its own custom TTL, then you probably also want other things too, like ensuring that that query doesn't get bumped from the release buffer (and therefore you might just retain it and manage its TTL manually)

I think buffer and TTL (even without custom ttl) should be used exclusively as they could cause behavior beyond your control
An example of use are for all queries that find data that rarely changes (for example data to populate select)
Or use different TTLs for queries that perform searches and queries that show the detail of a search

> My sense is that the release buffer and store-level TTL allows a good degree of control and can help ensure consistency across the app

Right, but why don't give more?

josephsavona:
@morrys I think we agree that it should be possible for advanced users to use Relay in a way that allows specifying a per-query TTL. That's why we were okay with your previous PR that passes through the CacheConfig: because it allows alternate `Store` implementations to provide per-query TTL. For all the reasons I specified above I'm not sure if supporting per-query TTL for the _default_ Store implementation is the right approach, but at this point it should now be possible to support it in an alternative Store implementation. Can you confirm that's the case? 

I would imagine us having a section in the docs about TTL and the release buffer, with a note along the lines that to use per-query TTLs developers should see <list> of 3rd-party Store implementations that support it.

sibelius:
docs about TTL and release buffer would be great

engmutabazi100 (approved):
This is working well

morrys:
Hi @josephsavona,
I would like to propose an alternative solution that also answers your question:

> Also, what happens if the same query is fetched with two different TTLs?

I have the changes ready to use a new directive instead of cacheConfig:
```
query ParentQuery @cache(ttl: 1010) {
  me {
    id
    name
  }
}
```
Let me know if you want me to push to this branch

maraisr:
I think the addition of a directive such as that would require more thought. I'm just cautious of not seeing that as a directive for server side caching perhaps, or a directive that should be passed on.

Personally though, I think cache on a per-query basis offers some advantages, but think it should be something like useMemo in that its not a guarantee. I feel a mission critical "this must be cached" mindset is riddled with problems in reality. 

So a TTL on the query im okay with, but knowing full well that it may very well evict sooner.

josephsavona:
Note that making this a directive doesn't solve the problem of fetching the same query with different variables: you could use a variable as the value, and invoke the query w different values for that variable. 

At this point it sounds like we're agreed that it _is_ possible to build a custom implementation of the `Store` API that supports per-query TTL. Given that, our preference would be to encourage @morrys and other folks in the community to try that out and report back. What is your experience? Then we can consider adding support in the default Store. It's generally easy to add new APIs and hard to remove them, so we are very reluctant to add a new API (especially one that affects internal behavior) without a clear use-case or, in the absence of that, experience to guide us.

josephsavona (changes requested):
Requesting changes because we don't plan to immediately move forward w this diff, but we are definitely interested in seeing folks in the community experiment w this and report back. Thank you, @morrys, for proposing this and kicking off the discussion!

morrys:
H @josephsavona, I pushed the changes to the directive both to show better how it solves the problem and because I find it cool 👍 

Integrating it into external stores is now easier and wora / relay-store will support it.

> so we are very reluctant to add a new API (especially one that affects internal behavior) without a clear use-case or, in the absence of that, experience to guide us

It is not clear to me what can help you

I think that wherever there is a TTL (see for example a CDN) it is normal to foresee the possibility of specifying one for a content and not just a global one

PR #3265, opened 2020-11-18
feat: Support AsyncGenerators as a result from fetchFn

maraisr:
This PR aims to allow an async generator be returned from the fetchFn to aid in things like incremental delivery.

Intended usage:

```js
async function* fetchFn(params, variables) {
    const result = await fetch(...).then(myMultipartFnButProbablyMeros);
    for await (const part of result) {
        yield part;
    }
}
```

and being able to feed that directly into Relay's `Observable.from` aka `Network.create`

### DIFF
--- a/packages/relay-runtime/network/RelayObservable.js
+++ b/packages/relay-runtime/network/RelayObservable.js
@@ -497,9 +497,9 @@ function fromValue<T>(value: T): RelayObservable<T> {
 function fromAsyncIterator<T>(obj: AsyncIterator<T>): RelayObservable<T> {
   return RelayObservable.create(sink => {
     (async () => {
-      while(true) {
+      while (true) {
         const {value, done} = await obj.next();
-        if (done) break;
+        if (done) {break;}
         sink.next(value);
       }
       sink.complete();
--- a/packages/relay-runtime/network/RelayObservable.js
+++ b/packages/relay-runtime/network/RelayObservable.js
@@ -468,7 +468,10 @@ declare function isAsyncIterator(p: mixed): boolean %checks(p instanceof
   AsyncIterator);
 
 function isAsyncIterator(obj) {
-  return typeof obj === 'object' && typeof obj[Symbol.asyncIterator] < 'u';
+  return typeof val < 'u' &&
+    val !== null &&
+      (val[Symbol.toStringTag] === 'AsyncGenerator' ||
+        (Symbol.asyncIterator && Symbol.asyncIterator in val));
 }
 
 function fromObservable<T>(obj: Subscribable<T>): RelayObservable<T> {
--- a/packages/relay-runtime/network/RelayObservable.js
+++ b/packages/relay-runtime/network/RelayObservable.js
@@ -74,7 +74,11 @@ export interface Subscribable<+T> {
 
 // Note: This should accept Subscribable<T> instead of RelayObservable<T>,
 // however Flow cannot yet distinguish it from T.
-export type ObservableFromValue<+T> = RelayObservable<T> | Promise<T> | T;
+export type ObservableFromValue<+T> =
+  | RelayObservable<T>
+  | Promise<T>
+  | AsyncGenerator<T>
+  | T;
 
 let hostReportError = swallowError;
 
@@ -143,11 +147,16 @@ class RelayObservable<+T> implements Subscribable<T> {
    * useful for accepting the result of a user-provided FetchFunction.
    */
   static from<V>(obj: ObservableFromValue<V>): RelayObservable<V> {
-    return isObservable(obj)
-      ? fromObservable(obj)
-      : isPromise(obj)
-      ? fromPromise(obj)
-      : fromValue(obj);
+    switch (true) {
+      case isObservable(obj):
+        return fromObservable(obj);
+      case isPromise(obj):
+        return fromPromise(obj);
+      case isAsyncIterator(obj):
+        return fromAsyncIterator(obj);
+      default:
+        return fromValue(obj);
+    }
   }
 
   /**
@@ -454,6 +463,14 @@ function isObservable(obj) {
   );
 }
 
+// Use declarations to teach Flow how to check isAsyncIterable.
+declare function isAsyncIterator(p: mixed): boolean %checks(p instanceof
+  AsyncIterator);
+
+function isAsyncIterator(obj) {
+  return typeof obj[Symbol.asyncIterator] < 'u';
+}
+
 function fromObservable<T>(obj: Subscribable<T>): RelayObservable<T> {
   return obj instanceof RelayObservable
     ? obj
@@ -474,6 +491,17 @@ function fromValue<T>(value: T): RelayObservable<T> {
   return RelayObservable.create(sink => {
     sink.next(value);
     sink.complete();
+  });
+}
+
+function fromAsyncIterator<T>(obj: AsyncIterator<T>): RelayObservable<T> {
+  return RelayObservable.create(sink => {
+    (async () => {
+      for await (const patch of obj) {
+        sink.next(patch);
+      }
+      sink.complete();
+    })().catch(sink.error);
   });
 }
 
--- a/packages/relay-runtime/network/RelayObservable.js
+++ b/packages/relay-runtime/network/RelayObservable.js
@@ -468,7 +468,7 @@ declare function isAsyncIterator(p: mixed): boolean %checks(p instanceof
   AsyncIterator);
 
 function isAsyncIterator(obj) {
-  return typeof obj[Symbol.asyncIterator] < 'u';
+  return typeof obj === 'object' && typeof obj[Symbol.asyncIterator] < 'u';
 }
 
 function fromObservable<T>(obj: Subscribable<T>): RelayObservable<T> {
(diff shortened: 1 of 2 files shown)


packages/relay-runtime/network/RelayObservable.js
```diff
@@ -477,6 +494,19 @@ function fromValue<T>(value: T): RelayObservable<T> {
...
 }
 
+function fromAsyncIterator<T>(obj: AsyncIterator<T>): RelayObservable<T> {
+  return RelayObservable.create(sink => {
+    (async () => {
+      while(true) {
+        const {value, done} = await obj.next();
+        if (done) break;
+        sink.next(value);
+      }
```
maraisr, line 504: this can alternately be:

```suggestion
    (async () => {
      for await (const patch of obj) {
        sink.next(patch);
      }
      sink.complete();
    })().catch(sink.error);
```

if babel wasnt getting unhappy.

sibelius (approved):

maraisr:
would appreciate some insihgts with this one @jstejada @kassens @josephsavona

packages/relay-runtime/network/RelayObservable.js
```diff
@@ -477,6 +494,19 @@ function fromValue<T>(value: T): RelayObservable<T> {
...
+  return RelayObservable.create(sink => {
+    (async () => {
+      while (true) {
+        const {value, done} = await obj.next();
+        if (done) {break;}
+        sink.next(value);
+      }
+      sink.complete();
+    })().catch(sink.error);
+  });
```
n1ru4l, line 507:
```suggestion
    return () => obj.return?.()
  });
```

The iterator should be returned when the RelayObservable is getting disposed of.

n1ru4l (commented):
AsyncIterableIterators/generators are easily composable and would allow writing "lean" fetcher functions for applying middlewares!

maraisr:
@kassens @josephsavona what do you think of landing this one?

packages/relay-runtime/network/RelayObservable.js
```diff
@@ -454,6 +463,17 @@ function isObservable(obj) {
...
+// Use declarations to teach Flow how to check isAsyncIterable.
+declare function isAsyncIterator(p: mixed): boolean %checks(p instanceof
+  AsyncIterator);
+
+function isAsyncIterator(obj) {
+  return typeof val < 'u' &&
+    val !== null &&
+      (val[Symbol.toStringTag] === 'AsyncGenerator' ||
+        (Symbol.asyncIterator && Symbol.asyncIterator in val));
+}
```
josephsavona, line 475: shouldn't this be checking `obj`, not `val`?

packages/relay-runtime/network/RelayObservable.js
```diff
@@ -454,6 +463,17 @@ function isObservable(obj) {
   );
 }
 
+// Use declarations to teach Flow how to check isAsyncIterable.
+declare function isAsyncIterator(p: mixed): boolean %checks(p instanceof
+  AsyncIterator);
+
+function isAsyncIterator(obj) {
+  return typeof val < 'u' &&
```
josephsavona, line 471: `typeof obj !== undefined`?

packages/relay-runtime/network/__tests__/RelayObservable-test.js
```diff
@@ -1444,6 +1444,28 @@ describe('RelayObservable', () => {
...
+        },
+        error: err => {
+          done(err);
+        },
+        complete: () => {
+          done();
+        },
+      });
+    });
+
```
josephsavona, line 1468: let's add a test for error handling and for cleanup

packages/relay-runtime/network/__tests__/RelayObservable-test.js
```diff
@@ -1444,6 +1444,28 @@ describe('RelayObservable', () => {
...
+        next: val => {
+          expect(val).toEqual(value);
+        },
+        error: err => {
+          done(err);
+        },
+        complete: () => {
+          done();
+        },
+      });
```
josephsavona, line 1466: This approach to testing isn't very reliable, because it doesn't guarantee that next() is actually called. Our approach is to store all the events into an array and then assert the final result of the array, which checks that all the right events occurred and in the right order. Something like:

```javascript
it('Converts an AsyncGenerator to an Observable', () => {
  const value = ...;
  // ... same
  const obs = RelayObservable.from(result);
  
  // changed part
  const events = [];
  obs.subscribe({
    next: val => events.push(...['next', val]),
    error: err => events.push(...['error', err]),
    complete: () => events.push('complete'),
});
jest.runAllTimers();
expect(events).toEqual(['next', value, 'complete']);
```

josephsavona:
@maraisr Seems reasonable to me, see comments

packages/relay-runtime/network/RelayObservable.js
maraisr: 🤦🏻 ya... i had it as obj, but copied the body from elsewhere when I realized the Symbol isnt a thing on Safari. Have fixed up.

packages/relay-runtime/network/__tests__/RelayObservable-test.js
maraisr: WOW, thank you so much for the guidance! Tricky stuff to test this on my pc. Nearly all snapshots fail:
image

---

but ya, love the approach.

maraisr: Maybe you can help me with this, because they are promises the `runAllTimers` does nothing in this case—so `events` is always an empty array. As the subscribe runs after the assertion.

I have amended the test to now also expect an assertion.

packages/relay-runtime/network/__tests__/RelayObservable-test.js
maraisr: I think cleanup is very tricky to unit test, i have "tested' this manually, but is very hard to actually test because you cannot step through the yields. 

One theory i had was to:

```js
async function* generate() {
    yield 'a';
    yield 'b';
}
```

but there is no way to expect a value, then unsubscribe. 

Calling .unsubscribe doesn't run `complete` (which is probably a bug). But that aside, if you know of a better way to test this, please let me know! Im stuck for ideas.

packages/relay-runtime/network/__tests__/RelayObservable-test.js
```diff
@@ -1444,6 +1444,27 @@ describe('RelayObservable', () => {
...
+
+      async function* generate() {
+        yield value;
+      }
+
+      // Something else will "run" this before, like the fetchFn
+      const result = generate();
+      const obs = RelayObservable.from(result);
+
+      expect.assertions(1);
```
josephsavona, line 1458: hmm really strange that runAllTimers() isn't working here, as we use that throughout this file for all Promise-related tests without issue. the `obs.subscribe()` call _must_ happen synchronously, it's strange that the earlier alternative i proposed wouldn't work.

maraisr: I think with the promise based stuff, you have the promise handy in scope in order to await, which guarantees the .next's have fired. You can stack awaits.

With generators though, once it's consumed the "iteration", it's consumed. The only way I can think is, to have an async-push-pull mechanism here. Have it such that the subscribe takes place, then emit the next event, kick of timers. Expect, and so on. 

I think what makes this special is that, in all other scenarios the promise is created in scope, whereas generators, although created in scope, the wrapping promise is created inside the sin fn.

maraisr:
@n1ru4l keen on helping me finish these tests off? I know you've been doing loads in this space, and I'm lost for ideas.

maraisr:
@josephsavona would be keen to hear how we can bring this across the line? Feel free to DM me on the GraphQL slack. Would be such a good feature.

josephsavona:
We're not opposed to this but we'd really like to have tests covering all the important cases. If we can figure out testing for cancellation etc (see comments above) then we can import.

maraisr:
Thank you for this one @josephsavona I have updated unit tests which confirm the feature, there is a little of be deduction there that confirms things. 

Fundamentally, yes the tests are a tad flaky, but without exposing the internal promise it creates there is no way to get it 100%.

maraisr:
Bump cc @alunyov @kassens

maraisr:
after some more reading on how Entry Points work, its become apparent that promise ticks harm UI's — so take this pr won't ever land, because of this tick?

packages/relay-runtime/network/RelayObservable.js
```diff
@@ -477,6 +500,23 @@ function fromValue<T>(value: T): RelayObservable<T> {
...
+        const {value, done} = await obj.next();
+        if (done) {break;}
+        sink.next(value);
+      }
+      sink.complete();
+    })().catch(error => {
+      sink.error(error, true /* isUncaughtThrownError */);
+    });
+
+    return () => obj.return?.();
```
josephsavona, line 516: will this cause the loop above to break? or do we have to do that manually? i'm not familiar w details of generators

maraisr: Yeah that's right, if you return the generator the next while iteration will have a done true.

maraisr:
bump @captbaritone @josephsavona

maraisr:
maybe in another time

PR #3266, opened 2020-11-18
feat(moved utility functions from relay-experimental to relay-runtime)

morrys:
this PR moves Direction type and utility functions _getRefetchMetadata_, _getPaginationMetadata_,
_getValueAtPath_ _getPaginationVariables_ from relay-experimental to relay-runtime as none of them have a dependency with react.

Also, with this PR it will be easier to integrate relay-runtime into other libraries such as [relay-angular](https://github.com/morrys/relay-angular)


sibelius (approved):

kassens:
Thanks, this makes sense to me on first look. Pulling for internal review as well.

kassens:
Actually a bit torn, for example the reference to `UNSTABLE_extraVariables` and the (kinda weird, complex) handling of variables it's maybe not something we want to necessarily make part of the public API directly?

morrys:
@kassens, especially because it is complex I see it well here. 
It is true that variable management seems more appropriate in relay-experimental and not relay-runtime but relay-runtime is what defines how relay should work (for react, for angular etc ...)

I would also include this piece of code inside the function:
https://github.com/facebook/relay/blob/master/packages/relay-experimental/useLoadMoreFunction.js#L190-L196

morrys:
@kassens, can I take advantage of asking the purpose of these functions?
https://github.com/facebook/relay/blob/master/packages/relay-runtime/query/GraphQLTag.js#L107-L133

I have not seen any use of them, can they be deleted?

morrys:
Hi @kassens,
I hope I don't bother you too much. 
Now I have seen that together with these functions this too would have to be brought into relay-runtime [getConnectionState](https://github.com/zth/relay/blob/master/packages/relay-experimental/useLoadMoreFunction.js#L276-L352)

If you're okay with it, give me some feedback so I can make the necessary changes

Thanks

morrys:
Hi @kassens,
happy New Year!

any news on this PR? 👍

morrys:
@kassens friendly ping 👍

morrys:
hi @kassens,
I think we can close this PR as it is imported with commit https://github.com/facebook/relay/commit/3b6f70d85fbcfc920713d83ae94cb3d6a84191d3

Issue #3030, opened 2020-02-22
Relay Compiler --config option not being read


tazsingh:
Hi there,

I've upgraded to `relay-compiler` 9.0.0 and it appears to no longer respect the `--config` option. I've tracked it down to PR #2999 which seems to change the behaviour of this.

Based on that PR, I've additionally installed the `relay-config` package, but the problem persists.

--------------------------

**To reproduce:**

Create a `config/relay.config.js` file:

```js
const { resolve } = require('path')

module.exports = {
  src: resolve(__dirname, '../src'),
  schema: resolve(__dirname, './schema.graphql')
}
```

_Note: I don't actually think you need real values in the file, the entire file won't get picked up. Provided the above for demonstration purposes._

Then run the `relay-compiler`:

```sh
relay-compiler --config config/relay.config.js
```

You should see that it doesn't pick up the config file and says that it's `Missing required arguments: schema, src` after printing out the full arg list (i.e. same as if you'd run `relay-compiler --help`).

Note that this worked without issue on `relay-compiler` 8.0.0.

---------------------------

**Potential Workaround**

Based on my `console.log` debugging, the problem exists because [`cosmicconfig` that's being used by `RelayConfig` internally](https://github.com/facebook/relay/blob/master/packages/relay-config/index.js#L20) is [only reading from `process.cwd()`](https://github.com/davidtheclark/cosmiconfig#explorersearch). Therefore, I can move the `relay.config.js` to the `process.cwd()` (i.e. root of the project in my case), and it does work. However, I'm filing this issue as I believe it to have introduced unintentionally different behaviour. If this is now desirable, then I'm happy to help document it!

------------------------------

Thank you all for the awesome project! Let me know how I can be of assistance with this 😄

mrtnzlml:
@tazsingh Hi, this was a bug in Relay Compiler: there was never supposed to be a `--config` options. Relay Config is being auto-discovered via [cosmiconfig](https://github.com/davidtheclark/cosmiconfig), so you can safely remove the `--config` option and it should work.

More info here: https://mrtnzlml.com/docs/relay#relay-config-relayconfigjs

> However, I'm filing this issue as I believe it to have introduced unintentionally different behaviour.

It was an intentional change since the `--config` option was actually from Yargs and not from Relay Compiler and therefore the previous state was unintentional. I believe that a nice solution would be to add the `--config` option into Relay Compiler explicitelly (new feature).

tazsingh:
[...]

PR #3267, opened 2020-11-19
Add --config option to specify relay config file for relay-compiler

rightjelkin:
You can use few different graphql schemas for your projects, and there is no possibility to manage them via config files. This PR adds this possibility with using yargs config option. Just specify path to relay config file with --config cli option

Some discussions in #3030


eMerzh:
i think this has been fixed in v13

PR #3278, opened 2020-11-30
@append/prependNode - explicitly set cursor to null on created edge

zth:
This PR explicitly sets the `cursor` field to `null` on the edge that's synthetically created by `@appendNode`/`@prependNode`. This is because Relay otherwise yields a warning about missing data (`cursor` being `undefined`). 

My understanding is that this is fine because Relay doesn't look at `cursor` for pagination (just `pageInfo`, right), and it's not realistic that the user can "guess" a correct cursor for an added node anyway. So, better to just null it to get rid of the warning.

Thoughts?


zth:
Closing this as a fix that'll cover this case has been implemented in master.

jstejada:
Fixed by a7b8ab61cefd29ba42f2565aac9900026ecbeb8d

PR #3280, opened 2020-12-01
[rust] Better support for abstract types (unions and interfaces)

MaartenStaa:
This is the Rust equivalent of #3178. What I have added compared to that is I made the future proofness of enums and abstract types optional and configurable.

Since the feedback there was positive but the main objection was that the old Relay compiler is basically in support mode until the Rust version is released, I ported this.

> Copy paste from the other PR: Basically I have included some changes to support more accurate type output, especially regarding union types and type names.
>
> This should also fix the case where duplicated __typename property definitions are generated, depending on where in the query or fragment the __typename property is requested.

Please let me know what you think! I'm personally not that fond of the current setup I have with testing different combinations of future proof enums/abstract types and how this is captured in the fixtures, maybe someone else has a better idea?

/cc @kassens @josephsavona @tyao1


compiler/crates/relay-typegen/tests/generate_flow/fixtures/typename-on-union.expected
```diff
@@ -80,22 +80,346 @@ fragment TypenameAliases on Actor {
...
-  +$refType: TypenameAlias$ref,
 |} | {|
   // This will never be '%other', but we need some
   // value in case none of the concrete values match.
-  +_typeAlias: "%other",
-  +$refType: TypenameAlias$ref,
+  +_typeAlias: "%other"
+|} & {|
+  +$refType: TypenameAlias$ref
 |};
```
kassens, line 423: Unfortunately, this doesn't work with flow, while it appears to work with Typescript.

What `{| a: number |} & {| b: number |}` expresses in Flow is: an object with exactly key `a`, nothing more and at the same time and object with exactly the key b and nothing more. That object can obviously not exist, so access to properties fails. I'm not sure what TS makes of this, not a type system expert.

- [flow example](https://flow.org/try/#0C4TwDgpgBAKuEDsCGBbCB1AlsAFgYQHsUUCEBlCAGwgGNhNSBnKAXigG8AfAKCigGoA+oNCRkaAFxQARAFVGEAE7SANLwEAzTIsbAAcqghSA-LsWYEAczWcAvlE4cefISPjijMgApJLEVer8AK4Kih4mZhbW3HZQAGROgeFQpsDmVja2ANzc3BpBCHQMCFAaABQAHlJwYoZYuITEpBTURUwAlBzqmBpQlQB0wqKIhqxsAOTySuOd7Op8Ff1aOvqG6rbctkA)
- [typescript example](https://www.typescriptlang.org/play?#code/C4TwDgpgBAKuEDsCGBbCB1AlsAFgYQHsUUCEBlCAGwgGNhNSBnKAXigAoBvAKCigCcISACalKIKAH1JoSMjQAuKACIAqowj9lAGl4ChohOKgAzTP0bAAcqghLL-TAgDmUAD5QEAV0qVdAX3coHj5BETEJaVlEWyVlAAUkZwgdPTDDYy8Nfnk7KAcnVw9vXwCASigAMmC0gwjPWPzgRxcgkr9ufwBubm4TLwQ6BgRTdgAPJTg5WyxcQmJSCmohpgqQqEwTDjGAOij4XNY2AHJ1TWO1vT5dswtrWz1-TqA)

compiler/crates/relay-typegen/src/flow.rs
```diff
@@ -14,6 +14,20 @@ pub struct FlowPrinter {
 }
 
 impl Writer for FlowPrinter {
+    fn write_ast(&mut self, ast: &AST) -> String {
```
kassens, line 17: We should keep the previous way of not generating intermediate `String`s. What this implementation is doing is:
1. Allocate a new `String`.
2. Write into that string.
3. Write the result onto the larger result.

The updated version just directly wrote to the target, skipping the intermediate string.

I think this probably got lost in a rebase you did, as I updated this recently.


The other lint fixes in this file (`&[StringKey]`, instead of `&Vec<StringKey>` are still good)

compiler/crates/relay-typegen/src/lib.rs
```diff
@@ -142,7 +149,8 @@ impl<'schema, 'config> TypeGenerator<'schema, 'config> {
         let input_variables_type = self.generate_input_variables_type(typegen_operation);
 
         let selections = self.visit_selections(&typegen_operation.selections);
-        let mut response_type = self.selections_to_babel(selections, false, None);
+        let mut response_type =
+            self.selections_to_babel(Some(typegen_operation.type_), selections, false, None);
```
kassens, line 153: Haha, just noticed we still have the `babel` name in here from my original 1:1 translation from the JS code. This is some of the most confusing part of the compiler to me, so it was just a direct line-by-line translation where possible…

compiler/crates/relay-typegen/src/lib.rs
```diff
@@ -643,55 +669,137 @@ impl<'schema, 'config> TypeGenerator<'schema, 'config> {
...
-                        read_only: true,
-                        optional: false,
-                        value: AST::OtherTypename,
-                    })
-                    .collect(),
-            );
-        } else {
-            let mut selection_map = selections_to_map(hashmap_into_value_vec(base_fields), false);
+            // It might be some other type then the listed concrete types. We try to
+            // figure out which types remain here.
```
kassens, line 737: Is there benefit to this?

This kindof blew up with many pages of types for some huge interfaces. We would need to make this an option, or maybe only do if "no future proof types" is set. There's probably no point in combining the listing out of typenames when also including ther `%other` sentinel value?

```
`'%other' | 'SomeOtherType' | 'YetAnotherType'
```

kassens (changes requested):
Left some comments on parts. There's a bit work left here to make it possible to merge. We have a huge existing codebase, so this affects a lot of generated files internally that break them. We might have to make the advanced code generation behind config options.

I'm unsure about the combinational explosion in the test outputs. What do you think of generating the vast majority of the tests only in 1 configuration option and then having a few tests that test the differences the options produce? A way I can think of to do that would be either to create additional fixture folders or a thing we did in other places is adding a "header" with options to the file, i.e.

```
{
  futureProofTypename: false
}
%%%
query Test {
  # ...
}
```
then in the fixture transform function, if the file contains %%% split on that and use
```
let typegen_options = serde_json::from_string(header);
``
to parse the header as json.

(if you have questions for any of this, feel free to reach out!)

File: compiler/crates/relay-typegen/src/lib.rs (hunk 4 above)
Comment (MaartenStaa) [reply compiler/crates/relay-typegen/src/lib.rs]: You're right, probably not. I changed to only produce either %other (future proof) or the possible typenames (no future proofness).

File: compiler/crates/relay-typegen/src/lib.rs (hunk 3 above)
Comment (MaartenStaa) [reply compiler/crates/relay-typegen/src/lib.rs]: Yeah, same. As mentioned, I made these same changes earlier in Flow (and TypeScript before that), but it's not a function that's easy to read. In any case, I renamed it to `selections_to_ast` to get rid of the Babel name.

File: compiler/crates/relay-typegen/src/flow.rs (hunk 2 above)
Comment (MaartenStaa) [reply compiler/crates/relay-typegen/src/flow.rs]: Whoops yeah, that got lost in either a merge or rebase. Fixed.

File: compiler/crates/relay-typegen/tests/generate_flow/fixtures/typename-on-union.expected (hunk 1 above)
Comment (MaartenStaa) [reply compiler/crates/relay-typegen/tests/generate_flow/fixtures/typename-on-union.expected]: TypeScript doesn't have this concept of "exact" objects, so it's just a regular intersection. I made some changes so typegen writers that support exact objects get fed a regular union of exact objects, with the shared properties being duplicated in each union member. For TypeScript (which does not support the exact objects) or any other future languages that do not support exact objects, it still issues an intersection AST.

**Comment by MaartenStaa** [comment]
Thanks for the feedback @kassens! Was a bit busy lately so didn't get around earlier to make changes based on your comments. Please take a look when you have time, and happy holidays :)

**Comment by MaartenStaa** [comment]
@kassens Did you have a chance to look at this PR again?

**Comment by MaartenStaa** [comment]
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contributions.

Pinging @kassens again, or maybe @alunyov? It would be great if this PR can be included in #3180. Right now it's in limbo and I don't really know where it stands. Does it need more work? Is it out of scope, and if so, why?

Any word at all since it's been more than a year since I opened this PR.

**Review by alunyov — CHANGES_REQUESTED** [review]
Thank you for working on this, and sorry we dropped the ball here. But let's revive the discussion.

I know there was a high-level discussion with @josephsavona and @kassens.
What is the latest update on this? Did we figure out the rollout plan, where all their questions addressed from the initial discussion in #3178?

Also, I don't think this was discussed before, but listing all possible implementations in the `node` (for example) may produce a very long type. Consider an example where you may have more than 10K objects in the schema that implement a `node` interface. (and this is a totally real example). Will all of these be included in the type?

Requesting changes just to get more context on the PR.

File: compiler/crates/relay-typegen/tests/generate_flow/fixtures/typename-on-union.expected [hunk 5]
```diff
@@ -185,14 +166,25 @@ import type { FragmentReference } from "relay-runtime";
...
 export type TypenameOutsideWithAbstractType = {|
-  +__typename: string,
+  +__typename: "User",
+  +firstName: ?string,
+  +address: ?{|
+    +street: ?string,
+    +city: ?string,
+  |},
+  +username?: ?string,
+  +address?: ?{|
```
alunyov, line 176: hmm, why `address` is included twice here?

MaartenStaa:
Thanks for getting back to me @alunyov! I've moved to a different apartment the other weekend, so didn't have much time in the past weeks to dig deeper into where the code stands. Looking at the old discussion over at #3178, I think the feedback there can be summarized as follows:

- `__typename: "%other"` is intentional, in case the client is outdated in relation to the server. This has been addressed by making the future proofness opt-in/opt-out.
- Rollout is a breaking change. I believe this depends on the default value of the two future proofness config settings (for enums and for abstract types). The current default is for future proofness to be enabled (same as the old compiler), so I think this resolves the breaking change. Having said that, I have not double-checked this (comparing output between the old and the new compiler).
- @jacob-indieocean mentioned an issue where a type name may not be resolved correctly. I have not yet looked into this, but will see if I can add a test here based on his example.

Have I missed anything?

Regarding your own points, I have dug into the issue with the duplicated `address` field in the `typename_on_union` test (the `username` field as was actually also being duplicated) and found the cause, so this has been resolved. I have also introduced a new option to exclude certain types from the listing of all implementations, which I believe takes care of your concern regarding the `Node` interface. I'm not very happy with the name of the option though (`exclude_from_typename_unions`), so I'm open to suggestions.

If a rollout plan has been discussed, I am not aware of it.

compiler/crates/relay-typegen/tests/generate_flow/fixtures/typename-on-union.expected
MaartenStaa: Fixed!

rbalicki2:
Hi @MaartenStaa — thanks again for the awesome work on this diff. And I'm sorry that I didn't review this sooner, and have different feedback than @kassens et al.

First, the diff changes the order of generated fields and unions. This is going to make this very hard to land internally. Can you figure out a way to keep the order the same, or else could you split this into two PRs, one of which alphabetizes the order of fields, unions, etc., and the other which is this PR? Alternatively, I could work on that internally, since I am planning on making that change myself, anyway.

Second, I think this shouldn't be set by global configuration. Let's look for the presence of a schema directive on the type/enum to determine whether to make the type/enum future safe.

A sketch of how this might work is:
* Add `directive Closed on INTERFACE | UNION | ENUM` in relay-extensions.graphql. (Feel free to think of a better name. Maybe NotExtendable?)
* Then (e.g. for testing), add that directive to `testschema.graphql`.
* Then, that directive should be accessible as `union_or_enum_or_interface.directives.named(string_key)` or the like

How does that sound?

Once we agree on the contours of the diff, I can review the implementation in more detail.

rbalicki2:
Okay, having thought about this diff some more, the typegen config is a good place to put a global default for the future-proofed-ness of these enums.

If there is a default, the type-level directives should be local overrides, in which maybe a good directive + parameter would be `@futureProof(isFutureProof: bool)`? Again, I'm not sure what a good name for this would be.

Anyway, I'll go ahead and review the implementation of this diff!

MaartenStaa:
Thanks @rbalicki2! Curious to hear what you think of the implementation. I'll try to look into the field ordering in the coming days. 

So for the `@futureProof` directive, I just want to make sure I understand the use case correctly. The global would be set to future proof, and then you could add the directive in your Relay extensions to disable future proofness for client only types, correct? Or can you also add directives to server types/unions/interfaces?
Continuing that thought, does the reverse (no future proofness as the global setting, and enabling it for a few client side types) even make sense? Or to take it a step further, is future proofness ever really needed for client only types? 

Looking forward to hearing your thoughts. I just want to make sure I understand which direction you'd like to go in before starting on an implementation.

rbalicki2:
I'm envisioning a global future-proofness setting, set e.g. in the typegen config. Then, developers would override that setting for individual types in their schema file.

Yeah, you're right — client only types should not be future proofed, ideally.

rbalicki2:
Hi @MaartenStaa , I started reviewing your diff. This is a lot! Thank you so much for taking the plunge. I'm sorry that you're dealing with relay-typegen/src/lib, and in particular, with the `to_babel` functions. They make no sense to begin with, which makes modifying them hard, and makes reviewing changes to them hard.

Anyway, so, because it's so hard to reason about this file, I think it'd be best if we paired so that I could understand exactly what's going on in your changes.

Can you DM me on Twitter? @statisticsftw or DM me on the GraphQL discord?

That being said, this needs to be broken up into smaller diffs. Here are some ideas for landing this in discrete chunks:

* Trivial renames, e.g. "to_babel" to "to_ast", renaming types to concrete types, etc. should go in a separate diff
* We should add the AST intersection variant in a separate diff. In this diff, lets modify the printing for AST::Union and AST::Intersection to be wrapped in parentheses if there is more than one, and remove the parentheses wrapping from write_nullable (and anywhere else).
* We need to alphabetize everything in a separate diff, and roll that out. It's impossible to reason about why things are ordered how they are right now, and (in OSS) you don't have access to all of the generated files, so you won't find all the weird edge cases :'(
* It **seems** (but I can't tell!) that the work to make enums non-future-proof is pretty separate from the work to make unions and intersections non-future-proof. Could those be split into separate diffs?

I think if we divide it up like that, it'll be a lot easier to review the meaty final diffs.

PR #3283, opened 2020-12-02
docs(installation): add relay-runtime

alexkrolick:
The environment section of the quick start guide imports from `relay-runtime`: https://relay.dev/docs/en/quick-start-guide#relay-environment
It's not a good idea to rely on transitive package dependencies in your app code, so I am adding `relay-runtime` to the top-level install section.

### DIFF
--- a/docs/Introduction-InstallationAndSetup.md
+++ b/docs/Introduction-InstallationAndSetup.md
@@ -8,7 +8,7 @@ title: Installation and Setup
 Install React and Relay using `yarn` or `npm`:
 
 ```sh
-yarn add react react-dom react-relay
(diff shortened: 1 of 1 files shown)


sibelius (approved):

alunyov:
Thank you @alexkrolick  

I agree that with this statement:

> It's not a good idea to rely on transitive package dependencies 

But the proposed change is not entirely correct. To accept it, we need to change `react-relay` dependencies and make `relay-runtime` it's peer dependency. And this is a breaking change. 

We discussed this internally and decided to work on this in the next major release (11.0.0).

PR #3284, opened 2020-12-04
Docs: Fix inter doc links in Relay Classic guides

JuanitoFatas:
Hi,

I found many of these guides link URL are missing `classic-` prefix. E.g.

https://relay.dev/docs/en/classic/guides-containers is 404 and the new location is https://relay.dev/docs/en/classic/classic-guides-containers.


alunyov:
Thank you, @JuanitoFatas, for this contribution!

I think the correct way would be to remove Classic documentation and all links to it.

JuanitoFatas:
> I think the correct way would be to remove Classic documentation and all links to it.

I am still using Relay Classic 😅, so hopefully the documentation could stick around 🙏

PR #3287, opened 2020-12-06
docs(pagination): add usage example for pagination containers

alexkrolick:


alexkrolick:
Any issues with this doc change? I found it a bit confusing without an example of using the pagination fragment and component.

PR #3290, opened 2020-12-08
Fix code example typo in homepage

RicCu:
Code sample in homepage imports fragment container `ArtistHeader` but returns nonexistent `Artist` component.

Alternatively the intention might have been to show `QueryRenderer` rendering an `Artist` component-that isn't shown in the examples-which presumably renders `ArtistHeader` at some point in its tree. Then the import statement should be changed instead.


kassens:
Thank you!

PR #3304, opened 2020-12-22
feat(removed null / false values from generated artifacts so as to reduce bundle size)

morrys:
This PR removed null / false values from generated artifacts so as to reduce bundle size

In the link below you can see the difference between an artifact generated before and after
https://editor.mergely.com/IJevFBgL/

### DIFF
--- a/packages/relay-compiler/index.js
+++ b/packages/relay-compiler/index.js
@@ -59,6 +59,7 @@ const getLiteralArgumentValues = require('./core/getLiteralArgumentValues');
 const getNormalizationOperationName = require('./core/getNormalizationOperationName');
 const getSchemaInstance = require('./runner/getSchemaInstance');
 const md5 = require('./util/md5');
+const removeNulls = require('./util/removeNulls');
 const writeRelayGeneratedFile = require('./codegen/writeRelayGeneratedFile');
 
 const {main} = require('./bin/RelayCompilerMain');
@@ -195,6 +196,7 @@ module.exports = {
     getSchemaInstance,
     md5,
     parseExecutableNode: extractAST.parseExecutableNode,
+    removeNulls,
     toASTRecord: extractAST.toASTRecord,
   },
 };
--- a/packages/relay-compiler/util/__tests__/dedupeJSONStringify-test.js
+++ b/packages/relay-compiler/util/__tests__/dedupeJSONStringify-test.js
@@ -131,8 +131,7 @@ test('extract recursive duplicates', () => {
     ],
     `(function(){
     var v0 = {
-      "name": "id",
-      "alias": null
+      "name": "id"
     },
     v1 = {
       "name": "id",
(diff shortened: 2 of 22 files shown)


josephsavona:
We intentionally don't do this to ensure that objects in the generated artifacts get consistent hidden classes.

josephsavona:
cc @kassens

kassens:
Keep in mind that chunks like `"alias":null` are repeated pretty frequently and therefore compress exceptionally well. That said, there's still overhead of course.

One option I was thinking to explore would be to introduce some more AST types. An obvious one would be to split up `LinkedField` into a `PluralLinkedField` and then remove the `plural: bool` flag.

Another one would be to introduce a `SimpleField` kind that has no arguments or no alias. Previous investigations into making that change and look at the effective bundle size reduction in production didn't really convince me though. IIRC it was a pretty small percentage after compression.

morrys:
Hi @josephsavona, could you explain yourself better? I haven't seen this connection between null / false values and hidden class in the code.

I made some changes to fix the tests and typing.
Now the PR is ready for Review.

As for the failed test case **ReactRelayPaginationContainer › fails if missing @connection directive**

This was fixed using a hack that is no longer valid now.
https://github.com/facebook/relay/blob/master/packages/react-relay/ReactRelayPaginationContainer.js#L275-L280
```js
// HACK: metadata is always set to `undefined` in classic. In modern, even
// if empty, it is set to null (never undefined). We use that knowlege to
// check if we're dealing with classic or modern
if (fragment.metadata !== undefined) {
  isRelayModern = true;
}
```

morrys:
@kassens, another change I want to try is to drop all static strings in favor of constants imported from relay-runtime.

kassens:
One thing I tried was replacing the objects (with known key order) with arrays where the first key is the kind and anything after the same keys as the objects. Even that didn't seem to have material output on the compressed output size. I'd recommend trying to see what happens when you minify and compress the generated files of some larger example app together and instead of making the runtime changes, see how the size of this example app might change.

morrys:
As with this change, I do some testing on a sample app but of course it's hard for me to understand if this improvement is acceptable to you.

It is difficult to find improvements that significantly decrease the bundle size. From my point of view, every bundle reduction improvement is to be evaluated as it is the sum that makes the total 👍

morrys:
Hi @kassens,
I noticed that this PR has conflicts, could you tell me if it is your intention to integrate it? So that I proceed with the resolution or closure of the PR.

thanks

josephsavona:
@morrys We don't intend to land this change. As noted above, JS engines have optimizations that attempt to optimize object representation and/or field access by detecting common object shapes. When some code always creates object with the same set of keys in the same order, the objects will all be assigned the same hidden class, which then allows the JIT to specialize code that always sees instances of a particular class. By ensuring that we always emit all fields on the AST objects, we ensure they all receive the same hidden class and that runtime code (RelayReader etc) that walk over the ASTs is at least more likely to JIT-compile well. 

That said, we _are_ looking at larger changes to how we represent generated artifacts. But we're thinking of more holistic changes that balance code size with runtime processing cost.

For changes like this, we definitely recommend posting an issue or discussing on slack first to see if the team is aligned on the direction before proceeding with a change. We're happy to collaborate, but we can't always accept every PR if the direction doesn't align.

PR #3305, opened 2020-12-28
Change interface to detect_changes so that we don't need to pass in current TypeSystemDefinitions.

aakshintala:
Differential Revision: D25717916


aakshintala:
@tyao1  @kassens Any idea on how to fix the CI/ Check Rust and CI/ Test Rust failures? Both seem like something orthogonal to my changes...

tyao1:
Closing as the commit has already landed.

PR #3308, opened 2021-01-02
Rust: js config loader crate

kassens:
This crate implements a similar config resolution mechanism as [cosmiconfig](https://www.npmjs.com/package/cosmiconfig). This could be used for the Rust based compiler to find the Relay compiler config.

### DIFF
--- a/compiler/crates/js-config-loader/tests/test.rs
+++ b/compiler/crates/js-config-loader/tests/test.rs
@@ -1,0 +1,184 @@
+use js_config_loader::{search, ConfigError, ErrorCode};
+use serde::Deserialize;
+use std::fs::create_dir_all;
+use tempfile::tempdir;
+
+#[derive(Debug, Deserialize)]
+struct TestConfig {
+    name: String,
+}
+
+#[test]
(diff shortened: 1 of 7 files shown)


compiler/crates/js-config-loader/tests/test.rs
```diff
@@ -0,0 +1,184 @@
...
+#[test]
+fn rc_json_invalid_syntax() {
+    let dir = tempdir().unwrap();
+    let dir_d = dir.path().join("a/b/c/d");
+    let dir_f = dir.path().join("a/b/c/d/e/f");
+    create_dir_all(&dir_f).unwrap();
+
+    std::fs::write(dir_d.join(".foorc"), "invalid").unwrap();
+
+    match search::<TestConfig>("foo", &dir_f) {
```
kassens, line 34: This is the API, this crate is independent of the shape of the config, but returns a typed result. It just needs the value to implement `Deserialize`.

I really love `serde` <3

zth:
Ooh, nice @kassens ! That's a really clever solution.

PR #3332, opened 2021-01-23
Add getConnectionID util to ConnectionHandler

kyarik:
This PR adds a `getConnectionID` util to `ConnectionHandler` to easily get the connection ID knowing the ID of the record that has the connection field, the connection key, and optionally filters used to identify the connection:

```js
getConnectionID(recordID: string, key: string, filters?: ?Variables): string
```

This util is especially useful when working with the `@appendNode` and `@prependNode` directives, which take connection IDs as argument.

### Usage example with `@appendNode`

This example shows how we could add a comment to a story:

```js
const mutation = graphql`
  mutation AddCommentMutation($input: AddCommentInput!, $connections: [ID!]!) {
    addComment(input: $input) {
      comment @appendNode(connections: $connections, edgeTypeName: "CommentEdge") {
        id
      }
    }
  }
`;

const variables = {
  input: {
    storyID: '<story-id>',
    text: 'Nice story!'
  },
  connections: [
    ConnectionHandler.getConnectionID('<story-id>', 'Story_story_comments')
  ]
};

commitMutation(
  environment,
  {
    mutation,
    variables,
    ...
  }
);
```

### DIFF
--- a/packages/relay-runtime/handlers/connection/ConnectionHandler.js
+++ b/packages/relay-runtime/handlers/connection/ConnectionHandler.js
@@ -19,6 +19,7 @@ const invariant = require('invariant');
 const warning = require('warning');
 
 const {generateClientID} = require('../../store/ClientID');
+const {getStableStorageKey} = require('../../store/RelayStoreUtils');
 
 import type {
   HandleFieldPayload,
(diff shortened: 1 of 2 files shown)


josephsavona:
I'm not opposed to this, but note that you can query the id directly from a fragment with, for example:

```
fragment Story_story on Story {
  comments(...) @connection(key: "Story_story_comments") {
    __id
    edges { ... }
  }
}
```

Then you can pass `story.comments.__id` as the connection id to your mutation.

sibelius:
__id feels like a private field

kyarik:
Hey @josephsavona, thanks for pointing that out!

That will work nice if the mutation is done by the same component that is rendering the comments.

However, in cases where the component rendering comments and the one rendering the comment editor are independent, we would need to query the connection ID in the `CommentEditor` like this:

```graphql
fragment CommentEditor_story on Story {
  id
  comments(...) @connection(key: "Story_story_comments") {
    __id
    edges {
      cursor
    }
  }
}
```

which works, but it's not ideal as we are required by the Relay compiler to also query for `edges`, which are not used in the end, and so we get a warnig about unused fields in the `CommentEditor` file.

kyarik:
Now that I thought about it more, I realized that we should keep in one single component both the fragment with the comments connection and the mutation logic to create a comment. Otherwise, we break component encapsulation because if the fragment with the comments connection is inside `Story` and the mutation logic is inside `CommentEditor`, then `CommentEditor` would need to know the connection key used in the fragment of `Story` as well as what fields are queried on the comments in the connection (in order to query those same fields on the newly created comment).

Therefore, it's probably best to keep the mutation logic inside `Story` and just pass a callback prop to `CommentEditor`, taking the new comment text as argument.

@sibelius `__id` definitely feels like a private field, but we could think of it as a meta field, just like `__typename`.

gtkatakura:
I have a use case how would be great to have this helper. I have one component called StoresList which has the connection and another component how updates one Store, but the components aren't close enough to share the __id of the connection without context or something like that. Would be great to have this helper or some way to access the RecordSourceSelectorProxy, which is passed to updater. Something like:


```tsx
const store = useRecordSourceSelectorProxy()

const handleUpdate = () => {
  const root = store.getRoot()

  const connection = ConnectionHandler.getConnection(
    root,
    'StoreList_stores',
    {
      where: { category: 'SHOPPING' },
    }
  )

  update({
    variables: {
      input,
      connections: [connection.getDataID()]
    }
  })
}
```

PR #3334, opened 2021-01-24
Document the @appendNode and @prependNode directives

kyarik:
This PR adds the documentation for the [`@appendNode` and `@prependNode` directives](https://github.com/facebook/relay/pull/3155).

When I first learned about them, i wasn't sure how to get the ID of a connection in order to pass it as part of the `connections` argument that `@appendNode` and `@prependNode` accept. So, I assumed that it needs to be computed from the record ID and connection key, which led me to open the [PR that adds the getConnectionID util to ConnectionHandler](https://github.com/facebook/relay/pull/3332). However, @josephsavona pointed out that the connection data ID can be queried directly on the connection, which made it clear to me that we probably don't need a `getConnectionID` util. Therefore, we should definitely document how to use the `@appendNode` and `@prependNode` directives.

Let me know if the explanation is not clear enough, is missing some important points, or it should be placed elsewhere in the documentation. Thanks!


alunyov:
Thanks @kyarik!

cc @tyao1  can you look at this?

kyarik:
@tyao1 Can you please review this when you find some time?

josephsavona:
cc @jstejada and @rbalicki2 who are working on docs updates

jstejada:
hey @kyarik, thanks so much for the PR. we have some internal docs for this that we are releasing as part the new docs updates we are working on. maybe after we land those changes we can revisit this PR to incorporate any improvements based  on the docs you added here ?

kyarik:
Hey @jstejada, I checked locally `website-v2` and it looks great! :muscle:
I noticed that the new docs are using Relay hooks, so I assume that the new docs will go live after the experimental Relay hooks enter stable. If that's going to happen soon, then there's no problem postponing this PR. :+1: If not, I think that we could add the documentation for `@appendNode` and `@prependNode` in the current docs so that people can start learning about them without having to wait a long time. Either way, I'll incorporate any improvements from this PR into the new docs :slightly_smiling_face:

jstejada:
hey @kyarik yes the plan is for them to be released very soon

kyarik:
Superseded by #3424

PR #3345, opened 2021-02-04
avoid an unnecessary rerender in useFragmentNode 

morrys:
With this PR I changed the useHasChange by using useRef instead of useState. In this way, when the stored value changes, no unnecessary rerender is performed

### DIFF
--- a/packages/relay-experimental/__tests__/useFragmentNode-test.js
+++ b/packages/relay-experimental/__tests__/useFragmentNode-test.js
@@ -501,10 +501,7 @@ it('should re-read and resubscribe to fragment when environment changes', () => 
     profile_picture: null,
     ...createFragmentRef('1', singularQuery),
   };
-  assertFragmentResults([
-    {data: expectedUser, shouldUpdate: true},
-    {data: expectedUser, shouldUpdate: false},
-  ]);
+  assertFragmentResults([{data: expectedUser, shouldUpdate: true}]);
 
   TestRenderer.act(() => {
     newEnvironment.commitPayload(singularQuery, {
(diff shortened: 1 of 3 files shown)


josephsavona:
Requesting review from @jstejada - my guess is that we're using state here instead of a ref to avoid concurrency bugs.

jstejada:
thanks @morrys. @josephsavona is correct here, using useRef here wouldn't be concurrent safe. in any case we will be removing those hooks in an upcoming commit. (i happen to be making some updates to hooks right now)

morrys:
@jstejada Is there a test case showing this case?
Because this hook also causes refetch rerenders (e.g. https://github.com/facebook/relay/blob/master/packages/relay-experimental/__tests__/useRefetchableFragmentNode-test.js#L695)

jstejada:
@morrys  what do you mean by refetch re-renders? yes, there will be a second render, but generally in concurrent mode a render may occur multiple times before the changes commit. 

there is no test case for verifying that using useRef is unsafe, but generally refs shouldn't be mutated during render in concurrent mode, for the same reason that renders may happen multiple times before an update commits.

in any case, as I mentioned, we are going to remove the usage of useHasChanged since we no longer need the should update information. we will probably still use useMemoVariables most likely.

is there any particular issue this is causing? or are you concerned about perf here? 

note that the additional render from that setState here wont cause an additional commit, i.e. we're not going to end up actually updating the dom with any changes (which would be the expensive operation), but instead keeping the state value in sync. this is similar to how you would use getDerivedStateFromProps with classes: https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops. in general calling the render function should be cheap, and especially in concurrent mode where it can be called an arbitrary number of times.

morrys:
@jstejada yes, looking at the tests, I was worried about performance, for example it is strange to see all these renders https://github.com/facebook/relay/blob/master/packages/relay-experimental/__tests__/useRefetchableFragmentNode-test.js#L1138-L1142

jstejada:
@morrys got it, as i mentioned, calling the render function a couple of times shouldn't be problematic for perf. the problematic thing would be if these triggered several commits, which is where react has to do the hard work of updating the dom.

as mentioned, this is the recommended way of keeping state in sync with hooks:  https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops.

and as mentioned, in practice, with concurrent mode on, the render function might end up being called more times than even appears in the test.

jstejada:
in any case, useHasChanged was removed here: https://github.com/facebook/relay/commit/20419f0127406bb6bae31cee4a26e37be0afe306

PR #3348, opened 2021-02-05
refactor RelayOperationTracker & getPromiseForActiveRequest

morrys:
This refactor optimizes memory management in order to store only the identifier of the owner and not the whole object.

### DIFF
--- a/packages/relay-experimental/FragmentResource.js
+++ b/packages/relay-experimental/FragmentResource.js
@@ -82,7 +82,7 @@ function getPromiseForPendingOperationAffectingOwner(
 ): Promise<void> | null {
   return environment
     .getOperationTracker()
-    .getPromiseForPendingOperationsAffectingOwner(request.identifier);
+    .getPromiseForPendingOperationsAffectingOwner(request);
 }
 
 class FragmentResourceImpl {
@@ -434,7 +434,7 @@ class FragmentResourceImpl {
   ): Promise<void> | null {
     const environment = this._environment;
     const networkPromise =
-      getPromiseForActiveRequest(environment, fragmentOwner.identifier) ??
+      getPromiseForActiveRequest(environment, fragmentOwner) ??
       getPromiseForPendingOperationAffectingOwner(environment, fragmentOwner);
 
     if (!networkPromise) {
--- a/packages/relay-experimental/FragmentResource.js
+++ b/packages/relay-experimental/FragmentResource.js
@@ -82,7 +82,7 @@ function getPromiseForPendingOperationAffectingOwner(
 ): Promise<void> | null {
   return environment
     .getOperationTracker()
-    .getPromiseForPendingOperationsAffectingOwner(request);
+    .getPromiseForPendingOperationsAffectingOwner(request.identifier);
 }
 
 class FragmentResourceImpl {
@@ -434,7 +434,7 @@ class FragmentResourceImpl {
   ): Promise<void> | null {
     const environment = this._environment;
     const networkPromise =
-      getPromiseForActiveRequest(environment, fragmentOwner) ??
+      getPromiseForActiveRequest(environment, fragmentOwner.identifier) ??
       getPromiseForPendingOperationAffectingOwner(environment, fragmentOwner);
 
     if (!networkPromise) {
--- a/packages/relay-runtime/query/__tests__/fetchQueryInternal-test.js
+++ b/packages/relay-runtime/query/__tests__/fetchQueryInternal-test.js
@@ -374,7 +374,10 @@ describe('isRequestActive', () => {
 
 describe('getPromiseForActiveRequest', () => {
   it('returns null if request is not in flight', () => {
-    const promise = getPromiseForActiveRequest(environment, query.request);
+    const promise = getPromiseForActiveRequest(
+      environment,
+      query.request.identifier,
+    );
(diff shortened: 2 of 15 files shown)


josephsavona:
Seems reasonable but requesting review from @jstejada and @rbalicki2 since they're iterating on hooks right now.

sibelius:
you should work on relay team

jstejada:
adding @alunyov since hw originally worked on this

alunyov:
Thanks, @morrys! Changes look good to me. Can you provide additional context on the memory usage? Did you measure the effect of the change?

I understand that storing strings versus storing references to the objects should not be that big. But maybe I'm wrong here. Do you have any information on that?

alunyov:
Just a quick update from my findings:

A quote from this article:
https://itnext.io/v8-deep-dives-understanding-map-internals-45eb94a183df#:~:text=V8%20uses%20the%20so%2Dcalled,stands%20for%20the%20hash%20table.

How the hash function works for strings/objects:

> For string-like values (strings and symbols), it calculates hash code based on the [string contents](https://github.com/nodejs/node/blob/238104c531219db05e3421521c305404ce0c0cce/deps/v8/src/objects/string.cc#L1338) and then caches it in the internal header.
>
> Finally, for objects, V8 calculates the hash code based on a [random number](https://github.com/nodejs/node/blob/238104c531219db05e3421521c305404ce0c0cce/deps/v8/src/execution/isolate.cc#L3785) and then caches it in the internal header.

So, I assume, in practice, it actually faster to generate a hash for an object than for a string because it will iterate over string characters.

If you have any benchmark data, it would be interesting to see what do we have in practice here.

alunyov (changes requested):
Thank you @morrys for this PR!

Overall, the changes looks good.

Please update the motivation part of the PR. Why and how exactly does it improve the memory usage, and what impact it has on the runtime.

morrys:
Hi @alunyov,
I try to better motivate my PR with a practical test and some considerations.

Below is the result of a test carried out on chrome using react in which in the first case the identifier of the request is inserted in a Map and in the second case the object of the request. 

| Cache Identifiers  | Cache References |
| ------------- | ------------- |
requestid | requestref

This is the code used to do the test: 

```ts
const maxMemory = Math.floor(performance.memory.jsHeapSizeLimit / 1000000) - 350;

const getMemory = () => {
    return Math.floor(performance.memory.usedJSHeapSize / 1000000);
};

const numRequest = 5000000;

const AppTest = (propsApp) => {
    const [mapState, setMap] = React.useState(new Map());
    const tot = Array.from(mapState.values()).reduce(
        (acc, cur) => acc + cur.size,
        0,
    );
    return (
        <>
            <button
                onClick={() => {
                    const map = new Map();
                    const start = new Date().getTime();
                    console.log('start time', start);
                    console.log('max memory MB:', maxMemory);
                    try {
                        [...Array(15)].forEach((_e, index) => {
                            const maprequest = new Map();
                            map.set(index, maprequest);
                            [...Array(numRequest)].forEach((e, i) => {
                                if (getMemory() < maxMemory) {
                                    const variables = { userId: i };
                                    const gqlQuery = getRequest(QueryApp);
                                    const request = createOperationDescriptor(
                                        gqlQuery,
                                        variables,
                                        {},
                                    );
                                    // CACHE IDENTIFIERS
                                    /*maprequest.set(
                                        request.request.identifier,
                                        request.request.identifier,
                                    );*/
                                    // CACHE REFERENCES
                                    maprequest.set(request, request)
                                } else {
                                    throw Error('Out of Memory, memory used MB:');
                                }
                            });
                            console.log(index + 1 + ' - loop time sec:', ((new Date().getTime() - start) / 1000));
                            console.log(index + 1 + ' - loop memory MB:', getMemory());
                        });
                    } catch (e) {
                        console.log(e.message, getMemory());
                    }
                    const end = new Date().getTime();
                    console.log('end', end);
                    console.log('end time sec:', ((end - start) / 1000));
                    console.log('end memory MB:', getMemory());
                    setMap(map);
                }}
                className="refetch"
            >
                memory
            </button>
            <div>Requests: {tot}</div>
        </>
    );
};
```


Furthermore, considering these two requests:

```ts
const variables = {userId: 0};
const gqlQuery = getRequest (QueryApp);
const request = createOperationDescriptor (gqlQuery, variables, {},);
const request2 = createOperationDescriptor (gqlQuery, variables, {},);
```

another important point of reflection is this:

**request === request2 is false
request.identifier === request2.identifier is true**


Finally, in my opinion, even the function signatures are more explanatory as it is immediately evident that the only property used within it is the identifier.

alunyov:
Oh, wow! I didn't expect this. Thanks you for the performance report @morrys, this looks amazing!

alunyov (approved):

alunyov:
@morrys  a minor feedback:
Can we keep the public interface of the OperationTracker intact. Still accept the `request` object, but internally store identifiers? This should reduce the number of changes in the tests, and it won't be a breaking change?

morrys:
> Finally, in my opinion, even the function signatures are more explanatory as it is immediately evident that the only property used within it is the identifier.
Since this is a public interface, it is better to use the Request and not the identifier to avoid BC in case you need to use other properties in the future

@alunyov soon I adapt the PR 👍

morrys:
@alunyov done 👍

PR #3353, opened 2021-02-09
feat: added id & dispose function in RelayModernEnvironment

morrys:
Hi to all,
with this PR I added:
 * a unique id to the environment
 * the function dispose to the environment in order to clear the cache of requests associated with the environment
 * deleteQueryResourceForEnvironment in QueryResource in order to delete the query resources associated with the environment from the cache
 * deleteFragmentResourceForEnvironment in FragmentResource in order to delete the fragment resources associated with the environment from the cache
 * useEffect in RelayModernProvider in order to dispose the environment and delete resources when the environment is changed

In this way it is possible to have:
  * a control over the management of the resources used in relay
  * eliminate the need to use weakmap
  * use the id in the useEffect instead of the object
  * evaluate the creation of a public function `getPromiseForAllRequest `in `fetchQueryInternal` in order to be able to use this information during the SSR (you can easily make a first render, perform the await of all promises and perform a new render)

Finally, I made the following changes to the tests:
  * replaced the use of ReactRelayContext in favor of RelayEnvironmentProvider
  * moved the creation of alternative environments to the beginning of the test as the jest resetModule reset the RelayModernEnvironment index and environments with the same identifier were generated

### DIFF
--- a/packages/relay-experimental/EntryPointContainer.react.js
+++ b/packages/relay-experimental/EntryPointContainer.react.js
@@ -67,7 +67,7 @@ function EntryPointContainer<
       profilerContext,
       rootModuleID,
     });
-  }, [environment, profilerContext, rootModuleID]);
+  }, [environment.id, profilerContext, rootModuleID]);
   return (
     <Component
       entryPoints={entryPoints}
--- a/packages/relay-experimental/FragmentResource.js
+++ b/packages/relay-experimental/FragmentResource.js
@@ -41,12 +41,6 @@ export type FragmentResource = FragmentResourceImpl;
 
 type FragmentResourceCache = Cache<Promise<mixed> | FragmentResult>;
 
-const WEAKMAP_SUPPORTED = typeof WeakMap === 'function';
-interface IMap<K, V> {
-  get(key: K): V | void;
-  set(key: K, value: V): IMap<K, V>;
-}
-
 type SingularOrPluralSnapshot = Snapshot | $ReadOnlyArray<Snapshot>;
 opaque type FragmentResult: {data: mixed, ...} = {|
   cacheKey: string,
@@ -497,23 +491,28 @@ function createFragmentResource(environment: IEnvironment): FragmentResource {
   return new FragmentResourceImpl(environment);
 }
 
-const dataResources: IMap<IEnvironment, FragmentResource> = WEAKMAP_SUPPORTED
-  ? new WeakMap()
-  : new Map();
+const dataResources: Map<number, FragmentResource> = new Map();
 
 function getFragmentResourceForEnvironment(
   environment: IEnvironment,
 ): FragmentResourceImpl {
-  const cached = dataResources.get(environment);
+  const cached = dataResources.get(environment.id);
   if (cached) {
     return cached;
   }
   const newDataResource = createFragmentResource(environment);
-  dataResources.set(environment, newDataResource);
+  dataResources.set(environment.id, newDataResource);
   return newDataResource;
+}
+
+function deleteFragmentResourceForEnvironment(
+  environment: IEnvironment,
+): void {
+  dataResources.has(environment.id) && dataResources.delete(environment.id);
 }
 
 module.exports = {
   createFragmentResource,
   getFragmentResourceForEnvironment,
+  deleteFragmentResourceForEnvironment
 };
--- a/packages/relay-experimental/LazyLoadEntryPointContainer_DEPRECATED.react.js
+++ b/packages/relay-experimental/LazyLoadEntryPointContainer_DEPRECATED.react.js
@@ -196,7 +196,7 @@ function LazyLoadEntryPointContainer_DEPRECATED<
     );
     // NOTE: stableParams encodes the information from params
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [environment, environmentProvider, getPreloadProps, entryPointParamsHash]);
+  }, [environment.id, environmentProvider, getPreloadProps, entryPointParamsHash]);
   const Component = useMemo(() => {
     return getComponent();
   }, [getComponent]);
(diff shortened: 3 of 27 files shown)


jstejada:
@morrys, thanks so much for submitting this!

I had a couple of questions before proceeding:

- Regarding the `id` on environment, is there any particular reason why passing the environment itself to the useEffect deps is not desirable? Is there any other reason for wanting an id?
- Why would we eliminate the need for a WeakMap here? Is it because we'd be using the `id` as the key instead of the environment?

cc @josephsavona in case you have any thoughts

morrys:
> Regarding the id on environment, is there any particular reason why passing the environment itself to the useEffect deps is not desirable? Is there any other reason for wanting an id?
  * similar to this PR https://github.com/facebook/relay/pull/3348
  * id is cool 👍 

> 
Why would we eliminate the need for a WeakMap here? Is it because we'd be using the id as the key instead of the environment?
  * avoid if like this https://github.com/facebook/relay/blob/master/packages/react-relay/relay-hooks/QueryResource.js#L580, so as to always have the same behavior and the same efficiency
  * a control over the management of the resources used in relay
  * WeakMap not support clear, size, keys, values
  * evaluate the creation of a public function getPromiseForAllRequest in fetchQueryInternal in order to be able to use this information during the SSR (you can easily make a first render, perform the await of all promises and perform a new render).

josephsavona:
It feels like this PR is trying to accomplish multiple things at once and those goals are not clearly defined, which makes it difficult to evaluate. Generally speaking, we encourage starting a discussion about the approach before proceeding with larger changes. We _really_ don't like having to say no to changes (especially when you've clearly put so much time and energy into this!), and we can avoid that by first discussing the goals, then coming to a consensus on the approach.

In this case I think we need to start from the beginning:
* Can you clarify why passing the environment instance as a hook dependency is problematic? Note that the environment should be retained for the lifetime of any hooks that reference it anyway, so I don't see a particular problem with our current approach. 
* Can you clarify why the use of WeakMap is insufficient or problematic? Again, memory usage should be fairly deterministic here, with environments being freed automatically. 
* We don't believe that a `getPromiseForAllRequest()` approach is the right way to go for SSR. The EntryPoint API allows applications to ensure that all data for a tree is fetched before starting to render that tree (for SSR). Also, the React Core team is currently working on a new, streaming server-rendering that will support suspense, which will allow fetching data more naturally during SSR.

Given the above it isn't immediately clearly that this PR is the right direction. The main piece that seems like it would be useful is `Environment::dispose()` in order to cleanup any remaining work on an environment, but i'm curious how often this is actually needed in practice.

morrys:
@josephsavona don't worry you can always say no, from a PR I always learn something whether it is accepted or not.
I think that a discussion with a proposal already made is easier both to explain and to understand..


> Can you clarify why passing the environment instance as a hook dependency is problematic? Note that the environment should be retained for the lifetime of any hooks that reference it anyway, so I don't see a particular problem with our current approach.

In the comment of this PR https://github.com/facebook/relay/pull/3348#issuecomment-775204516 I have created a practical test in which it shows the different memory usage between reference and id. 

> Can you clarify why the use of WeakMap is insufficient or problematic? Again, memory usage should be fairly deterministic here, with environments being freed automatically.

as you say: `should be fairly deterministic` (not to mention that if the WeakMap is not supported the memory is never freed)

with this PR it is always definitely deterministic

> We don't believe that a getPromiseForAllRequest() approach is the right way to go for SSR. The EntryPoint API allows applications to ensure that all data for a tree is fetched before starting to render that tree (for SSR). Also, the React Core team is currently working on a new, streaming server-rendering that will support suspense, which will allow fetching data more naturally during SSR

I am sure that the React team will give a solution to the SSR for this reason I wrote that the creation of the getPromiseForAllRequest function could be evaluated and I did not put it inside the PR. However it is a function that can be created at 0 cost that a user might find useful

josephsavona:
> In the comment of this PR #3348 (comment) I have created a practical test in which it shows the different memory usage between reference and id.

That benchmark is interesting, but the results don't really apply here. Specifically, your example tracked memory usage of OperationDescriptors, comparing a version where the code accumulates a map of identifier->identifier vs a map of OperationDescriptor -> OperationDescriptor. But note that the memory win is largely from the fact that the new version doesn't otherwise hold onto references to the OperationDescriptor - if something else was holding onto a reference anyway, then storing a map of ids likely wouldn't use substantially less memory.

For the environments, active environment instances will already be retained in memory anyway, so using identifiers as map keys isn't clearly a win. 

> as you say: should be fairly deterministic (not to mention that if the WeakMap is not supported the memory is never freed)

Right: it's reasonably deterministic, so the question is whether there is actually an observable difference between the memory usage w WeakMap or without. Generally speaking WeakMap should be supported in any version of Node.js that folks may be running, and it seems the main issue of where you'd be concerned about memory usage w lots of environment instances is on the server. That said we're fans of making performance and memory usage more predictable, so this approach is compelling. However, there's a downside in that it requires the user to clean up manually.

morrys:
> if something else was holding onto a reference anyway, then storing a map of ids likely wouldn't use substantially less memory.
> For the environments, active environment instances will already be retained in memory anyway, so using identifiers as map keys isn't clearly a win.

you are right, we need more reasons to justify the use of the id 

> Right: it's reasonably deterministic, so the question is whether there is actually an observable difference between the memory usage w WeakMap or without. Generally speaking WeakMap should be supported in any version of Node.js that folks may be running, and it seems the main issue of where you'd be concerned about memory usage w lots of environment instances is on the server. That said we're fans of making performance and memory usage more predictable, so this approach is compelling. However, there's a downside in that it requires the user to clean up manually.

yes, client side with RelayEnvironmentProvider is automatic, server side must be done manually (in the environment creation logic, before creating the new environment, the dispose of the previous one must be called)

PR #3359, opened 2021-02-17
Allow array of Uploadable in UploadableMap

AaronBuxbaum:
The current types of `UploadableMap` define a strict single-value pairing on `FormData`, despite the fact that with vanilla javascript, `FormData` values can be arrays:
```
const formData = new FormData();
formData.append('myKey', 1);
formData.append('myKey', 2);
formData.append('myKey', 3);
```
in this case, `myKey` is `[1, 2, 3]`

The example comes up primarily when using `multiple` file input tags -- the value of that tag is a `FileList`, which is often then converted to an array with `Array.from`. I'm looking for any good solution to multiple uploadables on the same `FormData` key; this just seems the easiest but I'm not tied to it. I'd appreciate any guidance as to the "correct" way to tackle this kind of problem!

Some libraries alternatively work by also appending a key which represents a mapping from form data key to pathname, enabling the backend to know which keys to look at. This works, but it's quite ugly/opaque, 
 and it pushes an implicit requirement on the consumer to generate such a mapping.

It's also worth noting that `react-relay-network-modern` (which is referenced by Relay, but not directly endorsed) allows a `FileList` or array of files via `extract-files`: https://github.com/relay-tools/react-relay-network-modern/blob/master/src/middlewares/upload.js#L19 and https://github.com/jaydenseric/extract-files/blob/master/src/public/extractFiles.js#L98

### DIFF
--- a/packages/relay-runtime/network/RelayNetworkTypes.js
+++ b/packages/relay-runtime/network/RelayNetworkTypes.js
@@ -120,7 +120,10 @@ export type SubscribeFunction = (
 ) => RelayObservable<GraphQLResponse>;
 
 export type Uploadable = File | Blob;
-export type UploadableMap = {[key: string]: Uploadable, ...};
+export type UploadableMap = {
+  [key: string]: Uploadable | Array<Uploadable>,
+  ...,
+};
 
 /**
  * React Flight tree created on the server.


alloy:
cc @rbalicki2

giautm:
Hey @AaronBuxbaum sorry about off-topic, but how do you handle Uploadble in your backend? Did you use any framework for that?

AaronBuxbaum:
> Hey @AaronBuxbaum sorry about off-topic, but how do you handle Uploadble in your backend? Did you use any framework for that?

In the uploadables case, we're not doing anything too fancy -- the request body gets assigned to a formData object, and sent along as normal to the backend. Once it's there, it's just regular Flask file management

We usually use a (better) async approach where we provide a presigned S3 URL and have the client upload directly as opposed to uploading as part of the request -- this is a nicer experience and better for our servers. But there are a couple of cases where direct "uploadable" uploading still makes sense for us

giautm:
Thanks @AaronBuxbaum for fast reply, I'm very appreciated.

I did both FormData upload and presigned S3 URL upload. But I want to references more ways that everyone did.

One more question: How you associated uploaded files with your mutation in the presigned S3 URL upload case?

AaronBuxbaum:
bumping on this -- any feedback on this?

PR #3375, opened 2021-03-03
docs: switch comment style to blockstring

rgbkrk:
When going through the quick start guide, I noticed this strange comment rendering:

image

This PR switches over the comment format to blockstring, as is currently used in the `schema.graphql` for `relay-examples`. At the very least, the docs are now consistent with the example! 😄 

It renders like this now:

image

-------

Note: this is not an issue with `website-v2` so the updated docusaurus set up must be working a-ok.

image


jstejada:
we've published new docs so this no longer applies. feel free to submit any improvements on top of the new docs!

PR #3380, opened 2021-03-05
generate refetch query with directives

bonafideyan:
still have more work to do,

if relay support use directive this way, it is kind and user are lucky.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "flow-bin": "^0.145.0",
     "glob": "^7.1.1",
     "graphql": "^15.0.0",
-    "gulp": "4.0.2",
+    "gulp": "^4.0.2",
     "gulp-babel": "8.0.0",
     "gulp-chmod": "3.0.0",
     "gulp-header": "2.0.9",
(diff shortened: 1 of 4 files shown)


josephsavona:
Can you please clarify what the purpose of this change is? It appears you're trying to copy all the directives from any operation that references a refetchable fragment onto the corresponding refetch query - which doesn't make sense. As I noted in the corresponding issue, refetch queries must be generated solely based on the fragment definition.

bonafideyan:
initially I have not written the directive on fragment, and expect the refetch bring the directive like the root query.
I thought this should be default, no case where query bring the directive but refetch not. writing the directive on fragment works, but is redundant.

josephsavona:
Thanks for confirming what you intended here. Reiterating my comment from above (and in the slack discussion), **generated refetch queries can only be derived from the fragment definition**. They _cannot_ use information from queries that may have referenced them. Imagine if there are two queries A and B that reference some refetchable fragment C. If A and B have different directives, then it would not be clear which directives to use for C's refetch query. Combining all of the directives is incorrect, as not all of them should apply. The only logical, predictable choice is to use the directives from the fragment itself.

Also note that **directives should not change the data returned**. Directives can affect _how_ data is returned (defer/stream), _whether_ it is returned (include/skip), add logging, trigger supplementary behavior, etc, but directives may not affect the data returned by a query. The example you provided - `@withUser` - definitely looks like a directive that breaks this invariant. We'd recommend finding a different way to handle authorization/identity.

PR #3381, opened 2021-03-06
Remove mutation from the server specification

suhaibmujahid:
As I understood, the mutation restrictions are not required anymore, thous I think it is cleaner to remove them from the server specification.


kassens:
Thanks makes sense to me. I'll merge this next week!

PR #3423, opened 2021-03-25
docs: Add Contra to the list of Relay users

kulshekhar:
Relay has been at the heart of the main app at Contra right from the beginning. So much so that it features in job listings right alongside React!


kulshekhar:
I have signed the CLA but it isn't reflecting here. Is there anything else I need to do?

Boeing787:
@sibelius Is it possible to get this in motion? Contra is big on relay, and we'd love to make it known to the world 💯

kulshekhar:
activity..

PR #3426, opened 2021-03-29
Fix documentation about RelayEnvironmentProvider

juffalow:
Several places in documentation uses:

```js
const {RelayEnvironmentProvider} = require('react-relay');
```

But this is not working because there is no export like this. It is missing
hooks "namespace".

### DIFF
--- a/website/versioned_docs/version-v11.0.0/api-reference/hooks/relay-environment-provider.md
+++ b/website/versioned_docs/version-v11.0.0/api-reference/hooks/relay-environment-provider.md
@@ -13,7 +13,7 @@ This component is used set a Relay environment in React Context. Usually, a *sin
 ```js
 const React = require('React');
 
-const {RelayEnvironmentProvider} = require('react-relay');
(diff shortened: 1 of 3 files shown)


juffalow:
I am trying it right now and it is not working.

```json
"react-relay": "^11.0.1",
```

Also in the [example project](https://github.com/relayjs/relay-examples/blob/master/issue-tracker/src/index.js) it is the same.

sibelius:
it is working well for me here

https://github.com/sibelius/relay-workshop/blob/master/packages/web/src/Providers.tsx#L2

image

I've also checked my node_modules, and it is there

juffalow:
Ah, yes, you are right. I didn't realize that after clicking on the error it took me to`@types/react-relay` and that package wasn't updated. Sorry for this :-( maybe worth updating the example project I mentioned.

juffalow:
Also I am closing this pull request as it is resolved and ok. Thank you @sibelius !

sibelius:
try to upgrade there and fix imports

I did like this https://github.com/sibelius/relay-workshop/pull/390

PR #3447, opened 2021-04-06
Fix typo in config.src doc string

eliperkins:
I think this is meant to be "where to search for source code," since "fo a" doesn't make sense.


kassens:
Thanks!

PR #3461, opened 2021-04-18
Remove misleading advice regarding single artifact directory

squidsoup:
The advice regarding the configuration of a single artifact directory to enable typing of fragment references appears to be outdated. Currently fragment references are typed appropriately when imported from a co-located `__generated__` path. The background information about Haste is interesting, but given this feature is no longer _required_, I suggest removing it here to avoid confusion.

### DIFF
--- a/website/versioned_docs/version-v11.0.0/guides/type-emission.md
+++ b/website/versioned_docs/version-v11.0.0/guides/type-emission.md
@@ -391,9 +391,7 @@ return <ExampleFragmentComponent artist={data.artist} />;
 
 ## Single artifact directory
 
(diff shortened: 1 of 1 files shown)


sibelius (approved):

zth:
Is this also true for the generated Flow types? I know it's true for the TS plugin, but for Flow...?

squidsoup:
> Is this also true for the generated Flow types? I know it's true for the TS plugin, but for Flow...?

A fair point, I have no idea. Are any Flow users able to confirm? @zth

sibelius:
I think flow never used a single artifact folder

PR #3464, opened 2021-04-19
Add new example project in learning resources documentation

biantris:
- Adding the Relay Workshop to Community Learning Resources as a Relay example projects, I think it will help.
- Relay-Workshop (by @sibelius): https://github.com/sibelius/relay-workshop


sibelius (approved):
can we add more description for each project?

Relay Workshop is an example of fullstack GraphQL Relay real world of a social network

it is also a set of 12 practical exercises to learn Relay hooks API

PR #3465, opened 2021-04-20
Fix: Check if global exists before calling ErrorUtils

mellson:
I was experiencing some `global is not defined`'s from `RelayPublishQueue.js` when using Relay in Vite. And I think this change will solve it for the cases where global is not yet defined.

### DIFF
--- a/packages/relay-runtime/store/RelayPublishQueue.js
+++ b/packages/relay-runtime/store/RelayPublishQueue.js
@@ -55,7 +55,7 @@ type PendingUpdater = {|
 |};
 
 const applyWithGuard =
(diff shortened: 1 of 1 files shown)


sibelius (approved):

asterikx:
I really hope this can be merged soon. I'm experiencing the same issues than @mellson with Vite 2.

mellson:
Until this gets fixed you can solve it by defining global in you vite config in dev mode.
But don't use it when building though.

```
export default defineConfig({
  define: {
    global: {
      ErrorUtils: null,
    },
  }
})
```

PR #3467, opened 2021-04-21
Fixed Repository Reference in monorepo package.json files

wyattjoh:
Based on the documentation for the `package.json` syntax on the `repository` field, this PR adds support for subdirectory information (See https://docs.npmjs.com/cli/v6/configuring-npm/package-json#repository).


wyattjoh:
Note that I only added this change to non-private packages.

PR #3469, opened 2021-04-23
chore: Docusaurus with webpack 5

slorber:
Use an unreleased version of Docusaurus to test the webpack5 support works for this site, detect potential problems, and perf improvements.

See also https://github.com/facebook/docusaurus/pull/4089

Perf improvements
- First build (local): 300s
- Second build (local): 90s 

To get the benefits, it requires caching `node_modules/.cache`. 
It is not possible in GH actions, so I suggest migrating the site to Netlify and use the appropriate toml config (example: https://github.com/react-navigation/react-navigation.github.io/pull/974/files)

--- 

Unrelated suggestion: Relay has a lot of versions and long build/start times

Assuming nobody ever edit Relay docs of 2 years ago, you can improve the DX by limiting the number of versions display in dev with this docs plugin config:

```
          onlyIncludeVersions:
            (isDev || isDeployPreview)
              ? ['current', ...versions.slice(0, 2)]
              : undefined,
```

For faster prod build, you can also move older docs to standalone deployments and link to them (see Jest for versions < 25 => https://jestjs.io/)


website/sitePlugin.js
```diff
@@ -0,0 +1,16 @@
+const webpack = require("webpack");
+
+module.exports.default = function sitePlugin() {
+  return {
+    name: "sitePlugin",
+    configureWebpack(config) {
+      return {
+        plugins: [
+          new webpack.EnvironmentPlugin({
+            FB_INTERNAL: process.env.FB_INTERNAL ?? false,
```
slorber, line 10: temporarily required, will likely be fixed in the fb custom preset by @jknoxville

website/sitePlugin.js
```diff
@@ -0,0 +1,16 @@
+const webpack = require("webpack");
+
+module.exports.default = function sitePlugin() {
```
slorber, line 3: Plugin to fix env variable error when visiting http://localhost:3000/docs/

image

This should be added to FB internal plugin by @jknoxville

jknoxville: Thanks very much for the heads-up and fix @slorber 
I've incorporated this change into the docusaurus plugin, so you can get it by upgrading:

`docusaurus-plugin-internaldocs-fb` to `^0.8.4`

slorber:
ready for review

jstejada:
@slorber when attempting to install these packages i get an error that version `2.0.0-alpha.75` does not exist. is this the right version to use?

jstejada:
nvm i was making a mistake, attempting to land this locally now

christianvuerings:
@jsteiner @slorber I think this change causes an issue with zIndexes on the relay docs site: 

image

slorber:
Hi

I'm sorry but I'm in holiday until 1st June

I believe this is some custom css on relay, as other docusaurus sites don't
have this problem

slorber:
Fixing website CSS issue in https://github.com/facebook/relay/pull/3505

PR #3471, opened 2021-04-27
[rust]: properly output typescript artifacts

maraisr:
- Producing ts artifacts were actually "typescript" typegen, but wrapped in flow things. So this will now instead print ts things
- introduce `@ts-nocheck`
- file categorizer didnt like seeing tsx ts files.

### DIFF
--- a/compiler/crates/relay-compiler/src/build_project/artifact_content.rs
+++ b/compiler/crates/relay-compiler/src/build_project/artifact_content.rs
@@ -306,17 +306,20 @@ fn generate_operation(
         )
         .unwrap();
     }
+
+    let node_request = printer.print_request(
+        schema,
+        normalization_operation,
+        &operation_fragment,
+        request_parameters,
+    );
+
     if let TypegenLanguage::Flow = project_config.typegen_config.language { 
         writeln!(content, "*/\n").unwrap();
         writeln!(
             content,
             "var node/*: ConcreteRequest*/ = {};\n",
-            printer.print_request(
-                schema,
-                normalization_operation,
-                &operation_fragment,
-                request_parameters,
-            )
+            node_request
         )
         .unwrap();
     } else {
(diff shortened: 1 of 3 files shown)


compiler/crates/relay-compiler/src/build_project/artifact_content.rs
```diff
@@ -434,22 +516,42 @@ fn generate_fragment(
...
     writeln!(content, "module.exports = node;").unwrap();
     sign_file(&content).into_bytes()
 }
 
-fn get_content_start(config: &Config) -> String {
+fn get_content_start(config: &Config, project_config: &ProjectConfig) -> String {
     let mut content = String::new();
+
+    if let TypegenLanguage::TypeScript = project_config.typegen_config.language {
+        writeln!(content, "// @ts-nocheck").unwrap();
```
kassens, line 552: I didn't really find documentation on this, the best I found were these release notes:

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#-ts-nocheck-in-typescript-files

Does this mean the file is ignored? Why would we want that?

kassens (commented):
Thanks, this looks pretty good. Just wondering why we need the no-check comment?

maraisr:
We currently do that in the relay compiler for typescript world already. Think sometimes the ast doesn't fall right with the types that exist, I suppose a better way could have been to just wrap the ast in a ts-ignore block?! 

Cannot quite recall exact cases.

compiler/crates/relay-compiler/src/build_project/artifact_content.rs
maraisr: Just means don't typecheck anything in the file—but is still made available to consumers to read. So exported Node, or types is assumed to be valid, rather than checked to be valid.

maraisr: Come to think of it, this doesn't need to exist "at the very very top" just needs to exist before any code. So we can move this back into the generate methods. Sorry, will amend!

maraisr:
@kassens this is ready again for your review.

compiler/crates/relay-compiler/src/build_project/artifact_content.rs
maraisr: This change has been made.

PR #3501, opened 2021-05-28
[docs] Add missing `uploadables` argument to `useMutation` docs

dariusk:
The [documentation page for `useMutation`](https://relay.dev/docs/api-reference/use-mutation/) lists the signature of the custom arguments object which is "almost the same" as the `commitMutation` API. It currently lacks `uploadables` -- I thought this was one of the differences between this object and the referenced API, but it turns out that [`useMutation`'s arguments object does have an `uploadables` property](https://github.com/facebook/relay/blob/4b78c7dd27b286f9ee8f5b993ccb160163956999/packages/react-relay/relay-hooks/useMutation.js#L54). It just isn't listed on the docs, so I've added it.

I purposely did not include direct links to types in order to keep my new bullet point consistent with the style of the rest of this page (which does not seem to be consistent with the rest of the docs).


sibelius (approved):

dariusk:
Just commenting here -- now that the CLA has been signed, can this be re-reviewed and merged?

poteto:
Thanks and apologies for the delay!

PR #3505, opened 2021-06-02
Fix website issues

slorber:
Docusaurus beta 0 has been released so let's use it!

Also, you are currently affected by an annoying regression in a Docusaurus transitive dependency (cssnano@5.0.3 => https://github.com/facebook/docusaurus/issues/4826)

Reported by @christianvuerings here: https://github.com/facebook/relay/pull/3469#issuecomment-845980982

image

This should now be fixed by re-generating the lockfile.

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -18,9 +18,9 @@
     "docusaurus": "docusaurus"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


slorber:
seems to work better now, thanks

PR #3512, opened 2021-06-05
Improve Persisted Queries documentation

kyarik:
This PR introduces the following improvements to the [Persisted Queries docs](https://relay.dev/docs/guides/persisted-queries/):
- Mention `doc_id` as an example for the persisted query ID parameter, rather than making it potentially sound as a requirement.
- Capitalize "ID" properly.
- Fix capitalization and punctuation in list about `pushQueries.js` possibilities.
- Capitalize sentences in the list about persisted queries usefulness.
- Improve formatting of persisted queries code examples.
- Fix verb in a sentence about persisted queries.
- Fix [`express-graphql`](https://github.com/graphql/express-graphql) import in persisted queries code example. We need to import the `graphqlHTTP` named export in the latest version of `express-graphql` since it no longer has a default export.
- Use [`express-graphql-persisted-queries`](https://github.com/kyarik/express-graphql-persisted-queries) in the simple server example. This is actually an open source library that I created because I couldn't find an existing solution flexible enough to: 
  - Allow to send HTTP GET requests with persisted queries (necessary to preload queries with `<link rel="preload">`, for example).
  - Allow a custom way to map a query ID to an actual query (not necessarily using an object, but a custom function that queries a DB, for example).
  - Allow to specify whether you want to allow only persisted queries for enhanced security (this is mentioned in the docs, yet I couldn't find a solution supporting this).
  - Allow to specify a custom query ID key (rather than forcing a specific query ID key, like `queryId`).
  - Be lightweight, with minimal dependencies.

Considering the flexibility that this library gives, I think it makes more sense to mention it in the example, as I think many people will find it useful.

### DIFF
--- a/website/docs/guides/persisted-queries.md
+++ b/website/docs/guides/persisted-queries.md
@@ -18,10 +18,10 @@ import {FbInternalOnly, OssOnly} from 'internaldocs-fb-helpers';
 
 The relay compiler supports persisted queries. This is useful because:
 
--   the client operation text becomes just an md5 hash which is usually shorter than the real
+-   The client operation text becomes just an md5 hash which is usually shorter than the real
     query string. This saves upload bytes from the client to the server.
 
--   the server can now whitelist queries which improves security by restricting the operations
+-   The server can now whitelist queries which improves security by restricting the operations
     that can be executed by a client.
 
 <OssOnly>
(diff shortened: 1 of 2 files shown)


sibelius (approved):

kyarik:
Hi @alunyov,
I just rebased this PR. Is there anything blocking it from being merged?

Issue #3499, opened 2021-05-26
[relay-compiler] Generated "require" calls include file extension


ascagnel:
When a Relay fragment is compiled and generates a `require` call (ie: `@refetchable` fragments), the path to the generated file includes its extension.

```graphql
        fragment MySample_data on Viewer @refetchable(queryName: "MySampleRefetchQuery") {
            id
        }
```

```ts
  "metadata": {
    "refetch": {
      "connection": null,
      "fragmentPathInResult": [
        "node"
      ],
      "operation": require('./MySampleRefetchQuery.graphql.ts'),
      "identifierField": "id"
    }
  },
```

This can cause issues later in the build chain.  In my case, the output from relay-compiler is in TypeScript, which will later be passed through `tsc` and deployed as a plain `.js` file, but the contents within will still be pointing to a `.ts` file that no longer exists.  Since `require` does not need the file extension to be included, can it be removed from the compile script?

(note: this was [flagged earlier](https://github.com/facebook/relay/issues/2713#issuecomment-693708450), but it looks like no action was taken)

PR #3518, opened 2021-06-15
fix: do not emit extension in compiler includes

ascagnel:
when a generated relay query includes a "require" statement, do not include the file extension in the "require" call.

fixes https://github.com/facebook/relay/issues/3499


ascagnel:
I have signed the CLA, as per the contributing guide.

ascagnel:
This fixes a mismatch between the babel plugin and the compiler, so both will emit the path without the file extension.

https://github.com/facebook/relay/blob/master/packages/babel-plugin-relay/__tests__/__snapshots__/BabelPluginRelay-modern-test.js.snap#L38

robrichard:
This issue blocks us from using `@refetchable` with typescript. When our code is deployed, all typescript files are transpiled and renamed to a `.js` extension, but the `require`s generated by relay-compiler have a `.ts` extension. This differs from the `require`s generated by `babel-plugin-relay`, which do not specify an extension.

PR #3521, opened 2021-06-20
Better wording

felipe-gdr:


poteto (changes requested):
Thanks @felipe-gdr, can you update https://github.com/facebook/relay/blob/master/website/docs/guided-tour/rendering/queries.md as well? That is the source for this doc, and will ensure that subsequent doc releases contain your fixes.

felipe-gdr:
> Thanks @felipe-gdr, can you update https://github.com/facebook/relay/blob/master/website/docs/guided-tour/rendering/queries.md as well? That is the source for this doc, and will ensure that subsequent doc releases contain your fixes.

Sure @poteto , I just pushed a commit with the fix in the source doc.

PR #3523, opened 2021-06-21
Fix: Check if global exists before using it

mellson:
Similar to https://github.com/facebook/relay/pull/3465 calling `global` can be troublesome since it doesn't always exist, for instance when using `vite`.


packages/relay-runtime/store/OperationExecutor.js
```diff
@@ -1558,7 +1558,7 @@ function validateOptimisticResponsePayload(
 }
 
 const isPerformanceNowAvailable =
-  global.performance != null && typeof global.performance.now === 'function';
+  global?.performance != null && typeof global.performance.now === 'function';
```
poteto, line 1561: Thanks @mellson! This got moved to https://github.com/facebook/relay/blob/master/packages/relay-runtime/util/withDuration.js, could you rebase and update your PR?

poteto (changes requested):

packages/relay-runtime/store/OperationExecutor.js
mellson: @poteto sure thing!

mellson: Hi @poteto I made a mess of the rebasing 🥴 So I recreated this much cleaner here https://github.com/facebook/relay/pull/3525

PR #3525, opened 2021-06-22
Fix: Check if global exists before using it

mellson:
Similar to #3465 calling global can be troublesome since it doesn't always exist, for instance when using vite.

### DIFF
--- a/packages/relay-runtime/util/withDuration.js
+++ b/packages/relay-runtime/util/withDuration.js
@@ -13,7 +13,7 @@
 'use strict';
 
 const isPerformanceNowAvailable =
-  global.performance != null && typeof global.performance.now === 'function';
+  global?.performance != null && typeof global.performance.now === 'function';
 
 function currentTimestamp(): number {
   if (isPerformanceNowAvailable) {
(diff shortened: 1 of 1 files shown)


packages/relay-runtime/util/withDuration.js
```diff
@@ -13,7 +13,7 @@
 'use strict';
 
 const isPerformanceNowAvailable =
-  global.performance != null && typeof global.performance.now === 'function';
+  global?.performance != null && typeof global.performance.now === 'function';
```
poteto, line 16: It might be safer to check for existence like this:

```js
const isPerformanceNowAvailable = 
  typeof window !== 'undefined' && 
  typeof window?.performance?.now === 'function';
```

1. `whatever?.foo` where `wherever` is not a defined identifier will throw an uncaught ReferenceError
2. `window` is more appropriate here since we're checking for a browser only API. Only node.js defines `global` AFAIK.

mellson: @poteto great input, I'll update the PR

mellson:
Can confirm it works again with Vite using today's npm package from master. Thanks 😊

PR #3528, opened 2021-06-23
Hide internal-only glossary item

kassens:
@relay_early_flush is something that's just used for a specifc internal
optimization specific to the static resource pipeline used on
facebook.com.


captbaritone:
LGTM

PR #3540, opened 2021-07-14
Run Rust tests on windows CI

kassens:


kassens:
There's still 1 tests at least failing. I think the path needs to be updated from a hardcoded `/tmp/` to something else (`report_diagnostic_test`)

PR #3553, opened 2021-07-30
Updated minimum @babel/runtime version for relay-compiler and relay-runtime

tapayne88:
The version of @babel/runtime required by relay-compiler and relay-runtime is incorrect. Both make use of `createForOfIteratorHelper` 
 which is only available in @babel/runtime `>7.9.0`.

I used unpkg.com to figure out which version introduced it.
- https://unpkg.com/@babel/runtime@~7.8/helpers/createForOfIteratorHelper.js
- https://unpkg.com/@babel/runtime@7.9.0/helpers/createForOfIteratorHelper.js

This just makes it awkward to incorporate into older projects with an existing dependency on @babel/runtime.

### DIFF
--- a/packages/relay-compiler/package.json
+++ b/packages/relay-compiler/package.json
@@ -18,7 +18,7 @@
     "@babel/core": "^7.14.0",
     "@babel/generator": "^7.14.0",
     "@babel/parser": "^7.14.0",
-    "@babel/runtime": "^7.0.0",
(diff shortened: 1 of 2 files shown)


poteto:
@tapayne88 Thanks! You'll need to run `yarn` as well to update the lockfile.

resessh:
@potato Hi, is there any update?? I'm really waiting for this update!

PR #3556, opened 2021-08-03
Docs: return Adeira Relay Example

mrtnzlml:
This example was already part of the docs for version 10 (see: https://relay.dev/docs/v10.1.3/community-learning-resources/). I would like to add it again for version 11 because we are keeping the code up-to-date to our best knowledge and possibilities for the OSS world.


mrtnzlml:
This is still valid 😎

PR #3564, opened 2021-08-15
Docs: fix simple typo in `use-preloaded-query.md`

mrtnzlml:


poteto:
@mrtnzlml Thanks! Could you update the [non-versioned](https://github.com/facebook/relay/blob/main/website/docs/api-reference/hooks/use-preloaded-query.md) doc as well? This will ensure the correction shows up in future docs releases too.

mrtnzlml:
@poteto Sure thing! Updated.

PR #3567, opened 2021-08-18
Fix typo

yujong-lee:
It's a simple typo.

`?PreloadedQuery` -> `PreloadedQuery`

Thanks.


yujong-lee:
I'm reading more doc now and I see `type State = {error: ?Error}`.

Maybe `?` is not a typo? Can somebody share some resources I should look at to understand this?

PR #3609, opened 2021-10-06
Add `toPromise()` call to Modern-fetchQuery.md

chnakamura:
If I call `then` directly on the `RelayObservable` returned by `fetchQuery` I get the following flow error: ```Cannot call `fetchQuery(...).then` because property `then` is missing in  `RelayObservable` [1].Flow(prop-missing)``` as well as the runtime error `Uncaught (in promise) TypeError: Object(...)(...).then is not a function`. 

I believe this is because `RelayObservable` doesn't define a `then()` function, and we have to call `toPromise()` before we call `then()`


sibelius (approved):

alunyov (changes requested):
Hey @chnakamura thank you for the fix. Sorry with for the delay with review. 

Can you also update the source for this document?

PR #3615, opened 2021-10-15
Fix conflicting field warning when using treatMissingFieldsAsNull

janicduplessis:
When using `treatMissingFieldsAsNull` the conflicting validation raises a false positive because the value is set using `null` but validated using `fieldValue` which at this point will be undefined. To fix this I changed the validation to use `null` instead of `fieldValue` so this matches the value that we actually set.

Before: `Warning: RelayResponseNormalizer: Invalid record. The record contains two instances of the same id: 'x' with conflicting field, y and its values: null and undefined. If two fields are different but share the same id, one field will overwrite the other.`

After: no warning

### DIFF
--- a/packages/relay-runtime/store/__tests__/RelayResponseNormalizer-test.js
+++ b/packages/relay-runtime/store/__tests__/RelayResponseNormalizer-test.js
@@ -1642,69 +1642,6 @@ describe('RelayResponseNormalizer', () => {
       expect(recordSource.toJSON()).toEqual(result);
     });
 
-    it('skips client fields not present in the payload but present in the store when treatMissingFieldsAsNull is true', () => {
-      const recordSource = new RelayRecordSource({
-        '1': {
-          __id: '1',
-          __typename: 'User',
-          id: '1',
-          firstName: 'Alice',
-          nickname: 'ecilA',
-        },
-        'client:root': {
-          __id: 'client:root',
-          __typename: '__Root',
-          'node(id:"1")': {__ref: '1'},
-        },
-      });
-      // TODO: This warning was detected when we started to enforce warnings in this test (D28091790). Payload needs to be updated.
-      expectWarningWillFire(
-        'RelayResponseNormalizer: Invalid record. The record contains two instances of the same id: `1` with conflicting field, firstName and its values: Alice and Bob. If two fields are different but share the same id, one field will overwrite the other.',
-      );
-      normalize(
-        recordSource,
-        createNormalizationSelector(
-          getRequest(StrippedQuery).operation,
-          ROOT_ID,
-          {
-            id: '1',
-            size: 32,
-          },
-        ),
-        payload,
-        {...defaultOptions, treatMissingFieldsAsNull: true},
-      );
-      const result = {
-        '1': {
-          __id: '1',
-          __typename: 'User',
-          id: '1',
-          firstName: 'Bob',
-          nickname: 'ecilA',
-        },
-        'client:root': {
-          __id: 'client:root',
-          __typename: '__Root',
-          'node(id:"1")': {__ref: '1'},
-        },
-      };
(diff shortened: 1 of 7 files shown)


scripts/config.tests.json
```diff
@@ -23,15 +23,14 @@
       "jsModuleFormat": "commonjs",
       "featureFlags": {
         "enable_flight_transform": true,
-        "enable_required_transform_for_prefix": "",
+        "enable_required_transform": true,
```
janicduplessis, line 26: Had to make these changes to run the relay compiler on tests

packages/relay-runtime/store/__tests__/RelayResponseNormalizer-test.js
```diff
@@ -3248,6 +3241,207 @@ describe('RelayResponseNormalizer', () => {
     });
   });
 
+  describe('when treatMissingFieldsAsNull is true', () => {
```
janicduplessis, line 3244: Grouped the treatMissingFieldsAsNull tests

packages/relay-runtime/store/__tests__/RelayResponseNormalizer-test.js
```diff
@@ -3248,6 +3241,207 @@ describe('RelayResponseNormalizer', () => {
...
+    it('set undefined fields to null', () => {
```
janicduplessis, line 3245: Not really related to this fix, but basic test that was missing.

packages/relay-runtime/store/__tests__/RelayResponseNormalizer-test.js
```diff
@@ -3248,6 +3241,207 @@ describe('RelayResponseNormalizer', () => {
...
+          __id: 'client:root',
+          __typename: '__Root',
+          'node(id:"1")': {
+            __ref: '1',
+          },
+        },
+      });
+    });
+
+    it('skips client fields not present in the payload but present in the store', () => {
```
janicduplessis, line 3302: Moved test

packages/relay-runtime/store/__tests__/RelayResponseNormalizer-test.js
```diff
@@ -3248,6 +3241,207 @@ describe('RelayResponseNormalizer', () => {
...
+            size: 32,
+          },
+        ),
+        payload,
+        defaultOptions,
+      );
+      expect(recordSource.toJSON()).toEqual(result);
+    });
+
+    it('does not warn if a single response contains the same fields with the same id', () => {
```
janicduplessis, line 3390: This tests repro the bug this PR fixes.

packages/relay-runtime/store/__tests__/RelayResponseNormalizer-test.js
```diff
@@ -2713,6 +2650,62 @@ describe('RelayResponseNormalizer', () => {
     );
   });
 
+  it('does not warn if a single response contains the same fields with the same id', () => {
```
janicduplessis, line 2653: Same test but with treatMissingFieldsAsNull=false

janicduplessis:
@alunyov Added a test that reproduces the bug this PR fixes, as well as a few other ones related to `treatMissingFieldsAsNull`

alunyov (commented):
Wow, fantastic work! Thank you for adding these tests!

thank you

I have one small question about config changes, but overall this looks great!

scripts/config.tests.json
```diff
@@ -23,15 +23,14 @@
...
         "no_inline": {
           "kind": "enabled"
         },
         "actor_change_support": {
           "kind": "enabled"
         }
-      },
-      "haste": false
```
alunyov, line 34: I think we no longer have this option in the config. So the OSS test config is invalid. This change makes sense! Thank you!

alunyov (approved):

PR #3618, opened 2021-10-29
Motivation for using @arguments and @argumentDefinitions

rafasanmartinez:
Where it says "in order to prevent bloating queries with global variable declarations" I would propose "in order to reutilize fragments whose variable names may differ to their corresponding ones in the query global variables declaration". 

As far as I have been able to find out, by using this feature you still have to declare at least one global variable per each argument in a fragment. The way it is currently written, leaded me to think that you may not need to do so.


rbalicki2:
Hi! Thanks for this PR. Sorry that we let it sit. I'll go ahead and import this, tweak it slightly and merge it. Thanks!

PR #3620, opened 2021-11-01
[docs] refer to correct fragment

RemyBordignon:
I believe the explanation section should be referring to `...UserComponent_user` rather than `... UsernameSection_user`.


alunyov (changes requested):
Thanks for PR!

Can you also update a source (non-versioned) docs: https://github.com/facebook/relay/blob/main/website/docs/guided-tour/rendering/fragments.md

Issue #3622, opened 2021-11-03
ERROR: Invalid AST Node when following Step-by-step Guide


skleeschulte:
I followed [the Step-by-step Guide for v12.0.0](https://relay.dev/docs/getting-started/step-by-step-guide/) using the yarn-commands. Everything works fine, except when executing the very last command - `yarn start` in step 5 - I get this error output:

    node ➜ /workspaces/relay-test (master ✗) $ yarn start
    yarn run v1.22.5
    $ yarn relay && react-scripts start
    $ yarn run relay-compiler --schema schema.graphql --src ./src/ --watchman false $@
    $ /workspaces/relay-test/node_modules/.bin/relay-compiler --schema schema.graphql --src ./src/ --watchman false

    Writing js
    ERROR:
    Invalid AST Node: { kind: "Root", operation: "query", loc: { kind: "Source", start: 3, end: 105, source: [Source] }, metadata: null, name: "AppRepositoryNameQuery", argumentDefinitions: [], directives: [], selections: [[Object]], type: Query }.
    error Command failed with exit code 100.
    info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
    error Command failed with exit code 100.
    info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
    error Command failed with exit code 100.
    info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

I double- and triple-checked all steps, verified the schema is saved in UTF-8, stripped out everything from the App.js file except the graphql query, ran `yarn run relay-compiler` directly, successfully tested the query string with [the GitHub GraphQL API Explorer](https://docs.github.com/en/graphql/overview/explorer), added and removed whitespace in the query, tested other queries composed with the GitHub GraphQL API Explorer, modified the name of the query, changed App.js to `export default App;` instead of `AppRoot`, downgraded the related packages from `^12.0.0` to `^11.0.0`, but all to no avail. The error remains:

**ERROR:
Invalid AST Node: { kind: "Root", operation: "query", loc: { kind: "Source", start: 3, end: 105, source: [Source] }, metadata: null, name: "AppRepositoryNameQuery", argumentDefinitions: [], directives: [], selections: [[Object]], type: Query }.**

skleeschulte:
[...]

PR #3623, opened 2021-11-04
Set graphql version to ^15.0.0 in install commands

skleeschulte:
Fix https://github.com/facebook/relay/issues/3622


alunyov:
I think we should figure out why exactly the node is invalid and fix that issue, but we should not fix the version in the docs.

alunyov (approved):

alunyov (approved):

PR #3624, opened 2021-11-04
Set graphql version to ^15.0.0 in yarn command

skleeschulte:
Fix https://github.com/facebook/relay/issues/3622

### DIFF
--- a/website/versioned_docs/version-v12.0.0/getting-started/installation-and-setup.md
+++ b/website/versioned_docs/version-v12.0.0/getting-started/installation-and-setup.md
@@ -52,7 +52,7 @@ module.exports = {
 Relay Modern requires a Babel plugin to convert GraphQL to runtime artifacts:
 
 ```sh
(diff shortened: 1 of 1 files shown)


alunyov:
Thanks for looking into this.

I think we should figure out why exactly the node is invalid and fix that issue, but we should not fix the version in the docs.

alunyov (approved):

website/versioned_docs/version-v12.0.0/getting-started/installation-and-setup.md
```diff
@@ -52,7 +52,7 @@ module.exports = {
 Relay Modern requires a Babel plugin to convert GraphQL to runtime artifacts:
 
 ```sh
-yarn add --dev babel-plugin-relay graphql
+yarn add --dev babel-plugin-relay graphql@^15.0.0
```
Streeterxs, line 55: Cool, but not a fix.

PR #3626, opened 2021-11-07
Remove unnecessary null values from metadata

tomgasson:
Hey folks 👋

We at Atlassian have some bundle size tooling that was flagging a larger query contributing about 200KB to our page load. That tooling _isn't_ compression aware, so the actual size is more like 9KB when gzipped so it's not too big of a concern.

But it got me looking at what contributes to that size, and what we can do to reduce it. I noticed that the majority of metadata is null values. Whilst these compress really well, they still contribute a lot of unnecessary extra bytes over the wire. Removing these nulls reduces our gzipped metadata size by about 1/3.

The code in `relay-runtime` uses loose equality, either as `field.X != null`, or `field.X ?? default`, or `if (field.X) {`, which means that we can safely remove these fields over the wire without runtime changes. 

As you can see by the stats on this PR, this removes a lot of LOC. Even though each individual impact is small, the number of queries / fragments involved mean it will probably have a decent production impact. For you at Meta, with persisted queries, it should look even better than our 1/3 reduction.

This PR addresses the following fields:
 - `"alias": null`
 - `"args": null`
 - `"concreteType": null`
 - `"storageKey": null`
 - `"abstractKey": null`
 - `"metadata": null`
 - `"plural": false`. Most fields are non-plural, so we can use non-plural as a default

### Rust compiler
In the rust compiler, this was quite easy to achieve. Instead of building objects as `object(items: Vec<ObjectEntry>)`, I've replaced these with `object_filter_none(items: Vec<Option<ObjectEntry>>)` which then filters out None values. Replacing `Primitive::Null` values with a None just means inverting where the test happens.

### JS compiler
In the JS compiler, I've currently maintained the field order so that the snapshot diffs are clear. But I've also currently scattered a lot of $FlowFixMes there. Just want to check the approach first.

The JS compiler uses the same types as the runtime. Those types are all currently read-only, exact objects. This makes it a little difficult to "build" the fields / nodes

```
const field = {
    alias: null // works with read-only
}
if (needs_alias){
    field.alias = alias // needs to be writable now
}
```

There's a few options here:
 - Move the `read-only` from field variance (`+alias: ?string`) to the utility `$ReadOnly<>`.
 - Move all the `$ReadOnly` annotations off the shared types and into the runtime (so the runtime can't write fields, but the shared types become writable for `relay-compiler` to use)
 - Use `$Shape<>` to build the objects in `relay-compiler` and then assert their output types after building them (don't let non-exact or writable types escape the functions in ReaderCodeGenerator, NormalizationCodeGenerator, but treat them as in-exact / writable within).
 - Use object spreads (as some fields like handleArgs already do) and maybe stableSort the snapshots so the fields output consitently

### Questions
This is a large PR right now. I'll break this down into clearer steps in order to ship it, but I wanted feedback before I go ahead with that. Questions:
 - Are you happy to accept a contribution to remove these `null`s from the metadata sent to the client?
 - Which fields does this make sense for? I.e. `plural` might be stretching it a bit far.
 - Would you like me to feature flag this?
 - Any suggestions on how you'd like the Normalization* and Reader* types to look if I change them to be more amenable to allowing field writing in the JS compiler.

### DIFF
--- a/compiler/crates/relay-codegen/src/ast.rs
+++ b/compiler/crates/relay-codegen/src/ast.rs
@@ -73,6 +73,9 @@ pub enum Primitive {
     RawString(String),
     GraphQLModuleDependency(StringKey),
     JSModuleDependency(StringKey),
+    // Don't include the value in the output when
+    // skip_printing_nulls is enabled
+    NullOrNothing,
 }
 
 impl Primitive {
--- a/compiler/crates/relay-codegen/src/printer.rs
+++ b/compiler/crates/relay-codegen/src/printer.rs
@@ -192,6 +192,7 @@ pub struct JSONPrinter<'b> {
     eager_es_modules: bool,
     js_module_format: JsModuleFormat,
     top_level_statements: &'b mut TopLevelStatements,
+    skip_printing_nulls: bool,
 }
 
 impl<'b> JSONPrinter<'b> {
@@ -207,6 +208,10 @@ impl<'b> JSONPrinter<'b> {
             builder,
             js_module_format: project_config.js_module_format,
             eager_es_modules: project_config.typegen_config.eager_es_modules,
+            skip_printing_nulls: project_config
+                .feature_flags
+                .skip_printing_nulls
+                .is_fully_enabled(),
         }
     }
 
@@ -306,6 +311,10 @@ impl<'b> JSONPrinter<'b> {
                     let next_indent = indent + 1;
                     f.push('{');
                     for ObjectEntry { key, value } in object {
+                        match value {
+                            Primitive::NullOrNothing if self.skip_printing_nulls => continue,
+                            _ => (),
+                        }
                         f.push('\n');
                         print_indentation(f, next_indent);
                         write!(f, "\"{}\": ", key).unwrap();
@@ -333,6 +342,10 @@ impl<'b> JSONPrinter<'b> {
                     f.push('[');
                     let next_indent = indent + 1;
                     for value in array {
+                        match value {
+                            Primitive::NullOrNothing if self.skip_printing_nulls => continue,
+                            _ => (),
+                        }
                         f.push('\n');
                         print_indentation(f, next_indent);
                         self.print_primitive(f, value, next_indent, is_dedupe_var)
@@ -356,7 +369,7 @@ impl<'b> JSONPrinter<'b> {
         is_dedupe_var: bool,
     ) -> FmtResult {
         match primitive {
-            Primitive::Null => write!(f, "null"),
+            Primitive::Null | Primitive::NullOrNothing => write!(f, "null"),
             Primitive::Bool(b) => write!(f, "{}", if *b { "true" } else { "false" }),
             Primitive::RawString(str) => {
                 f.push('\"');
@@ -545,7 +558,7 @@ fn write_constant_value(f: &mut String, builder: &AstBuilder, value: &Primitive)
                 }
             }
         }
-        Primitive::Null => {
+        Primitive::Null | Primitive::NullOrNothing => {
             f.push_str("null");
             Ok(())
         }
--- a/compiler/Cargo.lock
+++ b/compiler/Cargo.lock
@@ -791,9 +791,9 @@ dependencies = [
 
 [[package]]
 name = "indexmap"
-version = "1.7.0"
+version = "1.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bc633605454125dec4b66843673f01c7df2b89479b32e0ed634e43a91cff62a5"
+checksum = "282a6247722caba404c065016bbfa522806e51714c34f5dfc3e4a3a46fcb4223"
 dependencies = [
  "autocfg",
  "hashbrown",
--- a/compiler/crates/common/src/feature_flags.rs
+++ b/compiler/crates/common/src/feature_flags.rs
@@ -49,6 +49,9 @@ pub struct FeatureFlags {
 
     #[serde(default)]
     pub enable_provided_variables: FeatureFlag,
+
+    #[serde(default)]
+    pub skip_printing_nulls: FeatureFlag,
 }
 
 #[derive(Debug, Deserialize, Clone, Serialize)]
--- a/compiler/crates/relay-codegen/src/build_ast.rs
+++ b/compiler/crates/relay-codegen/src/build_ast.rs
@@ -190,7 +190,7 @@ impl<'schema, 'builder> CodegenBuilder<'schema, 'builder> {
                     &fragment.used_global_variables),
             kind: Primitive::String(CODEGEN_CONSTANTS.fragment_value),
             metadata: if skip_metadata {
-                    Primitive::Null
+                    Primitive::NullOrNothing
                 } else {
                     self.build_fragment_metadata(fragment)
                 },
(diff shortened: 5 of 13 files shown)


josephsavona:
Thanks for the PR! We have also considered this as a potential optimization. The reason we haven't shipped this yet though is a _hunch_ that this will not play well with JavaScript engines, which depend quite a bit on hidden classes for performance. The current output ensures that each node kind has a stable set of fields and field ordering, which means that each "instance" of a given node kind will (should) always get the same hidden class. This in turn allows the engine to more effectively monomorphize (specialize) code consuming those objects, ie Relay Runtime). As I mentioned though this is definitely a hunch - we don't currently have an easy way to test the performance impact of changes to the AST format at a scale representative of our production workloads. We're currently in the middle of building out a benchmark suite to help answer questions like this though.

In terms of next steps: we're supportive of adding an off-by-default _option_ to the compiler that enables omitting null fields. We are not actively using the JavaScript compiler, so our recommendation would be to make this change only in the Rust compiler, which we are hoping to make a supported option in OSS relatively soon. Thoughts?

josephsavona:
In terms of next steps on implementation:
* Revert changes to the JS compiler (or move them to a later PR, lets start with Rust)
* Make this a [compiler feature flag](https://github.com/facebook/relay/blob/main/compiler/crates/common/src/feature_flags.rs#L16) that defaults to off. Note that this will allow you to revert tests, since by default nothing is getting the behavior.
* Add a new test that enables the feature, with a small selection of fixtures.

This should dramatically reduce the size of the PR :-)

alunyov:
RE: JS Changes. I think we can just ignore JS compiler at this point. The release of Rust-based version is coming soon: https://github.com/facebook/relay/issues/3180

tomgasson:
Thanks for the feedback.

I've updated this PR to only make this change to the rust compiler. Feature flagged under `enable_reduced_metadata`.

I agree this might be a bytesize vs runtime tradeoff -  interested to see the result.



Nice to hear the new compiler is getting close to release. Some small contribution meta:
 - It's not clear what rust version I should be on (so my Cargo.lock changes are probably not correct here)
 - It's not clear if I should `fmt` files as I touch them
 - `realpath` doesn't exist on macos. I had to `brew install coreutils` which might want to be documented, but could alternatively change `scripts/compile-tests.sh` to not rely on it

alunyov:
Thanks for updates to the PR @tomgasson and for contributors feedback!

I think we'll need to figure out some of these steps (as we're don't really know some of them): correct rust version, fmt, running tests, etc. I've a task for that to update the contributors guide to reflect some of these steps: https://github.com/facebook/relay/issues/3630

alunyov (approved):

compiler/crates/relay-codegen/src/printer.rs
```diff
@@ -27,16 +28,18 @@ pub fn print_operation(
     schema: &SDLSchema,
     operation: &OperationDefinition,
     js_module_format: JsModuleFormat,
+    enable_reduced_metadata: &FeatureFlag,
 ) -> String {
-    Printer::without_dedupe(js_module_format).print_operation(schema, operation)
+    Printer::without_dedupe(js_module_format).print_operation(schema, operation, enable_reduced_metadata)
```
alunyov, line 33: can we add this flag to the `Printer` itself, so we do not pass it around all methods?

compiler/crates/relay-compiler/src/build_project/artifact_content.rs
```diff
@@ -436,7 +437,7 @@ fn generate_split_operation(
         &mut content,
         "node",
         "NormalizationSplitOperation",
-        &printer.print_operation(schema, normalization_operation),
+        &printer.print_operation(schema, normalization_operation, &project_config.feature_flags.enable_reduced_metadata),
```
alunyov, line 440: maybe we can create an instance `printer` with the `enable_reduced_metadata` option, so we don't have to pass it around?

alunyov:
Hey @tomgasson sorry for delay with this. So, I could not land this internally, some tests were failing, and once I start looking into these we have a big refactoring with the `intern` package.

I think we can still add this PR, can you try rebasing it?

tomgasson:
Sorry for the delay on coming back to this. Had to complete our move to the move compiler first.


I've taken a slightly different approach now that a few things have changed like the config (and therefore flags) being easier to access.

* Add a feature flag `enable_reduced_metadata`
* Added a `Primitive::Elide` type to indicate that the value should be removed rather than output `null`
* Filter out `Elide` values from arrays and objects when printing and the feature is enabled (otherwise continues to output `null` as before)
* Added a snapshot test of this feature

packages/relay-runtime/store/__tests__/__generated__/RelayModernEnvironmentExecuteSubscriptionWithStreamTestCommentCreateSubscription.graphql.js
```diff
@@ -176,9 +176,7 @@ return {
   "params": {
     "cacheID": "c13dedba1d0c8b17a192b735d141de99",
     "id": null,
-    "metadata": {
```
alunyov, line 179: Hmm, why this has changed?

compiler/crates/relay-codegen/src/ast.rs
```diff
@@ -69,6 +69,7 @@ pub enum Primitive {
     Int(i64),
     Bool(bool),
     Null,
+    Elide,
```
alunyov, line 72: Oh, love this idea!

This is super minor nit: I wonder if there is a different name for this: `MaybeNull`, `NullOrNothing`, `OptionalNull`? 

As an ESL person, I have to look up this word, and still not entirely sure what it means in my language, but that's me. Great word!

packages/relay-runtime/store/__tests__/__generated__/RelayModernEnvironmentExecuteSubscriptionWithStreamTestCommentCreateSubscription.graphql.js
tomgasson: `scripts/compile-tests.sh` was broken on `main` for me.

I added `"enable_provided_variables": {"kind": "enabled"}` to fix, and running it generated this change.

(maybe I was doing something wrong?)

compiler/crates/relay-codegen/src/printer.rs
```diff
@@ -192,6 +192,7 @@ pub struct JSONPrinter<'b> {
     eager_es_modules: bool,
     js_module_format: JsModuleFormat,
     top_level_statements: &'b mut TopLevelStatements,
+    elide: bool,
```
alunyov, line 195: nit: maybe a more descriptive name? `should_skip_printing_nulls`?

alunyov (changes requested):
Thank you for the follow up.

Can we figure out what happened with test artifacts? Is it something we need to fix separately? I don't think changes to `metadata` in generated artifacts are expected here.

compiler/crates/relay-codegen/src/printer.rs
tomgasson: 👍 . I might make the feature flag name `skip_printing_nulls` so it's consistent with this too

tomgasson:
Fixed nits, and removed the commit which allowed `scripts/compile-tests.sh` to work on master (I needed it to manually check this worked, but it's not related to this change)

josephsavona:
@tomgasson Thank you for the contribution! I'm really excited that we can provide this as an option.

PR #3633, opened 2021-11-11
[relay-compiler] Inline test-utils in JS relay compiler, update dependencies in test-utils/relay-compiler pacakges  

alunyov:
- This diff removed internal dependencies on `relay-compiler` and `graphql` in Relay Test Utils and Internal Test Utils.
- TestSchema/parseGraphQLText added to `relay-compiler`. We eventually remove them, once rust-version is fully released
- `relay-compile` now depends on `graphql@15.3.0` (not as `peerDependency`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "fbjs": "^3.0.0",
     "flow-bin": "^0.163.0",
     "glob": "^7.1.1",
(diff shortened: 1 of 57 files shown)


maraisr:
So good, was just about to raise this issue!! haha

PR #3638, opened 2021-11-13
Allow `Node` interface `id` field to be configurable.

kesne:
This adds a new option to the Rust compiler, `nodeInterfaceIdField`, which allows customization of the `Node` interface's `id` field. I initially tried to determine this via inference, but it's entirely possible that the `Node` interface has multiple `id` fields, so I figured explicitly specifying it made more sense.


orta:
( +1-ing the desire for this, lots of OSS JS/graphQL tooling takes over `id` or makes it hard to get the right table to query for an `id`. At Artsy there was so much interacting with legacy systems we had `id` (table), `_id`(db) and _then_ `__id` which was finally the one we could use for GQL object identification  )

maraisr:
We'd also need to let the relay-runtime know about that property, as so much hangs on that ID field 😅

kesne:
@maraisr From what I can tell, `getDataID` should be able to be used to accomplish that: https://github.com/facebook/relay/blob/38f1c96e7696378b55f68509a320a36846457bbe/packages/relay-runtime/store/RelayModernEnvironment.js#L79

maraisr:
Just a ripe ol pain, coz in places where there is no id field, relay still tries to normalise like serialising some input arguments and such. I've not looked too close, but would suspect this method to have to implement all of those avenues as well?

compiler/crates/relay-compiler/src/compiler_state.rs
```diff
@@ -341,7 +341,7 @@ impl CompilerState {
                 .any(|sources| !sources.processed.is_empty())
     }
 
-    fn is_change_safe(&self, sources: &SchemaSources) -> bool {
+    fn is_change_safe(&self, sources: &SchemaSources, project_config: &ProjectConfig) -> bool {
```
alunyov, line 344: can we add `node_interface_id_field` to the compiler_state itself?

compiler/crates/relay-compiler/src/compiler_state.rs
```diff
@@ -363,7 +363,7 @@ impl CompilerState {
                 &current,
                 &Vec::<(&str, SourceLocationKey)>::new(),
             ) {
-                Ok(schema) => schema_change.is_safe(&schema),
+                Ok(schema) => schema_change.is_safe(&schema, project_config.node_interface_id_field),
```
alunyov, line 366: then here we can pass `self. node_interface_id_field`

compiler/crates/relay-compiler/src/config.rs
```diff
@@ -843,6 +851,9 @@ struct ConfigFileProject {
 
     #[serde(default)]
     js_module_format: JsModuleFormat,
+
+    #[serde(default)]
+    pub node_interface_id_field: Option<StringKey>,
```
alunyov, line 856: can we add comment here, and also to the compiler README.md that explains the purpose of this field

alunyov: Also, I think this is something from the internal discussion about this PR.

I think we should add this field to` FeatureFlags` here:
https://github.com/facebook/relay/blob/59c9bc01c1d215b75d6c683f6ee1074a8110fd7b/compiler/crates/common/src/feature_flags.rs

And this PR (or this can be a separate PR) should just add `feature_flags` to SingleProjectConfigFile.

compiler/crates/relay-transforms/src/apply_transforms.rs
```diff
@@ -130,6 +134,7 @@ fn apply_common_transforms(
     program: Arc<Program>,
     connection_interface: &ConnectionInterface,
     feature_flags: Arc<FeatureFlags>,
```
alunyov, line 136: the benefits of adding `node_interface_id_field` to `FeatureFlags` is that you already have them in `apply_transforms` so it's less changes.

alunyov (changes requested):
Thanks for adding this! Sorry, we dropped the ball on this PR and did not review it earlier.

The main feedback is that let's try re-using `FeatureFlags` for this `id` field configuration.

compiler/crates/relay-compiler/src/config.rs
kesne: I can make the change to move this to feature flags. I didn't originally since it seems like all of the feature flag config was resolved to boolean values, whereas this is a string configuration value, which didn't seem to me like a feature flag.

alunyov: > this is a string configuration value, which didn't seem to me like a feature flag.

Fair point. 

I general, I think this `field` is similar with `ConnectionInterface`. I think both of these should be defined on the `project_config` here:

https://github.com/facebook/relay/blob/59c9bc01c1d215b75d6c683f6ee1074a8110fd7b/compiler/crates/relay-compiler/src/config.rs#L252-L276

Something like `schema_configuration` field which is a struct:

```rust
struct SchemaConfiguration {
  connection_interface: ConnectionInterface,
  node_interface_id_field: Option<StringKey>,
}
```

But this refactoring can be done independently from this PR.

alunyov: or `schema_config` for short, as we already have `typegen_config`.

compiler/crates/relay-compiler/src/compiler_state.rs
kesne: I tried to do this, but the config that's used to create the compiler_state is the root config, not the project config, so this seemed challenging to try to move.

compiler/crates/relay-compiler/src/config.rs
kesne: @alunyov I moved it to a FeatureFlag, just so you can see what that would be like. Let me know which approach you'd like to move forward with!

alunyov: Thanks @kesne! I importing this internally.

alunyov:
So, moving this to `feature_flags` didn't make it cleaner.

And I totally agree with you  @kesne that it doesn't seem right to have these settings in the feature_flags. All these suggestions were with the idea that we somehow add this `id_field_name` without passing this individual property around too much.

One problem was, we could not use `ProjectConfig` type in the `relay-transforms`, as it was defined in the `relay-compiler` crate. I've moved `ProjectConfig` to a separate crate so we can use this type both in `relay-compiler` and in `relay-transforms`.  See: https://github.com/facebook/relay/commit/5baad8ca2147bb494532d527fe8a9797ba7802ae

I'm working on updates to `apply_transform` that will make them accept `project_config`, these change should land soon.

I will also add new property on the `ProjectConfig` - `schema_config` where we can put `node_interface_id_field`.

alunyov:
The changes to the `project_config` landed. I think we can add the `id_field_name` to the `schema_config` here:

https://github.com/facebook/relay/blob/main/compiler/crates/relay-config/src/project_config.rs#L42-L44

alunyov:
A quick update. I think there will be another iteration on top of the `schema_config`. We need it also here: https://github.com/facebook/relay/pull/3121#discussion_r771476496 - so I think I'll add it directly to `program`: it would be easier to access these configs in various transforms.

Lalitha-Iyer:
@alunyov There could be pagination bug Related to this pr. If we configure the compiler to use a custom id, I found pagination `loadNext` fails at runtime, as it queries with node.id and not the custom configured node id.  
Looking at the source we probably need to change ` paginationVariables.id` to  `paginationVariables.identifierField` 
https://github.com/facebook/relay/blob/main/packages/react-relay/relay-hooks/useLoadMoreFunction.js#L212

Let me know if I should have a more sandboxed example and a new issue.

Emilios1995:
> @alunyov There could be pagination bug Related to this pr. If we configure the compiler to use a custom id, I found pagination `loadNext` fails at runtime, as it queries with node.id and not the custom configured node id. Looking at the source we probably need to change ` paginationVariables.id` to `paginationVariables.identifierField` https://github.com/facebook/relay/blob/main/packages/react-relay/relay-hooks/useLoadMoreFunction.js#L212
> 
> Let me know if I should have a more sandboxed example and a new issue.

@Lalitha-Iyer I opened an [issue](https://github.com/facebook/relay/issues/3897) for this already. Sadly I haven't had time to debug and the issue hasn't received much attention.

Lalitha-Iyer:
@Emilios1995 Appreciate you opening an issue. I will follow along.

Issue #3473, opened 2021-04-27
[rust] relay-compiler to support template literals across lines.


maraisr:
with the new rust compiler if a template literal spans more than one line it will refuse to output artifacts—this is prevliant when used with Prettier that decides its own formatting:

```ts
// ❌ No produced documents
const Thing = ({ search: searchRef }) => {
	const search = useFragment(
		graphql`
			fragment search_data on SearchItem {
				name
			}
		`,
		searchRef,
	);

	return <h1>{ search.name } < /h1>;;
};

// ✅ Produces documents
const Thing = ({ search: searchRef }) => {
	const search = useFragment(
		graphql`fragment search_data on SearchItem {
				name
			}`,
		searchRef,
	);

	return <h1>{ search.name } < /h1>;;
};
```

Like i suggested, this is more than okay to enforce — but with a lack of control to when prettier does stuff it makse it hard to enforce that?

---

More than happy to help rectify this, but will need to just be pointed at the correct files to go to impl this.

josephsavona:
Hmm, that is not the case internally so I don't believe this is a general issue with the new compiler. Are you using Flow or TypeScript (or JS)?

maraisr:
I'm using typescript yeah — I'll do some further digging locally and try find a root cause!

kassens:
To avoid full JS parsing including different variants like TypeScript, Flow or latest babel features, we do a simple "tokenization". Internally we've seen 2 or 3 failures across many many files. Cases like "graphql`" appearing in another string might trip up the analyzer.

Search for "`" in your code and remove that part to find the offending part in the file. Might be useful to get a repro, but it's hard :( So far we've gotten away with this pretty successfully. Maybe your or TypeScript have different patterns that we could potentially encode.

Code for that is here: https://github.com/facebook/relay/blob/f73417979344c9e5b61dbabcc80c0414e5d495e9/compiler/crates/extract-graphql/src/lib.rs#L71

maraisr:
Yeah can surely understand "my code broken fix it" not right! Thanks for the tips, will keep at it! 👌🏻 can probably close this issue, and will reopen if i find something! Thanks people 🥰

maraisr:
Going to reopen this — am in a brand new space, and issue still there: Have a check at => 
https://github.com/maraisr/meros/tree/main/examples/relay-with-helix

and run `pnpm format` in the root of the monorepo.

PR #3643, opened 2021-11-15
Support tab-indented queries

### LINKED ISSUE #3642: Rust compiler does not parse tab-indented queries
When using the new rust compiler, it looks like it fails to parse any queries that have leading tabs before the operation type. It looks like newlines and spaces are handled fine though.

For example, this results in the Relay compiler generating no code:
```js
graphql`
	query AppQuery {
		node(id: "test") { id }
	}
`;
```

Meanwhile if we just dedent before the operation type, the compiler will pick it up fine:
```js
graphql`
query AppQuery {
		node(id: "test") { id }
	}
`;
```

kesne:
I noticed that the tab character was not being handled when parsing files, causing it to immediately exit and not parse any operations.

This fixes #3642.

### DIFF
--- a/compiler/crates/extract-graphql/tests/extract/fixtures/tabbed.js
+++ b/compiler/crates/extract-graphql/tests/extract/fixtures/tabbed.js
@@ -1,0 +1,12 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+graphql`
+	fragment Test on User {
+		name
+	}
+`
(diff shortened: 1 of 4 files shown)


maraisr:
Probably even fixes https://github.com/facebook/relay/issues/3473

alunyov (changes requested):
Thank you for fixing this!

Can you add a test for this case?

compiler/crates/extract-graphql/tests/extract/fixtures/tabbed.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+graphql`
+	fragment Test on User {
```
kesne, line 9: This is one of the few times I'm glad Github displays tabs super wide, makes it obvious this is a tab character! 😄

PR #3644, opened 2021-11-15
Correctly resolve module names for TypeScript index files

kesne:
I noticed that the module name resolution had hard-coded the assumption that the `index` files would have a `js` extension. This updates this to include checks for every extension that the compiler currently supports (`js`, `jsx`, `ts`, `tsx`).


maraisr (commented):
👏🏻

josephsavona:
This makes sense, thanks for the contribution!

PR #3646, opened 2021-11-16
Check for existence of `global` before usage

rrdelaney:
An unchecked usage of `global` is causing failures in environments where it may not be polyfilled (see https://github.com/vercel/next.js/issues/30968). This is fixed by adding a check for `global` in the same fashion [`registerEnvironmentWithDevTools.js`](https://github.com/facebook/relay/blob/87696f9cbf61739cb7474be447cf1d88e6eec0c8/packages/relay-runtime/util/registerEnvironmentWithDevTools.js#L21-L26) already does. The tests seem to still run fine after this, and the compiled code still references `global` so it should be a safe change 😄


kesne:
I imagine it'd make sense to also check the standard `globalThis` property here?

packages/relay-runtime/store/RelayPublishQueue.js
```diff
@@ -57,8 +57,15 @@ type PendingUpdater = {|
   +updater: StoreUpdater,
 |};
 
+const _global =
+  typeof global !== 'undefined'
+    ? global
+    : typeof window !== 'undefined'
+    ? window
+    : undefined;
```
maraisr, line 65: i feel like this is used in quite a number of other places, and very idiomatic. 

I suspect, just `global` as it is is fine, and handle this in your own eco with webpack overrides or babel plugins.

rrdelaney: Just checking `global` without testing for existence breaks in non-Node.js environments. In this case Relay is broken on Vercel's new runtime that doesn't use Node.js and has no access to `global`. Using `globalThis` would fix the issue as well but I'm not sure that would have been a simple direct swap. This assignment to `_global` is copied directly from another place in the Relay codebase that also looks for global variables.

Regarding handling this in Webpack or Babel plugins: that would require additional processing to `node_modules` which can be tricky to set up in metaframeworks. In the case of `relay-nextjs` I would also have to add instructions on how to do that in the onboarding guide, which adds another step to an already lengthy process.

rrdelaney:
@kesne Could definitely add `globalThis` if requested, I went with a simple approach and copied some other code I knew would work in this codebase and kept the changes minimal.

packages/relay-runtime/store/RelayPublishQueue.js
shuding: My 2 cents here: many environments such as Web Worker don't have `window` or `global`. And there're also non-Node.js and non-browser environments like RN and Deno. I think using the `globalThis` standard will make it a lot easier.

One potential problem is, `globalThis` was introduced since Node.js 12 (same for browsers). But it should be good if we check it with `typeof`.

rrdelaney: @shuding I think replacing all usages of `global` in the Relay codebase might be out of scope of this PR 😅 I agree that it's probably better but there's a good chance that a flat-out replacement may not work in FB's internal system.

rrdelaney:
Just a quick ping on this 🙂 Also would like to know if `globalThis` can be used here. I think it would be a simple swap but there may be other factors in play.

shicholas:
is there anything non-fb employees can do here since we don't have access to your CI tool? thank you

alunyov:
There are some weird flow errors that surfaced internally. I'll take a look at this again

PR #3654, opened 2021-11-25
Change 'three' assumptions to 'two'

jonrose-dev:
Seems like a typo. Only two assumptions are enumerated in the list above


alunyov (changes requested):
Thank you!

Can you also update the source here: https://github.com/facebook/relay/blob/main/website/docs/guides/graphql-server-specification.md

jonrose-dev:
@alunyov Changes made 👍

alunyov (approved):

PR #3656, opened 2021-11-28
Adds a README to relay-config

orta:
I couldn't find any docs on how to use relay-config, or what config fields supported - so I figured I'd add them to the README as that was the first place I looked.


maraisr:
We should probably delete this package now? It has no more relevancy

orta:
It does? It works for me or was support upstreamed into the compiler

alunyov (approved):

PR #3657, opened 2021-11-28
Adds the TypeScript setup to the search index

orta:
Today a search in the site for "TypeScript" doesn't give super useful results. This PR should make it so that 'typescript language plugin' will get added to the aloglia search index and people setting up a TS environment will be able to use search to figure out where to go from the 'get started' path


maraisr:
The legacy js compiler has been deprecated; and thus the typescript language plugin along with it. 

I'd say, we shouldn't be pushing people there. Rather we should update the typescript docs around the new compiler. `language: typescript` and done. Just so easy now.

Obviously; there is a slight bug with latest compiler, but that aside point remains. Put efforts into directing people to new/current things.

orta:
Agree, closing

PR #3658, opened 2021-11-28
Adds Viewer as a TODO in the glossary

orta:
I have my own sense of it (in Artsy we didn't necessarily use `viewer` if I recall) but it's a good item for the glossary IMO


rbalicki2:
Awesome! I'll import this and write up a blurb about the viewer.

PR #3666, opened 2021-12-01
Add a JSON file query persister strategy to relay-compiler

vincentriemer:
**NOTE**: This is essentially the first time I've developed in Rust so feel free to eviscerate it because...

I have no idea what I'm doing

# Summary

Previously in the new rust relay-compiler your only option for persisting queries is to use the remote persister which makes POST requests to a given endpoint with the operation text to be persisted. This PR adds the option to instead persist your queries to a local JSON file in a very similar way to how the JS compiler originally did it.

If you had a basic remote persister config like this:

```js
persistConfig: {
  url: "http://localhost:3000",
  params: {}
}
```

changing it to a local JSON file persister means updating the config to:

```js
persistConfig: {
  filePath: "./queries.json"
}
```

When this is enabled and a new query is persisted the compiler will first attempt to read the local file if it exists as a hash map (creating a new one if there is no existing data), hashes the operation text, adds the operation text to the hashmap keyed by the hashed version of the operation, and then serializes & writes the resulting map to the configured path.

# Test Plan

Leveraging my own [simple repo](https://github.com/vincentriemer/new-relay-compiler-issues-repro) I've been using for reproducing issues with the new relay compiler, installed the relay-compiler to a version including this PR's changes, and updated the `relay.config.js` file as such:

```diff
module.exports = {
  "src": "./src",
  "schema": "./schema.graphql",
  "artifactDirectory": "./src/__generated__",
  "language": "typescript",
  "persistConfig": {
-    "url": "http://localhost:3000",
-    "params": {}
+    filePath: "./queries.json"
  }
};
```

Then after running the relay compiler I verified that it created a `queries.json` file and it contained the expected persisted query json output:

```json
{
  "00dc69b2cfc4c1da3528e9aaf65a05091740fc71": "query AppQuery {\n  viewer {\n    submission(name: \"t3_qvyl2j\") {\n      ...Post_submission\n      id\n    }\n  }\n}\n\nfragment Post_submission on Submission {\n  title\n  author\n  selftext {\n    html\n  }\n}\n"
}
```

### DIFF
--- a/compiler/crates/relay-compiler/src/config.rs
+++ b/compiler/crates/relay-compiler/src/config.rs
@@ -12,6 +12,8 @@ use crate::build_project::{
 };
 use crate::compiler_state::{ProjectName, SourceSet};
 use crate::errors::{ConfigValidationError, Error, Result};
+use crate::remote_persister::{RemotePersistConfig, RemotePersister};
+use crate::file_persister::{FilePersistConfig, FilePersister};
 use crate::saved_state::SavedStateLoader;
 use crate::status_reporter::{ConsoleStatusReporter, StatusReporter};
 use async_trait::async_trait;
@@ -88,9 +90,6 @@ pub struct Config {
     pub saved_state_config: Option<ScmAwareClockData>,
     pub saved_state_loader: Option<Box<dyn SavedStateLoader + Send + Sync>>,
     pub saved_state_version: String,
-
-    /// Function that is called to save operation text (e.g. to a database) and to generate an id.
-    pub operation_persister: Option<Box<dyn OperationPersister + Send + Sync>>,
 
     pub post_artifacts_write: Option<PostArtifactsWriter>,
 
@@ -202,6 +201,14 @@ impl Config {
             projects,
             ..
         } = config_file;
+
+        let config_file_dir = config_path.parent().unwrap();
+        let root_dir = if let Some(config_root) = config_file.root {
+            config_file_dir.join(config_root).canonicalize().unwrap()
+        } else {
+            config_file_dir.to_owned()
+        };
+
         let projects = projects
             .into_iter()
             .map(|(project_name, config_file_project)| {
@@ -245,6 +252,19 @@ impl Config {
                         }],
                     })?;
 
+                let operation_persister: Option<Box<dyn OperationPersister + Send + Sync>> =
+                    match config_file_project.persist {
+                        Some(config) => match config {
+                            PersistConfig::Remote(remote_config) => {
+                                Some(Box::new(RemotePersister::new(remote_config)))
+                            },
+                            PersistConfig::File(file_config) => {
+                                Some(Box::new(FilePersister::new(file_config.file_path, root_dir.clone())))
+                            }
+                        },
+                        _ => None,
+                    };
+
                 let project_config = ProjectConfig {
                     name: project_name,
                     base: config_file_project.base,
@@ -256,7 +276,7 @@ impl Config {
                     shard_strip_regex,
                     schema_location,
                     typegen_config: config_file_project.typegen_config,
-                    persist: config_file_project.persist,
+                    operation_persister: operation_persister,
                     variable_names_comment: config_file_project.variable_names_comment,
                     extra: config_file_project.extra,
                     test_path_regex,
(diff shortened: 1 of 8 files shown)


alunyov:
Wow! This is great @vincentriemer! As for the first time rust contribution you very much know what you're doing!

I think we also need to test the case with `--watch` mode. I think we may be adding more queries to this file, essentially for every intermediate, but valid change in the watch mode these queries would get persisted to this file.

Maybe we also need to pass CLI flags to FilePersister - and error if `--watch` flag is provided (as a simple workaround)?

compiler/crates/relay-compiler/src/config.rs
```diff
@@ -845,14 +860,11 @@ struct ConfigFileProject {
...
-#[derive(Debug, Serialize, Deserialize)]
-#[serde(deny_unknown_fields)]
-pub struct PersistConfig {
-    /// URL to send a POST request to to persist.
-    pub url: String,
-    /// The document will be in a POST parameter `text`. This map can contain
-    /// additional parameters to send.
-    pub params: FnvIndexMap<String, String>,
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(untagged)]
```
alunyov, line 864: I like this `serde` feature, the only problem with it: obscure error messages if both/mixed values are provided.

compiler/crates/relay-compiler/src/remote_persister.rs
```diff
@@ -5,21 +5,40 @@
...
+
+pub struct RemotePersister {
+    config: RemotePersistConfig,
+}
+
+impl RemotePersister {
+    pub fn new(config: RemotePersistConfig) -> Self {
+        Self { config }
+    }
+}
```
alunyov, line 35: Nice!

alunyov (approved):
Fill free to import this to fbsource, I think there also maybe some internal formatting issues that needs to be fixed.

But this looks great!

josephsavona:
I just quickly skimmed the code but what is the strategy for handling the fact that queries are persisted concurrently?

alunyov:
Also, alternative approach is to implement this persisting the way you did here: https://github.com/vincentriemer/new-relay-compiler-issues-repro/blob/main/mockQueryPersistServer.js

But just make this `mockQueryPersistServer` store the map in memory, and flush it to the file system (periodically, maybe)?

vincentriemer:
> I just quickly skimmed the code but what is the strategy for handling the fact that queries are persisted concurrently?

I'm sorry I'm a JS dev what does "concurrently" mean \s 

> But just make this `mockQueryPersistServer` store the map in memory, and flush it to the file system (periodically, maybe)?

But more seriously: I think this is the right strategy and I'll update the PR to do this — make it so that it reads the existing file once at initialization time and keeps the persisted query map in memory, flushing back to the file only when a new query is added.

wyattjoh:
It seems as though the hashing mechanism here is locked to SHA1:

https://github.com/facebook/relay/blob/c955577147d75bdda386dea9f973af24ae19bdf6/compiler/crates/relay-compiler/src/file_persister.rs#L35-L39

Our team has been looking to utilize the generated ID's to use with Apollo Server's Automatic Persisted Queries (or [APQ](https://www.apollographql.com/docs/apollo-server/performance/apq/)), and they stipulate a required SHA256 hash.

As it seems y'all are already changing this from MD5 to SHA1, is it possible to either:

1. Change this to SHA256 to support APQ directly
2. Support changing the hashing algorithm directly using this new `FilePersister` config

janicduplessis:
Tested this and works well! Needed to make some changes to fix conflicts (https://github.com/janicduplessis/relay/commit/ae1f7830b9efaf48322802fa183fc0170b3283ce).

This would help a lot migrating from the js compiler. There is actually a lot of complexity that goes into running an http server just to generate a queries json file, especially if you want to abstract away starting the server in background (start server in bg, wait for it to start before running compiler, close server once compiler done). Also making the server somewhat performant by batching writes.

tbezman:
@vincentriemer My team could definitely benefit from this. Are you going to continue work or should I drive this to completion?

alunyov:
@tbezman @vincentriemer yes, we will include these changes to one of the upcoming releases

PR #3667, opened 2021-12-01
Add MacOS ARM64 build target to relay-compiler for better M1 Mac support

vincentriemer:
idk what the performance difference is since I don't have a project large enough where I can notice a difference but I figured there's no harm in at least providing an M1 native executable for the new relay-compiler.

Verified on my M1 MacBook Pro by manually downloading the new macOS-arm64 artifact built from the GH workflow and replacing a local executable which still runs successfully.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,7 +90,7 @@ jobs:
             os: macos-latest
             build-name: relay
             artifact-name: relay-compiler-macos-arm64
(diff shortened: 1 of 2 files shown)


alunyov:
Thanks @vincentriemer! I think it's fine to add. 

The only concern I can think of is the size of the package itself. Because when we'll download this form NPM we will download all 4 files now.

PR #3676, opened 2021-12-06
Updated fragments.md

apremalal:
Removed duplicated word(is)


alunyov (changes requested):
Thank you,

Can you also update the source?
https://github.com/facebook/relay/blob/main/website/docs/guided-tour/rendering/fragments.md

PR #3677, opened 2021-12-06
Remove relay-conifg

alunyov:
This package were used with in v12 with `relay-compiler`. Now we have this config parser (or similar) implemented in rust. So we don't need this.

Test Plan:
```
yarn
```


orta:
Can you ship an npm release before this is removed, it's still empty and people won't know that it isn't valid for 13+ - https://www.npmjs.com/package/three-svg-loader

alunyov:
@orta do you mean ship README changes to NPM?

orta:
Yeah,  https://github.com/facebook/relay/pull/3656 added a README but never got deployed to npm

maraisr:
Why bother? There is zero support for these libraries anymore. Wether they have a readme or not 🤷🏼‍♂️

alunyov:
Oh, I'll try to publish a custom 12.0.1 tomorrow.

But what's interesting all our packages are missing README on NPM, so there's probably something we need to fix in our bundler/publish script to include these files.

orta:
ah, that makes sense - ha

alunyov:
I published 12.0.1 with README https://www.npmjs.com/package/relay-config

orta:
Ace, thanks!

PR #3684, opened 2021-12-10
Watch js,jsx files with watchman with typescript

robrichard:
Updates the watchman query builder to match what the glob file source already does: https://github.com/facebook/relay/blob/main/compiler/crates/relay-compiler/src/file_source/glob_file_source.rs#L49-L50

It's useful in projects that are not 100% typescript


maraisr:
Could the compiler just always watch for both?

alunyov (approved):

Issue #3686, opened 2021-12-12
mockClear does not clean the recordsource


artola:
The `environment` returned by `createMockEnvironment` has a method `mockClear` which does not reset the recordsource to the [initial state](https://github.com/facebook/relay/blob/5c1045dcc05f522644fa5079ba5880d169546357/packages/relay-test-utils/RelayModernMockEnvironment.js#L185).
`mockClear` does care about cleaning the [`cache`](https://github.com/facebook/relay/blob/5c1045dcc05f522644fa5079ba5880d169546357/packages/relay-test-utils/RelayModernMockEnvironment.js#L569) but does not the same with the recordsource. Specially this difference between them it seems to me just like missing.

```javascript
import {createMockEnvironment} from 'relay-test-utils';

const environment = createMockEnvironment();

afterEach(() => {
  environment.mockClear(); // <= does not clear the recordsource
});
```

Of course we could force to create a new environment between tests, but it seems not intended such case else whole `mockClear` makes no sense.

```javascript
import {createMockEnvironment} from 'relay-test-utils';

let environment;

beforeEach(() => {
  environment = createMockEnvironment();
});
```

Other possibilities which require extra scaffolding are:

- by passing own `store` to the `createMockEnvironment` and handling cleanup with `recordsource.clear()`.
- by making a "snapshot" (`environment.getStore().snapshot()`) before the test and "restore" after the test.

sibelius:
The recommended approach is to have 1 environment per test

artola:
> The recommended approach is to have 1 environment per test

Until now we have being able to setup the Provider using RTL wrapper (on jest setup), and doing the cleanup between tests.
We have several thousands of tests and just 1 wrapper for all of them.
As said above `mockClear` cleans the `cache`, but it does not the recordsource. It feels just like missing.

sibelius:
I think it makes sense a pull request to improve the current behavior

artola:
@sibelius Thanks. PR #3687

PR #3687, opened 2021-12-12
clean the recordsource in mockClear

artola:
Solves #3686

### DIFF
--- a/packages/relay-test-utils/RelayModernMockEnvironment.js
+++ b/packages/relay-test-utils/RelayModernMockEnvironment.js
@@ -182,8 +182,8 @@ export interface RelayMockEnvironment extends MockEnvironment, IEnvironment {}
 function createMockEnvironment(
   config?: $Shape<EnvironmentConfig>,
 ): RelayMockEnvironment {
-  const recordSource = config?.store ? undefined : new RecordSource();
-  const store = config?.store ?? new Store(recordSource);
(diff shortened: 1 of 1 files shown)


sibelius (approved):

alunyov:
@artola another approach would be to clear the store:

```js
// in the `afterEach(...)`
environment.getStore().getSource().clear();
```

As my colleague @kassens pointed out - this change will clear `recordSource` only if no record sources are passed. Which may not be an entirely consistent behavior.

artola:
> ```js
> .getStore().getSource().clear();
> ```

@alunyov I agree. But if `environment.mockClear()` cleans the `cache`, then it makes sense also cleans the `recordsource`.
The problem is that [`Store.getSource()`](https://github.com/facebook/relay/blob/30995390857adca421a89f0a58b098bf5ef1c839/packages/relay-runtime/store/RelayStoreTypes.js#L271) returns a `Recordsource`, and there is no `clear` [there](https://github.com/facebook/relay/blob/a8b9c2c1d6a33df5b71d80a6bdc850429afb2403/packages/relay-runtime/store/RelayStoreTypes.js#L230).

Therefore, if the `store` is passed in the config, there is no guarantee that will be provided a `clear` method as it is not part of the contract.
It is different when you have no config, then it will be provided a store. I know that this will be the case 99% of the time.

PR #3693, opened 2021-12-15
[relay 13] Add default exclude paths for multi-project config

bigfootjon:
This follows up on the changes I made in https://github.com/facebook/relay/pull/3688 to ensure that we properly add default excludes in multi-project mode to match the one set in single-project mode

Test plan:
`FORCE_NO_WATCHMAN=1 cargo run --bin relay -- ../../../myproject/js/relay.config.json` runs correctly without an `excludes` line in my `relay.config.json` file


alunyov (approved):
Thank you!

Issue #3438, opened 2021-04-02
Rust compiler: always purging __generated__ folders


mrtnzlml:
Hi, I have a feature request (or proposal/discussion) for the new Rust compiler. I noticed that the new compiler always deletes the whole content of `__generated__` folder. It's even explained here: https://github.com/facebook/relay/blob/63c4e99687c26a9e05b42e6e68aa55782a45a135/compiler/crates/relay-compiler/src/config.rs#L667-L672

I see the `__generated__` folder to be a nice convention together with `__tests__`, `__fixtures__` and so on. Nothing screams from it "Relay" (but rather "it's generated content"). So I used this pattern for some other use-cases where I generate some kind of non-Relay content into `__generated__`. This obviously brings some troubles now since Relay Rust compiler purges these files (even though it has nothing to do with Relay and there are no Relay artifacts).

What is the recommended strategy here? Or perhaps, would you please reconsider this behavior? Some other solutions could be:

- target only `*.graphql.js` (or new `*.relay.js`) files under `__generated__/…`
- have the possibility to put Relay artifacts under `__generated__/relay/…`
- have the possibility to keep subfolders like `__generated__/something-different/…`

Have you encountered this situation inside Facebook? How did you solve it?

Thanks! 😎

kassens:
For most code we use a fixed output directory, but in the places we use the relative directory I don't think we had anything other then Relay in them.

I think it'd be reasonable to match on `*.graphql.js` or `*.graphql.ts`.

Not sure I feel good about adding yet another output directory version as our options are already kindof hard to all test and it'd also add more confusion to the config as it would also need that configuration in the Babel plugin, I think.

kassens:
But assuming the existing JS compiler only removes the Relay files, we should fix this.

kassens:
@mrtnzlml : Actually, I was wrong:
Rust also does `md5(print(definition))` here:

https://github.com/facebook/relay/blob/b87e6b12aa42b76da5e0a88eafd1fd58d2e587b2/compiler/crates/relay-compiler/src/build_project/build_ir.rs#L33

which should match this code

https://github.com/facebook/relay/blob/b87e6b12aa42b76da5e0a88eafd1fd58d2e587b2/packages/babel-plugin-relay/compileGraphQLTag.js#L117-L120

[...]

PR #3700, opened 2021-12-21
do not automatically remove files not generated by Relay

zth:
Fixes https://github.com/facebook/relay/issues/3438

This ensures that the Rust compiler does not automatically remove anything from the artifact folders that wasn't generated by Relay itself. This restores the behavior of the JS compiler, that didn't remove non-Relay files (or at least could be configured so that it didn't).

I'm unsure if this is the best place to put a check like this. Ideally I guess the non-Relay files shouldn't be picked up at all by the compiler. But here I had access to the config I needed, which is why it ended up here.

Comments appreciated!

### DIFF
--- a/compiler/crates/relay-compiler/src/build_project/mod.rs
+++ b/compiler/crates/relay-compiler/src/build_project/mod.rs
@@ -40,6 +40,7 @@ use log::{debug, info, warn};
 use rayon::{iter::IntoParallelRefIterator, slice::ParallelSlice};
 use relay_codegen::Printer;
 use relay_transforms::{apply_transforms, find_resolver_dependencies, DependencyMap, Programs};
+use relay_typegen::TypegenLanguage;
 use schema::SDLSchema;
 pub use source_control::add_to_mercurial;
 use std::iter::FromIterator;
@@ -203,6 +204,14 @@ pub fn build_programs(
     )?;
 
     Ok((programs, Arc::new(source_hashes)))
+}
+
+fn is_relay_file_path(language: &TypegenLanguage, path: &PathBuf) -> bool {
+    match (path.to_str(), &language) {
+        (None, _) => false,
+        (Some(path), TypegenLanguage::Flow) => path.ends_with(".graphql.js"),
+        (Some(path), TypegenLanguage::TypeScript) => path.ends_with(".graphql.ts"),
+    }
 }
 
 pub fn build_project(
@@ -373,7 +382,9 @@ pub async fn commit_project(
                     break;
                 }
                 let path = config.root_dir.join(remaining_artifact);
-                config.artifact_writer.remove(path)?;
+                if is_relay_file_path(&project_config.typegen_config.language, &path) {
+                    config.artifact_writer.remove(path)?;
+                }
             }
             log_event.stop(delete_artifacts_time);
             ArtifactMap::from(artifacts)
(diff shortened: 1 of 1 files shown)


compiler/crates/relay-compiler/src/build_project/mod.rs
```diff
@@ -205,6 +206,14 @@ pub fn build_programs(
     Ok((programs, Arc::new(source_hashes)))
 }
 
+fn is_relay_file_path(language: &TypegenLanguage, path: &PathBuf) -> bool {
+    match (path.to_str(), &language) {
```
alunyov, line 210: do we need to convert this `to_str()`

can you use this: https://doc.rust-lang.org/std/path/struct.Path.html#method.ends_with?

zth: Yes, I think so, because it seems ends_with there only checks for entire chunks of the path. So that ends_with would mean the entire last chunk (the full file name in this case), which isn't what we're after. I tried using that function first. But maybe there's a better alternative still.

compiler/crates/relay-compiler/src/build_project/mod.rs
```diff
@@ -373,7 +382,9 @@ pub async fn commit_project(
                     break;
                 }
                 let path = config.root_dir.join(remaining_artifact);
-                config.artifact_writer.remove(path)?;
+                if is_relay_file_path(&project_config.typegen_config.language, &path) {
```
alunyov, line 385: This is mostly nits (and not sure if it makes it better). But can we move this function to the artifact_writer?

We already have `should_write`: https://github.com/facebook/relay/blob/e72ea12c746735f8bfa8e59e1c9704fcd6cd2179/compiler/crates/relay-compiler/src/build_project/artifact_writer.rs#L21

I think we can also add `should_remove`?

alunyov: `should_remove` can just return a boolean instead of `Result<...>`.

alunyov (approved):

compiler/crates/relay-compiler/src/build_project/mod.rs
zth: I'd need to make sure that the typegen config is added to the artifact_writer if so, but maybe it's ok to add that? Or I could do a typegen unaware check and just check for .graphql.js or .graphql.ts always. What do you think?

alunyov: I think it's fine to do a check without typegen config.

What do you think about alternative approach, where we would filter these non-relay artifacts in the `file_categorizer`?

Somewhere here: https://github.com/facebook/relay/blob/a07750f252e409e419e6886ec4a3d2a20cc4c48c/compiler/crates/relay-compiler/src/file_source/file_categorizer.rs#L237-L239

alunyov:
Hey @zth I have a slightly different fix for this: https://github.com/facebook/relay/commit/0b5a8b97b4c6bc938998885873bc127964c50785 

We now can ignore files in generated directories if they are not "potentially" created with `relay-compiler`. Will it work for your use case?

zth:
@alunyov that's looking great! Sorry for dropping the ball on this, I was trying to do something similar to that approach, but got stuck.

I'm closing this in favor for yours, thank you for solving it!

PR #3701, opened 2021-12-22
Make compiler transforms customizable and applied as a Vec of Transforms

leoasis:
Some advanced usages of Relay benefit from customizing the compiler so that users can provide custom transformations to the documents to affect how artifacts are generated. This PR adds the capability to specify these transformations in the `ProjectConfig` as a vector of `Transform`s.

In order to support this flexibility, I refactored the `apply_transforms` module in `relay-transforms` to apply transforms from a vector, and create default vectors for each of the steps in the execution pipeline (common, reader, operation, typegen, etc). These are used when there's no configuration for custom transforms, and are also exposed in the `relay-transforms` crate so they can be used and customized before setting them in the configuration.

The `Transform` is a new struct that takes a name, and a transform function, which takes a `project_config`, the `program`, and `base_fragment_names`, and returns a `DiagnosticResult<Program>`. I created closures for each of the current transforms to accommodate to this signature (wrapping the result in an `Ok` whenever the transform returned just the `Program`). Maybe it could be useful to enforce this signature in the transforms themselves at some point, to avoid the closure.

Now we have a generic function that takes the transforms vector, and applies it in order. It also wraps each transform call to log the time with the perf logger, and conditionally prints stats. Given that only the normalization step was explicitly passing the configuration to print stats, I made it so that the other steps explicitly pass `None` to this param.

One thing to note, is that I made a slight modification in how stats are printed in the normalization step. Before, we were printing the stats before starting the process with the string "normalization start", but now, since we have a generic function that does this, this is printed with the step function name, so this will show up as "apply_normalization_transforms start". Hope this change is acceptable, otherwise we will need to pass an extra string param there.

Also, this is my first time using Rust, so there's probably a lot of things that could be done in a more efficient/idiomatic way.

_Note: This ended up being a big PR, I'm happy to split it up into smaller chunks if needed, just wanted to get feedback on the approach_

### DIFF
--- a/compiler/crates/relay-compiler/src/build_project/log_program_stats.rs
+++ b/compiler/crates/relay-compiler/src/build_project/log_program_stats.rs
@@ -17,7 +17,7 @@ lazy_static! {
     static ref LOG_AST_STATS: bool = std::env::var("RELAY_LOG_AST_STATS").is_ok();
 }
 
-pub fn print_stats(extra_info: &'static str, program: &Program) {
+pub fn print_stats(extra_info: &str, program: &Program) {
     if *LOG_AST_STATS {
         let mut visitor = IRStatsVisitor::default();
         let stats = visitor.visit_program(program);
--- a/compiler/crates/relay-compiler/src/build_project/mod.rs
+++ b/compiler/crates/relay-compiler/src/build_project/mod.rs
@@ -41,7 +41,9 @@ use intern::string_key::{StringKey, StringKeySet};
 use log::{debug, info, warn};
 use rayon::{iter::IntoParallelRefIterator, slice::ParallelSlice};
 use relay_codegen::Printer;
-use relay_transforms::{apply_transforms, find_resolver_dependencies, DependencyMap, Programs};
+use relay_transforms::{
+    apply_transforms, find_resolver_dependencies, CustomTransformsConfig, DependencyMap, Programs,
+};
 use schema::SDLSchema;
 pub use source_control::add_to_mercurial;
 use std::iter::FromIterator;
@@ -120,6 +122,7 @@ pub fn transform_program(
     base_fragment_names: Arc<StringKeySet>,
     perf_logger: Arc<impl PerfLogger + 'static>,
     log_event: &impl PerfLogEvent,
+    custom_transforms_config: Option<&CustomTransformsConfig>,
 ) -> Result<Programs, BuildProjectFailure> {
     let timer = log_event.start("apply_transforms_time");
     let result = apply_transforms(
@@ -128,6 +131,7 @@ pub fn transform_program(
         base_fragment_names,
         perf_logger,
         Some(print_stats),
+        custom_transforms_config,
     )
     .map_err(|errors| BuildProjectFailure::Error(BuildProjectError::ValidationErrors { errors }));
 
@@ -192,6 +196,7 @@ pub fn build_programs(
         Arc::new(base_fragment_names),
         Arc::clone(&perf_logger),
         log_event,
+        config.custom_transforms.as_ref(),
     )?;
 
     Ok((programs, Arc::new(source_hashes)))
--- a/compiler/crates/relay-compiler/src/config.rs
+++ b/compiler/crates/relay-compiler/src/config.rs
@@ -28,6 +28,7 @@ use relay_config::{
     FlowTypegenConfig, JsModuleFormat, SchemaConfig, TypegenConfig, TypegenLanguage,
 };
 pub use relay_config::{PersistConfig, ProjectConfig, SchemaLocation};
+use relay_transforms::CustomTransformsConfig;
 use serde::de::Error as DeError;
 use serde::{Deserialize, Deserializer, Serialize};
 use serde_json::Value;
@@ -104,6 +105,9 @@ pub struct Config {
 
     /// Type of file source to use in the Compiler
     pub file_source_config: FileSourceKind,
+
+    // Provide extra custom transforms to apply
+    pub custom_transforms: Option<CustomTransformsConfig>,
 }
 
 pub enum FileSourceKind {
@@ -309,6 +313,7 @@ impl Config {
             additional_validations: None,
             is_dev_variable_name: config_file.is_dev_variable_name,
             file_source_config: FileSourceKind::Watchman,
+            custom_transforms: None,
         };
 
         let mut validation_errors = Vec::new();
@@ -458,7 +463,11 @@ impl fmt::Debug for Config {
         } = self;
 
         fn option_fn_to_string<T>(option: &Option<T>) -> &'static str {
-            if option.is_some() { "Some(Fn)" } else { "None" }
+            if option.is_some() {
+                "Some(Fn)"
+            } else {
+                "None"
+            }
         }
 
         f.debug_struct("Config")
--- a/compiler/crates/relay-compiler/tests/compile_relay_artifacts/mod.rs
+++ b/compiler/crates/relay-compiler/tests/compile_relay_artifacts/mod.rs
@@ -117,6 +117,7 @@ pub fn transform_fixture(fixture: &Fixture<'_>) -> Result<String, String> {
         Default::default(),
         Arc::new(ConsoleLogger),
         None,
+        None,
     )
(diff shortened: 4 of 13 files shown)


alunyov:
Thanks for this PR @leoasis!

What type of transforms do you need to add to your build script? Would it make sense to add them to Relay itself, maybe behind the feature flag? I'm not saying we should add them, but just curious to know about the use-case, and maybe we can find a better place to inject these.

Specifically, for this PR. I'm not sure if we need this level of customization in the `apply_transfrom`. For example, changing the order of a few of them can break some assumptions we have in the `apply_transfrom`. Ideally,  these transforms should not depend on each other, but they are. 

Also, alternatively, if you need to customize `apply_transfom` at this level, it's probably easier to create your own `apply_transfrom` and pass it via config (or some other way) to the compiler.

Again, let's discuss the actual use-case maybe we can find a different way.

leoasis:
Thanks for reviewing this PR! We have various use cases where we were leveraging the JS compiler, and we were inserting transforms at different points. Sometimes we needed to add them after built in ones, sometimes before, and in the right transform steps (common, normalization, reader, etc). We have use cases like transforming specific argument values with something different, extra validations specific to our schema, transforms to provide default values to fields, and perhaps most importantly we also have some transformations for our custom pagination mechanism, as we don't use connections in our schema.

I think having a way to inject a custom `apply_transforms` function in the config would also work for us, though it would lead to some duplication, as we want to continue using the steps like common, normalization, operation, etc. and the parallelization you're already doing. We only need to add extra transforms in the right places, which is at the beginning or end of a group.

Before working on this, I originally tried to extract an `apply_transforms` so that it can be specified in the config, but had a hard time making it work because of the PerfLogger trait and dealing with generics in the config. I think this is quite advanced Rust for me at this stage and couldn't figure out how to do that.

alunyov:
> Sometimes we needed to add them after built in ones, sometimes before, and in the right transform steps (common, normalization, reader, etc).

Would something like `before_apply_transforms(programs: &Programs) -> Programs` and `after_apply_transforms(programs: &Programs) -> Programs` work for this case?

This way we can preserve the correct order of the `apply_transforms`, but it is also flexible enough to handle the cases you've mentioned?

leoasis:
Oh interesting, yeah that sounds like it would be useful for our use cases. I didn't consider the `Programs` struct but it is a neat way to expose the programs for each step.

I don't think these are a big deal, but to be explicity, a couple of tradeoffs with this approach are that we would need to do manual perf logging of the transforms (that's ok), and also that this wouldn't allow to add transforms in the middle of the pipeline (we haven't had this use case yet, though).

But it seems this should be sufficient for our use cases. I'll try to take a stab at this.

leoasis:
After some more thought, I am thinking this may be challenging for the "before" scenario, because of how transformation is parallelized and continued from a previous step.

For example, we want to add a transform before any of the normalization transforms, but these are only applied after the operation and common transforms.

So, having a single function `before_apply_transforms` that takes `Programs` won't work, we'll need to have a way to specify a function for each step so that it can be called at the right time.

leoasis:
How about having separate `before_apply_*_transforms` and `after_apply_*_transforms` for each step? This would allow us to call each function at the right place.

Given it's a lot of new options, maybe they can be grouped in a struct passed to the config.

Curious about your thoughts @alunyov!

alunyov:
@leoasis 

Sorry for the delay with the response.

Individual before/after methods seems like a good idea. I'm not sure what is the best implementation could be...

Maybe we can add a new enum (not sure what is the correct name would be). And it is something similar to your initial version:

```rust
type TransformFn = fn(&Program) -> Program;

enum CustomTransform {
	BeforeCommon(TransformFn),
	AfterCommon(TransformFn),
	// ...
	// all other variants
}
```

and in the `config`, we would have a `Vec<CustomTransform>`, and in the apply_transform we can split this array into specific before/after steps.


With this version: 1) we preserve the original `apply_transfrom` structure, 2) it is flexible enough to customize the compilation process to your needs.

leoasis:
@alunyov I've updated this PR to ensure we can only add transforms either before or after the core ones. I went with a struct that explicitly allows users to configure `apply_before` and `apply_after` for each transform pipeline. I think this is more intuitive than the enum in a Vec as that would allow things such as `[BeforeCommon(fn), AfterOperation(fn), BeforeOperation(fn), ...]` which would make it confusing to understand what the behavior would be. Let me know what you think! This is how it would look like now:

```
common_transforms: Some(CustomTransforms { apply_before: before_vec, apply_after: after_vec }),
reader_transforms: Some(CustomTransforms { apply_before: before_vec, apply_after: after_vec }),
...
```

I've kept the refactor to make the transforms be applied as `Vec`s, as I think it would allow us to reuse the same functions to apply the customized ones via the config.

Thank you in advance!

alunyov (changes requested):
Thanks @leoasis!

There are a few suggestions I have:

1) Higlevel: lets a separate refactoring of `apply_transforms` and new functionality, prioritizing new functionality (custom transforms) - it is easier to review, with fewer chances of accidentally breaking something.

Specific things:
I don't think we can add `graphql-ir` as a dependency in `relay-config`. So these custom_transforms types/configs should be I think in `apply_transnforms`:
```rust
/// this is apply_transforms.rs
/// and we can export `CustomTransformConfig` from `relay-transforms` ///crate


#[derive(Clone)]
pub struct CustomTransform {
    pub name: &'static str,
    pub transform: fn(&ProjectConfig, &Program, &StringKeySet) -> DiagnosticsResult<Program>,
}

impl std::fmt::Debug for CustomTransform {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.debug_struct("Transform")
            .field("name", &self.name)
            .finish()
    }
}

#[derive(Debug)]
pub struct CustomTransforms {
    pub apply_before: Vec<CustomTransform>,
    pub apply_after: Vec<CustomTransform>,
}

#[derive(Debug)]
pub struct CustomTransformConfig {
    pub common_transforms: Option<CustomTransforms>,
    pub reader_transforms: Option<CustomTransforms>,
    pub operation_transforms: Option<CustomTransforms>,
    pub normalization_transforms: Option<CustomTransforms>,
    pub operation_text_transforms: Option<CustomTransforms>,
    pub typegen_transforms: Option<CustomTransforms>,
}

```

let's make `CustomTransformConfig` part of the `Config` and not a `ProjectConfig` - I think it is a reasonable limitation for all projects to have a similar set of transforms. If some transforms are not relevant to the project - the function can decide to skip the transformation, it has access to `&ProjectConfig`.

I think the **main change should be this**: adding new argument to the `apply_transform`:
```rust
pub fn apply_transforms<TPerfLogger>(
    project_config: &ProjectConfig,
    program: Arc<Program>,
    base_fragment_names: Arc<StringKeySet>,
    perf_logger: Arc<TPerfLogger>,
    print_stats: Option<fn(extra_info: &'static str, program: &Program) -> ()>,
    custom_transforms_config: Option<&CustomTransformConfig>,
) -> DiagnosticsResult<Programs>
```

and then calling these custom transform before/after each step (without migrating all steps to an array of transforms - again the main reason to focus only on the new feature. Later, we can re-consider a refactoring of `apply_transforms` .

Let me know what do you think?

maraisr:
Wondering actually if @leoasis you could provide an example of how this is leveraged? Do we `cargo build` our own binary??

alunyov:
> Wondering actually if @leoasis you could provide an example of how this is leveraged? Do we `cargo build` our own binary??

Yeah, that's the idea. For the custom builds you can create your own `main.rs`. And then you can pass these extra configurations via `config` object:
 
https://github.com/facebook/relay/blob/0c4108d630d67356f0b65e526201b5e7ccd78d29/compiler/crates/relay-compiler/src/main.rs#L132-L133

leoasis:
@alunyov Sounds good, I'll give that a shot and revert the changes on the existing transforms.

@maraisr yeah, what @alunyov said above is what we're planning to do on our end.

leoasis:
@alunyov I've applied all your suggestions, let me know what you think now! I've reverted the changes to the core transforms, and only apply the custom transforms as a Vec, in the beginning and end of each step.

leoasis:
Thanks @alunyov for merging this! 🎉

PR #3724, opened 2022-01-07
Fix eagerEsModules flag in babel-plugin-relay

TrySound:
The flag was renamed along with new compiler implementation,
fixed in docs but forgotten in babel plugin.

Using old flag makes compiler to panic and the only workaround now is
overriding flag in babel config.


TrySound:
cc @alunyov @kassens

alunyov (approved):

PR #3736, opened 2022-01-10
Use Semaphore to limit number of concurrent requests when persisting

### LINKED ISSUE #3728: [v13] Throttle persisted query uploads
I'm setting up my codebase to use relay v13 with persisted queries. I wrote a simple node http server based on the example here: https://relay.dev/docs/guides/persisted-queries/ but am running into the following issue.

```
[DEBUG] project: myproject
[DEBUG] persist_documents: 1647
[DEBUG] worker_count: 0
[DEBUG] done persisting
[DEBUG] error: Persisting operation(s) failed:
 - Network error: error trying to connect: tcp connect error: Connection reset by peer (os error 54)
 - Network error: error trying to connect: tcp connect error: Connection reset by peer (os error 54)
 - Network error: error trying to connect: tcp connect error: Connection reset by peer (os error 54)
 - Network error: error trying to connect: tcp connect error: Connection reset by peer (os error 54)
 - Network error: error trying to connect: tcp connect error: Connection reset by peer (os error 54)
 // repeated many times
```

It appears that relay-compiler is sending a concurrent http connection for each document (1,647 in my case) to the node server and I am hitting either an OS or node connection limit. Is it possible to throttle these connections to avoid this type of issue?

robrichard:
Fixes #3728

### DIFF
--- a/compiler/crates/relay-compiler/src/remote_persister.rs
+++ b/compiler/crates/relay-compiler/src/remote_persister.rs
@@ -20,7 +20,10 @@ impl OperationPersister for RemotePersister {
     ) -> Result<String, PersistError> {
         let params = &persist_config.params;
         let url = &persist_config.url;
-        persist(&operation_text, url, params).await
+        let permit = &persist_config.semaphore.acquire().await.unwrap();
+        let result = persist(&operation_text, url, params).await;
+        drop(permit);
+        result
     }
 
     fn worker_count(&self) -> usize {
--- a/compiler/Cargo.lock
+++ b/compiler/Cargo.lock
@@ -1433,6 +1433,7 @@ dependencies = [
  "regex",
  "serde",
  "serde_json",
+ "tokio",
 ]
 
 [[package]]
--- a/compiler/crates/relay-config/Cargo.toml
+++ b/compiler/crates/relay-config/Cargo.toml
@@ -14,3 +14,4 @@ intern = { path = "../intern" }
 regex = "1.5.4"
 serde = { version = "1.0.126", features = ["derive", "rc"] }
 serde_json = { version = "1.0.64", features = ["float_roundtrip", "unbounded_depth"] }
+tokio = { version = "1.10", features = ["full", "test-util", "tracing"] }
--- a/compiler/crates/relay-config/src/project_config.rs
+++ b/compiler/crates/relay-config/src/project_config.rs
@@ -13,7 +13,8 @@ use fnv::FnvBuildHasher;
 use indexmap::IndexMap;
 use intern::string_key::{Intern, StringKey};
 use regex::Regex;
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
+use tokio::sync::Semaphore;
 
 use crate::{connection_interface::ConnectionInterface, JsModuleFormat, TypegenConfig};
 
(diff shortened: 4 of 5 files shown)


compiler/crates/relay-compiler/src/remote_persister.rs
```diff
@@ -8,8 +8,22 @@
...
 use persist_query::{persist, PersistError};
+use tokio::sync::Semaphore;
 
-pub struct RemotePersister;
+pub struct RemotePersister {
+    semaphore: Semaphore,
+}
+
+impl RemotePersister {
+    pub fn new(mut permits: usize) -> Self {
```
alunyov, line 18: nit: can we make this immutable? We usually prefer not to mutate params, unless there is no other way.

compiler/crates/relay-compiler/src/main.rs
```diff
@@ -45,6 +45,9 @@ struct Opt {
     /// Run the persister even if the query has not changed.
     #[structopt(long)]
     repersist: bool,
+
+    #[structopt(default_value = "0", long)]
+    persist_concurrency: usize,
```
alunyov, line 50: I think we should move this to `persist_config`. Adding this to CLI is spreading persist configuration across these configs.

Our general approach is to reduce number of CLI flags as much as possible, and move everything to config files.

compiler/crates/relay-compiler/src/remote_persister.rs
```diff
@@ -8,8 +8,22 @@
...
+        if permits == 0 {
+            permits = usize::MAX >> 3
+        }
```
alunyov, line 21: can we have this value from defaults on persist_config?

alunyov (changes requested):
Thanks for working on this @robrichard!

compiler/crates/relay-compiler/src/main.rs
robrichard: @alunyov I agree it makes sense to add to `persist_config`. I had trouble with that approach because it doesn't seem like `persist_config` is available [when the RemotePersister is instantiated](https://github.com/robrichard/relay/blob/persist_concurrency/compiler/crates/relay-compiler/src/main.rs#L112) because it's part of the project config. I'm still very much a rust beginner so it's likely I'm missing something.

How would you recommend I approach this?

compiler/crates/relay-compiler/src/remote_persister.rs
robrichard: fixed

compiler/crates/relay-compiler/src/remote_persister.rs
```diff
@@ -20,7 +35,10 @@ impl OperationPersister for RemotePersister {
     ) -> Result<String, PersistError> {
         let params = &persist_config.params;
         let url = &persist_config.url;
-        persist(&operation_text, url, params).await
+        let permit = &self.semaphore.acquire().await.unwrap();
```
alunyov, line 38: @robrichard can we make `semaphore` part of the `persist_config` (but ignore it in `serde`)?

Then, this line became `&persist_config.semaphore.acquire().await.unwrap()`?

robrichard: @alunyov updated. `semaphore` is now part of the `persist_config`.

alunyov: Thanks! This looks great! One last minor request:

Can you add this option to the docs here:
https://github.com/facebook/relay/blob/main/packages/relay-compiler/README.md

robrichard: @alunyov docs are updated. Thanks!

alunyov (approved):

PR #3740, opened 2022-01-11
Use Array instead of ReadOnlyArray on input types

robrichard:
I'm curious why ReadonlyArrays are needed on input objects. We have mutations with complex input types and use array methods like `push` to generate the input object. The types generated by `relay-compiler-language-typescript` did not use ReadonlyArrays on input objects.

### DIFF
--- a/compiler/crates/relay-typegen/src/flow.rs
+++ b/compiler/crates/relay-typegen/src/flow.rs
@@ -42,6 +42,7 @@ impl Writer for FlowPrinter {
             AST::Identifier(identifier) => write!(&mut self.result, "{}", identifier),
             AST::RawType(raw) => write!(&mut self.result, "{}", raw),
             AST::Union(members) => self.write_union(members),
+            AST::Array(of_type) => self.write_array(of_type),
             AST::ReadOnlyArray(of_type) => self.write_read_only_array(of_type),
             AST::Nullable(of_type) => self.write_nullable(of_type),
             AST::ExactObject(props) => self.write_object(props, true),
@@ -174,6 +175,12 @@ impl FlowPrinter {
             write!(&mut self.result, "{}$fragmentType", fragment)?;
         }
         Ok(())
+    }
+
+    fn write_array(&mut self, of_type: &AST) -> FmtResult {
+        write!(&mut self.result, "Array<")?;
+        self.write(of_type)?;
+        write!(&mut self.result, ">")
     }
 
     fn write_read_only_array(&mut self, of_type: &AST) -> FmtResult {
--- a/compiler/crates/relay-typegen/src/lib.rs
+++ b/compiler/crates/relay-typegen/src/lib.rs
@@ -1506,7 +1506,7 @@ impl<'a> TypeGenerator<'a> {
     fn transform_non_nullable_input_type(&mut self, type_ref: &TypeReference) -> AST {
         match type_ref {
             TypeReference::List(of_type) => {
-                AST::ReadOnlyArray(Box::new(self.transform_input_type(of_type)))
+                AST::Array(Box::new(self.transform_input_type(of_type)))
             }
             TypeReference::Named(named_type) => match named_type {
                 Type::Scalar(scalar) => self.transform_graphql_scalar_type(*scalar),
--- a/compiler/crates/relay-typegen/src/typescript.rs
+++ b/compiler/crates/relay-typegen/src/typescript.rs
@@ -45,6 +45,7 @@ impl Writer for TypeScriptPrinter {
             AST::Identifier(identifier) => write!(&mut self.result, "{}", identifier),
             AST::RawType(raw) => write!(&mut self.result, "{}", raw),
             AST::Union(members) => self.write_union(members),
+            AST::Array(of_type) => self.write_array(of_type),
             AST::ReadOnlyArray(of_type) => self.write_read_only_array(of_type),
             AST::Nullable(of_type) => self.write_nullable(of_type),
             AST::ExactObject(props) => self.write_object(props),
@@ -146,6 +147,12 @@ impl TypeScriptPrinter {
             self.write(member)?;
         }
         Ok(())
+    }
+
+    fn write_array(&mut self, of_type: &AST) -> FmtResult {
+        write!(&mut self.result, "Array<")?;
+        self.write(of_type)?;
+        write!(&mut self.result, ">")
     }
 
     fn write_read_only_array(&mut self, of_type: &AST) -> FmtResult {
--- a/compiler/crates/relay-typegen/src/writer.rs
+++ b/compiler/crates/relay-typegen/src/writer.rs
@@ -11,6 +11,7 @@ use std::fmt::{Result as FmtResult, Write};
 #[derive(Debug, Clone)]
 pub enum AST {
     Union(Vec<AST>),
+    Array(Box<AST>),
     ReadOnlyArray(Box<AST>),
     Nullable(Box<AST>),
     Identifier(StringKey),
--- a/compiler/crates/relay-typegen/tests/generate_flow/fixtures/mutation-input-has-array.expected
+++ b/compiler/crates/relay-typegen/tests/generate_flow/fixtures/mutation-input-has-array.expected
@@ -9,7 +9,7 @@ mutation InputHasArray($input: UpdateAllSeenStateInput) @raw_response_type {
 ==================================== OUTPUT ===================================
 export type UpdateAllSeenStateInput = {|
   clientMutationId?: ?string,
-  storyIds?: ?$ReadOnlyArray<?string>,
+  storyIds?: ?Array<?string>,
 |};
 export type InputHasArray$variables = {|
   input?: ?UpdateAllSeenStateInput,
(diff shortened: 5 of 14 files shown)


josephsavona:
Hmm, in general Flow/TS allow mutably constructing a value and then passing it where a readonly is expected. Can you share a code snippet of how you construct values and what the error is?

robrichard:
@josephsavona it's mostly something like this:

https://www.typescriptlang.org/play?#code/PTAEHEFMDtIJwIYBdIBNQCMCeoBKkAbBLAKCSwAdJQBlBAN0gAkB7AW0goQHNIBJaBQCuSUAF5QAbxKhZoLrwByCDgH4AXKADOSOAEto3UAB9Q0IQQInQQ6KkgAzA2gDcMuRgTRYcLZvwIqCzQBFgAgnCIWAA8AEJePgLCoqbmlgB8bgC+biQgEDDwyGiYOPhEpORUoPHe8Eki4lLuslWQGtq6BkapFlamtvZOsKjZuQ62AMZIesGgALJYAMLsFMEwSAAUAJTNcqD5DnDsoA4scGydxS2gk8E6VyhNANoARGcsrwA0oK+ecK8ALq5fZ3aAPAzJTR0RisDgKfiCRoSaT7fYI5QcTSvAAW7E4PEg3xu+08dV8mmegJuWRINzOcFAmzBDzaoBYDkekF2qLRskhIgAdGSfFpBcItDjNpI2lltm59rSskA

maraisr:
My 2 cents is to leave it "immutable". Or config it... 

But to me having a pojo without "features" is less prone to errors. If the input object is derived, you can always unwrap the input type and cast in your local area.

That being said though, the entire object should also be `Readonly<{ banners }>` — so despite the first sentence, it's either entirely "immutable", or not at all??

robrichard:
@maraisr [as far as I can tell](https://github.com/facebook/relay/blob/v12.0.0/packages/relay-compiler/language/javascript/__tests__/__snapshots__/RelayFlowGenerator-test.js.snap#L733), non-array properties on input objects have never been typed as readonly in flow. I'm not sure why only the arrays are readonly, but the other properties are not.

It seems like there was a discrepancy between the flow output and `relay-compiler-language-typescript` output. In the old typescript output, all of the properties on input objects were not marked as readonly. This change breaks some of our code when attempting to update to relay 13. 

I can certainly work around it if necessary, but it's a bit awkward since there is no built in utility type to un-readonly something. I'm curious to know what benefits we get from the types being generated this way.

robrichard:
@alunyov @josephsavona any thoughts on this?

alunyov:
First of all, thank you for reporting this, and submitting this PR!

I think I agree with @maraisr here

>  the entire object should also be Readonly<{ banners }>

Probably the fact that some parts of these objects are not marked as immutable is just historical oversight (same as the discrepancy between flow and typescript in v12 and older versions), and it wasn't an intentional decision. But maybe I'm wrong.

I'm also not a big fan of making a new configuration option for these types of things. These configs will grow over time and will make things more complex, decreasing mailability, adding additional mental overhead, adding more confusion for future (and existing) uses on why we even have all these config options.

As for generated types: maybe at some point, in the next major version, we will make a big change and made all these read-only. 

As for this specific PR, it will be very hard to land this internally, as it will change a lot of files, and the only way to do this is to add a config/feature flag. 

But again, I think this change isn't in the right direction, we want to keep things immutable. For example, this code will be a thing (and probably already are in typescript, and maybe with objects in flow):  https://www.typescriptlang.org/play?#code/PTAEHEFMDtIJwIYBdIBNQCMCeoBKkAbBLAKCSwAdJQBlBAN0gAkB7AW0goQHNIBJaBQCuSUAF5QAbxKhZoLrwByCDgH4AXKADOSOAEto3UAB9Q0IQQInQQ6KkgAzA2gDcMuRgTRYcLZoCCcIhYADwAQl4+AsKipuaWAHxuAL5uJCAQMPDIaJg4+ESk5FSgEd7w0SLiUu6yxZAa2roGRnEWVqa29k6wqClpDrYAxkh6LNA2WpAAikLwoQAqAGoI+ggYBJBaCQAU9Kt665t+oMsHR1sAlDVyoEgAFnAsAO5mkK8AokEscDsA5IoWKI9GwKJsONAUKg-pcUiQSINoCMxhMALJYADC7Ao4xgSB212ktwyDiebFADh+5J0OVqoCG4x0TRy1QA2gAiSksdkAGlA7M8cHZAF00rcGdAmQYYpo6IxWBwFPxBFUJETbnIlcoOJp2fd2JweJBeXTbp5yr5NKzhXTkvDbpS4KAdhKmfVQCwHMyUITTXJpSIAHTmnxaQPCLT3HaSerJWG2+1yV2iVDIBDVIRTWbzEJy5gGpWVJC7ANIeN00vByLwMMRqMxyiQTR-XGgNg-agPFp-OMpIA

robrichard:
@alunyov thanks a lot for taking a look. I think in your example you are demonstrating that it's possible to mutate the input object after it's been passed to a relay function, potentially even before relay has read those variables. I agree that's definitely bad, but I don't think readonly types prevent that issue. As @josephsavona mentioned, Typescript allows mutably constructing a value and then passing it where a readonly is expected. [So even with the readonly types, mutating input objects is allowed.](https://www.typescriptlang.org/play?#code/PTAEHEFMDtIJwIYBdIBNQCMCeoBKkAbBLAKCSwAdJQBlBAN0gAkB7AW0goQHNIBJaBQCuSUAF5QAbxKhZoLrwByCDgH4AXKADOSOAEto3UAB9Q0IQQInQQ6KkgAzA2gDcMuRgTRYcLZvwIqCzQBFgAgnCIWAA8AEJePgLCoqbmlgB8bgC+biQgEDDwyGiYOPhEpORUoPHe8Eki4lLuslWQGtq6BkapFlamtvZOsKjZuQ62AMZIesE2WpAAikLwMQAqAGoI+ggYBJBa6QAU9Nt6u-t+oJtnFwcAlM1yoEgAFnAsAO5mkN8AopEWHAjgByRQsUR6NgUfYcaAoVAg+7ZEgkCbQaazaCgACyWAAwuwKMEYEgjo9pM98g4PmxQA4gXSdMUWqBJsEdJ1ik0ANoAIgZLD5ABpQHzPHA+QBdXLPdnQTkGZJNSnPOQKSDKDiaPmvdicHiQEWs56eOq+TQ8yRtTQgwUgrJS1lZVFyjmiVDIBBNIQLZaraJ0RisDgahpIY5KkTI11yKNIAB0Zp8WgTwi0ryO1sokFtJNAbCB1De3QdMayQA)

I guess I still don't understand the value of readonly input type. I think there's two options in the type of code that generates complex input objects:
1. Replace array manipulation with immutable versions (concat into a new array instead of push). Seems like this is less efficient because new objects/arrays are created unnecessarily. IMO the code is less readable too.
2. Write a type utility to strip readonly as suggested by @maraisr. If this is ok, then why not make the types mutable to start? I'll likely go down this path in our codebase to unblock the relay 13 upgrade.

For anyone else who runs into this, the util that's needed looks like this:

```ts
type Writable<T> = 
  T extends ReadonlyArray<infer U>
  ? Array<Writable<U>>
  : T extends {}
  ? { -readonly [K in keyof T]: Writable<T[K]> }
  : T;
```

The only other alternative I can think of is to make this change only for the Typescript types to maintain compatibility with v12 and minimize FB internal changes by leaving the Flow output as is.

alunyov:
> Typescript allows mutably constructing a value and then passing it where a readonly is expected

Yeah, I'm not really good with TypeScript:-) Is this a structural typing thing?

A similar [flow example](https://flow.org/try/#0PTAEHEFMDtIJwIYBdIBNQCMCeoBKkAbBLAKCSwAdJQBlBAN0gAkB7AW0goQHNIBJaBQCuSUAF5QAbxKhZoLrwByCDgH4AXKADOSOAEto3UAB9Q0IQQIBuGXIwJosOFs0ASfAlQB5aASwBBOEQsAB4AIQcnAWFRU3NLAD4bAF8bEhAIGHhkNEwcfCJScipQCMd4aJFxKVtZYsgNbV0DIziLAlAUtIAzIWgAYyQ9FmhQIS1IAEUheFCAFQA1BH0EDAJILQSACnplvVX1l1BFvYONgEoauVAkAAs4FgB3M0hngFEgljgtgHJFFlEejYFHWHGgKFQP3OKRIJF6AyGI1AAFksABhdgUEYwJBbS7Sa4ZboPNigbpfUk6HK1UD9EY6Jo5aoAbQAROSWKyADSgVn2OCsgC6aWudOgDIMMWqBOucgUkGUHE0rNu7E4PEg3Jp13s5WcmmZknqmh+HJ+yUFNOSsNF9NEqGQCGq4ymMzgoTojFYHHllSQ20lImhNMDSAAdLqnFow8ItLctkbKJATdjQGwvtQ7i1zcHkkA) where this works.

> I guess I still don't understand the value of readonly input type.

My understanding is that it will help enforcing some of the assumptions we have in Relay: the variables should not be mutated https://github.com/facebook/relay/blob/77073a22a963aa0a28fb8fae9564cde4287bf02f/packages/relay-runtime/store/RelayModernOperationDescriptor.js#L91 (I think the example we discussing here would throw a runtime error in the dev mode).

So the overall direction should be: make the whole variables object `ReadOnly`.

PR #3747, opened 2022-01-13
Remove unused import from example

xrash:
When following the step-by-step guide, there is a code snippet with an unused import. This PR removes that.


alunyov:
Thanks @xrash!

PR #3759, opened 2022-01-17
[relay-compiler] Publish compiler binaries to platform-specific binaries

alunyov:
Based on this suggestion https://twitter.com/IAmTrySound/status/1479191806420078598 (@TrySound)

I'm using similar to `esbuild` approach:  publish binaries to a standalone packages, and use them as platform-specific optional dependencies in the `relay-compiler`.

This should reduce the size `relay-compiler` is using on the `node_nodules`.

Test Plan:
- Tested locally on mac with the `todo` app (manually creating `artifacts` directory, running `gulp release`) and referencing local dependencies for `relay-compiler` in the `todo` app.


TrySound:
Awesome, thanks!

kassens:
Each binary is 10-20mb. Is this really a concern that warrants adding complexity and room for breakage?

packages/relay-compiler-linux-x64/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "relay-compiler-linux-x64",
```
maraisr, line 2: I feel like we an autogenerated this as part of build?

alunyov:
@kassens 

> Each binary is 10-20mb. Is this really a concern that warrants adding complexity and room for breakage?

My concern is that we potentially will need to add support for more platforms. Look at this list in the`esbuild` 
https://github.com/evanw/esbuild/tree/master/npm 

I think it is a reasonable improvement in OSS, the storage may not be not free.

alunyov:
We've decided not to land this for now. Both solutions: `optionalDependencies`  and postInstall script have issues (see the `esbuild` discussion: https://github.com/evanw/esbuild/issues/1647). 

But I'll keep this open for now, maybe we will need to land this if the size will keep regressing.

Issue #3743, opened 2022-01-11
relay-compiler 13.0.1 writes to `stderr`


hanford:
The new compiler seemingly writes useful information to `stderr` which can be problematic as some CI environments will error if `stderr` is written to

This can be verified by using a redirect operator and redirecting `stderr` to a file

```
relay-compiler &> foo
```

.. if this is intentional, feel free to close

https://linuxize.com/post/bash-redirect-stderr-stdout/#redirecting-stderr-to-stdout

ch1ffa:
By default, env_logger logs to stderr. It can be changed by `target` option, but errors will be redirected to stdout either. So we need to use another logger to split to stdout/err, for example `simplelog`.

I've tried simplelog::TermLogger and it looks good. Just added
```rust
let config = ConfigBuilder::new()
    .set_time_level(LevelFilter::Off)
    .build();

TermLogger::init(
    LevelFilter::Info,
    config,
    TerminalMode::Mixed,
    ColorChoice::Auto,
).unwrap();
```
instead of env_logger, and output was quite similar to the previous one and split to stdout/err:
<img width="587" alt="Screenshot 2022-01-17 at 00 26 00" src="https://user-images.githubusercontent.com/17417010/149678725-63fb7ad1-ff3b-4428-896d-70b98dc52859.png">

alunyov:
This fix was added in #3763. Thanks @ch1ffa!

hanford:
Thanks, feel free to close out whenever is appropriate @alunyov

Issue #3362, opened 2021-02-22
Add option to only print errors but nothing else (i.e. `--quietExceptErrors`)


LukasKalbertodt:
When executing `relay-compiler` (and there are no errors), it writes something like this to stdout:

```

Writing ts
Unchanged: 4 files
```

In some situations, I want to avoid this output as it adds noise to a build script output. There is of course the `--quiet` flag, but it also leads to `relay-compiler` not printing errors! I would like to have a way to have `relay-compiler` print errors (if any occur) but nothing else. 

On a related note, shouldn't errors printed to stderr instead of stdout?

LukasKalbertodt:
Still a problem.

alunyov:
This option was added in https://github.com/facebook/relay/pull/3763. Thanks @ch1ffa!

PR #3763, opened 2022-01-18
Replace env_logger by simplelog

ch1ffa:
Possible fix for #3743 and #3362

`evn_logger` has been replaced by `simplelog` which allows splitting stdout/stderr output.

I'm not sure about desired API, so added new options only as command line arguments.


alunyov (approved):
Amazing, thank for looking into this!

PR #3768, opened 2022-01-21
Use $PropertyType instead of indexed access for backwards Flow compat

leoasis:
There are some Flow type declarations that make use of [indexed access](https://flow.org/en/docs/types/indexed-access/), but that's only available from Flow 0.155 and older versions can't parse the syntax.

`$PropertyType` works in the same way and it's supported in older versions, so updating usages of indexed access for this so that Relay can be used with older versions of Flow.

### DIFF
--- a/packages/react-relay/ReactRelayTestMocker.js
+++ b/packages/react-relay/ReactRelayTestMocker.js
@@ -41,7 +41,7 @@ export type DataWriteConfig = {
 export type NetworkWriteConfig = {
   query: ConcreteRequest,
   variables?: Variables,
-  payload: GraphQLSingularResponse | (Variables => GraphQLSingularResponse),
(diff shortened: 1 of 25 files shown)


alunyov:
Hey @leoasis it looks like it will take some time for me to land this internally. We already have internal lint rules that are forcing the usage of indexed-access vs. `$PropertyType` with automatic replacement.

How hard it would be for you to update flow?

leoasis:
We're ~20 versions behind from 0.155 (which is the one that adds indexed access). Based on the release notes of some versions, sounds like it may be a while before we can get there 😔

leoasis:
@alunyov we've found a workaround to this issue, by adding a post install script that transforms the code with babel and converts indexed property access occurrences back to $PropertyType. At least this is a good workaround until we get to 0.155. So, closing this one.

Thank you for your help!

PR #3777, opened 2022-01-26
support providing schemaConfig in single project config files

zth:
This adds support for configuring `schemaConfig` in single project config files. Previously, defining `schemaConfig` in a single project config file would panic the compiler.

### DIFF
--- a/compiler/crates/relay-config/src/project_config.rs
+++ b/compiler/crates/relay-config/src/project_config.rs
@@ -63,6 +63,7 @@ pub enum SchemaLocation {
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct SchemaConfig {
+    #[serde(default = "default_connection_interface")]
     pub connection_interface: ConnectionInterface,
 
     /// The name of the `id` field that exists on the `Node` interface.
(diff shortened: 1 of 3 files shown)


compiler/crates/relay-config/src/project_config.rs
```diff
@@ -74,6 +75,10 @@ fn default_node_interface_id_field() -> StringKey {
     "id".intern()
 }
 
+fn default_connection_interface() -> ConnectionInterface {
+    ConnectionInterface::default()
+}
+
```
zth, line 81: Needed to add this default to avoid always having to supply `connectionInterface` in the config file.

alunyov (changes requested):
Thanks @zth!

The `schemaConfig` is still optional in the config file, right?

Also, can we you this option to the README here: https://github.com/facebook/relay/blob/main/packages/relay-compiler/README.md

zth:
Yes, still optional. I'll add it to the readme - what's the readme formatted with? Seems to have a special formatting, is it via prettier?

alunyov:
> Seems to have a special formatting, is it via prettier?

No, I just manually formatted it. But prettier sounds like a good idea to add for these .md files.

zth:
@alunyov I updated the readme. Hope I didn't butcher the formatting, had to add some spacing!

PR #3800, opened 2022-02-17
Allow using native flow/typescript enums.

### LINKED ISSUE #3596: Compiler: Flow Enums with Unknown Members instead of `"%future added value"`
I've been reading through the new [Flow Enums](https://flow.org/en/docs/enums/) documentation and one thing that caught my attention was [enums with unknown members](https://flow.org/en/docs/enums/defining-enums/#toc-flow-enums-with-unknown-members). In short:

```js
enum Status {
  Active,
  Paused,
  Off,
  ...
}
```

Specifically the last sentence:

> One use case for this would be the JS output of [GraphQL Enums](https://graphql.org/learn/schema/#enumeration-types): Flow Enums with unknown members could be used instead of the added `'%future added value'` member.

So my question/feature request is: are you considering this? Would you consider it?

Thanks! 😎

MaartenStaa:
Fixes #3596

Basically, allow using native [Flow enums](https://flow.org/en/docs/enums/) or [Typescript enums](https://www.typescriptlang.org/docs/handbook/enums.html). I removed the `flow_enums` config key, as it seems to be unused. That one was a set, rather than a boolean. I think this was only because of Flow enums not being allowed to have values starting with a lowercase letter, but instead I just transform all enum values to title case (aka pascal case), which I believe eliminates the need for a set.

@alunyov @mrtnzlml

### DIFF
--- a/compiler/Cargo.lock
+++ b/compiler/Cargo.lock
@@ -1421,6 +1421,7 @@ dependencies = [
  "common-path",
  "dashmap",
  "dependency-analyzer",
+ "docblock-syntax",
  "errors",
  "extract-graphql",
  "fixture-tests",
--- a/compiler/crates/relay-config/src/typegen_config.rs
+++ b/compiler/crates/relay-config/src/typegen_config.rs
@@ -7,12 +7,11 @@
 
 use common::Rollout;
 use fnv::FnvBuildHasher;
-use indexmap::{IndexMap, IndexSet};
+use indexmap::IndexMap;
 use intern::string_key::StringKey;
 use serde::{Deserialize, Serialize};
 
 type FnvIndexMap<K, V> = IndexMap<K, V, FnvBuildHasher>;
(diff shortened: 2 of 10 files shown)


alunyov:
Thanks @MaartenStaa!

I think this `flow_enums` is used internally (maybe just not exposed in the SingleProjectConfig) and the idea (but @kassens  have more context here, so I may be wrong) is not to write `export const enum ...` in the relay generated files, but import these allow-listed enums from generated enum files (another change you've been looking in to).

MaartenStaa:
Thanks @alunyov! If `flow_enums` is indeed already being used internally, I'll probably have to add it back. Maybe as a deprecated version, which sets `use_native_enums`. 

The import behavior you refer to is only activated when the `enum_suffix` option is set as far as I know—which this change should be compatible with.

fedotxxl:
is it going to be released?

lolopinto:
also ping about if this is going to be released and supported

PR #3805, opened 2022-02-18
Support index with suffixes when validating module names

lpalmes:
We have been moving to the new rust compiler with @leoasis and team, while doing so we found out that module names were failing for the new rust compiler while they were valid in the js compiler. 
In our case from `/path/__tests__/index.test.js` we were getting `index` in the rust compiler instead of `Tests`.

I found out that `extract_module_name` works correctly with files with one suffix (eg: `index.js`) but it didn't work with multiple suffixes hence this pr and two more assertions for those cases.

sidenote: The results in speed (and features) is incredible, thank you for working on the rust implementation!

### DIFF
--- a/compiler/crates/relay-transforms/src/validations/validate_module_names/extract_module_name.rs
+++ b/compiler/crates/relay-transforms/src/validations/validate_module_names/extract_module_name.rs
@@ -174,10 +174,6 @@ mod tests {
             Some("nonNumericEnd".to_string())
         );
         assert_eq!(
-            extract_module_name("/path/non-numeric-end-.js"),
-            Some("nonNumericEnd".to_string())
-        );
-        assert_eq!(
             extract_module_name("/path/__tests__/index.test.js"),
             Some("Tests".to_string())
         );
--- a/compiler/crates/relay-transforms/src/validations/validate_module_names/extract_module_name.rs
+++ b/compiler/crates/relay-transforms/src/validations/validate_module_names/extract_module_name.rs
@@ -14,24 +14,16 @@ pub fn extract_module_name(path: &str) -> Option<String> {
     let final_segment = get_final_non_index_js_segment(&path)?;
     let mut iter = final_segment.split('.');
     let mut first_segment = to_camel_case(iter.next()?.to_string());
-    // If the file name is index the module name is the directory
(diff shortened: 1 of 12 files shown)


compiler/crates/relay-transforms/src/validations/validate_module_names/extract_module_name.rs
```diff
@@ -161,6 +165,10 @@ mod tests {
             extract_module_name("/path/non-numeric-end-.js"),
             Some("nonNumericEnd".to_string())
         );
+        assert_eq!(
+            extract_module_name("/path/__tests__/index.test.js"),
+            Some("Tests".to_string())
+        );
```
josephsavona, line 171: can you add a test case for a path like `/path/button/indexButton.js`?

compiler/crates/relay-transforms/src/validations/validate_module_names/extract_module_name.rs
```diff
@@ -30,7 +30,7 @@ pub fn extract_module_name(path: &str) -> Option<String> {
 
 fn get_final_non_index_js_segment(path: &Path) -> Option<&str> {
     let file_stem = path.file_stem()?.to_str();
-    if file_stem == Some("index") {
+    if file_stem.map_or(false, |f| f.starts_with("index")) {
```
josephsavona, line 33: should this be `starts_with("index.")`? otherwise this would return the parent for a file like `indexFoo.js` which i don't think is intended

josephsavona (changes requested):
This makes sense and we're happy to support this, see comments though for being more precise about which files are matched

compiler/crates/relay-transforms/src/validations/validate_module_names/extract_module_name.rs
lpalmes: done

compiler/crates/relay-transforms/src/validations/validate_module_names/extract_module_name.rs
lpalmes: good catch, thanks

josephsavona (approved):
awesome, thanks!

PR #3808, opened 2022-02-22
Sort object members in typegen

MaartenStaa:
Trying to start split #3280 into smaller parts. When writing objects (whether exact or inexact) in typegen, sort the object's members by name. The used sorting is a bit subjective, so I'm open to suggestions.

@rbalicki2

### DIFF
--- a/compiler/crates/relay-typegen/src/flow.rs
+++ b/compiler/crates/relay-typegen/src/flow.rs
@@ -217,7 +217,7 @@ impl FlowPrinter {
         }
         self.indentation += 1;
 
-        for prop in props {
+        for prop in self.sort_props(props) {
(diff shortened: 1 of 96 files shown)


rbalicki2:
Love it! Thank you for doing this!!! This is going to make changes in the future **so** much easier to land. I might make some minor changes on my end — using descriptive constants in sort_props, and potentially avoiding a clone by sorting when the AST::[In]ExactObject is created, but overall, this looks great!

Can you also put up a diff to sort the items in union and intersection types, and to sort FragmentReference types?

I think these ordering diffs + adding parentheses around union and intersection types can all land at once, and be rolled out slowly (but as a unit) internally.

PR #3809, opened 2022-02-22
Typegen: Add intersection variant, and wrap unions and intersections in parentheses when they have more than one member

MaartenStaa:
Trying to start split #3280 into smaller parts. Not crazy about how many parentheses are being added as a result in the test fixtures. I could add a few more changes to reduce these when not needed. Thoughts?

@rbalicki2


MaartenStaa:
I think this one is no longer required.

PR #3810, opened 2022-02-22
Relax Viewer requirements for refetchable fragments

edigaryev:
Problem: Currently [`@refetchable` directive](https://relay.dev/docs/api-reference/use-refetchable-fragment/#arguments) can be used on `Viewer` type. However, real-world the APIs often utilize a `User` type for this use-case ([GitHub, for example](https://docs.github.com/en/graphql/reference/queries#user)). This causes a lot of pain because there's no easy way to make a `Viewer` → `User` type alias in GraphQL.

Solution: infer the actual viewer type from the `viewer: ...` field instead of hardcoding it.

### DIFF
--- a/compiler/crates/relay-transforms/src/refetchable_fragment/utils.rs
+++ b/compiler/crates/relay-transforms/src/refetchable_fragment/utils.rs
@@ -31,7 +31,6 @@ pub struct Constants {
     pub node_field_name: StringKey,
     pub node_type_name: StringKey,
     pub viewer_field_name: StringKey,
-    pub viewer_type_name: StringKey,
 }
 
 lazy_static! {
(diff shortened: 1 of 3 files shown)


josephsavona:
I'm curious - we have multiple strategies for generating refetch queries, one of which is the `Node` case which handles any type that implements Node and has an id. The GitHub case falls into that category — do you have a `viewer` field whose type doesn't implement Node?

edigaryev:
> I'm curious - we have multiple strategies for generating refetch queries, one of which is the `Node` case which handles any type that implements Node and has an id. The GitHub case falls into that category — do you have a `viewer` field whose type doesn't implement Node?

Hello! 👋

Apparently, yes, see [Cirrus CI GraphQL Schema](https://github.com/cirruslabs/cirrus-ci-web/blob/master/schema.gql), for example.

It also seems a bit odd to re-fetch a `viewer` using an `ID` because the `viewer` itself might change over the course of web UI session.

As Relay already offers multiple strategies around missing `Node`, this change just makes them easier to use.

PR #3830, opened 2022-03-08
Fix getting relative import path in windows for babel-plugin-relay

### LINKED ISSUE #3201: Unable to resolve module `..\\__generated__\\...` in React Native on Windows with artifactDirectory
Reproduction repository using the SWAPI server from here https://github.com/graphql/swapi-graphql.

https://github.com/benmurden/react-native-relay-windows-repro

It seems this issue is related to Windows directory traversal, because things seem to compile fine when generated artifacts are next to files where the query is used. I.e. when `artifactDirectory` is used to collect all relay artifacts, and components are located in another folder up the tree, Babel or Metro seem unable to correctly resolve the artifact file.

### Environment
```
OS: Windows 10
Node: v12.18.3
react: 16.13.1
react-native: 0.63.3
react-relay: 10.0.1
```

Screenshot_1601453909

Notice that the output suggests that too many upward directory traversals (`..\\..\\..\\`) are being prepended to the artifact directory.

### LINKED ISSUE #3272: Can't resolve graphl file when using Webpack 5 (and nextjs)
Hi.

Using:
- Next.js
- Babel relay plugin
- typescript (projects)
- Default webpack config (which is 4)
- Yarn workspaces
everything runs fine.

But when I upgrade to webpack 5, which is supported by nextjs I get the following error while starting development server or building:

`Module not found: Can't resolve '..\..\__generated__\relay\someMutation.graphql'`
this is the graphql file added by babel relay plugin.

I've been trying to debug and understand better babel's work but I confess I do not have much knowledge to debug it further by myself.

babel.config.json
```
{
  "presets": [
    "next/babel"
  ],
  "plugins": [
    [
      "relay",
      {
        "artifactDirectory": "./src/__generated__/relay/"
      }
    ],
    "@babel/plugin-transform-runtime",
    "babel-plugin-macros",
    [
      "styled-components",
      {
        "ssr": true,
        "displayName": true
      }
    ]
  ]
}
```
Note that I use babel plugin macros for tailwind's macro, not relay plugin macro. Although I tried using it and I got the same error too...

My generated files are indeed correctly generated at `'.src/__generated__/relay'`

It does work with webpack 4, and not with webpack 5.

I appreciate any help to try debugging it, I didn't find any public repo that uses typescript, relay and webpack 5, that's why I'm not sure I'm the one doing something wrong.

wKich:
Fix #3272

On windows there is a problem with building project, where `getRelativeImportPath` from `babel-plugin-relay` constructs replative path, which contains combination of `/` and `\` slashes, that leads to a webpack compile error. Because webpack can't find module with that path. Here is example of that error

```
ERROR in ./src/components/PersonDetailsDrawer/PersonDetailsDrawer.tsx 12:130-191
Module not found: Error: Can't resolve '..\..\__generated__\PersonDetailsDrawer.graphql' in 'C:\Users\nflaz\My Drive\AsusG14\hp\hpdev-graph\frontend\people-search\src\components\PersonDetailsDrawer'
Did you mean './..\..\__generated__\PersonDetailsDrawer.graphql'?
Requests that should resolve in the current directory need to start with './'.
Requests that start with a name are treated as module requests and resolve within module directories (node_modules).
If changing the source code is not an option there is also a resolve options called 'preferRelative' which tries to resolve these kind of requests in the current directory too.
```

I just replace all backslashes in a path to normal slashes.


packages/babel-plugin-relay/compileGraphQLTag.js
```diff
@@ -24,7 +24,7 @@ const {
   join: joinPath,
   relative: relativePath,
   resolve: resolvePath,
-} = require('path');
+} = require('path').posix;
```
giautm, line 27:
```suggestion
} = require('path/posix');
```

wKich:
@giautm Thanks for comment

sorenhoyer:
Ran into this problem yesterday as well when trying to change to a Single artifact directory setup instead of the default co-located artifacts.
Couldn't get it working with Webpack 5 - even with this patch applied. It did solve the problem you state, however it just got me to the next error related to paths in Webpack.
Switched to Vite today. Never tried it before, minimal configuration necessary (less than 10 lines) and it just worked like a charm with Relay.. This patch wasn't even necessary. Never looking back.
Not saying this should not be merged, but with Vite it's not really necessary. To me that indicates a problem with Webpack, so maybe you should make a PR to fix this in their end instead/as well :-)

wKich:
@sorenhoyer What's the next error you got?

giautm:
Hey @wKich, sorry for my suggestion. Can you revert it back to your changes, please? T___T

wKich:
Interesting it had been good until I merged the latest `main`

artola (commented):
@wKich  You should change  the import from `require('path/posix')` to `require('path').posix` to make Flow and old NodeJS happy.

PR #3839, opened 2022-03-15
Set inFlight to true before calling commitMutationFn

tinleym:
In environments with local resolvers (i.e. during tests), setMutationInFlight can lose a race to a commitMutationFn callback, causing inFlight to be stuck on true. Setting isMutationInFlight to true before calling commitMutationFn prevents this.

### DIFF
--- a/packages/react-relay/relay-hooks/useMutation.js
+++ b/packages/react-relay/relay-hooks/useMutation.js
@@ -97,6 +97,9 @@ function useMutation<TMutation: MutationParameters>(
 
   const commit = useCallback(
     (config: UseMutationConfig<TMutation>) => {
+      if (isMountedRef.current) {
+        setMutationInFlight(true);
+      }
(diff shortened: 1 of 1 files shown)


voideanvalue:
Thanks for the pull request! Could you please share an example of the issue? Or better yet, add a test case illustrating the issue that fails without this fix and validates this fix.

tinleym:
Sure! I encountered this in Storybook, would OK to link a repo w/ all that overhead included?

voideanvalue:
While something is better than nothing... by providing a repro test case along with the fix makes it easier for me to merge your pull request. Not having that creates more work for me 😛 

I took a stab at putting together a test case while procrastinating on something else this morning. It ended up being pretty straightforward: #3841

voideanvalue:
merged with eb0fc26f8483fb213d20b7a72288dce56e5b7385. thanks for reporting the issue and providing a fix!

PR #3841, opened 2022-03-17
Fix useMutation inFlight race condition

voideanvalue:
Basically #3839 but with a repro test case that fails without the fix and passes with it


voideanvalue:
bot didn't autoclose with eb0fc26f8483fb213d20b7a72288dce56e5b7385

PR #3851, opened 2022-03-24
Add support for the lsp in relay-bin

tbezman:
This PR aims to unblock an implementation of an open source VSCode extension to integrate with the Relay LSP server.

## High Level Changes
- Move our clap parser to a subcommand model so we can isolate the logic for different commands
  - For backwards compatibility and ease of use, compile is the default command.
 - Extract out some functions to help us configure a logger / get the relay config
   - When the LSP server is running, writing non LSP messages to stdout causes the VSCode extension to freak out. Writing to stderr works oddly enough and you can continue to see debug messages in while developing the extension
- Added a DummyDataProvider to appease the language server

Output of running `relay-compiler -h`
image

Output of running `relay-compiler lsp -h`
image

### DIFF
--- a/compiler/crates/relay-bin/src/main.rs
+++ b/compiler/crates/relay-bin/src/main.rs
@@ -75,15 +75,12 @@ struct CompileCommand {
 }
 
 #[derive(Parser)]
-#[clap(about = "Run the LSP server")]
+#[clap(about = "Run the LSP server", rename_all = "camel_case")]
 struct LspCommand {
     /// Run the LSP using this config file. If not provided, searches for a config in
     /// package.json under the `relay` key or `relay.config.json` files among other up
     /// from the current working directory.
     config: Option<PathBuf>,
-
-    #[clap(flatten)]
-    cli_config: CliConfig,
 
     /// Verbosity level
     #[clap(long, arg_enum, default_value = "verbose")]
@@ -141,7 +138,23 @@ impl CliConfig {
     }
 }
 
-fn get_config(config_path: Option<PathBuf>, cli_config: CliConfig) -> Config {
+#[tokio::main]
+async fn main() {
+    let opt = Opt::parse();
+
+    let command = opt.command.unwrap_or(Commands::Compiler(opt.compile));
+
+    match command {
+        Commands::Compiler(command) => {
+            handle_compiler_command(command).await;
+        }
+        Commands::Lsp(command) => {
+            handle_lsp_command(command).await;
+        }
+    }
+}
+
+fn get_config_backwards_compatible(config_path: Option<PathBuf>, cli_config: CliConfig) -> Config {
     let config_result = if let Some(config_path) = config_path {
         Config::load(config_path)
     } else if cli_config.is_defined() {
@@ -151,6 +164,21 @@ fn get_config(config_path: Option<PathBuf>, cli_config: CliConfig) -> Config {
                 cli_config.get_config_string(),
             ),
         })
+    } else {
+        Config::search(&current_dir().expect("Unable to get current working directory."))
+    };
+
+    let config = config_result.unwrap_or_else(|err| {
+        error!("{}", err);
+        std::process::exit(1);
+    });
+
+    config
+}
+
+fn get_config(config_path: Option<PathBuf>) -> Config {
+    let config_result = if let Some(config_path) = config_path {
+        Config::load(config_path)
     } else {
         Config::search(&current_dir().expect("Unable to get current working directory."))
     };
@@ -184,7 +212,7 @@ fn configure_logger(output: OutputKind, terminal_mode: TerminalMode) {
 async fn handle_compiler_command(command: CompileCommand) {
     configure_logger(command.output, TerminalMode::Mixed);
 
-    let mut config = get_config(command.config, command.cli_config);
+    let mut config = get_config_backwards_compatible(command.config, command.cli_config);
 
     if command.validate {
         config.artifact_writer = Box::new(ArtifactValidationWriter::default());
@@ -240,7 +268,7 @@ async fn handle_compiler_command(command: CompileCommand) {
 async fn handle_lsp_command(command: LspCommand) {
     configure_logger(command.output, TerminalMode::Stderr);
 
-    let config = get_config(command.config, command.cli_config);
+    let config = get_config(command.config);
 
     let perf_logger = Arc::new(ConsoleLogger);
     let extra_data_provider = Box::new(DummyExtraDataProvider::new());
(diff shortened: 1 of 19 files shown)


compiler/crates/relay-bin/src/main.rs
```diff
@@ -36,6 +41,10 @@ struct Opt {
     #[clap(long, short)]
     watch: bool,
 
+    /// Run the LSP server
+    #[clap(long, short)]
+    lsp: bool,
```
captbaritone, line 46: I think that rather than a flag, we should use an optional subcommand. That would allow us to have flags which are LSP specific. Will also allow for future compiler subcommands (I could imagine one that would dump stats about your codebase or something).

https://docs.rs/structopt/latest/structopt/#subcommands

I don't think StructOpt supports default subcommands, so we'll still need to handle the fact that some global options (intended for the compile) don't make sense in the context of the LSP, but I think that's okay.

compiler/crates/relay-bin/src/main.rs
```diff
@@ -49,8 +58,8 @@ struct Opt {
     repersist: bool,
 
     /// Verbosity level
-    #[clap(long, arg_enum, default_value = "verbose")]
```
captbaritone, line 52: A bit of a bummer here, I think the generated `--help` output will no-longer tell you what the default value is. Maybe there's some additional way to manually specify that with structopt?

compiler/crates/relay-bin/src/main.rs
```diff
@@ -101,11 +110,14 @@ impl From<CliConfig> for SingleProjectConfigFile {
...
-        OutputKind::QuietWithErrors => LevelFilter::Error,
-        OutputKind::Verbose => LevelFilter::Info,
+    let log_level = match &opt.output {
+        Some(output) => match output {
+            OutputKind::Debug => LevelFilter::Debug,
+            OutputKind::Quiet => LevelFilter::Off,
+            OutputKind::QuietWithErrors => LevelFilter::Error,
+            OutputKind::Verbose => LevelFilter::Info,
+        },
+        None => LevelFilter::Info,
```
captbaritone, line 120: I think the default before was `Verbose`, is this an intentional change?

compiler/crates/relay-bin/src/main.rs
```diff
@@ -115,13 +127,30 @@ async fn main() {
...
+        //
+        // We also don't want to litter existing projects with a relay_lsp.log file
+        // Let's only write out to a file if the LSP client specified an output level.
+        if opt.output.is_some() {
+            WriteLogger::init(
+                log_level,
+                log_config,
+                File::create("relay_lsp.log").unwrap(),
+            )
+            .unwrap();
```
captbaritone, line 143: Did you try writing to std err? Did that fail?

tbezman (commented):
@captbaritone I was able to get subcommands working by following along [this commit](https://github.com/clap-rs/clap/commit/db863de6c1e38df5ec6fab9e3491074011182df4). 

Writing to stderr does work and gives anyone working on the extension a better DX as they can see the output of the LSP directly in VSCode.

image


I think this addresses all of your feedback. I'm ready for a second round 💪

compiler/crates/relay-bin/src/main.rs
```diff
@@ -57,6 +74,28 @@ struct Opt {
...
+#[derive(Parser)]
+#[clap(about = "Run the LSP server")]
+struct LspCommand {
+    /// Run the LSP using this config file. If not provided, searches for a config in
+    /// package.json under the `relay` key or `relay.config.json` files among other up
+    /// from the current working directory.
+    config: Option<PathBuf>,
+
+    #[clap(flatten)]
+    cli_config: CliConfig,
```
captbaritone, line 86: I think we can omit this. I believe it's only left in the current version for backward compat, and we don't need to do that for this command.

compiler/crates/relay-bin/src/main.rs
```diff
@@ -197,6 +237,51 @@ async fn main() {
...
+        }
+        Err(err) => {
+            error!("Relay LSP unexpectedly terminated: {:#?}", err);
+            std::process::exit(1);
+        }
+    }
+}
+
+#[tokio::main]
+async fn main() {
```
captbaritone, line 270: Nit. Can we move this to the top, and have the supporting functions below (I know this is code preference, but I think it makes it easier to quickly understand what the module is trying to do.

compiler/crates/relay-bin/src/main.rs
```diff
@@ -28,9 +31,23 @@ use std::{
...
+    #[clap(subcommand)]
+    command: Option<Commands>,
+
+    #[clap(flatten)]
+    compile: CompileCommand,
+}
+
+#[derive(Parser)]
+#[clap(
+    rename_all = "camel_case",
```
captbaritone, line 47: Let's also do this for the LSP.

captbaritone (approved):
This looks good! A few details, but I think this is fine to merge after they are addressed.

compiler/crates/relay-compiler/src/errors.rs
```diff
@@ -17,6 +17,9 @@ pub type Result<T> = std::result::Result<T, Error>;
 
 #[derive(Debug, Error)]
 pub enum Error {
+    #[error("Unable to run the relay language server. Error details: \n{details}")]
+    LSPError { details: String },
```
captbaritone, line 21: Rather than adding a new variant here, let's crate a new `errors.rs` file in `relay-bin`. It's okay if it just has an enum with one or two variants.

compiler/crates/relay-bin/src/main.rs
```diff
@@ -57,6 +74,25 @@ struct Opt {
     validate: bool,
 }
 
+#[derive(Parser)]
+#[clap(about = "Run the LSP server", rename_all = "camel_case")]
```
captbaritone, line 78:
```suggestion
#[clap(about = "Run the language server. Used by IDEs.", rename_all = "camel_case")]
```

captbaritone: Might be useful to let people know they can ignore this option for their own manual use.

compiler/crates/relay-bin/src/main.rs
```diff
@@ -57,6 +74,25 @@ struct Opt {
...
+struct LspCommand {
+    /// Run the LSP using this config file. If not provided, searches for a config in
+    /// package.json under the `relay` key or `relay.config.json` files among other up
+    /// from the current working directory.
+    config: Option<PathBuf>,
+
+    /// Verbosity level
+    #[clap(long, arg_enum, default_value = "verbose")]
```
captbaritone, line 86: Does this show up in the "Output" panel in VSCode? If not, I think we should use a lower level here, like `info` or `quiet`, since 99% of the time nobody would see it, and it would just burn cycles.

captbaritone (commented):
Noticed a few more details.

compiler/crates/relay-bin/src/main.rs
tbezman: Great callout!

captbaritone (approved):
LGTM. I'll import and find an internal reviewer as well.

PR #3858, opened 2022-04-05
Init VSCode Extension

tbezman:
This PR aims to scaffold out an MVP for the Relay LSP VSCode Extension.

## High Level Changes
- Added new `vscode-extension` in the root dir containing all the extension code.
- We're using typescript in this new directory. Spoke this over with @captbaritone and it seems like we're cool with it.
  - Main benefit is we get all the types published by the VSCode team for free without having to wait for published flow defs.
- Added a basic LSP Client in `extension.ts` which looks for a relay binary and starts it up. Things seem to be working well in the Coinbase codebase.
- Added a `launch.json` so contributors can easily test the extension with `F5`. It also lets you see the debug output from the LSP Server directly in the opened window. It's really sweet.

## Follow Ups
I left a bunch of TODOs in the source if you want some more insight.

- [ ] Use VSCode config instead of environment variables from the launch json
- [ ] Support VSCode workspaces (don't use `workspace.rootPath`). Not exactly sure how to handle this but will be looking at the flow extension for more guidance.
- [ ] Better error message handling
- [ ] Status Bar support
- [x] Fix go to definition on fields / types
- [ ] Add command to manually restart the LSP Client.
- [x] Add a Github Action to do type checking / build check / prettier / linting


vscode-extension/.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
...
+  "configurations": [
+    {
+      "name": "Development Extension",
+      "type": "extensionHost",
+      "request": "launch",
+      "runtimeExecutable": "${execPath}",
+      "args": ["--extensionDevelopmentPath=${workspaceFolder}"],
+      "preLaunchTask": "ts-build",
+      "env": {
+        "RELAY_BINARY_PATH": "/path/to/your/binary",
```
captbaritone, line 12: I'm guessing we should omit this?

vscode-extension/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "relay-lsp-client",
+  "version": "1.0.0",
```
captbaritone, line 3: Maybe make this "0.0.1" for now.

vscode-extension/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "configuration": {
+      "type": "object",
+      "title": "Relay",
+      "properties": {
+        "relay.enabled": {
+          "scope": "window",
+          "type": "boolean",
+          "default": true,
+          "description": "Is relay enabled"
+        },
```
captbaritone, line 23: Do we need this? People can just enable/disable the extension I think? Is there some other value?

vscode-extension/package.json
```diff
@@ -0,0 +1,51 @@
...
+        "relay.outputLevel": {
+          "scope": "window",
+          "type": "string",
+          "default": "quiet-with-errors",
+          "description": "Can be one of the following. `quiet`, `quiet-with-errors`, `verbose`, `debug`"
```
captbaritone, line 28:
```suggestion
          "description": "Controls what is logged to the Output Channel."
```

Maybe we can use the enum/enum description as outlined here? https://code.visualstudio.com/api/references/contribution-points#Configuration-property-schema

vscode-extension/src/extension.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  if (!relayBinary) {
+    outputChannel.appendLine(
+      "Could not find relay binary in path. Maybe you're not inside of a project with relay installed."
+    );
+
+    return;
+  }
+
+  outputChannel.appendLine(`Using relay binary: ${relayBinary}`);
```
captbaritone, line 35: Love this.

vscode-extension/src/extension.ts
```diff
@@ -0,0 +1,130 @@
...
+      { scheme: "file", language: "typescript" },
+      { scheme: "file", language: "typescriptreact" },
+      { scheme: "file", language: "javascriptreact" },
+    ],
+
+    outputChannel,
+
+    // Since we use stderr for debug logs, the "Something went wrong" popup
+    // in VSCode shows up a lot. This tells vscode not to show it in any case.
+    revealOutputChannelOn: RevealOutputChannelOn.Never,
```
captbaritone, line 55: 👍

vscode-extension/src/extension.ts
```diff
@@ -0,0 +1,130 @@
...
+      // This happens when the LSP server stops running.
+      // e.g. Could not find relay config.
+      // e.g. watchman was not installed.
+      //
+      // TODO: Figure out the best way to handle this `closed` event
+      //
+      // Some of these messages are worth surfacing and others are not
+      // e.g. "Watchman is not installed" is important to surface to the user
+      // but "No relay config found" is not relevant since the user is likely
+      // just in a workspace where they don't have a relay config.
```
captbaritone, line 73: In that case, would they also not have a relay binary? Maybe we could assume that no Relay binary means not a Relay project, but if they have a Relay binary, we expect to find a config?

**edit** Ah, you pointed that out below.

vscode-extension/src/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+    binaryPathRelativeToPackageJson = path.join("macos-x64", "relay");
+  } else if (process.platform === "darwin" && process.arch === "arm64") {
+    binaryPathRelativeToPackageJson = path.join("macos-arm64", "relay");
+  } else if (process.platform === "linux" && process.arch === "x64") {
+    binaryPathRelativeToPackageJson = path.join("linux-x64", "relay");
+  } else if (process.platform === "win32" && process.arch === "x64") {
+    binaryPathRelativeToPackageJson = path.join("win-x64", "relay.exe");
+  } else {
+    binaryPathRelativeToPackageJson = null;
+  }
```
captbaritone, line 23: Maybe add a comment that this mirrors what we're doing in [`relay-compiler/index.js`](https://github.com/facebook/relay/blob/14550c8bd80e53369852db66d7681f13d6dd5f84/packages/relay-compiler/index.js)? If we make a change in one place we probably want to make it in both.

Maybe also add a comment in that file pointing back here?

vscode-extension/src/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+  rootPath: string
+): Promise<string | null> {
+  const binaryPathRelativeToPackageJson = getBinaryPathRelativeToPackageJson();
+
+  let currentPath = rootPath;
+  while (true) {
+    let possibleBinaryPath = path.join(
+      currentPath,
+      binaryPathRelativeToPackageJson
+    );
```
captbaritone, line 47: I like that this will correctly handle the case relay is installed at the root of a Yarn workspace!

vscode-extension/src/extension.ts
```diff
@@ -0,0 +1,130 @@
...
+      //
+      // We already bail early if there is no relay binary found. 
+      // So maybe we should just show all of these messages since it would
+      // be weird if you had a relay binary in your node modules but no relay
+      // config could be found. 🤷 for now.
+      closed() {
+        window
+          .showWarningMessage(
+            "Relay LSP client connection got closed unexpectdly.",
```
captbaritone, line 82:
```suggestion
            "Relay LSP client connection got closed unexpectedly.",

```

vscode-extension/src/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    if (await exists(possibleBinaryPath)) {
+      return possibleBinaryPath;
+    }
+
+    let nextPath = path.normalize(path.join(currentPath, ".."));
+
+    // Eventually we'll get to `/` and get stuck in a loop.
+    if (nextPath === currentPath) {
```
captbaritone, line 56: Is this copied from some extension? If not, it might be worthing seeing if we can find someone to test this on Windows just to double check. I assume it should work, but an infinite loop is also a bad error to have.

vscode-extension/.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Development Extension",
```
captbaritone, line 5: Is this string shown anywhere? Maybe this should be "Relay Extension"?

captbaritone (changes requested):
Overall this looks quite reasonable! First round of feedback left. I think addressing these would be enough to merge and then iterate.

vscode-extension/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "relay-lsp-client",
```
captbaritone, line 2: Is this the name that gets used in the VSCode Marketplace? I suspect this should be human readable. Maybe "Relay GraphQL"? Not sure what the best name is here. Might need to think about it.

vscode-extension/src/extension.ts
captbaritone: Oh, another thing we should explicitly handle is what happens if you are on an older version of Relay. When this extension comes out, it will only work with the very newest version of Relay. Ideally we could check if they are on an old version and give them a message letting them know they need to upgrade. Otherwise they will probably get a very confusing error where the binary just exits with an error code.

(We can tackle this in a followup PR)

vscode-extension/.vscode/launch.json
captbaritone: Confirmed, this causes the extension to crash (unless you know to edit it) but removing it allowed it to work!
Screen Shot 2022-04-04 at 7 49 18 PM

captbaritone:
Got it working on a little toy side project!

https://user-images.githubusercontent.com/162735/161670634-a193a85f-28f3-43fc-8a68-7d10d116cbf9.mov

vscode-extension/src/extension.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  outputChannel.appendLine(`Using relay binary: ${relayBinary}`);
+
+  const serverOptions: ServerOptions = {
+    command: relayBinary,
+    args: ["lsp", `--output=${outputLevel}`],
+  };
+
+  // Options to control the language client
+  const clientOptions: LanguageClientOptions = {
```
captbaritone, line 43: Can we try adding `markdown: {isTrusted: true}`? I'm noticing that hover items are not correctly rendering the content as markdown
Screen Shot 2022-04-04 at 9 01 14 PM
.

vscode-extension/.vscode/launch.json
tbezman: this is temporary and probably helped you run it locally? probably fine to keep it until we get proper vscode config setup.

vscode-extension/package.json
tbezman: yeah vscode can disable extensions per workspace now
image

vscode-extension/.vscode/launch.json
captbaritone: Once I removed it my server started correctly. I don't think we need it.

captbaritone: Oh, I see what you're saying. This lets you use a locally compiled version of the binary. I have ` "relay-compiler": "0.0.0-main-7a27722d",` in my package.json, so it's using a recent enough version that the compiler supports the LSP.

Maybe leave it but comment it out? (I think VSCode's JSON parser supports comments?)

vscode-extension/src/utils.ts
tbezman: yeah I'm definitely making an assumption here. i'll try to run some tests on my gaming pc tonight

vscode-extension/package.json
tbezman: So the flow one has some hyphenated name https://github.com/flow/flow-for-vscode/blob/main/package.json

I think they're like bundle identifiers in iOS. I'm fine with anything. I'll defer to you 
image

vscode-extension/src/utils.ts
captbaritone: If you have any question about it, you could add a counter and increment each loop and bail if it reaches something absurd like 5,000.

vscode-extension/src/extension.ts
tbezman: okay I think that did the trick. does this look right? 
image

captbaritone: Perfect

vscode-extension/package.json
captbaritone: Oh interesting. I wonder where the human readable bits come from.

orta: Use `displayName` in the `package.json` - https://code.visualstudio.com/api/references/extension-manifest

vscode-extension/src/utils.ts
tbezman: good call

vscode-extension/package.json
tbezman: Thanks @orta !

vscode-extension/src/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+export async function findRelayBinary(
+  rootPath: string
+): Promise<string | null> {
+  const binaryPathRelativeToPackageJson = getBinaryPathRelativeToPackageJson();
+
+  let counter = 0;
+  let currentPath = rootPath;
+  while (true) {
+    if (counter >= 5000) {
+      throw new Error("Could not find Relay binary after 5000 traversals. This is likely a bug in the extension code and should be reported to https://github.com/facebook/relay/issues");
```
captbaritone, line 49: ❤️

vscode-extension/src/utils.ts
tbezman: image

vscode-extension/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "displayName": "Relay for VSCode",
```
captbaritone, line 3: I think the fact that it's in the VSCode Marketplace is enough to communicate that it's for VSCode. I'll chat with the team about what we want here. I'm thinking "Relay GraphQL". Fine to merge this as is and update once we decide.

vscode-extension/src/utils.ts
tbezman: looks like it works on windows. but i'll keep the counter in as a fail safe just in case.

vscode-extension/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "relay-lsp-client",
```
alunyov, line 2: can we just use `relay` here?

vscode-extension/package.json
```diff
@@ -0,0 +1,47 @@
...
+          "scope": "window",
+          "type": "string",
+          "description": "Absolute path to the relay binary. If not provided, the extension will look in the nearest node_modules directory"
+        }
+      }
+    }
+  },
+  "scripts": {
+    "watch": "tsc -w"
+  },
```
alunyov, line 36: do we have any plans integration typecheker with Github actions?

Also we probably need to add lint/prettier?

tbezman: yeah i'll add those as followups

tbezman: @alunyov any preference on including these changes in this PR vs a follow up?

alunyov: Follow up is good! Thank you!

tbezman: @alunyov I went ahead and added the following to this PR

- prettier config which is basically the same as the flow one (just had to change the parser)
- eslint config from airbnb with some slight modifications
- typecheck

All this is passing in CI

tbezman: rough timing, totally cool if you want me to pull these commits out and add them in a follow up instead.

alunyov: Oh, this is great! Thanks, it's fine to keep it here

vscode-extension/src/extension.ts
```diff
@@ -0,0 +1,134 @@
+import {workspace, window} from 'vscode';
```
captbaritone, line 1: Can you add this header?

```
/**
 * Copyright (c) Meta Platforms, Inc. and affiliates.
 *
 * This source code is licensed under the MIT license found in the
 * LICENSE file in the root directory of this source tree.
 */
 ```

File: vscode-extension/src/utils.ts [hunk 20]
```diff
@@ -0,0 +1,78 @@
+import * as path from 'path';
```
captbaritone, line 1: Same header as above:

```
/**
 * Copyright (c) Meta Platforms, Inc. and affiliates.
 *
 * This source code is licensed under the MIT license found in the
 * LICENSE file in the root directory of this source tree.
 */
 ```

File: vscode-extension/.eslintrc [hunk 21]
```diff
@@ -0,0 +1,14 @@
...
+  "ignorePatterns": "out/*",
+  "parserOptions": {
+    "project": "./tsconfig.json"
+  },
+  "extends": ["airbnb-base", "airbnb-typescript/base"],
+  "rules": {
+    "operator-linebreak": "off",
+    "@typescript-eslint/object-curly-spacing": "off",
+    "import/prefer-default-export": "off",
+    "no-await-in-loop": "off"
```
captbaritone, line 12: Can wee add https://eslint.org/docs/rules/linebreak-style set to "unix"?

Maybe it's already fixed by Prettier, but when I first imported this PR I got an error from an internal linter:

Screen Shot 2022-04-05 at 9 28 25 PM

.github/workflows/ci.yml
```diff
@@ -8,6 +8,57 @@ name: CI
 on: [push, pull_request]
 
 jobs:
+  vscode-extension-lint:
```
alunyov, line 11: I think we can just have a single job here:
- vscode-extension

.github/workflows/ci.yml
```diff
@@ -8,6 +8,57 @@ name: CI
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 16.x
+          cache: 'yarn'
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile --ignore-scripts
+      - name: Typecheck
+        run: yarn run typecheck
```
alunyov, line 60: and then we won't need this

vscode-extension/src/extension.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  // TODO: Support multi folder workspaces by not using rootPath.
+  // Maybe initialize a client once for each workspace?
+  const relayBinary =
+    // TODO: Use VSCode config instead of process.env
+    process.env.RELAY_BINARY_PATH ??
+    (await findRelayBinary(workspace.rootPath));
+
+  // TODO: Use VSCode config instead of process.env
+  const outputLevel = process.env.RELAY_LSP_LOG_LEVEL ?? 'debug';
```
alunyov, line 26: same with the log-level? Should we use the value from the `configuration`?

vscode-extension/.eslintrc
tbezman: yeah I think this should be fixed with the latest & greatest

vscode-extension/src/extension.ts
tbezman: def in the follow up PR

tbezman:
@alunyov I've consolidated the CI jobs into one. lemme know if that looks alright.

alunyov (approved):
Thank you @tbezman!

PR #3859, opened 2022-04-05
Expose Relay schema extensions

### LINKED ISSUE #3787: [v13 relay-compiler] Exposing Relay Schema Extensions.
Pre v13 I know that `relay-compiler` exposed some of the schema extensions that are specific to Relay (ie directives) so that other third party libraries and tools could easily add them (ex: IDE GraphQL hinting/tooling, etc).

I noticed that since v13 that `relay-compiler` has been simplified (obviously) down to just exposing the underlying binary.

Are there any plans re-expose the [schema extensions](https://github.com/facebook/relay/blob/main/compiler/crates/relay-schema/src/relay-extensions.graphql) in some way?

I'm also curious why the underlying `relay-extensions.graphql` file seems to omit `@arguments` and `@argumentDefinitions`?

christianvuerings:
fixes #3787

# Context

From @erictaylor:

> Pre v13 I know that `relay-compiler` exposed some of the schema extensions that are specific to Relay (ie directives) so that other third party libraries and tools could easily add them (ex: IDE GraphQL hinting/tooling, etc).
> 
> I noticed that since v13 that `relay-compiler` has been simplified (obviously) down to just exposing the underlying binary.

From @josephsavona

> It seems reasonable to expose the schema extensions for other tools to pick up, we'd be happy to support a PR to that effect.

# Test

Run `RELEASE_COMMIT_SHA=212d8d66d9031180f08acf76f43c88162c6208e0 yarn gulp mainrelease`

## Before
Screen Shot 2022-04-04 at 5 42 54 PM

## After
Screen Shot 2022-04-04 at 5 43 40 PM

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -339,6 +339,20 @@ const relayCompiler = gulp.parallel(
       .src(['LICENSE'])
       .pipe(gulp.dest(path.join(DIST, 'relay-compiler')));
   },
+  function copyGraphQLExtensions() {
+    return gulp
+      .src(
+        path.join(
+          '.',
+          'compiler',
+          'crates',
+          'relay-schema',
+          'src',
+          'relay-extensions.graphql',
+        ),
+      )
+      .pipe(gulp.dest(path.join(DIST, 'relay-compiler')));
+  },
   function copyPackageFiles() {
     return gulp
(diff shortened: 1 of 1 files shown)


captbaritone:
Thanks!

PR #3860, opened 2022-04-05
Add locations to server sdls in build_schema_with_extensions

tbezman:
This enables the VSCode Extension to jump to definition on fields / typenames.

### DIFF
--- a/compiler/crates/schema/src/lib.rs
+++ b/compiler/crates/schema/src/lib.rs
@@ -39,23 +39,20 @@ const BUILTINS: &str = include_str!("./builtins.graphql");
 pub use flatbuffer::serialize_as_flatbuffer;
 
 pub fn build_schema(sdl: &str) -> DiagnosticsResult<SDLSchema> {
-    build_schema_with_extensions::<_, &str>(&[sdl], &[])
+    build_schema_with_extensions::<_, &str>(&[(sdl, SourceLocationKey::generated())], &[])
 }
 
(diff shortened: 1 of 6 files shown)


captbaritone:
Looks like CI is reporting some errors in `crates/relay-test-schema/src/lib.rs`. Let me know if you are having trouble reproducing locally.

compiler/crates/schema/src/lib.rs
```diff
@@ -39,23 +39,20 @@ const BUILTINS: &str = include_str!("./builtins.graphql");
...
+    server_sdls: &[(T, SourceLocationKey)],
     extension_sdls: &[(U, SourceLocationKey)],
 ) -> DiagnosticsResult<SDLSchema> {
     let mut server_documents = vec![builtins()?];
-    let mut combined_sdl: String = String::new();
-    for server_sdl in server_sdls {
-        combined_sdl.push_str(server_sdl.as_ref());
-        combined_sdl.push('\n');
+    for (server_sdl, location_key) in server_sdls {
+        server_documents.push(graphql_syntax::parse_schema_document(
```
captbaritone, line 51: @mofeiZ noticed that our internal schema files are not semantically complete on their own (they are split on arbitrary lines rather than between definitions). I've updated this internally to fall back to the old behavior using a generated location if there are multiple schema files.

tbezman: 🤯

PR #3862, opened 2022-04-06
fix: goto definition on typename in schema

tbezman:
Seems like the `locate_type_definition` function does not currently look at the schema. I updated the function to look at the schema first.

We still try to use the extra data provider, but if it fails, we fallback to the schema location.


compiler/crates/relay-lsp/src/goto_definition/mod.rs
```diff
@@ -120,22 +124,48 @@ fn locate_type_definition(
...
     project_name: StringKey,
     type_name: StringKey,
+    schema: &Arc<SDLSchema>,
+    root_dir: &std::path::Path,
 ) -> Result<GotoDefinitionResponse, LSPRuntimeError> {
+    let type_ = schema.get_type(type_name);
+
+    let schema_location = type_.and_then(|type_| {
+        match type_ {
+            // Why don't these other types have locations?
```
tbezman, line 134: @captbaritone / @alunyov can you speak to this?

captbaritone: I think we just haven't added it yet, but we should! Will require adding the location field to the correct struct within the schema, and then collecting it during parse.

tbezman: okay that makes sense. this sounds pretty deep to fix maybe? you think I could take a shot at this?

captbaritone: Yeah. It can be a pretty tedious update, since you need to touch every place that creates or updates the struct field that you choose to update to be `WithLocation`, but it tends to be pretty mechanical if you have Rust Analyzer setup correctly.

1. Update the field to be `WithLocation<_>`
2. Fix all errors by either accessing `.item` off the field, or using some version of `WithLocation::new` etc.

Once cargo check returns cleanly, you are done (I'll have some other work internally to update our other crates, but i"m happy to do that).

Want to take a stab at one?

compiler/crates/relay-lsp/src/goto_definition/mod.rs
```diff
@@ -120,22 +124,48 @@ fn locate_type_definition(
...
+            Type::InputObject(_) => None,
+            Type::Enum(_) => None,
+            Type::Interface(_) => None,
+            Type::Scalar(_) => None,
+            Type::Union(_) => None,
+            Type::Object(object_id) => Some(schema.object(object_id).name.location),
+        }
+    });
```
captbaritone, line 142: If the external data provider as a valid response, we will not end up using this. Would it be possible to mimic what we do in `locate_field_definition` and try to use the provided response if one is given, and then fallback to finding the value in the schema?

compiler/crates/relay-lsp/src/goto_definition/mod.rs
tbezman: yup! I'll take a stab and let you know if anything gets in the way

compiler/crates/relay-lsp/src/goto_definition/mod.rs
tbezman: @captbaritone I tried doing what you said. the nested options / results got pretty hairy. is what I did ok? is there a more idiomatic way of going from `Option<Result<T, E>>` => `Result<T, E>`?

captbaritone:
A few nits, but overall this looks great.

tbezman:
@captbaritone Rust analyzer made that a breeze. Should be good now. I think there's another issue where `get_graphql_definition_description` isn't properly matching input objects on mutations but I'll try to address that after we land this.

captbaritone (approved):
This looks great to me. I'll import in the morning.

Thanks for taking the time to add location support for more entities in the schema!

PR #3863, opened 2022-04-07
VSCode Extension restructure + Restart LSP command

tbezman:
This PR starts off by giving the extension a little bit of structure to support the list of follow ups described in #3858 .

## Features
- Restart language server command
  - Update the LSP side not to exit when we receive the shutdown request as per the [spec](https://microsoft.github.io/language-server-protocol/specification#shutdown)

### Relay Extension Context
```
// Mutable object to pass around to command handlers so they
// can reference the current state of the extension
//
// Things like configuration should not be stored in this context
// as they should be resolved contextually (workspace, file, etc).
// This state is global to the entire extension.
// At some point, we'll most likely have a map of Workspace => LSP Client
// for downstream callers to use.
//
// We're not using the VSCode's API for this since that state must be serializable
// as it's persisted to disk.
// https://code.visualstudio.com/api/extension-capabilities/common-capabilities#data-storage
```

### Splitting up logic
I basically split some of the chunkier parts into their own modules. All commands get their own module which get access to the relay extension context.

## Little things
- Enabled typescript strict mode (I cannot live without this)
- Configuration moved from env vars => vscode config


compiler/crates/relay-lsp/src/shutdown.rs
```diff
@@ -16,7 +16,7 @@ pub(crate) fn on_shutdown(
     _state: &impl GlobalState,
     _params: <Shutdown as Request>::Params,
 ) -> LSPRuntimeResult<<Shutdown as Request>::Result> {
-    std::process::exit(0);
+    LSPRuntimeResult::Ok(())
```
tbezman, line 19: @captbaritone this might break things internally?

vscode-extension/src/extension.ts
```diff
@@ -1,141 +1,23 @@
...
-            }
-          });
-
-        return ErrorAction.Continue;
-      },
-    },
+export async function activate(extensionContext: ExtensionContext) {
+  relayExtensionContext = {
+    client: null,
+    outputChannel: window.createOutputChannel('Relay Language Server'),
```
captbaritone, line 12: Do we need to ensure that the output channel gets its `.dispose` method called when the extension is closed?

Maybe add it to the context's subscriptions?

vscode-extension/src/config.ts
```diff
@@ -0,0 +1,15 @@
...
+export type Config = {
+  pathToRelay: string | null;
+  outputLevel: string;
+};
+
+export function getConfig(scope?: ConfigurationScope): Config {
+  const configuration = workspace.getConfiguration('relay', scope);
+
+  return {
+    pathToRelay: configuration.get('pathToRelay') ?? null,
```
captbaritone, line 12: Just curious: Does the fact that we declared `null` as the default in our `package.json` mean that VSCode will provide us with that default value when we call `.get` here? Or is that simply a documentation description, which we must actually implement ourselves here?

vscode-extension/src/context.ts
```diff
@@ -0,0 +1,20 @@
...
+// can reference the current state of the extension
+//
+// Things like configuration should not be stored in this context
+// as they should be resolved contextually (workspace, file, etc).
+// This state is global to the entire extension.
+// At some point, we'll most likely have a map of Workspace => LSP Client
+// for downstream callers to use.
+//
+// We're not using the VSCode's API for this since that state must be serializable
+// as it's persisted to disk.
```
captbaritone, line 14: Where does this get persisted to disk?

compiler/crates/relay-lsp/src/shutdown.rs
captbaritone: Can you explain why this is needed? Since the LSP is supposed to run forever, I think it makes idomatic sense to have it always return an error exit code, but I'm open to learning why we might not want to do that.

captbaritone (changes requested):
This looks great! So much cleaner, and it gives us a great foundation on which to build!

A few small bits of feedback/questions, but I think we're really close.

vscode-extension/src/config.ts
tbezman: I assume anyone can open their `settings.json` and put whatever they want. Let me test this out to see what the behavior is and how much safety we actually need to put in.

vscode-extension/src/context.ts
tbezman: i figured someone coming across this might think "Why not just use the data storage api from vscode" which is like a bag of data you can keep in the VSCode `ExtensionContext`. I just wanted to point out that we can't use that because it writes to disk.

tbezman:
> Can you explain why this is needed? Since the LSP is supposed to run forever, I think it makes idomatic sense to have it always return an error exit code, but I'm open to learning why we might not want to do that.

@captbaritone so according to the [spec](https://microsoft.github.io/language-server-protocol/specification#shutdown), we're not supposed to exit the program in the shutdown request. Without this, the LSP client hangs when we try to shutdown as it never receives a response from the server saying "the shutdown went well". 

I think the flow is something like shutdown => clean up stuff if needed => tell client shutdown is cool => client tells server to finally exit.

captbaritone:
Got it. Makes sense!

vscode-extension/src/context.ts
captbaritone: Oh got it. I read this backwards. Makes sense.

PR #3865, opened 2022-04-09
VSCode Extension | Status Bar Support

tbezman:
A lot of this PR was inspired by the work done in the [Flow for VSCode extension](https://sourcegraph.com/github.com/flow/flow-for-vscode/-/blob/lib/utils/LanguageClient/StatusFeature/StatusFeature.js?L10:36).

## Changes

- Add a `StaticFeature` which listens intercepts the `onRequest` events (specifically `window/showStatus`)
- Recreate all the types from the [`lsp_types` crate](https://github.com/gluon-lang/lsp-types/blob/master/src/window.rs#L15).
- Add a command to open the output console (used by the status bar click handler)

## Demo

https://user-images.githubusercontent.com/6754223/162557878-3f512ca3-323e-44a0-8811-d9d8468113e9.mov


vscode-extension/src/languageClient.ts
```diff
@@ -70,13 +72,32 @@ export async function createAndStartClient(context: RelayExtensionContext) {
...
+        (params) => {
+          handleShowStatusMethod(context, params);
+        },
+      );
+    },
+
+    dispose() {
+      showStatusDisposable?.dispose();
+    },
+  });
```
captbaritone, line 98: See my comment in `statusBar.ts` I wonder if we could move this logic into that file.

vscode-extension/src/statusBar.ts
```diff
@@ -0,0 +1,127 @@
...
+ */
+
+import {StatusBarAlignment, StatusBarItem, window} from 'vscode';
+import {RelayExtensionContext} from './context';
+
+export function createStatusBar(): StatusBarItem {
+  const statusBar = window.createStatusBarItem(StatusBarAlignment.Right, 100);
+  statusBar.command = 'relay.showOutput';
+
+  return statusBar;
```
captbaritone, line 15: Does this ever get disposed? If we make this a private property of a new `RelayStatusBar` class, we can do it there, if not, we probably need to do that wherever we clean up the context.

captbaritone (changes requested):
A few subscription things, but mostly I'm curious if we can reorganize the code a bit to make things a little cleaner.

vscode-extension/src/statusBar.ts
```diff
@@ -0,0 +1,127 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import {StatusBarAlignment, StatusBarItem, window} from 'vscode';
+import {RelayExtensionContext} from './context';
+
```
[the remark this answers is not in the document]
tbezman: @captbaritone good feedback. I think I landed at a healthy middle ground. I put all of the logic relevant to intercepting the `window/showStatus` command & turning those commands into a status bar message into a `LSPStatusBarFeature` class. 

I kept the management of the status bar separate since we might want to use that before the LSP gets booted up.

Left this comment on top of the `LSPStatusBarFeature` class

```
// This StaticFeature is solely responsible for intercepting
// window/showStatus commands from the LSP Server and displaying
// those messages on the client status bar.
//
// The StatusBarItem creation does not happen here since we may
// want to use the status bar to display messages before we
// get messages from the LSP server.
// e.g. Looking for Relay binary...
```

orta:
It's been a while since I've used flow-for-vscode now, so I'm not sure what this look like but you can also consider using the relay logo in the status bar now. I set this up for the graphql extension: http://orta.io/notes/apps/vscode-icon-contributions

PR #3868, opened 2022-04-13
Add new option to support compiling a single project

tbezman:
Add new option to support compiling a single project #3868

## Usage

`yarn relay-compiler -p projectA -p projectB`


compiler/crates/relay-compiler/src/config.rs
```diff
@@ -510,7 +523,11 @@ impl fmt::Debug for Config {
         } = self;
 
         fn option_fn_to_string<T>(option: &Option<T>) -> &'static str {
-            if option.is_some() { "Some(Fn)" } else { "None" }
+            if option.is_some() {
+                "Some(Fn)"
+            } else {
+                "None"
+            }
```
tbezman, line 530: rust fmt

compiler/crates/relay-bin/src/main.rs
```diff
@@ -205,6 +209,10 @@ async fn handle_compiler_command(command: CompileCommand) -> Result<(), Error> {
 
     let mut config = get_config(command.config)?;
 
+    if let Some(project_to_retain) = command.project {
+        config.retain_project(project_to_retain);
+    }
+
```
captbaritone, line 215: It looks like we can just set the project as `enabled` in the config. Here's what we do in the internal version:

```rust
/// Update Config if the `project` flag is set
fn set_project_flag(config: &mut Config, project: Vec<String>) {
    if project.is_empty() {
        return;
    }
    for project_config in config.projects.values_mut() {
        project_config.enabled = false;
    }
    for selected_project in project {
        let selected_project = selected_project.intern();
        if let Some(project_config) = config.projects.get_mut(&selected_project) {
            project_config.enabled = true;
        } else {
            println!(
                "Project `{}` not found, available projects: {}.",
                selected_project,
                config
                    .projects
                    .keys()
                    .map(|name| name.lookup())
                    .collect::<Vec<_>>()
                    .join(", ")
            );
            std::process::exit(1);
        }
    }
}
```

compiler/crates/relay-bin/src/main.rs
```diff
@@ -56,6 +57,11 @@ struct CompileCommand {
     #[clap(long, short)]
     watch: bool,
 
+    /// A list of projects to compile.
+    /// If excluded, all projects will be compiled
+    #[clap(long)]
+    projects: Vec<String>,
```
captbaritone, line 63: I would expect this to just be `project` and that you would specify it multiple times to specify multiple projects. Can you confirm what it actually looks like to pass multiple projects? I think it would be not great to have to pass multiple projects using one flag (how are you supposed to delineate them? commas? spaces?)

compiler/Cargo.lock
```diff
@@ -1798,9 +1814,9 @@ dependencies = [
 
 [[package]]
 name = "serde"
-version = "1.0.126"
+version = "1.0.136"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ec7505abeacaec74ae4778d9d9328fe5a5d04253220a85c4ee022239fc996d03"
+checksum = "ce31e24b01e1e524df96f1c2fdd054405f8d7376249a5110886fb4b658484789"
```
captbaritone, line 1819: I'm surprised to see all these version changes. Are these expected for some reason?

compiler/crates/relay-bin/src/main.rs
tbezman: yeah I think it doesn't do any space / command delimiting by default. I updated the language a bit to better explain.

compiler/Cargo.lock
tbezman: Hmmm, this is unintentional but I can't figure out how to make this not happen. Any tips here? 

Like there's no way for me to add `intern` as a dep to this crate without those other ones updating.

compiler/crates/relay-bin/src/main.rs
captbaritone: Cool. Can just make this `--project`?
```suggestion
    project: Vec<String>,
```

compiler/Cargo.lock
captbaritone: Strange. I think it's probably fine if it's happening automatically.

compiler/crates/relay-bin/src/main.rs
tbezman: @captbaritone Incase it isn't clear from the code. I kept the rust name as `projects` since it truly is a list of projects to compile. The clap name is `project` since the user can pass projects like so `--project projectA --project projectB`

PR #3869, opened 2022-04-14
Minor vscode package tweaks

orta:
Thought I'd take a look at whether I could use this locally, but I got hit with

```
Starting the Relay GraphQL extension...
Using relay binary: /home/orta/dev/app/node_modules/relay-compiler/linux-x64/relay
[Error - 4:44:06 PM] Connection to server got closed. Server will not be restarted.
[0m[31m[ERROR] [0mFailed to read config file `lsp`.

```

Which I'm guessing is because that command comes from unreleased code (I upgraded to 13.2.0 just in case). Either way, this is probably some quality of life changes

### DIFF
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -2,7 +2,13 @@
   "name": "relay",
   "displayName": "Relay GraphQL",
   "version": "0.0.1",
-  "description": "Relay LSP Client for VSCode",
+  "description": "Relay-powered IDE experience",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/facebook/relay",
+    "directory": "vscode-extension"
+  },
+  "license": "MIT",
   "main": "./out/extension.js",
   "activationEvents": [
     "onLanguage:plaintext",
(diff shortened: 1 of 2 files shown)


vscode-extension/package.json
```diff
@@ -2,7 +2,13 @@
   "name": "relay",
   "displayName": "Relay GraphQL",
   "version": "0.0.1",
-  "description": "Relay LSP Client for VSCode",
+  "description": "Relay-powered IDE experience",
```
orta, line 5: Most big extensions end up on vscodium and others via https://open-vsx.org/ - so it doesn't hurt to remove the vs code mention now

captbaritone:
Thanks for this @orta (good to see you!). To try out the LSP, you'll need an unreleased version of the compiler. Try `relay-compiler@0.0.0-main-ba2080a3`.

cc @tbezman since the error message here gets to one of the items on our TODO list: What happens if you run the LSP with an older version of the compiler. We are planning on having a more helpful error message here.

PR #3874, opened 2022-04-15
Fix windows tests

tbezman:
I'm not sure how cloning this repo causes the `builtins.graphql` to replace `\n` with `\r\n` but that's seem like what's happening. I updated the code to replace the `\r\n` with `\n` only once using `lazy_static`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+compiler/**/fixtures/* text eol=lf
--- a/compiler/.gitattributes
+++ b/compiler/.gitattributes
@@ -1,1 +1,1 @@
-**/fixtures/* text eol=lf
+* text eol=lf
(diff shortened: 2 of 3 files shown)


captbaritone:
TIL Git will "helpfully" translate line endings to match your system https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

I wonder if instead we should add a `.gitattributes` file which states that all `.graphql` files should keep their unix line endings as outlined [here](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#per-repository-settings).

alunyov:
Thanks @tbezman for looking into this!

I agree with @captbaritone here, and maybe we should try find a way either: change the file itself, or add `. gitattributes`.

tbezman:
@captbaritone @alunyov I'll try the gitattributes thing. That makes the most sense.

PR #3878, opened 2022-04-16
support sha256 and md5 algos for local file persister

tbezman:
Coinbase needs to use SHA256 . We temporarily have been using a custom persist server that devs run locally. Would be nice to have it in the compiler. 

I also threw MD5 in here while I was at it.

I updated the docs to tell more of a story around local persisting. I removed the old local persist server example since those use cases should be covered by the local file persister


alunyov (changes requested):
Thanks @tbezman!

I like the rust part.

What do you think about keeping the example persist script/server? But still adding the new section `Local Persisted Queries`.

tbezman:
@alunyov Okay, I've added back the local persist server example. Let me know if we need any other changes?

alunyov (approved):

eMerzh:
Sorry for chiming in after the fact ... 

But this is really great :) 

Do you think it would be possible to do that without using the "persistor" thing ? 
like changing the cacheId in the opreations, but still have access to the operation's text ?

thanks

Issue #3574, opened 2021-08-27
Docs: Link for `@defer` directive (and others) points to intern


garyolsen:
In the most recent Glossary page in the docs, the _"See the documentation"_ link for [@defer](https://relay.dev/docs/glossary/#defer) points to the wiki: https://www.internalfb.com/intern/wiki/Relay/Web/incremental-data-delivery-defer-stream/#defer

It'd be helpful to get some more information about how the directive works into the public docs.

---

There are other links to intern in the docs, though it looks like a few of them are intentional—it may be worth auditing this list to see if they should be changed. If not, I can open a PR updating the docs to at least indicate them as FB-only.
```
docs % grep -R '/intern/' .
./api-reference/hooks/use-query-loader.md:            * `fetchPolicy`: _*[Optional]*_ Determines if cached data should be used, and when to send a network request based on the cached data that is currently available in the Relay store (for more details, see our [Fetch Policies](https://www.internalfb.com/intern/wiki/Relay/guided-tour-of-relay/reusing-cached-data-for-rendering/#fetch-policies) and [Garbage Collection](https://www.internalfb.com/intern/wiki/Relay/guided-tour-of-relay/reusing-cached-data-for-rendering/#garbage-collection-in-re) guides):
./api-reference/entrypoint-apis/use-entrypoint-loader.md:For more information, see the [Loading EntryPoints](https://www.internalfb.com/intern/wiki/Relay/Guides/entry-points/#loading-entrypoints) guide.
./guided-tour/introduction.md:* [Javascript](https://our.internmc.facebook.com/intern/wiki/JavaScript/)
./guided-tour/introduction.md:* [React](https://our.internmc.facebook.com/intern/wiki/ReactGuide/)
./guided-tour/introduction.md:* [GraphQL](https://our.internmc.facebook.com/intern/wiki/GraphQL/) and our internal [GraphQL Server](https://our.internmc.facebook.com/intern/wiki/Graphql-for-hack-developers/)
./guided-tour/updating-data/graphql-mutations.md:* Check out the [Hack documentation on writing mutations](https://www.internalfb.com/intern/wiki/Graphql-for-hack-developers/mutation-root-fields/) for information on how to add a mutation field to your back-end code.
[...]

PR #3884, opened 2022-04-21
Replace some links pointing to internalfb.com with public equivalents

astj:
In https://relay.dev/docs/, there're several links pointing to internalfb.com, which is not public (maybe related: https://github.com/facebook/relay/issues/3574 )
In this p-r, I replaced such links in https://relay.dev/docs/next/api-reference/use-query-loader/ with public equivalents.
I believe they point to intended documents, since https://relay.dev/docs/next/api-reference/use-lazy-load-query/ already has  similar links. (But I cannot confirm it, since I cannot access internalfb.com 😅).

### DIFF
--- a/website/docs/api-reference/hooks/use-query-loader.md
+++ b/website/docs/api-reference/hooks/use-query-loader.md
@@ -90,7 +90,7 @@ A tuple containing the following values:
     * Parameters
         * `variables`: the variables with which the query is loaded.
         * `options`: `LoadQueryOptions`. An optional options object, containing the following keys:
-            * `fetchPolicy`: _*[Optional]*_ Determines if cached data should be used, and when to send a network request based on the cached data that is currently available in the Relay store (for more details, see our [Fetch Policies](https://www.internalfb.com/intern/wiki/Relay/guided-tour-of-relay/reusing-cached-data-for-rendering/#fetch-policies) and [Garbage Collection](https://www.internalfb.com/intern/wiki/Relay/guided-tour-of-relay/reusing-cached-data-for-rendering/#garbage-collection-in-re) guides):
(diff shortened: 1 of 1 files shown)


website/docs/api-reference/hooks/use-query-loader.md
```diff
@@ -90,7 +90,7 @@ A tuple containing the following values:
     * Parameters
         * `variables`: the variables with which the query is loaded.
         * `options`: `LoadQueryOptions`. An optional options object, containing the following keys:
-            * `fetchPolicy`: _*[Optional]*_ Determines if cached data should be used, and when to send a network request based on the cached data that is currently available in the Relay store (for more details, see our [Fetch Policies](https://www.internalfb.com/intern/wiki/Relay/guided-tour-of-relay/reusing-cached-data-for-rendering/#fetch-policies) and [Garbage Collection](https://www.internalfb.com/intern/wiki/Relay/guided-tour-of-relay/reusing-cached-data-for-rendering/#garbage-collection-in-re) guides):
+            * `fetchPolicy`: _*[Optional]*_ Determines if cached data should be used, and when to send a network request based on the cached data that is currently available in the Relay store (for more details, see our [Fetch Policies](../../guided-tour/reusing-cached-data/fetch-policies) and [Garbage Collection](../../guided-tour/reusing-cached-data/presence-of-data) guides):
```
astj, line 93: Same links as https://github.com/facebook/relay/blob/b17907beb3b46b05a4639af1d2af2382aa56ab0e/website/docs/api-reference/hooks/use-lazy-load-query.md?plain=1#L47 
I also confirmed those links by running websites locally and browsing http://localhost:3000/docs/next/api-reference/use-query-loader/ .

PR #3888, opened 2022-04-27
Copy edits to testing documentation

talum:
👋🏼 I noticed some typos, missing words, and opportunities for simplifying the prose in this set of documentation, so I wanted to make some suggestions!

In addition, this line is a bit confusing to me, but I wasn't sure about the intent:

> Collection of Mock Resolvers it's an object where **keys are names of GraphQL types (ID, String, User, Feedback, Comment, etc),** and values are functions which will return the default data for the type.


kassens:
This these are great edits, thanks for this touch up!

The mock resolver sentence should probably be something like:

> An object of Mock resolvers can be provided where the keys are GraphQL type names and the values functions that provide test data for these types.

I'll see that I update that sentence when merging the PR internally.

PR #3895, opened 2022-05-03
[bug report] refetchables don't include required metadata

tbezman:
I created a failing test which shows the issue.

I think the root cause is the following:

- The generated refetchable query doesn't include the required directives used by the source fragment.

### DIFF
--- a/compiler/crates/relay-compiler/tests/compile_relay_artifacts/fixtures/required-under-connection.expected
+++ b/compiler/crates/relay-compiler/tests/compile_relay_artifacts/fixtures/required-under-connection.expected
@@ -1,0 +1,25 @@
+==================================== INPUT ====================================
+fragment requiredUnderConnection on User @refetchable(queryName: "SomeNewQuery") {
+  topLevelComments(first: 10) @connection(key: "AnythingHere__topLevelComments") {
+    edges {
+      node @required(action: THROW) {
+        id
+      }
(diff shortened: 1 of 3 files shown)


captbaritone:
Sorry, I'm not sure I follow. Could you add a comment on the diff showing what you were expecting to see, or maybe an example of what incorrect behavior this leads to in the runtime?

captbaritone:
Would it be possible to reconstruct this failing tests with something that does not use `@connection`. The compiler error I see is actually related to an issue with connections not refetchable.

tbezman:
@captbaritone I'm also a bit confused. When I remove the `@required` directive from the query, the test at least generates artifacts. What is leading you towards the connection bit?

captbaritone:
The `@connection` transform adds some fields to the selection. These fields can conflict with `@required`, since the added version of the fields are not `@requried` and you are not allowed to have the same field both `@requried` and not required in the same selection.

I think the error you are seeing might be caused by that issue.

tbezman:
@captbaritone I see. So is the accepted workaround to just put your selections in another fragment? Do y'all advocate for this internally?

captbaritone:
Usually I just advocate for doing a manual null check inside your component/hook in these cases.

tbezman:
okay, thanks for the guidance!

PR #3898, opened 2022-05-05
VSCode Extension Readme & Semver Range

tbezman:
This should be the last few things we need to ship this extension.

### Extension README
- I added a README heavily inspired by the Flow for VSCode extension. We've got a nice little gif demo using the https://github.com/gallery-so/gallery repo. 
- I also added a `LICENSE` file which is copy pasted from the root of this repo

### Semver range

Since each shipped version of this extension will only support some range of the `relay-compiler` package, we have a check to ensure they're on the right version. If they aren't, they get a helpful message explaining what we expect, and where they should be.


vscode-extension/src/utils.ts
```diff
@@ -89,3 +89,50 @@ export async function findRelayBinary(
...
+  while (true) {
+    if (counter >= 5000) {
+      throw new Error(
+        'Could not find Relay binary after 5000 traversals. This is likely a bug in the extension code and should be reported to https://github.com/facebook/relay/issues',
+      );
+    }
+
+    counter += 1;
+
+    const possiblePacakgeJsonPath = path.join(
```
captbaritone, line 109:
```suggestion
    const possiblePackageJsonPath = path.join(
```

vscode-extension/src/utils.ts
```diff
@@ -89,3 +89,50 @@ export async function findRelayBinary(
 
   return null;
 }
+
+export async function findRelayCompilerVersion(
```
captbaritone, line 93: What do you think about having one function which returns both the binary and the version? Right now we have these theoretical issues where we have to consider what happens if we find the compiler binary but not the version, or the version but not the binary, when in reality we should always either find both or neither. Having a single function would ensure that:

1. We always get the version of the binary that we found, and we never somehow mess up and get the version from one place and the binary from another.
2. We can reason locally about how we want to handle the potential edge cases of finding one but not the other.

I think we should

1. traverse until we find `node_modules/relay-compiler` 
2. Check the version
3. If the version is modern enough, return that binary path

So maybe the unified function would return:

```
{kind: "compilerFound", path: string}
| {kind: "versionDidNotMatch", path: string, version: string} 
| {kind: "compilerNotFound" }
```

vscode-extension/README.md
```diff
@@ -0,0 +1,66 @@
...
+  - We support standard config file formats (`.yml`, `.js`, `.json`), and the the `relay` field in your `package.json`
+- Make sure you have the `relay-compiler` installed in your project. The bare minimum for this extension is v13.
+- Make sure you are able to run the `relay-compiler` command from the command line. If `yarn relay-compiler` works, it's very likely that the extension will work. See more on [debugging below](#debugging).
+- Remove any conflicting GraphQL extensions you have installed. We recommend using [GraphQL for VSCode](https://marketplace.visualstudio.com/items?itemName=kumar-harsh.graphql-for-vscode) since this one seems to have the least conflicts with the Relay extension.
+
+- If you install Flow via `flow-bin` (recommended), then we will by default try to use this installation of flow by looking for it in `node_modules`. This behavior can be disabled by setting the `flow.useNPMPackagedFlow` setting to `false`.
+- If you install Flow globally (for example, `npm install -g flow-bin`), make sure `flow` is available on your `PATH`. If neither are true, then a version bundled with this extension will be used, but this is discouraged as it may change over time.
```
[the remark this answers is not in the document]
tbezman: yah, big woops

vscode-extension/src/utils.ts
```diff
@@ -34,43 +36,28 @@ function getBinaryPathRelativeToPackageJson() {
...
-    return path.join(
-      '.',
-      'node_modules',
-      'relay-compiler',
-      binaryPathRelativeToPackageJson,
-    );
-  }
-
-  return null;
+  return binaryPathRelativeToPackageJson;
```
captbaritone, line 39: If we don't need to do any processing of `binaryPathRelativeToPackageJson` after deciding the path, we could probably just avoid the variable all together and just early return in each condition.

captbaritone (approved):
One nit, otherwise this looks good to me!

tbezman:
@captbaritone The linter wouldn't let me use `else if` so we've just got a bunch of ifs now with early returns.

captbaritone:
@tbezman Sounds like a lint rule worth disabling. But this is fine with me as is. Would you delete the empty CONTRIBUTING.md file? That's the only thing left that I can see then we can import/merge this!

PR #3905, opened 2022-05-11
Non Haste Module Resolution for Relay Resolvers.

tbezman:
This PR aims to introduce a pattern to support non-haste module resolution.

- add `pathdiff` crate to help us do path diffing to construct a relative import
- move `path_for_artifact` and friends into `ProjectConfig` as a long term home so other crates like `relay-typegen` and `relay-codegen` can access.
-  `typegen` and `codegen` both use `path_for_artifact` to figure out what path they're about to write to. we then use that path construct a relative import against the resolver's import path.
- add tests to `relay-runtime` to ensure this is working for cjs (query, and refetchable fragments)


compiler/crates/relay-codegen/src/build_ast.rs
```diff
@@ -142,10 +222,12 @@ impl<'schema, 'builder> CodegenBuilder<'schema, 'builder> {
     }
 
     fn build_operation(&mut self, operation: &OperationDefinition) -> AstKey {
+        let definition = ExecutableDefinition::Operation(operation.clone());
```
tbezman, line 225: cloning is no good. what are some ways to avoid this clone?

- create the artifact location here and pass that down instead? feels a little bit more targeted but it's nice to have the definition available to you. 
- use an Arc instead?

compiler/crates/relay-transforms/src/relay_resolvers.rs
```diff
@@ -264,6 +264,8 @@ impl<'program> RelayResolverFieldTransform<'program> {
                     }
                     let parent_type = field_type.parent_type.unwrap();
 
+                    println!("Found import_path: {}", import_path);
```
captbaritone, line 267: 😄

compiler/crates/relay-codegen/src/build_ast.rs
captbaritone: How about we take just the source location key and set it as a property on the `CodegenBuilder` struct?

So `CodegenBuilder` would have a new field `definition_source_location: Option<WithLocation<StringKey>>` which `build_operation` and `build_fragment` would be responsible for setting/unsetting before/after traversing their selections.

This would avoid the noisy threading of `definition` through all the methods.

compiler/crates/relay-codegen/src/build_ast.rs
```diff
@@ -744,15 +849,41 @@ impl<'schema, 'builder> CodegenBuilder<'schema, 'builder> {
...
+            match definition {
+                ExecutableDefinition::Operation(operation) => operation.name.item,
+                ExecutableDefinition::Fragment(fragment) => fragment.name.item,
+            },
+        );
+
+        let module_location = PathBuf::from_str(module.lookup()).unwrap();
+
+        let module_path = module_location.parent().unwrap();
+        let definition_artifact_location_path = definition_artifact_location.parent().unwrap();
```
captbaritone, line 873: ...and here?

compiler/crates/relay-codegen/src/build_ast.rs
```diff
@@ -744,15 +849,41 @@ impl<'schema, 'builder> CodegenBuilder<'schema, 'builder> {
...
+            },
+            match definition {
+                ExecutableDefinition::Operation(operation) => operation.name.item,
+                ExecutableDefinition::Fragment(fragment) => fragment.name.item,
+            },
+        );
+
+        let module_location = PathBuf::from_str(module.lookup()).unwrap();
+
+        let module_path = module_location.parent().unwrap();
```
captbaritone, line 872: Why do we need `.parent()` here?

compiler/crates/relay-codegen/src/build_ast.rs
```diff
@@ -744,15 +849,41 @@ impl<'schema, 'builder> CodegenBuilder<'schema, 'builder> {
...
+
+        println!(
+            "Diffing paths: {:?} : {:?}",
+            definition_artifact_location_path, module_path
+        );
+
+        let resolver_module_location =
+            pathdiff::diff_paths(module_path, definition_artifact_location_path).unwrap();
+
+        let import_path = resolver_module_location.join(module_location.file_name().unwrap());
```
captbaritone, line 883: Obviously we need to check the config and use the Haste name if that's how the project is configured, but I assume that's outside of the scope of this sketch.

compiler/crates/relay-codegen/src/build_ast.rs
tbezman: Nice, feels much better. I `definition_source_location` doesn't even have to be optional since we construct a new builder for each definition.

compiler/crates/relay-codegen/src/build_ast.rs
tbezman: this `pathdiff` library strictly works for directories. If you don't strip the file off the end, we'll end up with some funky (incorrect) paths

tbezman:
@rbalicki2 I made some changes to the typegen crate that may have stepped on your toes a bit. I want to run these changes by you and ensure they're up to your expectations.

- I made `TypegenOptions` a bit more baggy by removing some of the fields and replacing them with direct calls to `project_config`. 
  - The exception here is `has_unified_outputs` which I think warrants its own field since there is a context change in the significance of calling `project_config.output.some()` directly.
- Added `definition_source_location` to the `TypegenOptions` since I wanted to to be available to `visit_relay_resolver` without threading the value through a bunch of function calls.

All of this is making me think maybe we should rename `TypegenOptions` to `TypegenContext` and make all of the `visit` functions impls on that context so they can all reach into self for the variables instead of threading them through.

Curious to hear your thoughts on this.

captbaritone (commented):
Amazing work here! This looks spot on to me, just a couple smaller nits.

compiler/crates/relay-compiler/src/config.rs
```diff
@@ -517,7 +517,11 @@ impl fmt::Debug for Config {
         } = self;
 
         fn option_fn_to_string<T>(option: &Option<T>) -> &'static str {
-            if option.is_some() { "Some(Fn)" } else { "None" }
+            if option.is_some() {
+                "Some(Fn)"
+            } else {
+                "None"
+            }
```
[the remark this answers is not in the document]
tbezman: @captbaritone do we have a something in CI to enforce formatting consistency on the rust side? is there precedent for this?

relay.code-workspace
```diff
@@ -0,0 +1,7 @@
+{
+    "folders": [
+        {"path": "./compiler"},
+        {"path": "./vscode-extension"},
+        {"path": "./packages"}
+    ]
+}
```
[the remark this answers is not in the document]
tbezman: whoops, didn't mean to commit this. this is just something I use locally and haven't put much deep thought into.

compiler/crates/relay-compiler/src/config.rs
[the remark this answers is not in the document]
captbaritone: I think we expect rustc 1.60.0 (`rustc version) `rustc --version`

compiler/crates/relay-codegen/src/build_ast.rs
tbezman: I'll leave a comment inline about this

compiler/crates/relay-compiler/src/config.rs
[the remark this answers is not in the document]
tbezman: how do you do rust formatting? I have this config in my vscode
image

compiler/crates/relay-config/src/project_config.rs
```diff
@@ -258,3 +266,87 @@ impl Debug for ProjectConfig {
...
+                TypegenLanguage::Flow | TypegenLanguage::JavaScript => {
+                    format!("{}.graphql.js", definition_name)
+                }
+                TypegenLanguage::TypeScript => format!("{}.graphql.ts", definition_name),
+            }
+        };
+        self.create_path_for_artifact(source_file, filename)
+    }
+
+    pub fn relative_import_between_definition_and_js_module(
```
tbezman, line 319: @captbaritone better name ?

compiler/crates/relay-config/src/project_config.rs
```diff
@@ -258,3 +266,108 @@ impl Debug for ProjectConfig {
...
```
captbaritone, line 319:
```suggestion
    /// Generates a relative import path in Common JS projects, and a module name in Haste projects.
    pub fn js_module_import_path(
```

packages/relay-runtime/util/__tests__/resolvers/__generated__/RelayReaderResolverTest1FragmentRefetchableQuery.graphql.js
```diff
@@ -0,0 +1,142 @@
+/**
```
captbaritone, line 1: Cool. This only generates a normalization artifact, so there's not reference to the actual resolver module. 👍

rbalicki2:
Re: changes to typegen. Looks great!
* Renaming sounds good
* Re: adding fields to `TypegenContext`/restructuring it. Go for it. As long as we keep `TypegenContext` immutable, I'm happy. The only reason to thread things through is is they mutated and then eventually consumed (e.g. `EncounteredEnums` is mutated when visiting, then enum imports are printed.)

captbaritone:
⏩ ⏩ ⏩

PR #3906, opened 2022-05-11
VSCode Extension | Support custom root directory / config path

tbezman:
To enable projects who have their relay config / node modules in a nested directory, we provide two new options. 

- `relay.rootDirectory` (default: VSCode project root) A path relative to the root of your VSCode project for the extension to work from. The default value is the root of your project. This will change where we start looking for the relay-compiler node module. This will also affect where the LSP server is started, therefore affecting how the relay config is found. This is helpful if your project is in a nested directory.

- `relay.pathToConfig` (default: null) Path to a relay config relative to the `rootDirectory`. This is helpful if your relay project is in a nested directory.

I also added some more logging to help debug.

### DIFF
--- a/vscode-extension/README.md
+++ b/vscode-extension/README.md
@@ -31,7 +31,7 @@ Search for "Relay for VSCode" in the VS Code extensions panel or install through
 
 - `relay.rootDirectory` (default: VSCode project root) A path relative to the root of your VSCode project for the extension to work from. The default value is the root of your project. This will change where we start looking for the relay-compiler node module. This will also affect where the LSP server is started, therefore affecting how the relay config is found. This is helpful if your project is in a nested directory.
 
-- `relay.pathToConfig` (default: null) Path to a relay config relative to the `rootDirectory`. This is helpful if your relay project is in a nested directory.
+- `relay.pathToConfig` (default: null) Path to a relay config relative to the `rootDirectory`. Without this, the compiler will search for your config. This is helpful if your relay project is in a nested directory.
 
 ## Features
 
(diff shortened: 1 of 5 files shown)


vscode-extension/src/languageClient.ts
```diff
@@ -113,6 +130,11 @@ export async function createAndStartClient(context: RelayExtensionContext) {
 
   client.registerFeature(new LSPStatusBarFeature(context));
 
+  context.outputChannel.appendLine(
+    `Starting the Relay Langauge Server with these options: ${JSON.stringify(
+      serverOptions,
+    )}`,
```
captbaritone, line 136: ❤️

vscode-extension/README.md
```diff
@@ -29,6 +29,10 @@ Search for "Relay for VSCode" in the VS Code extensions panel or install through
 
 - `relay.pathToBinary` (default: null) A path relative to the Relay binary relative to the root of your project. If this is not specified, we will try to find one in your `node_modules` folder.
 
+- `relay.rootDirectory` (default: VSCode project root) A path relative to the root of your VSCode project for the extension to work from. The default value is the root of your project. This will change where we start looking for the relay-compiler node module. This will also affect where the LSP server is started, therefore affecting how the relay config is found. This is helpful if your project is in a nested directory.
+
+- `relay.pathToConfig` (default: null) Path to a relay config relative to the `rootDirectory`. This is helpful if your relay project is in a nested directory.
```
captbaritone, line 34: Can we add a note here (and in package.json) saying that without this the compiler will attempt to locate the config on it's own?

captbaritone (changes requested):

PR #3908, opened 2022-05-12
Ignore BOM when lexing

### LINKED ISSUE #3798: [V13] UTF-8 with BOM encoded schema file causes 'unsupported character' compiler error
Since I've upgraded all my `relay` packages to v13, I've been getting the following error when running the `relay-compiler`:

image

The configuration in my `package.json` file is the following:

```json
"relay": {
    "language": "typescript",
    "src": "./src",
    "schema": "./src/schema.graphql",
    "exclude": [
      "**/node_modules/**",
      "**/__mocks__/**",
      "**/__generated__/**"
    ],
    "artifactDirectory": "./src/__generated__"
  },
```

I have also setup the `babel-plugin-relay` as instructed with the `artifactDirectory`.

How would I go about debugging this? I tried to find whether the compiler has a verbose flag that would help me diagnose the issue, but I haven't been successful.

tobias-tengler:
If a GraphQL schema file should be encoded as `UTF-8 with BOM`, the lexer will skip the [BOM](https://en.wikipedia.org/wiki/Byte_order_mark) (`\ufeff`) character.

I've also added a regression snapshot test.

Closes #3798


captbaritone:
Rather than a snapshot test (where the BOM is very easy to overlook) could we create a unit test that uses an explicit escaped string?

tobias-tengler:
> Rather than a snapshot test (where the BOM is very easy to overlook) could we create a unit test that uses an explicit escaped string?

Done :)
I moved my test out of the schema document parsing tests and added it as a unit test directly to the lexer tests. I hope that's alright!

PR #3912, opened 2022-05-17
[VSCode Extension] GraphQL syntax higlighting

tbezman:
Look ma, it works.

image

I ripped all of this configuration from https://github.com/graphql/graphiql/tree/main/packages/vscode-graphql. They're licensed under MIT so we should be golden here.

Some weird thought that came up while copying the config. We "support" syntax highlighting on vue / svelte / markdown files. These are all files that the compiler won't pick up on right? Do we want to explicitly remove these or keep them just because why not?

### Conflicting Extensions

I tried enabling a few extensions which support graphql syntax. I didn't see any errors or oddities. I googled around and didn't find anything from the VSCode docs explaining what will happen. Maybe we should just 🤷 this one off.


vscode-extension/README.md
```diff
@@ -36,9 +36,10 @@ Search for "Relay for VSCode" in the VS Code extensions panel or install through
 ## Features
 
 - IntelliSense
-- Go to Definition for fragments, fields, GraphQL types, etc.
-- Diagnostics (Errors, Warnings)
 - Hover type information
+- GraphQL syntax highlighting
```
captbaritone, line 40:
```suggestion
- GraphQL syntax highlighting within .graphql and JavaScript/TypeScript files.
```

vscode-extension/README.md
```diff
@@ -16,7 +16,7 @@ Search for "Relay for VSCode" in the VS Code extensions panel or install through
   - We support standard config file formats (`.yml`, `.js`, `.json`), and the the `relay` field in your `package.json`
```
captbaritone, line 16: Can we add a "Credits" section at the bottom of this file that acknowledges where we got these files? That will help future contributors know where they came from and also make it clear that this is not our original work.

Also, let's drop this PR in the `#graphql-lsp` discord channel to let them know our plans. Since we're taking their code, I'd like to make sure they feel good about it and don't get the impression that we are trying to pass their code off as our own.

captbaritone (changes requested):
This is great. You're fast! I think we should omit the syntax higlighting for file types that Relay can't parse/understand. (Reason, Markdown, svelte etc). Open to push back if you think there's some good reason to keep them.

I'm fine with YOLOing the issue of potential conflicts with other syntax highlighting. I also didn't observe any conflicts when running multiple.

vscode-extension/README.md
```diff
@@ -16,7 +16,7 @@ Search for "Relay for VSCode" in the VS Code extensions panel or install through
...
 - Make sure you have the `relay-compiler` installed in your project. The bare minimum for this extension is v13.
 - Make sure you are able to run the `relay-compiler` command from the command line. If `yarn relay-compiler` works, it's very likely that the extension will work. See more on [debugging below](#debugging).
-- Remove any conflicting GraphQL extensions you have installed. We recommend using [GraphQL for VSCode](https://marketplace.visualstudio.com/items?itemName=kumar-harsh.graphql-for-vscode) since this one seems to have the least conflicts with the Relay extension.
+- Remove / disable any conflicting GraphQL extensions you have installed.
```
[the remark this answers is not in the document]
tbezman: @captbaritone there's definitely conflicts with jump to definition. well, at least you get two options presented which I think is really annoying.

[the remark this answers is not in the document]
captbaritone: Got it. Should we leave the link in then?

vscode-extension/README.md
tbezman: https://discord.com/channels/625400653321076807/863688312756371476/976322813650665573

vscode-extension/grammars/graphql.js.json
```diff
@@ -0,0 +1,67 @@
...
+    "jsx",
+    "mjs",
+    "cjs",
+    "es6",
+    "es",
+    "esm",
+    "ts",
+    "tsx",
+    "vue",
+    "svelte"
```
captbaritone, line 13: Should we drop these?

vscode-extension/grammars/graphql.rescript.json
```diff
@@ -0,0 +1,25 @@
+{
+  "fileTypes": ["res", "resi"],
```
captbaritone, line 2: I don't think regular Relay can parse these files.

captbaritone (changes requested):

vscode-extension/grammars/graphql.rescript.json
tbezman: @zth what are the valid file types for rescript? i'm not familiar.

vscode-extension/grammars/graphql.js.json
tbezman: oh, didn't realize they were declared in both places.

vscode-extension/grammars/graphql.rescript.json
zth: @tbezman it's the ones you've got there, but like @captbaritone says regular Relay can't parse that, and rescript-relay has its own dedicated extension that takes care of this + integrating the lsp etc: https://github.com/zth/vscode-rescript-relay, so there's no need to do anything for ReScript specifically here.

Thanks for thinking of me though! 😃

captbaritone:
Once we remove the rescript stuff I'll import.

vscode-extension/README.md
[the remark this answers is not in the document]
tbezman: Well, now you don't need another extension installed. The only reason we recommended the other one was for the syntax highlighting.

vscode-extension/README.md
Kennysong55: is not good...

vscode-extension/package.json
```diff
@@ -23,6 +23,51 @@
...
+      {
+        "injectTo": [
+          "source.rescript"
+        ],
+        "scopeName": "inline.graphql.res",
+        "path": "./grammars/graphql.rescript.json",
+        "embeddedLanguages": {
+          "meta.embedded.block.graphql": "graphql"
+        }
+      }
```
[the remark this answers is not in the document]
tbezman: you're right, this worked. why don't we need this? isn't this what's required to get the graphql`` to work?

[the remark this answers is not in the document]
captbaritone: I think that block was specific to rescript

vscode-extension/README.md
alunyov: is there a way to define extension dependencies somehow? Saying that in order for relay extension to work the `graphql` extension is also have to be installed? Or maybe create an extension pack?

alunyov: or can we install this grammar via `npm` somehow?

tbezman: I mean we could definitely pull it off through an npm dep. Requiring it as an extension would cause conflicts with the Relay extension potentially. We strictly want the grammars.

Can we elaborate on that "is not good..." comment?

PR #3913, opened 2022-05-17
[VSCode Extension] Start the compiler in watch mode

tbezman:
Alright, so the main goal here is to have the compiler run in watch mode alongside the lsp when a user opens a project. A couple of things had to change to support this.

- Two configs for the compiler output level. one for the lsp, one for the compiler.
- New config `autoStartCompiler` which runs the compiler in watch mode (default false)
- Changed `restartLanguageServer` to `restart` since you definitely want to restart both if you're restarting one.
  - This will restart the compiler if it was previously running.
- Added `relay.startCompiler`, and `relay.stopCompiler`
- Rearchitected `RelayExtensionContext` to include the relay binary path since it's needed in other contexts.
  - You need it when you're restarting the compiler.
  - You need it when you're restarting the LSP
  - I didn't want to tie together starting / restarting the lsp / compiler with looking for the relay binary. Now we can update the binary path / root dir in isolation.

### DIFF
--- a/vscode-extension/.eslintrc
+++ b/vscode-extension/.eslintrc
@@ -13,6 +13,8 @@
 
     // Opinionated
     "import/prefer-default-export": "off",
-    "no-await-in-loop": "off"
+    "no-await-in-loop": "off",
+    "no-use-before-define": "off",
+    "@typescript-eslint/no-use-before-define": ["error", { "functions": false }]
   }
 }
--- a/vscode-extension/README.md
+++ b/vscode-extension/README.md
@@ -20,6 +20,8 @@ Search for "Relay for VSCode" in the VS Code extensions panel or install through
 
 ## Configuration
 
+- `relay.startCompiler` (default: `false`) Whether or not we should automatically start the Relay Compiler in watch mode when you open a project.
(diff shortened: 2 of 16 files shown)


captbaritone (commented):
This is so good (and fast!). My original idea was to run it in an embedded terminal (I assume an extension can do that?) rather than running it as a process and piping the output to the output channel. Output channel seems cleaner in some ways, but there are two downsides I'm tracking:

1. I assume we don't get colored output in the output view? (Or do we?)
2. Users have to know the special command to stop/restart the compiler, rather than just using their existing knowledge of how to stop a process in the terminal

On the flip side, using the output channel probably gives us more control, like the ability to prompt users to switch to our output channel under certain circumstances, and track if it's already running.

tbezman:
@captbaritone good callout on using the terminal. Looks like we can keep the same level of control (roughly) while getting the benefits of the terminal. Only a few small things had to change. This feels a lot better.

https://user-images.githubusercontent.com/6754223/168951370-84b52bc4-3c82-4df5-ae31-bee7ed3f22ac.mov

PR #3914, opened 2022-05-18
Example of bug with client edges file location

captbaritone:
Differential Revision: D36464645


captbaritone:
Oops. This didn't pick up the generated file for some reason.

PR #3917, opened 2022-05-19
Allow persisting to also include the query text for safe migration

tomgasson:
Hi folks.

We're attempting to start using persisted queries. However, we're not 100% confident in our ability to switch over to them in one big step safely.

Fetching the persisted documents is a new capability for our GraphQL server. If something goes wrong fetching the document, or if it introduces unexpected latency to our existing production queries we want to be able to fallback to the query text until we're more confident in its maturity.

Relay currently supports outputting the query text, or outputting the persisted document id. This PR introduces a middle state, where the text and id are both output. When both are sent with the GraphQL request from the client to the server, the server is able to control the rollout of persisted queries via a feature flag.

This PR introduces a `safeMigration` bool to the remote persist config, which when set will also output the query text in the operation params as well as the id
```js
persistConfig: {
  url: ...,
  params: ...,
  includeQueryText: true
}
```

The benefits of persisted queries (saving the network bytes) won't come until we've progressed past the migration stage, but it makes it possible to do the move in the first place.

### DIFF
--- a/compiler/crates/relay-codegen/src/build_ast.rs
+++ b/compiler/crates/relay-codegen/src/build_ast.rs
@@ -1870,20 +1870,16 @@ impl<'schema, 'builder, 'config> CodegenBuilder<'schema, 'builder, 'config> {
         // Construct metadata object
         let mut params_object = vec![];
 
-        match (&request_parameters.text, request_parameters.id) {
-            (Some(ref text), _) => {
-                params_object.push(ObjectEntry {
-                    key: CODEGEN_CONSTANTS.cache_id,
-                    value: Primitive::RawString(md5(&text)),
-                });
-            }
-            (None, None) => {
-                params_object.push(ObjectEntry {
-                    key: CODEGEN_CONSTANTS.cache_id,
-                    value: Primitive::RawString(md5(operation.name.item.0.lookup())),
-                });
-            }
-            _ => (),
+        if let Some(ref text) = &request_parameters.text {
+            params_object.push(ObjectEntry {
+                key: CODEGEN_CONSTANTS.cache_id,
+                value: Primitive::RawString(md5(&text)),
+            });
+        } else if request_parameters.id.is_none() {
+            params_object.push(ObjectEntry {
+                key: CODEGEN_CONSTANTS.cache_id,
+                value: Primitive::RawString(md5(operation.name.item.0.lookup())),
+            });
         }
 
         params_object.push(ObjectEntry {
--- a/compiler/crates/relay-codegen/src/build_ast.rs
+++ b/compiler/crates/relay-codegen/src/build_ast.rs
@@ -1868,24 +1868,25 @@ impl<'schema, 'builder, 'config> CodegenBuilder<'schema, 'builder, 'config> {
         metadata_items.sort_unstable_by_key(|entry| entry.key);
 
         // Construct metadata object
-        let metadata_prop = ObjectEntry {
-            key: CODEGEN_CONSTANTS.metadata,
-            value: Primitive::Key(self.object(metadata_items)),
-        };
-        let name_prop = ObjectEntry {
-            key: CODEGEN_CONSTANTS.name,
-            value: Primitive::String(request_parameters.name),
-        };
-        let operation_kind_prop = ObjectEntry {
-            key: CODEGEN_CONSTANTS.operation_kind,
-            value: Primitive::String(match request_parameters.operation_kind {
-                OperationKind::Query => CODEGEN_CONSTANTS.query,
-                OperationKind::Mutation => CODEGEN_CONSTANTS.mutation,
-                OperationKind::Subscription => CODEGEN_CONSTANTS.subscription,
-            }),
-        };
+        let mut params_object = vec![];
 
-        let id_prop = ObjectEntry {
+        match (&request_parameters.text, request_parameters.id) {
+            (Some(ref text), _) => {
+                params_object.push(ObjectEntry {
+                    key: CODEGEN_CONSTANTS.cache_id,
+                    value: Primitive::RawString(md5(&text)),
+                });
+            }
+            (None, None) => {
+                params_object.push(ObjectEntry {
+                    key: CODEGEN_CONSTANTS.cache_id,
+                    value: Primitive::RawString(md5(operation.name.item.0.lookup())),
+                });
+            }
+            _ => (),
+        }
+
+        params_object.push(ObjectEntry {
             key: CODEGEN_CONSTANTS.id,
             value: match request_parameters.id {
                 Some(QueryID::Persisted { id, .. }) => Primitive::RawString(id.clone()),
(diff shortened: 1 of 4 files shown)


maraisr:
Love it @tomgasson good moment this.

compiler/crates/relay-codegen/src/printer.rs
```diff
@@ -91,7 +92,11 @@ pub fn print_request_params(
     top_level_statements: &mut TopLevelStatements,
 ) -> String {
     let mut request_parameters = build_request_params(operation);
-    request_parameters.id = query_id;
+    request_parameters.request_format = if let Some(query_id) = &query_id {
+        Some(RequestFormat::ID(query_id.clone()))
+    } else {
+        None
+    };
```
captbaritone, line 99: Nit, could use `Option::map` here:

```suggestion
    request_parameters.request_format = query_id.map(|id| {
      RequestFormat::ID(id.clone())
    });
```

(untested)

compiler/crates/relay-codegen/src/build_ast.rs
```diff
@@ -1850,64 +1850,64 @@ impl<'schema, 'builder, 'config> CodegenBuilder<'schema, 'builder, 'config> {
...
+                key: CODEGEN_CONSTANTS.id,
+                value: Primitive::Null,
+            }
         };
 
+        let mut params_object = vec![];
+
+        match request_parameters.request_format {
+            Some(
+                RequestFormat::Text(Some(ref text)) | RequestFormat::Migration(_, Some(ref text)),
```
captbaritone, line 1881: Can `RequestFormat` have a method `text()` to get the text from the underlying variant?

compiler/crates/relay-codegen/src/build_ast.rs
```diff
@@ -1850,64 +1850,64 @@ impl<'schema, 'builder, 'config> CodegenBuilder<'schema, 'builder, 'config> {
...
+        }
+
+        params_object.push(id_prop);
+        params_object.push(metadata_prop);
+        params_object.push(name_prop);
+        params_object.push(operation_kind_prop);
+
+        match request_parameters.request_format {
+            Some(
+                RequestFormat::Text(Some(ref text)) | RequestFormat::Migration(_, Some(ref text)),
```
captbaritone, line 1900: See above: Can RequestFormat have a method text() to get the text from the underlying variant?

compiler/crates/relay-codegen/src/build_ast.rs
```diff
@@ -1850,64 +1850,64 @@ impl<'schema, 'builder, 'config> CodegenBuilder<'schema, 'builder, 'config> {
...
+                key: CODEGEN_CONSTANTS.cache_id,
+                value: Primitive::RawString(md5(operation.name.item.0.lookup())),
+            }),
+            _ => (),
+        }
+
+        params_object.push(id_prop);
+        params_object.push(metadata_prop);
+        params_object.push(name_prop);
+        params_object.push(operation_kind_prop);
```
captbaritone, line 1896: Could we unconditionally initialize the params object with the ones we know we are going to include and then just have one push for the request_format case?

compiler/Cargo.lock
```diff
@@ -1580,6 +1580,7 @@ dependencies = [
  "graphql-test-helpers",
  "intern",
  "lazy_static",
+ "relay-config",
```
captbaritone, line 1583: Is this intentional?

compiler/crates/relay-codegen/src/ast.rs
```diff
@@ -156,15 +156,20 @@ pub enum QueryID {
     External(StringKey),
 }
 
-pub struct RequestParameters<'a> {
-    pub id: &'a Option<QueryID>,
+pub enum RequestFormat {
+    ID(QueryID),
+    Text(Option<String>),
+    Migration(QueryID, Option<String>),
```
captbaritone, line 162: Can we rename this to something more concrete? `TextAndID` maybe?

captbaritone (changes requested):
Thanks for your PR and for your patience! Seems like a reasonable idea! I left some feedback.

packages/relay-runtime/util/RelayConcreteNode.js
```diff
@@ -43,30 +43,20 @@ export type ProvidedVariablesType = {+[key: string]: {get(): mixed}};
...
-      +name: string,
-      +operationKind: 'mutation' | 'query' | 'subscription',
-      +providedVariables?: ProvidedVariablesType,
-      +metadata: {[key: string]: mixed, ...},
-    };
+export type RequestParameters = {
+  +cacheID?: string,
+  +id: string | null,
+  +text: string | null,
+  // common fields
```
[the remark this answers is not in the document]
tomgasson: This is needed in the implementation of the `fetch` function, so that you have correct types when sending both values (feature flagged) to the server as you migrate

compiler/crates/relay-codegen/src/build_ast.rs
tomgasson: I've done it like this in order to maintain the existing artifact ordering. If I change this, it will invalidate all existing artifacts (even without this feature on). If that's okay to for you at Meta, I can do this - but wanted to reduce your churn

compiler/crates/relay-codegen/src/ast.rs
tomgasson: I've simplified the structure to no longer need this additional struct

Lalitha-Iyer:
@tomgasson and @captbaritone Thanks a ton for seeing this PR through. Do you know when we expect this to be merged?

packages/relay-runtime/util/RelayConcreteNode.js
```diff
@@ -43,30 +43,20 @@ export type ProvidedVariablesType = {+[key: string]: {get(): mixed}};
...
-      +cacheID: string,
-      +id: null,
-      +text: string | null,
-      // common fields
-      +name: string,
-      +operationKind: 'mutation' | 'query' | 'subscription',
-      +providedVariables?: ProvidedVariablesType,
-      +metadata: {[key: string]: mixed, ...},
-    };
+export type RequestParameters = {
```
captbaritone, line 50: I've imported this, and it triggers a number of Flow errors because previously people could unconditionally get a string by doing `query.params.id === null ? query.params.cacheID : query.params.id` now the types don't let you do that.

Could we update these types to communicate that we always have a cacheID if id is `null`? I suspect that means going back to a union.

compiler/crates/relay-config/src/project_config.rs
```diff
@@ -62,6 +62,9 @@ pub struct RemotePersistConfig {
         deserialize_with = "deserialize_semaphore_permits"
     )]
     pub semaphore_permits: Option<usize>,
+
+    #[serde(default)]
+    pub include_query_text: bool,
```
captbaritone, line 67: Should we support this for `LocalPersistConfig` as well for symmetry?

captbaritone: If so, what about adding a method to `PersistConfig`?


```rust
impl PersistConfig {
    pub fn include_query_text(&self) -> bool {
        match self {
            PersistConfig::Remote(remote_config) => remote_config.include_query_text,
            PersistConfig::Local(local_config) => local_config.include_query_text,
        }
    }
}
```

compiler/crates/relay-compiler/src/artifact_content/content.rs
```diff
@@ -178,11 +180,24 @@ pub fn generate_operation(
...
+            request_parameters.id = id_and_text_hash;
+            request_parameters.text = text.clone();
+        }
+        (Some(_), _) => {
+            request_parameters.id = id_and_text_hash;
+        }
+        (None, _) => {
+            request_parameters.text = text.clone();
+        }
+    }
```
captbaritone, line 199: I find these tuple matches really hard to follow. I _think_ this is the same logic? (See my comment elsewhere for the `include_query_text` method.

```suggestion
    if id_and_text_hash.is_some() {
        request_parameters.id = id_and_text_hash;
        if project_config
            .persist
            .as_ref()
            .map_or(false, |config| config.include_query_text())
        {
            request_parameters.text = text.clone();
        }
    } else {
        request_parameters.text = text.clone();
    }
```

compiler/crates/relay-codegen/src/build_ast.rs
```diff
@@ -1919,24 +1919,25 @@ impl<'schema, 'builder, 'config> CodegenBuilder<'schema, 'builder, 'config> {
...
+                });
+            }
+            (None, None) => {
+                params_object.push(ObjectEntry {
+                    key: CODEGEN_CONSTANTS.cache_id,
+                    value: Primitive::RawString(md5(operation.name.item.0.lookup())),
+                });
+            }
+            _ => (),
+        }
```
captbaritone, line 1938: I find these tuple matches hard to follow. I think this simpler if/else if encodes the same logic?

```suggestion
        if let Some(ref text) = &request_parameters.text {
            params_object.push(ObjectEntry {
                key: CODEGEN_CONSTANTS.cache_id,
                value: Primitive::RawString(md5(&text)),
            });
        } else if request_parameters.id.is_none() {
            params_object.push(ObjectEntry {
                key: CODEGEN_CONSTANTS.cache_id,
                value: Primitive::RawString(md5(operation.name.item.0.lookup())),
            });
        }
```

captbaritone (changes requested):
I've imported the diff and it uncovered some Flow errors. I also left some requests to avoid the tuple matches and replace with nested if/else statements

captbaritone:
This has landed. Folks can try it out by using the @main NPM tag. Make sure you upgrade both `relay-runtime`, `relay-compiler` and `react-relay`.

compiler/crates/relay-config/src/project_config.rs
eMerzh: @tomgasson  @captbaritone  discovering the feature with the release note, but it's me or local persist isn't supported? 
is there a reason for that?

tomgasson: It should be:
```js
persist: {
  file: "foo.json",
  includeQueryText: true
}
```
What issue are you having?

reckter:
FYI, if you use local persist query you have to use the field name in snake_case, not camelCase.
it has to be `include_query_text: true`.

PR #3923, opened 2022-05-24
Support CommonJS for @no_inline

tbezman:
The goal of this PR is to support `@no_inline` in CommonJS, but I did a little clean up along the way.

We used to clone the `no_inline` directive to all of the FragmentSpreads that used that source fragment. That felt like a hack so I switched it over to use `NoInlineMetadata` which uses the cool and very hip associated data stuff.


packages/relay-runtime/store/__tests__/no_inline/NoInline-test.js
```diff
@@ -0,0 +1,56 @@
...
+'use strict';
+
+const {getRequest} = require('../../../query/GraphQLTag');
+const {createReaderSelector} = require('../../../store/RelayModernSelector');
+const {graphql} = require('relay-runtime/query/GraphQLTag');
+const {
+  createOperationDescriptor,
+} = require('relay-runtime/store/RelayModernOperationDescriptor');
+const {createMockEnvironment} = require('relay-test-utils-internal');
+const SomeDeeplyNestedFragment = require('./some/deeply/nested/__generated__/SomeDeeplyNestedFragment.graphql');
```
captbaritone, line 21: Maybe a comment explaining why we've used this strange directory structure?

PR #3930, opened 2022-06-01
[Compiler Explorer] Persist explorer state in the URL/local storage

captbaritone:
This should allow the compiler explorer to be used as a mechanism for sharing minimal reproducible compiler bugs and other examples of compiler behavior.

The general idea is to consolidate all the state in the compiler explorer into a single object (Redux style):

* Schema
* Document
* Feature Flags
* Language
* Selected tab

And then (json) serialize and compress that state and persist it to the URL hash and local storage. Note that by using the URL hash we ensure the content of the explorer is never sent to the server, and thus the privacy of the content is improved.

## Test Plan

https://user-images.githubusercontent.com/162735/171328459-a675c8e5-a9a6-4a27-af2e-1230245b84d2.mov

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -25,6 +25,7 @@
     "classnames": "^2.2.6",
     "clsx": "^1.1.1",
     "docusaurus-plugin-internaldocs-fb": "0.10.7",
+    "lz-string": "^1.4.4",
     "monaco-editor": "^0.24.0",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
--- a/website/src/compiler-explorer/ExplorerState.js
+++ b/website/src/compiler-explorer/ExplorerState.js
@@ -1,0 +1,184 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ */
+
+import * as React from 'react';
+import * as LZString from 'lz-string';
+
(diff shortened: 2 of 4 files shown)


captbaritone:
@orta Curious to hear if you have any thoughts/suggestions here.

orta:
First of all, I love it (and kinda didn't know this existed) 

There's three things I'd think about:

- A change which is a bit more effort but felt worth it for me, is to not merge all the URL's into a single blob, but to encode each individual part as its own query param from the diff of the defaults. It makes the URLs quite a lot more readable and more obvious that it represents the users state to folks seeing the URL for the first time. 

e.g. from `'compiler-explorer/#asdadadsdsfs'` to `compiler-explorer/#code=asdadadsdsfs?flag=on&other=false' 

-  Add a 'share' button on the page somewhere, even though the URL is there and obvious it is something which gets used _far_ more than you would expect.

- Using `pushState` (or `replaceState`) on keychange will trash the users browser history, there are people who care about this. After a few years of push back I eventually opted to only update the state when monaco loses focus in the TS playground. This is enough for when you are going to copy the URL, and rare enough that it doesn't annoy history carers

orta:
Worth thinking about from the other side too, the relay vscode extension could be able to have a way to generate a url of your schema (or the relevant subset) and the document you're looking at too

orta:
Yeah, it doesn't seem to unreasonable to me that you could take a query, then only get all relevant types _used in that query_ and jam that into a URL, which could _probably_ be OK - there's a limit, but memory serves that is around 4k of text. I'm sure F will have queries with types bigger than that even after gzip, but I would be surprised if that was the majority of queries

PR #3933, opened 2022-06-01
Support query variables in `@inline` fragments

tbezman:
`[ERROR] ✖︎ Variables are not yet supported inside @inline fragments.`

This PR will enable query variables to be used inside of `@inline` fragments. I will work to get `@argument` style variables working in a follow up PR. I'll try to outline what work I think is required to support fragment variables.

## Query Variables
Easy to turn on, I wrote a test to confirm that this works as expected.

## Fragment Variables with `@arguments`
Not so easy, I'll try to layout what I found while diving deep.

### How `useFragment` works today

When reading the parent operation definition, fragment pointers are created. If the reader AST for that fragment has an `args` field (this comes from `@argumentDefinitions`), the translated variables are inlined in the fragment pointer object.

When the useFragment for the fragment with `@argumentDefinitions` is used, the variables from the parent's pointer are used in conjunction with the query's variables.

### What needs to change?
It seems like `InlineDataFragmentSpread` should have an `args` field in the reader AST. If we had that, we could temporarily swap out `this._variables` with the translated variables for the duration of `_createInlineDataOrResolverFragmentPointer` using `getArgumentValues`.


captbaritone:
Awesome! This looks right to me. Thanks for including a runtime test!

PR #3935, opened 2022-06-01
Support `@argumentDefinitions` in `@inline` fragments

### LINKED ISSUE #3106: handle arguments on inline fragments
When I try to pass some arguments to my inline fragment I got (that I imagined would happen)

```
ERROR:
Variables are not yet supported inside @inline fragments.
```

My question is whether inline fragments will support variables or any information about it.

tbezman:
This PR is the followup to #3933. Now we support both query & fragment variables inside of `@inline` directives.

Fixes #3106 

### What changed?

- Added `args` to the `InlineDirectiveMetadata` struct
- Piped those args through to the AST in `build_ast`
- Updated `RelayReader` to temporarily swap out the variables while traversing an inline fragment


daniloab:
Wow, awesome!

packages/relay-runtime/store/__tests__/readInlineData-test.js
```diff
@@ -63,16 +64,120 @@ test('unwrap inline fragment data', () => {
 });
 
 describe('integration test with variables', () => {
+  test.only('works with fragment variables & query variables', () => {
```
captbaritone, line 67:
```suggestion
  test('works with fragment variables & query variables', () => {
```

packages/relay-runtime/store/RelayReader.js
```diff
@@ -1111,11 +1111,35 @@ class RelayReader {
...
+
     this._traverseSelections(
       fragmentSpreadOrFragment.selections,
       record,
       inlineData,
     );
+
+    // Put the old variables back
+    this._variables = oldVariables;
+
```
captbaritone, line 1142:
```suggestion

    let oldVariables = this._variables;
    
    // If the inline fragment spread has arguments, we need to temporarily
    // switch this._variables to include the fragment spread's arguments
    // for the duration of its traversal.
    if (fragmentSpreadOrFragment.args) {
      const spreadArguments = getArgumentValues(
        fragmentSpreadOrFragment.args,
        this._variables,
      );

      // Inherit the operation's arguments and give local fragment arguments higher priority
      this._variables = {...this._variables, ...spreadArguments};
    }
    
    this._traverseSelections(
      fragmentSpreadOrFragment.selections,
      record,
      inlineData,
    );

    // Put the old variables back
    this._variables = oldVariables;

```

packages/relay-runtime/util/ReaderNode.js
```diff
@@ -33,6 +33,7 @@ export type ReaderInlineDataFragmentSpread = {|
   +kind: 'InlineDataFragmentSpread',
   +name: string,
   +selections: $ReadOnlyArray<ReaderSelection>,
+  +args?: ?$ReadOnlyArray<ReaderArgument>,
```
captbaritone, line 36: I think ideally this would be:

```suggestion
  +args: ?$ReadOnlyArray<ReaderArgument>,
```

But having it optional is a nice way to make it safe to land runtime and compiler changes separately internally. Maybe add a comment that we should come back and make the property non-optional (the value can continue to be optional).

captbaritone (changes requested):
This looks great overall! Just one issue with some debug stuff still left in the tests, and one style suggestion for the runtime change.

packages/relay-runtime/util/ReaderNode.js
tbezman: @captbaritone so we'll basically follow this pr up right after with the non-optional?

captbaritone: Yeah, that would be perfect.

captbaritone (changes requested):

captbaritone:
Also, if you can rebase on `main` you should see some ESLint/Prettier/rustfmt errors coming from CI. I've made a few PRs which allows those to run on external PRs similarly to how they run on internal PRs.

packages/relay-runtime/store/RelayReader.js
```diff
@@ -1111,11 +1111,31 @@ class RelayReader {
...
+    // switch this._variables to include the fragment spread's arguments
+    // for the duration of its traversal.
+    if (fragmentSpreadOrFragment.args) {
+      const spreadArguments = getArgumentValues(
+        fragmentSpreadOrFragment.args,
+        this._variables,
+      );
+
+      // Inherit the operation's arguments and give local fragment arguments higher priority
+      this._variables = {...this._variables, ...spreadArguments};
```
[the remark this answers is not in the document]
tbezman: @captbaritone is the idea here to expose `argumentDefinitions` on the `InlineDataFragmentSpread` so we can re-use that function?

```diff
return {
  "fragment": {
    "argumentDefinitions": (v0/*: any*/),
    "kind": "Fragment",
    "metadata": null,
    "name": "readInlineDataTestFragmentAndQueryVariablesQuery",
    "selections": [
      {
        "alias": null,
        "args": null,
        "concreteType": "User",
        "kind": "LinkedField",
        "name": "me",
        "plural": false,
        "selections": [
          {
            "kind": "InlineDataFragmentSpread",
            "name": "readInlineDataTestFragmentAndQueryVariables",
+           "argumentDefinitions": GET THESE HERE SOME HOW
            "selections": [
              {
                "alias": "fragmentVariable",
                "args": [
                  {
                    "kind": "Variable",
                    "name": "scale",
                    "variableName": "theScale"
                  }
                ],
                "concreteType": "Image",
                "kind": "LinkedField",
                "name": "profile_picture",
                "plural": false,
                "selections": (v1/*: any*/),
                "storageKey": null
              },
              (v2/*: any*/)
            ],
            "args": [
              {
                "kind": "Variable",
                "name": "theScale",
                "variableName": "scale2"
              }
            ]
          }
        ],
        "storageKey": null
      }
    ],
```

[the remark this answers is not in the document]
captbaritone: I believe so. I think the `argumentDefinitions` include information about default values, which I believe is required to correctly compute the actual value of the arguments.

captbaritone:
Let me know if the rustfmt CI job is causing you grief. I just added it. I don't yet know a good way to enforce that folks developing locally are using the expected version of rustfmt.

tbezman:
@captbaritone no worries, yeah I can't seem to align my vscode with the version y'all are using. i think it's a noob thing but running `rustfmt {path/to/file}` always does the trick for me.

captbaritone:
Is this ready for another round? No hurry, just let me know when it is.

tbezman:
@captbaritone yup!

packages/relay-runtime/store/RelayReader.js
```diff
@@ -1111,11 +1113,37 @@ class RelayReader {
...
+      );
+
+      RelayConcreteVariables.getFragmentVariables(
+        fragmentSpreadOrFragment,
+        this._variables,
+        argumentVariables,
+      );
+
+      // Inherit the operation's arguments and give local fragment arguments higher priority
+      this._variables = {...this._variables, ...argumentVariables};
```
captbaritone, line 1135:
```suggestion
       // Inherit the operation's arguments and give local fragment arguments higher priority
      this._variables = RelayConcreteVariables.getFragmentVariables(
        fragmentSpreadOrFragment,
        this._variables,
        argumentVariables,
      );
```

packages/relay-runtime/store/__tests__/readInlineData-test.js
```diff
@@ -63,16 +64,120 @@ test('unwrap inline fragment data', () => {
 });
 
 describe('integration test with variables', () => {
+  test('works with fragment variables & query variables', () => {
+    const Fragment = graphql`
+      fragment readInlineDataTestFragmentAndQueryVariables on User
+      @inline
+      @argumentDefinitions(theScale: {type: "Float"}) {
```
captbaritone, line 71: Can we add a test that includes a fragment argument with a default value?

compiler/crates/relay-compiler/tests/compile_relay_artifacts/fixtures/inline-data-fragment-local-args.expected
```diff
@@ -19,53 +18,191 @@ fragment inlineDataFragmentLocalArgsProfile on User
     uri
   }
 }
-==================================== ERROR ====================================
-✖︎ Variables from @argumentDefinitions are not yet supported inside @inline fragments.
```
captbaritone, line 23: So exciting! Love seeing "not supported" tests turning green.

compiler/crates/graphql-ir/src/ir.rs
```diff
@@ -326,7 +326,7 @@ impl Condition {
 // Associated Types
 
 /// @ Name Arguments?
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
```
captbaritone, line 329: Just out of curiosity. Why do we need to derive `Hash` for these structs?

captbaritone (changes requested):
Just one change in Reader which I suspect was just a simple oversight, and then a test to validate that default values work.

compiler/crates/graphql-ir/src/ir.rs
tbezman: Started at `InlineDirectiveMetadata` needing `Hash`. Going down the chain from `InlineDirectiveMetadata` => `VariableDefinition` => `Directive`

tbezman:
@captbaritone we found a bug at Coinbase. Query => Regular Fragment => Inline Fragment. I was using `this._variables` instead of `this._owner.variables`. This is now inline with how `getSingularSelector` works. I added a unit test to confirm the bug and fixed the issue. Feeling really good about this now.

packages/relay-runtime/store/RelayReader.js
```diff
@@ -1111,11 +1113,40 @@ class RelayReader {
...
+        this._owner.variables,
+        argumentVariables,
+      );
+    } else {
+      this._variables = RelayConcreteVariables.getFragmentVariables(
+        fragmentSpreadOrFragment,
+        this._owner.variables,
+        {},
+      );
+    }
```
captbaritone, line 1139: Style nit: Would this work?
```suggestion
    const argumentVariables = fragmentSpreadOrFragment.args ? getArgumentValues(
      fragmentSpreadOrFragment.args,
      this._variables,
    ) : {};

    this._variables = RelayConcreteVariables.getFragmentVariables(
      fragmentSpreadOrFragment,
      this._owner.variables,
      argumentVariables,
    );
```

captbaritone (changes requested):
One style nit (sorry!)

captbaritone:
(Flow CI error is not caused by this diff)

tbezman:
alright @captbaritone . looking good?

packages/relay-runtime/util/ReaderNode.js
captbaritone: Actually, I'll do this internally.

PR #3953, opened 2022-06-07
Unify Prettier config and enforce Prettier in GitHub CI

captbaritone:
This PR starts with an intentional error to confirm that it reports errors in CI.


captbaritone:
Confirmed that with the existing error I trigger a CI failure:

Screen Shot 2022-06-07 at 4 06 08 PM

captbaritone:
Confirmed that after fixing the one error (`yarn prettier`), CI is back to green:
Screen Shot 2022-06-07 at 4 08 52 PM

PR #3954, opened 2022-06-08
Resolvers without fragments don't have dependencies

captbaritone:
Previously we would mark a resolver as having a dependency on the parent record. If the parent record was updated, would invalidate the resolver.

On a subsequent render, we would check the resolver's validity, and trigger a warning here: https://github.com/facebook/relay/blob/baa2e6d728672b59e31ffc564a89c7be960aa7e9/packages/relay-runtime/store/experimental-live-resolvers/LiveResolverCache.js#L405


captbaritone:
Discovered when updating https://github.com/captbaritone/redux-to-relay-with-live-resolvers-example to not use fragments.

PR #3960, opened 2022-06-09
Fix VSCode extension semver range

### LINKED ISSUE #3957: Relay VSCode Extension v0.0.7 does not support Relay v14.0.0
Congrats and thank you all for the v14 release!

It looks like the VSCode Extension has not been updated to support the latest Relay stable version of v14.0.0

Got this error in VS Code (extension version: v0.0.7):
>The installed version of the Relay Compiler is version: '14.0.0'. We found this version in the package.json at the following path: [path]/node_modules/relay-compiler/macos-arm64/relay This version of the extension supports the following semver range: '>=13.3.x <14'. Please update your extension / relay-compiler to accommodate the version requirements.

tomgasson:
I guess this was initially expected to be released as a 13.x release rather than as part of a new major

When a breaking change is made, it's worth bumping up the minimum.
Not sure if you do actually want to limit forward compatibility here. It's pretty hard to guess at what will break. Instead it probably makes more sense to warn that they're using an old version of the extension on a newer version of relay and that they can always upgrade the extension to suit

fixes #3957

### DIFF
--- a/vscode-extension/src/constants.ts
+++ b/vscode-extension/src/constants.ts
@@ -5,4 +5,4 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-export const SEMVER_RANGE = '>=13.3.x <14';
+export const SEMVER_RANGE = '>=13.3.x';


vscode-extension/src/constants.ts
```diff
@@ -5,4 +5,4 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-export const SEMVER_RANGE = '>=13.3.x <14';
+export const SEMVER_RANGE = '>=13.3.x';
```
richardguerre, line 8: There are no `13.3.x` releases, the last release in v13 was `13.2.0`. Also the README says the VSCode extension should work with relay-compiler v13 and above.

```suggestion
export const SEMVER_RANGE = '>=13.x';
```

captbaritone:
Sorry for the bad experience and thanks for the PR. We ended up having an internal fix in parallel. https://github.com/facebook/relay/commit/7c2200f1c79efd27406c96bdae7e8045c0fa3585

This change has been pushed out as 1.0.1 of the VSCode Extension.

PR #3967, opened 2022-06-10
Remove note on Haste requirement for resolvers

bigfootjon:
I tried this out (with a CommonJS-based project) and it works fine!

The generated query resolver correctly `require`s based on the relative path:

```
...
"kind": "RelayResolver",
"name": "greeting",
"resolverModule": require('./../user/UserGreetingResolver.js'),
"path": "viewer.greeting"
...
```

And it works at runtime

Not entirely sure if I'm modifying the docs in the right place


captbaritone:
Thanks for flagging! This limitation was removed by @tbezman  in #3967. In addition the restriction in the line above (about query variables) has been lifted as well by @alunyov. Would you mind removing that line too?

Could you also remove the lines from https://github.com/facebook/relay/blob/main/website/versioned_docs/version-v14.0.0/guides/relay-resolvers.md?plain=1#L250-L251

tbezman:
Thanks @bigfootjon !

PR #3969, opened 2022-06-10
VSCode extension pathToRelay doesn't allow relative paths

leoasis:
This PR is an attempt at allowing the `pathToRelay` configuration option in VSCode to have relative paths. When this happens, the path will be resolved relative to the `rootPath`.

This is to support shared settings across many developers in the same workspace, while also having a custom compiler instead of the default one coming from the relay-compiler package. Because the extension currently doesn't allow `pathToRelay` to be relative, this would mean that each developer in our repository would need a different value for this setting.

I'm not too familiar with how VSCode extensions or the LSP client works, so I'm sorry in advance if the change I'm proposing doesn't make sense. Happy to follow up with improvements based on your feedback.

I'm not sure why the `LanguageClient` doesn't allow a relative path in `serverOptions.command`, so what I did was convert that to an absolute path by leveraging `path.resolve` and using the `rootPath` value from context. I tried this in my setup, with the extension installed in my VSCode and it works as expected.

### DIFF
--- a/vscode-extension/src/languageClient.ts
+++ b/vscode-extension/src/languageClient.ts
@@ -11,6 +11,7 @@ import {
 } from 'vscode-languageclient';
 import { ServerOptions, LanguageClient } from 'vscode-languageclient/node';
 import { window } from 'vscode';
+import * as path from 'path';
 import { RelayExtensionContext } from './context';
 import { createErrorHandler } from './errorHandler';
 import { LSPStatusBarFeature } from './lspStatusBarFeature';
@@ -33,7 +34,10 @@ export function createAndStartLanguageClient(context: RelayExtensionContext) {
     options: {
       cwd: context.relayBinaryExecutionOptions.rootPath,
     },
-    command: context.relayBinaryExecutionOptions.binaryPath,
+    command: path.resolve(
+      context.relayBinaryExecutionOptions.rootPath,
+      context.relayBinaryExecutionOptions.binaryPath,
+    ),
     args,
   };
 


vscode-extension/src/languageClient.ts
```diff
@@ -33,7 +34,10 @@ export function createAndStartLanguageClient(context: RelayExtensionContext) {
     options: {
       cwd: context.relayBinaryExecutionOptions.rootPath,
     },
-    command: context.relayBinaryExecutionOptions.binaryPath,
+    command: path.resolve(
+      context.relayBinaryExecutionOptions.rootPath,
+      context.relayBinaryExecutionOptions.binaryPath,
+    ),
```
tbezman, line 40: @leoasis doesn't this break down if you specify an absolute path to a binary in your config? 

the use case for the absolute path is when I'm testing a locally built binary

leoasis: According to https://nodejs.org/api/path.html#pathresolvepaths it shouldn't:

_The given sequence of paths is processed from right to left, with each subsequent path prepended until an absolute path is constructed. For instance, given the sequence of path segments: /foo, /bar, baz, calling path.resolve('/foo', '/bar', 'baz') would return /bar/baz because 'baz' is not an absolute path but '/bar' + '/' + 'baz' is._

In the case you mentioned, the binary path would already be absolute so it should return it as is.

Haven't tried this though, will confirm on Monday or you can give it a go if you have a way to try this out before then

leoasis: Ok I just tried this out locally with an absolute path and as expected it continues to work.

tbezman (commented):
lgtm. nice @leoasis

captbaritone:
Thanks! And thanks @tbezman for doing review! ❤️

captbaritone:
Just shipped version 1.0.2 of the VSCode extension which includes this.

PR #3971, opened 2022-06-11
Fix prettier config in VSCode Extension

tbezman:
Some of the config brought over in #3953 borked some stuff.

I "fixed" the config and re ran prettier against the files.

### DIFF
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -137,10 +137,8 @@
     "arrowParens": "avoid",
     "bracketSameLine": true,
     "bracketSpacing": false,
-    "requirePragma": true,
     "singleQuote": true,
-    "trailingComma": "all",
-    "parser": "flow"
+    "trailingComma": "all"
   },
   "scripts": {
     "typecheck": "tsc --noEmit",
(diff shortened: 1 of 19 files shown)


vscode-extension/package.json
```diff
@@ -137,10 +137,8 @@
     "arrowParens": "avoid",
     "bracketSameLine": true,
     "bracketSpacing": false,
-    "requirePragma": true,
```
tbezman, line 140: I don't think we want this here, it seems tedious and unnecessary.

vscode-extension/package.json
```diff
@@ -137,10 +137,8 @@
...
     "singleQuote": true,
-    "trailingComma": "all",
-    "parser": "flow"
```
tbezman, line 143: Let prettier infer the language.

vscode-extension/package.json
captbaritone: Given that this is TS, I'd agree.

vscode-extension/package.json
captbaritone: Oh right. Thanks!

captbaritone (approved):
Just one nit

PR #3982, opened 2022-06-20
Fix typo in the example

ddanielsantos:


website/docs/guides/relay-resolvers.md
```diff
@@ -58,7 +58,7 @@ function MyGreeting({userKey}) {
     fragment MyGreeting on User {
       greeting
     }`, userKey);
-  return<h1>{user.greeting}<h1>;
+  return<h1>{user.greeting}</h1>;
```
voideanvalue, line 61: thanks! could you also add a space after return while you're here?

PR #3959, opened 2022-06-09
Provide a good VSCode setup for contributors

captbaritone:
Knowing all the extensions and settings you should have in order to be an effective contributor takes a ton of effort. These should help people be more effective right away.

@tbezman any recommendations that I've missed?

PR #3983, opened 2022-06-21
Compact GraphQL query text

tomgasson:
We noticed that some of our query text contains a significant amount of unnecessary whitespace. Unlike server->client requests, client->server requests generally aren't compressed and would require a js compression library to do so.

This PR adds a feature flag to compact the query text. I saw the pre-existing `compact` mode wasn't wired up to anything, so I've co-opted it.

As an example of the benefit, we have one query which this shrinks from `84kb` to `31kb`

We're slowly [making our way to persisted queries](https://github.com/facebook/relay/pull/3917), but this is an immediate improvement and I imagine we're not the only ones who will benefit from smaller network payloads

Happy to make this a config rather than a feature flag, if you don't think this should be the default output

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,6 +8,9 @@
       "editor.defaultFormatter": "esbenp.prettier-vscode"
     },
     "prettier.enable": true,
+    "rust-analyzer.rustfmt.extraArgs": [
+      "+nightly"
+    ],
 
 
     // Enable these to get Relay GraphQL editor support working on test files.
--- a/compiler/Cargo.lock
+++ b/compiler/Cargo.lock
@@ -157,9 +157,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "3.2.10"
+version = "3.2.22"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "69c5a7f9997be616e47f0577ee38c91decb33392c5be4866494f34cdf329a9aa"
+checksum = "86447ad904c7fb335a790c9d7fe3d0d971dc523b8ccd1561a520de9a85302750"
 dependencies = [
  "atty",
  "bitflags",
@@ -177,9 +177,9 @@ dependencies = [
 
 [[package]]
 name = "clap_derive"
-version = "3.2.7"
+version = "3.2.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "759bf187376e1afa7b85b959e6a664a3e7a95203415dba952ad19139e798f902"
+checksum = "ea0c8bce528c4be4da13ea6fead8965e95b6073585a2f05204bd8f4119f82a65"
 dependencies = [
  "heck 0.4.0",
  "proc-macro-error",
@@ -403,6 +403,27 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e78d4f1cc4ae33bbfc157ed5d5a5ef3bc29227303d595861deb238fcec4e9457"
 
 [[package]]
+name = "errno"
+version = "0.2.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f639046355ee4f37944e44f60642c6f3a7efa3cf6b78c78a0d989a8ce6c396a1"
+dependencies = [
+ "errno-dragonfly",
+ "libc",
+ "winapi",
+]
+
+[[package]]
+name = "errno-dragonfly"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "aa68f1b12764fab894d2755d2518754e71b4fd80ecfb822714a1206c2aab39bf"
+dependencies = [
+ "cc",
+ "libc",
+]
+
+[[package]]
 name = "errors"
 version = "0.0.0"
 dependencies = [
@@ -490,9 +511,9 @@ checksum = "3a471a38ef8ed83cd6e40aa59c1ffe17db6855c18e3604d9c4ed8c08ebc28678"
 
 [[package]]
 name = "futures"
-version = "0.3.13"
+version = "0.3.24"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f55667319111d593ba876406af7c409c0ebb44dc4be6132a783ccf163ea14c1"
+checksum = "7f21eda599937fba36daeb58a22e8f5cee2d14c4a17b5b7739c7c8e5e3b8230c"
 dependencies = [
  "futures-channel",
  "futures-core",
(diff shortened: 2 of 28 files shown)


captbaritone:
This seems reasonable to me! I'd love to see some fixture tests which validate this behavior. I think a set of fixture tests similar to the three sets we have here: https://github.com/facebook/relay/tree/main/compiler/crates/graphql-text-printer/tests.

Ideally we could have tests that look like this (this would be the implementation of the `mod.rs` file:

1. Parse the fixture file to AST
2. Convert the AST to IR
3. Convert the IR to a program
4. Print the program (use this as the fixture test output, but also...)
5. Reparse the fixture output to a new AST
6. Convert the new AST to a new IR
7. Use our [location agnostic IR equality tests](https://github.com/facebook/relay/blob/52af6a65c52618d5e4707e81f9883b5066b6438b/compiler/crates/graphql-ir/src/node_identifier.rs#L251) to assert that the two IRs are identical

I think there's one issue here: Which is that we don't yet implement the location agnostic equality test for all IR node types. Maybe you could look into how difficult it would be to add the trait for the missing types?

compiler/crates/graphql-ir/src/node_identifier.rs
```diff
@@ -405,6 +409,16 @@ impl LocationAgnosticPartialEq for Argument {
     }
 }
 
+impl LocationAgnosticPartialEq for Option<&Argument> {
```
tomgasson, line 412: This impl should be redundant from `Option<T>`, but I couldn't work out the right ref incantation

compiler/crates/graphql-text-printer/src/print_to_text.rs
```diff
@@ -638,17 +596,99 @@ impl<'schema, 'writer, W: Write> Printer<'schema, 'writer, W> {
...
+        self.space(true)
+    }
+
+    fn item_separator(&mut self) -> FmtResult {
+        if self.options.compact {
+            write!(self.writer, ",")
+        } else {
+            self.next_line(false)
+        }
+    }
```
tomgasson, line 693: These separator utils are the main change here, they have conditional behaviour in them which avoids the need for `self.options.compact` checks everywhere

compiler/crates/graphql-text-printer/src/print_to_text.rs
```diff
@@ -638,17 +596,99 @@ impl<'schema, 'writer, W: Write> Printer<'schema, 'writer, W> {
...
+    fn print_default_value(&mut self, value: &ConstantValue) -> FmtResult {
+        self.space(true)?;
+        self.text("=")?;
+        self.space(true)?;
+        self.print_constant_value(&value)
+    }
+
+    fn print_type(&mut self, type_: &TypeReference) -> FmtResult {
+        write!(self.writer, "{}", self.schema.get_type_string(&type_))
+    }
```
tomgasson, line 653: I pulled out a few duplicated prints as I found them

compiler/crates/graphql-text-printer/tests/compact/fixtures/compact_test.expected
```diff
@@ -0,0 +1,19 @@
...
+
+fragment FX on Query {
+    aliased : username(name : "argVal") {
+        thin:  emailAddresses
+    }
+}
+==================================== OUTPUT ===================================
+query Test{aliasOfFoo:username(name:"val1")@customDirective(level:2){emailAddresses,firstName(if:false,unless:false)},...FX}
+
+fragment FX on Query{aliased:username(name:"argVal"){thin:emailAddresses}}
```
tomgasson, line 19: This test captures the same behaviour as GraphQLJava https://github.com/graphql-java/graphql-java/pull/2872

compiler/Cargo.lock
```diff
@@ -1144,9 +1145,9 @@ dependencies = [
 
 [[package]]
 name = "once_cell"
-version = "1.8.0"
+version = "1.12.0"
```
tomgasson, line 1148: Stale from https://github.com/facebook/relay/commit/ee99ea4bbbe9979eb52685b25b393ac93bf98489

.vscode/settings.json
```diff
@@ -8,6 +8,9 @@
       "editor.defaultFormatter": "esbenp.prettier-vscode"
     },
     "prettier.enable": true,
+    "rust-analyzer.rustfmt.extraArgs": [
+      "+nightly"
+    ],
```
tomgasson, line 13: Additional to https://github.com/facebook/relay/pull/3959, this helps contributors use the same setup as [the github rust lint action](https://github.com/facebook/relay/blob/068608c7f799a7558a3b4abd666f06851f1285b8/.github/workflows/ci.yml#L109)

compiler/Cargo.lock
```diff
@@ -175,16 +175,16 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "3.1.18"
+version = "3.2.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
```
tomgasson, line 179: The noise here is because automatic updates https://github.com/facebook/relay/commit/8e25d86ab184d0f656a075fc4678dec4c42727b6, https://github.com/facebook/relay/commit/3cfc85ff0b89440361f91ad26406768cdb76cf12, https://github.com/facebook/relay/commit/89357466f26735f7c5d1798f9ad4b1ed0a14e056 etc aren't also updating the lock file.

It would be nice if they were either kept in sync, or have this file ignored

tomgasson:
@captbaritone I implemented those fixture tests. Does this look alright to you?

captbaritone:
@tomgasson Sorry I lost track of this. I'll work on importing/rebasing now.

rbalicki2:
Hey @tomgasson — this PR looks great! Thanks for taking it on. I'm in the process of reviewing it now. We'll fix it up internally and land it, thank you for doing this work!!

Some questions for you:
* Is the space before `fragment` necessary? e.g. is this legal?
```
query MyQuery{blah}fragment UserFragment on User{}
```
* Why is the next line not skippable in `print_condition`? TBH, I'm not really sure what that code is doing. I couldn't figure out how to test that in a fixture. Do you know what is going on in `print_condition`? If we can't figure it out, keeping it as an unskippable line-break seems right to me.

tomgasson:
> Is the space before fragment necessary? e.g. is this legal?

I think that is legal. Looks like an oversight

> Why is the next line not skippable in print_condition

I think it probably is. The code to iterate through the conditions' selections should probably share more with the `print_selections` code, and use `item_separator` rather than `next_line`. This was [initially separated here](https://github.com/tomgasson/relay/commit/bfabdf94474dd98714a2ac5e79a5ca4b1b57ff72#diff-567805733efe02d051473661ab4b04d5e8ba63d624dd7dd5bbf3ef0e643f8001R280)

It will be hard to test this one, because I believe a `Condition {}` is only ever _created_ as a single element

https://github.com/tomgasson/relay/blob/91251aaa04a9749de5d2ce1946974bc395b5ba2b/compiler/crates/graphql-ir/src/build.rs#L1893-L1901

And sometimes combined via a transform, but I can't work out how to trigger that behavior

https://github.com/tomgasson/relay/blob/91251aaa04a9749de5d2ce1946974bc395b5ba2b/compiler/crates/graphql-ir/src/transform.rs#L295-L299

rbalicki2:
Okay, cool. I guess a future person can figure out what's going on with that condition code, and clean it up. I'll remove the space internally, no need to change the PR. Thanks!

PR #3991, opened 2022-06-22
[Prototype][RFC] Explore allowing fragments to read from query root

captbaritone:
This PR is just to show a sketch of how we might allow any fragment to also query data off of `Query`. If we decide this is something we wanted to support, there are a number of things that would be needed in order to take it across the finish line. This is just intended to start the conversation.

The idea here is that any fragment can also read from a field `__query` which re-exposes the query root. This would allow deeply nested components to add data dependencies that are not easy to access from their existing fragment. For example, a component on a Comment might want to look at some information about the current user to decide how to render. currently that requires threading two fragments through the React hierarchy. One referencing the comment and one referencing the viewer/user.

## How it Looks

```JavaScript
const comment = useFragment(graphql`
  fragment Foo on Comment {
    __query {
      me {
        preferences {
          auto_expand_comments
        }
      }
    }
    text
  }`, key);

if (comment.__query.me.preference.auto_expand_comments) {
  return comment.text;
} else {
  return comment.text.slice(0, 100)
}
```

## How it Works

We teach the schema about a special field `__query` that exists on every object and exposes the root Query type. Before generating the query text (and normalization artifact) we traverse the query and collect all the selections made on `__query` and inline them into the query root selection.

When generating the reader artifacts, we treat the `__query` field as if it really exists, trusting that the data was put into the store by the original query.

## Next Steps?

Possibly none. However, if people are interested in this, we should start to collect a list of other Relay features that may play poorly with this approach, or would need special handling.

---
This came up in a conversation with @tbezman and Chris yesterday and it got me thinking about how it could be done.

### DIFF
--- a/compiler/Cargo.lock
+++ b/compiler/Cargo.lock
@@ -175,16 +175,16 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "3.1.8"
+version = "3.1.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "71c47df61d9e16dc010b55dba1952a57d8c215dbb533fd13cdd13369aac73b1c"
+checksum = "d2dbdf4bdacb33466e854ce889eee8dfd5729abf7ccd7664d0a2d60cd384440b"
 dependencies = [
  "atty",
  "bitflags",
  "clap_derive",
+ "clap_lex",
  "indexmap",
  "lazy_static",
- "os_str_bytes",
  "regex",
  "strsim",
  "termcolor",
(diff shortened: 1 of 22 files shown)


tbezman:
@captbaritone I'm having trouble poking holes in this. I want this behind an experimental flag sooooooo bad lol

captbaritone:
@zth Called out another potential risk in chat:

> One thing would be to reaccess the same root field with the same filters deep into the component structure, for "convenience"

The risk is that these filters really ought to have the same source of truth, but allowing you to fetch deep in the tree would temp you to have multiple sources of truth.

> Instead of adding a fragment on the correct source of truth, and prop drill that, you'd replicate what the root does, but accidentally coupling to just that use case instead

tbezman:
> The main tradeoffs I'm tracking:
> 
> 1. Would take work to implement/maintain (opportunity cost)
> 2. How does this compose with all other Relay features? Will that complicate our compiler/runtime?
> 3. What happens if you use a fragment like this within a mutation?
> 4. Not great for discoverability/somewhat magical.
> 5. Harder to debug perhaps? Since the fragment looks different than the network response.
> 6. It might make field collisions more confusing? (If distant fragments try to read data off of query using arguments that differ)

1. For sure
2. Seems pretty spicy regarding the "What happens if you spread in a mutation / subscription". I think you and the rest of the Relay team have the best insight on how complexity may or may not increase.
3. I think we just want to block that
4. I think this is something we can document well in a top level topic in the docs
5. This is true for all Relay features I think (`@requied`, etc)
6. I think we can accommodate with high quality error messages.

chriserickson:
Small note: I wonder if the use of `__root` instead of `__query` for the alias might be more intuitive? Mostly for parity with keys in the Relay store?

PR #3993, opened 2022-06-24
[WIP] Transform to convert @requried to Client Controlled Nullability

captbaritone:
Sketch (code is quite rough) of a transform to attempt to convert @required directives to the equivalent [Client Controlled Nullability](https://github.com/graphql/graphql-spec/issues/867) semantics.

Some @required patterns cannot be expressed. For these we create an error.

Much more work to do here before it can be useful (or even correct), but I got to a good stopping place and wanted to with some folks.


scripts/build-fixture-tests.sh
```diff
@@ -0,0 +1,100 @@
...
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)
+GITHUB_ROOT=$(dirname "$SCRIPT_DIR")
+
+cd "$GITHUB_ROOT"/compiler || exit
+cargo run --bin fixture_tests_bin --quiet --release -- \
```
captbaritone, line 11: We need to add a version of this to open source.

PR #3994, opened 2022-06-24
update rustfmt used by ci

voideanvalue:
GitHub CI broke after 5998146e0ddd4843b137638f966487dc54cfd9eb. Matching the rustfmt toolchain version to the one used internally.


captbaritone:
Thanks! I wonder if we can setup an internal test to ensure these stay in sync.

Issue #3989, opened 2022-06-22
Relay Resolver has module resolution problem in TypeScript


Beingbook:
<img width="542" alt="Screen Shot 2022-06-22 at 15 34 22" src="https://user-images.githubusercontent.com/2682865/174959781-bff75b48-e711-47b9-a14d-8271db429c2e.png">

<img width="865" alt="Screen Shot 2022-06-22 at 15 37 58" src="https://user-images.githubusercontent.com/2682865/174960251-2636af9e-f835-4e04-8ca1-4bfa71f8c683.png">

As above images, generated relay artifacts import relay resolver with extension.
I think using ".js" extension or skipping extension are correct solution at this moment. So we can inference return type of RelayResolver function in TypeScript.

captbaritone:
Thanks for the report @Beingbook. If you're interested in working on a fix, I suspect stripping the extension would be the right approach and the I think the code that derives this path lives here: https://github.com/facebook/relay/blob/main/compiler/crates/relay-config/src/project_config.rs#L360

cc @tbezman

Beingbook:
oops, I'm too late. thanks!

chriserickson:
This was not fixed in 14.1 - the dynamic imports in the AST omit extensions, however the type imports still have the extensions.

It appears that 
https://github.com/facebook/relay/blob/639abe1b23ed2187500514c331bed3aebede2f02/compiler/crates/relay-typegen/src/typescript.rs#L97
and
https://github.com/facebook/relay/blob/639abe1b23ed2187500514c331bed3aebede2f02/compiler/crates/relay-typegen/src/typescript.rs#L101

Both also need to strip extensions, as typescript requires that extensions not be provided.

Is it possible to re-open this? I'm happy to work on a fix.

chriserickson:
I went ahead and created a new issue:

https://github.com/facebook/relay/issues/4056

PR #4002, opened 2022-07-06
Remove `.js` and `.ts` from require / import

tbezman:
Fixes https://github.com/facebook/relay/issues/3989#issuecomment-1175374119

This PR updates `relay-codegen`'s printer not to include the file's extension when writing imports /  requires. This might be a naive implementation, maybe there's a better, more holistic approach.

### DIFF
--- a/compiler/crates/relay-codegen/src/printer.rs
+++ b/compiler/crates/relay-codegen/src/printer.rs
@@ -39,6 +39,7 @@ use indexmap::IndexMap;
 use intern::string_key::StringKey;
 use std::fmt::Result as FmtResult;
 use std::fmt::Write;
+use std::path::Path;
 
 pub fn print_operation(
     schema: &SDLSchema,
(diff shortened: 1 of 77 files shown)


compiler/crates/relay-codegen/src/printer.rs
```diff
@@ -470,7 +470,15 @@ impl<'b> JSONPrinter<'b> {
...
             Primitive::JSModuleDependency(key) => match self.js_module_format {
                 JsModuleFormat::CommonJS => {
-                    self.write_js_dependency(f, key.to_string(), format!("./{}", key))
+                    let mut path = format!("./{}", key);
+
+                    let extension = &path[path.len() - 3..path.len()];
+
+                    if extension == ".js" || extension == ".ts" {
+                        path = path[0..path.len() - 3].to_string();
+                    }
```
captbaritone, line 479: > the Path/PathBuf types have utilities for extracting the extension, let's use those? there is also the camino crate, which is basically utf8-only versions of Path/PathBuf which might help since the string is already utf8.

-- @josephsavona (internal review)

sibelius: missing `tsx`

sibelius: fixed https://github.com/facebook/relay/pull/4303

PR #4004, opened 2022-07-08
Added script to update version during release process

### LINKED ISSUE #3999: [Relay Compiler] `--version` doesn't respect actual version
### Problem

- `yarn relay-compiler --version` returns "0.0.0"

### Expected behavior
- `yarn relay-compiler --version` returns "14.0.0"

### My exploration
- I believe this problem starts to manifest due to this contribution https://github.com/facebook/relay/commit/42de87ec0ffcbe051e93c06621819540be53fee6#diff-d83140edf1339f1f1917b24d81d44dfeb57a0b548db4518577096b333dab305f which moved CLI core features (like version command) out of relay-compiler crate. 
- The version in `relay-bin` is not updated as part of the release "commit" (https://github.com/facebook/relay/commit/f85538779e68e56dcd29f5c7a2b79ccf9a7d4611#diff-43717d1e161bdd2969b45b9451668f0cff213c82f636d261cc11103b80eab128 – https://github.com/facebook/relay/blob/main/compiler/crates/relay-bin/Cargo.toml#L4) though. 
- This outdated (not set) version is then consumed here during the build https://github.com/facebook/relay/blob/main/compiler/crates/relay-bin/src/main.rs#L37 and thus the issue. 

Unfortunately, I don't know how to fix it easily, if the deployment is automated, I would suggest just propagating the version to Cargo.toml next to relay-bin, that should do the trick. Anyway, I'd like to learn. 🙏

jukben:
Fixes #3999

### DIFF
--- a/scripts/update-version.js
+++ b/scripts/update-version.js
@@ -1,0 +1,105 @@
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+
+const RELAY_NPM_PACKAGES = [
+  'babel-plugin-relay',
+  'react-relay',
+  'relay-compiler',
+  'relay-runtime',
+  'relay-test-utils',
+  'relay-test-utils-internal',
+];
+
+const CARGO_FILES = [
+  `compiler/crates/relay-bin/Cargo.toml`,
+  `compiler/Cargo.lock`,
+];
+
+/**
+ * Updates the version of relay across all package.json files
+ * Based on previous internal version @see https://gist.github.com/captbaritone/3351af94fb0821b93e26e23fbbc0c4c5
+ * @param {string} version The version to update to
+ * @example
+ * node scripts/update-version.js 1.0.0
+ */
+(function main() {
+  const version = process.argv[2] // first argument
+
+  updateVersion(version);
+})();
+
+function resolveRootPath(filePath) {
+  return path.resolve(__dirname, '..', filePath);
+}
+
+function getCurrentVersion() {
+  const packageJsonPath = resolveRootPath("./package.json");
+  const packageJson = require(packageJsonPath);
+
+  return packageJson.version;
+}
+
+function isVersionValid(version) {
+  return version && /^\d+\.\d+\.\d+(\-[\d\w\.]+)?$/.test(version)
+}
+
+function updatePackageJson(packageJsonPath, version) {
+  const packageJson = require(packageJsonPath);
+
+  packageJson.version = version;
+
+  for (const dependenciesKey of [
+    'dependencies',
+    'devDependencies',
+    'peerDependencies',
+  ]) {
+    for (const dep in packageJson[dependenciesKey]) {
+      if (RELAY_NPM_PACKAGES.includes(dep)) {
+        packageJson[dependenciesKey][dep] = version;
+      }
+    }
+  }
+
+  fs.writeFileSync(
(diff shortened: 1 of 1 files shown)


scripts/update-version.js
```diff
@@ -0,0 +1,105 @@
...
+const RELAY_NPM_PACKAGES = [
+  'babel-plugin-relay',
+  'react-relay',
+  'relay-compiler',
+  'relay-runtime',
+  'relay-test-utils',
+  'relay-test-utils-internal',
+];
+
+const CARGO_FILES = [
```
jukben, line 15: It's missing `const COMPILER_TARGETS_FILE = `${ROOT_DIR}/fbcode/relay/oss/crates/relay-compiler/TARGETS`;` from the original script, but I'm not sure if a problem.

scripts/update-version.js
```diff
@@ -0,0 +1,105 @@
...
+];
+
+/**
+ * Updates the version of relay across all package.json files
+ * Based on previous internal version @see https://gist.github.com/captbaritone/3351af94fb0821b93e26e23fbbc0c4c5
+ * @param {string} version The version to update to
+ * @example
+ * node scripts/update-version.js 1.0.0
+ */
+(function main() {
```
jukben, line 27: Is this OK? I'm not a fan of this IIFE, any recommendations? Should I call `main()` at the end I'm not sure if I like either 😀 ? Or the goal was to keep it "compatible" with "js1" (no idea about it :D ).

```js
exports.command = 'update-version VERSION';
exports.describe = 'Updates the version of relay across all package.json files';
exports.handler = argv => {
   //...
}
```

scripts/update-version.js
```diff
@@ -0,0 +1,105 @@
...
+
+function getCurrentVersion() {
+  const packageJsonPath = resolveRootPath("./package.json");
+  const packageJson = require(packageJsonPath);
+
+  return packageJson.version;
+}
+
+function isVersionValid(version) {
+  return version && /^\d+\.\d+\.\d+(\-[\d\w\.]+)?$/.test(version)
```
jukben, line 45: I keept the same logic here

scripts/update-version.js
```diff
@@ -0,0 +1,105 @@
...
+    const text = fs.readFileSync(resolveRootPath(file), 'utf8');
+    const withUpdatedVersion = text.replace(
+      `version = "${getCurrentVersion()}"`,
+      `version = "${nextVersion}"`,
+    );
+    fs.writeFileSync(resolveRootPath(file), withUpdatedVersion);
+  }
+}
+
+function updateVersion(nextVersion) {
```
jukben, line 89: Basically the same behavior like in the original script

scripts/update-version.js
```diff
@@ -0,0 +1,105 @@
...
+  'babel-plugin-relay',
+  'react-relay',
+  'relay-compiler',
+  'relay-runtime',
+  'relay-test-utils',
+  'relay-test-utils-internal',
+];
+
+const CARGO_FILES = [
+  `compiler/crates/relay-bin/Cargo.toml`,
```
jukben, line 16: Changed to `relay-bin` because that's the CLI part which is relevant I hope

jukben:
@captbaritone What do you think? 🙏

PR #4013, opened 2022-07-15
Exposing cacheConfig for the hooks which also fetch queries

gplopes:
At the moment only a few hooks can access the cache config, this PR extends also this option to the two hooks `usePagination, useRefetchableFragment` to also have access to the cache config.

Use case: with the `cacheConfig.metadata`, I am able to send extra data to my NetworkLayer (i.e: custom header information), which I can easily do with all the other hooks, unfortunately, pagination and refetchable have hardcoded cache config.


djonesz:
We just ran into this, because we're using cache config to switch between graphql endpoints (we have to hit two of them in our app).

Can someone who is authorized take a look and merge this if possible?

gilles-yvetot:
Thanks a lot @gplopes for that!

@alunyov would you have time to take a look at this PR? It seems to be low-risk and would help a lot of people. If it is not ready as-is, I'd love to help make it good :)

PR #4018, opened 2022-07-18
fix yarn.lock

voideanvalue:
Not sure what's going on but running `yarn` on a fresh checkout leads to `yarn.lock` getting updated with these changes. This doesn't happen if I go one commit behind [dec019d](https://github.com/facebook/relay/commit/dec019d3290233e134b603124001814a77832271). Eyeballing the deletions... I'm guessing that the additions to `yarn.lock` from that commit somehow added stuff related to website to the root `yarn.lock` as well...

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,134 +2,6 @@
 # yarn lockfile v1
 
 
-"@algolia/autocomplete-core@1.7.1":
-  version "1.7.1"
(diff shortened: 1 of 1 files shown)


voideanvalue:
addressed by 48ef3e236fcd09c0e73c0fc75e060fa2b2b83d71 and 4264743dc383bdd6b1cca22a0d0e1fe6c30a1c60

PR #4025, opened 2022-07-22
[vscode-extension] Adds a way to quickly create a component fragment 

orta:
**Opening this to start a discussion** - I'd like to have the extension ship an easy way to get started with new files. This PR is an _example_ which shows how to do that with all the high-level vscode workspace APIs. It takes into account JS vs TS, but it more or less the lower bounds for what could be mergable.

---

You can right-click in the explorer, or run the command "Relay: Create a new fragment component" :
image

You are then asked about what the file name is:
image

It then generates the file;
image

### DIFF
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -153,8 +153,7 @@
     "bracketSameLine": true,
     "bracketSpacing": false,
     "singleQuote": true,
-    "trailingComma": "all",
-    "printWidth": 100
+    "trailingComma": "all"
   },
   "scripts": {
     "typecheck": "tsc --noEmit",
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -70,8 +70,20 @@
       {
         "command": "relay.stopCompiler",
         "title": "Relay: Stop Compiler"
+      },
+      {
+        "command": "relay.createNewFragmentComponent",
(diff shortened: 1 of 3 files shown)


orta:
Things worth thinking about:
 
- How clever do we want codegen to be?
 - Can the extension grab info from the LSP like: 
    - Path to the generated types from the current file
    - Copyright headers?

vscode-extension/package.json
```diff
@@ -70,8 +70,20 @@
...
+        "command": "relay.createNewFragmentComponent",
+        "title": "Relay: Create New Fragment Component"
       }
     ],
+    "menus": {
+      "explorer/context": [
+        {
+          "command": "relay.createNewFragmentComponent",
+          "group": "Relay@1"
+        }
```
orta, line 84: Ideally when the extension has decided that it is in a relay project somehow, it would set an context variable so we can [use "when" to not show these](https://code.visualstudio.com/api/references/when-clause-contexts) unless useful

damassi:
Was just thinking about making something like this the other day 👍

orta:
Opinions?

orta:
Poke

orta:
Alright, I'll eventually get around to making an extension which does this as a separate thing.

PR #4029, opened 2022-07-26
Use `import type` in typescript for all imports

tomgasson:
This makes the following change to typescript artifacts:

```diff
-import { ConcreteRequest, Query } from "relay-runtime";
+import type { ConcreteRequest, Query } from "relay-runtime";
 import type { FragmentRefs } from "relay-runtime";
```

Using the existing `useImportTypeSyntax` config.

As stated in the config, the `import type` syntax introduced in Typescript version 3.8 and prevents warnings from `importsNotUsedAsValues`. This setting also assists naive bundlers from taking a dependency on `relay-runtime`, when only the types are needed.

I kept this simple. Other paths for implementation would be
 - There's a mix of distinct typegen printers e.g. compiler/crates/relay-typegen/src/typescript.rs, and the inline use of conditionals on the language directly in compiler/crates/relay-compiler/src/artifact_content/content.rs. Consolidating these would allow sharing the existing `write_import_type` function
 - The imports could instead be consolidated and printed together, rather than potentially having multiple imports from the same file as it is now

### DIFF
--- a/compiler/Cargo.lock
+++ b/compiler/Cargo.lock
@@ -157,9 +157,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "3.2.10"
+version = "3.2.22"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "69c5a7f9997be616e47f0577ee38c91decb33392c5be4866494f34cdf329a9aa"
+checksum = "86447ad904c7fb335a790c9d7fe3d0d971dc523b8ccd1561a520de9a85302750"
 dependencies = [
  "atty",
  "bitflags",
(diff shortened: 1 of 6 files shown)


alunyov (changes requested):
Thank you for PR! This looks good to me! 

Can you add a test for these changes?

tomgasson:
Removed the missing newline at EOF, because I'm guessing that was the internal lint warning

alunyov:
Hmm, @tomgasson for some reason this breaks one of the internal typescript tests (we should probably move these to OSS)

It removes the import type of `ConcreteRequest` and `Query`...

```
/* eslint-disable */
 // [@ts](https://www.internalfb.com/intern/profile/ts)-nocheck
 
-import { ConcreteRequest, Query } from 'relay-runtime';
```

tomgasson:
Hmmm. Are those tests using `use_import_type_syntax`? Is the output _after_ stripping types / building typescript artifacts? That kind of the purpose of this change - to not have a _value_ import in the code actually sent to browsers

 It's a bit hard to see how this change would remove something if it's only the relay generated output

PR #4030, opened 2022-07-27
adding support for client extensions in mock payload generator

stewartavery:
## Problem
At Twitter we've been hoping to use MockPayloadGenerator to allow us to scaffold data in tests/[stories](https://storybook.js.org/) while our components use client side schema extensions. One of the engineers on my team discovered that this support is explicitly disallowed in the `relay` repo. I reached out to the discord channel and received feedback that enabling this would make sense.

## Solution
Within `RelayMockPayloadGenerator.js`, modify the logic within `_traverseSelections` to recursively traverse the nested selections. This places the logic alongside `DEFER` and `STREAM`. 

A snapshot was updated as a result of this change —  local client extension `client_name` and `client_code` now have generated mocked fields.

### DIFF
--- a/packages/relay-test-utils/__tests__/RelayMockPayloadGenerator-test.js
+++ b/packages/relay-test-utils/__tests__/RelayMockPayloadGenerator-test.js
@@ -21,9 +21,13 @@ const {FIXTURE_TAG} = require('relay-test-utils-internal');
 function testGeneratedData<TVariables: Variables, TData, TRawResponse>(
   query: Query<TVariables, TData, TRawResponse>,
   mockResolvers: ?MockResolvers,
+  options: ?{mockClientData?: boolean},
 ): void {
   const operation = createOperationDescriptor(query, {});
-  const payload = RelayMockPayloadGenerator.generate(operation, mockResolvers);
+  const mockClientData = options?.mockClientData ?? false;
+  const payload = mockClientData
+    ? RelayMockPayloadGenerator.generateWithClientData(operation, mockResolvers)
+    : RelayMockPayloadGenerator.generate(operation, mockResolvers);
 
   expect({
     [FIXTURE_TAG]: true,
@@ -1432,22 +1436,6 @@ describe('with @relay_test_operation', () => {
     );
   });
 
-  test('generate mock for client extensions', () => {
-    testGeneratedData(
-      graphql`
-        query RelayMockPayloadGeneratorTest43Query @relay_test_operation {
-          node(id: "my-id") {
-            ... on User {
-              id
-              client_name
-              client_code
-            }
-          }
-        }
-      `,
-    );
-  });
-
   test('should generate data for @module', () => {
     graphql`
       fragment RelayMockPayloadGeneratorTestNameRendererFragment on User {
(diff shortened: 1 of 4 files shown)


sibelius (approved):

alunyov (changes requested):
Thank you for submitting PR!

I think it make sense to enable the payload generation for client extension for your use case.
From the discord:

> In our case we're populating using environment.commitPayload before the query. So that works fine when we're writing tests for the component that is executing the query (the top level component). Where we have a tougher time is when there's a sub component that we may want to create a test for that also uses the client side field. We'll then create a query wrapper to create a story/test for that component. But at that point we need to be aware of the client side field and do the hydration manually in the story/test. 

In order to preserve the current behavior (see https://www.hyrumslaw.com/) of the API what do you think about adding a new method to the API:  `generateWithClientData(..)`?

We also need to update compiler transform to include client fields in the @relay_test_operation: https://github.com/facebook/relay/blob/df2c9965896d670d59ddaf08132b0ab778424b6a/compiler/crates/relay-transforms/src/test_operation_metadata.rs#L221 (this can be done in a separate PR)

stewartavery:
@alunyov You're totally right! I'll see if I can find some time to follow up to make that change

stewartavery:
What are your thoughts on extending the private methods in the class, something like:

```js
generateWithClientData(
    selections: $ReadOnlyArray<NormalizationSelection>,
    operationType: string,
  ): MockData {
    const defaultValues = this._getDefaultValuesForObject(
      operationType,
      null,
      null,
      [], // path
      {},
    );
    return this._traverse(
      {
        selections,
        typeName: operationType,
        isAbstractType: false,
        name: null,
        alias: null,
        args: null,
      },
      [], // path
      null, // prevData
      defaultValues,
      true, // isClientExtensionGenerated <--- __traverse would receive a new argument here
    );
  }
```

and then we'd check that condition above when evaluating `__traverse` at the `CLIENT_EXTENSION` evaluation?

stewartavery:
Or I could add an optional arg to `generate`. and have `generateWithClientData` call `generate` with a truthy value in the arg?

alunyov:
I would keep public version of `generate` unchanged (without adding options), and add a new public method `generateWithClientData`.

And internally, in this module we an pass this flag to the `generateData` function here:

https://github.com/facebook/relay/blob/d59b18320aa5656ca907ed11c71477bc160fed09/packages/relay-test-utils/RelayMockPayloadGenerator.js#L849-L854

And then in can be a property of the class:

https://github.com/facebook/relay/blob/d59b18320aa5656ca907ed11c71477bc160fed09/packages/relay-test-utils/RelayMockPayloadGenerator.js#L176-L180

something like `_shouldGenerateClientData`.

stewartavery:
Hey @alunyov, finally managed to get around to making this change! let me know if this is what you had in mind

packages/relay-test-utils/__tests__/RelayMockPayloadGenerator-test.js
```diff
@@ -21,9 +21,12 @@ const {FIXTURE_TAG} = require('relay-test-utils-internal');
 function testGeneratedData<TVariables: Variables, TData, TRawResponse>(
   query: Query<TVariables, TData, TRawResponse>,
   mockResolvers: ?MockResolvers,
+  isClientDataMocked: ?boolean = false,
```
alunyov, line 24: nit: can we make this argument an object `options` or something like that...


```
options: ?{
   mockClientData?: boolean
}
```

alunyov (changes requested):
Looks good!
I have one small suggestion.

stewartavery:
@alunyov cool, just pushed another change! also standardized on the `clientDataMocked` name across other files

alunyov (approved):

PR #4032, opened 2022-07-31
Switch to use `GraphQL Syntax` as an extension dependency

tbezman:
The Graphiql folk extracted out a syntax highlighting extension which we can rely upon instead of just copy pasting their grammars.


acao (commented):
Looks good from my side! these were exactly the changes I wanted to confirm.

acao:
@tbezman finally published 1.0.3! enjoy :)

alunyov:
Thanks @tbezman and @acao!

PR #4046, opened 2022-08-15
Move tslint-disable to below eslint-disable

leonardodino:
## Summary:
This PR moves the `eslint-disable` comment on typescript generated files to be above the `tslint-disable` one.

Solves the following:

Screenshot 2022-08-15 at 20 20 00

## Reasoning:
**given that**:
- `tslint` is deprecated
- some [popular presets](https://github.com/typescript-eslint/typescript-eslint/blob/921cdf17e548845311d0591249616ec844503926/packages/eslint-plugin/src/configs/strict.ts#L9) on `typescript-eslint` project forbids the usage of `tslint`

**then**: makes sense to reverse the order the disable comments are hidden.

---

**downside**: typescript projects that commit this generated files will have a large diff when upgrading
(but I guess compiler updates may generate this sorts of diffs anyways).

**fun fact**: generated TS files pass all rules in the `strict` `typescript-eslint` preset, except this one 😄

### DIFF
--- a/compiler/crates/relay-compiler/src/artifact_content/content.rs
+++ b/compiler/crates/relay-compiler/src/artifact_content/content.rs
@@ -765,8 +765,8 @@ fn generate_disable_lint_section(language: &TypegenLanguage) -> Result<GenericSe
     let mut section = GenericSection::default();
     match language {
         TypegenLanguage::TypeScript => {
-            writeln!(section, "/* tslint:disable */")?;
             writeln!(section, "/* eslint-disable */")?;
+            writeln!(section, "/* tslint:disable */")?;
             writeln!(section, "// @ts-nocheck")?;
         }
         TypegenLanguage::Flow | TypegenLanguage::JavaScript => {


sibelius:
should we just remove tslint ?

leonardodino:
yeah, I think the set of projects running both `tslint` and the latest major of `relay-compiler` should be small enough to make removing it a valid solution 😄 

I written it this way since I was afraid someone still using `tslint` would then open a issue to add it back 😆
up to the maintainers to choose

compiler/crates/relay-compiler/src/artifact_content/content.rs
```diff
@@ -765,8 +765,8 @@ fn generate_disable_lint_section(language: &TypegenLanguage) -> Result<GenericSe
     let mut section = GenericSection::default();
     match language {
         TypegenLanguage::TypeScript => {
-            writeln!(section, "/* tslint:disable */")?;
             writeln!(section, "/* eslint-disable */")?;
+            writeln!(section, "/* tslint:disable */")?;
```
maraisr, line 769: tslint has been deprecated since 2019. Better to just remove it imo

leonardodino: done! 😄

PR #4047, opened 2022-08-19
Modifying @relay_test_operation to write metadata for client fields

stewartavery:
## Problem

There were recent changes to allow support for mocking client side fields with MockPayloadGenerator. Now that these fields can be used we need to make sure `@relay_test_operation` writes associated metadata for them during artifact generation. 

## Solution

Add a new `FeatureFlag`, `enable_mock_client_data_metadata`, to control whether or not client data gets extra metadata generated when `@relay_test_operation` is applied to a query. 

Then, modify the `!field.is_extension` checks in both `ScalarField` and `LinkedField` patterns to also look for the feature flag value before skipping client data. 

Finally, add new test cases to support the new variations with client side schema fields. The `mod.rs` file was changed to create a local schema instance that incorporates client side schema extensions.


sibelius (approved):

compiler/crates/relay-transforms/tests/relay_test_operation_test.rs
```diff
@@ -1,11 +1,5 @@
-/*
- * Copyright (c) Meta Platforms, Inc. and affiliates.
- *
- * This source code is licensed under the MIT license found in the
- * LICENSE file in the root directory of this source tree.
- *
- * @generated SignedSource<<13a74be1bf3601ce595319e9b12ebb4b>>
- */
+// @generated SignedSource<<21d345448b5d3e67d26e86d61a161b22>>
+// Generated by $ cargo run -p fixture-tests -- relay-transforms/tests/relay_test_operation
```
stewartavery, line 2: I imagine this will need to be regenerated once the PR is imported

stewartavery:
Going to take another pass at this to add an argument to the @relay_test_operation directive.

compiler/crates/relay-schema/src/relay-extensions.graphql
```diff
@@ -3,7 +3,7 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-directive @relay_test_operation on QUERY | MUTATION | SUBSCRIPTION
+directive @relay_test_operation(mock_client_data: Boolean = false) on QUERY | MUTATION | SUBSCRIPTION
```
stewartavery, line 6: I experimented around with the default value here, not sure if it was working. Is the purpose more to show intent?

compiler/crates/relay-transforms/tests/relay_test_operation/mod.rs
```diff
@@ -5,14 +5,54 @@
...
 use regex::Regex;
+use relay_test_schema::get_test_schema_with_extensions;
 use relay_transforms::generate_test_operation_metadata;
 
 pub fn transform_fixture(fixture: &Fixture<'_>) -> Result<String, String> {
-    let test_path_regex = Some(Regex::new(r#"^test"#).unwrap());
-    apply_transform_for_test(fixture, |program| {
-        generate_test_operation_metadata(program, &test_path_regex)
-    })
+    let parts: Vec<_> = fixture.content.split("%extensions%").collect();
```
stewartavery, line 24: I referenced other `mod.rs` files in order to get this working

stewartavery:
other alternatives to this design — should I use an internal feature flag to control whether or not mock data is generated (rather than having to do `@relay_test_operation(mock_client_data_true: true)` every time)? 

and perhaps in the next major version the default behavior could be inverted?

stewartavery:
@voideanvalue 

> What's the rationale for doing this conditionally with a FeatureFlag instead of doing this unconditionally?

It was along the same line as the feedback I got [here](https://github.com/facebook/relay/pull/4030#pullrequestreview-1054434608) in a previous PR with client field changes. The metadata exposed to MockPayloadGenerator is now more rich for client fields — this could break systems(like snapshot tests or other assertions) that implicitly depend on the behavior today. 

I was thinking the flag could be removed in the next major version of Relay and at that point we could make this behavior the new default!

compiler/crates/relay-transforms/tests/relay_test_operation/fixtures/test_client_fields_query.expected
```diff
@@ -0,0 +1,37 @@
...
+extend type User {
+  client_info: ClientInfo
+}
+
+type ClientInfo {
+  name: String
+  description: String
+}
+==================================== OUTPUT ===================================
+query ClientFieldsQuery @__metadata(relayTestingSelectionTypeInfo: {node: {enumValues: null, nullable: true, plural: false, type: "Node"}, node.id: {enumValues: null, nullable: false, plural: false, type: "ID"}, node.name: {enumValues: null, nullable: true, plural: false, type: "String"}, node.client_info: {enumValues: null, nullable: true, plural: false, type: "ClientInfo"}, node.client_info.name: {enumValues: null, nullable: true, plural: false, type: "String"}, node.client_info.description: {enumValues: null, nullable: true, plural: false, type: "String"}}) {
```
stewartavery, line 26: this is the new behavior with the feature flag enabled. This extra metadata will allow `MockPayloadGenerator` to auto-mock client fields when they're not provided in the associated resolver

voideanvalue:
@stewartavery

If we don't add a feature flag and do this unconditionally... all that'd do is add additional metadata about to the generated artifact for any existing operations that have the `@relay_test_operation` directive and includes client fields.

That in itself isn't something I'd consider a breaking change.

The output from calling `RelayMockPayloadGenerator.generate` on any such operation will only change the output if `mockClientData` option added in 8a44673ef671e6e06eae0a4df5fe52da23fde703 is used. I doubt anyone is already using the `mockClientData` option. I'd also argue that this is finishing an incomplete feature and not breaking existing behavior... specially since the last relay release was before 8a44673ef671e6e06eae0a4df5fe52da23fde703 was committed.

That's why I find the FeatureFlag to be an overkill here.

Am I missing something?

stewartavery:
@voideanvalue 

> The output from calling RelayMockPayloadGenerator.generate on any such operation will only change the output if mockClientData option added in https://github.com/facebook/relay/commit/8a44673ef671e6e06eae0a4df5fe52da23fde703 is used

I'm not sure that's true since my last commit only gave support for providing _manual_ client data to a resolver. I never made any changes to `test_operation_metadata`. So technically someone could've been using a client field in their production code but just not relying on MockPayloadGenerator to make any assertions against it within their tests. 

But now with this latest changes those artifacts will update will update at compile time — which doesn't care about what options are passed to MockPayloadGenerator since that's a runtime thing.

voideanvalue:
Yes the generated artifacts will get changed.

https://github.com/facebook/relay/blob/7a12c67f80d4104332d621817ece5539cd496ed9/packages/relay-test-utils/RelayMockPayloadGenerator.js#L900

this is the only place we ever read `relayTestingSelectionTypeInfo`

If I'm reading `RelayMockPayloadGenerator` correctly... any additional selections that will get added to `relayTestingSelectionTypeInfo` in generated artifacts after this change... will not be traversed unless `mockClientData` option is `true`.

So the only cases where runtime behavior changes is when `RelayMockPayloadGenerator` is explicitly passed `{mockClientData: true}`.

`mockClientData` was added in https://github.com/facebook/relay/commit/8a44673ef671e6e06eae0a4df5fe52da23fde703. the latest relay release predates that commit. even if this was a breaking change relative to https://github.com/facebook/relay/commit/8a44673ef671e6e06eae0a4df5fe52da23fde703, it's not a breaking change relative to the latest release.

(it's possible I'm being oblivious to something that you're seeing. so please do call me out on anything I'm asserting incorrectly or overlooking.)

stewartavery:
@voideanvalue nope that's correct! I had in mind the [hyrum's law](https://www.hyrumslaw.com/) feedback from the last PR. But yeah I can update things next week 👍

alunyov (approved):
Thank you!

PR #4053, opened 2022-08-27
Fix refetching and pagination when using a custom nodeInterfaceIdField

### LINKED ISSUE #3897: Refetching doesn't work well with `nodeInterfaceIdField`
Setting a custom ID field with `nodeInterfaceIdField` breaks the normal behavior of refetching fragments. Namely, upon calling `refetch`,  the GraphQL server responds with an error complaining about the id missing in the `node` query variables.

The only way of working around that is passing the id explicitely as part of the `refetch` variables, which of course, is not necessary when using the default `id` field.

This also breaks the Relay pagination helpers.

### Bug Reproduction

I edited the official TodoMVC example to reproduce the bug. Please check out [this repo ](https://github.com/Emilios1995/relay-node-id-bug) and look at the reproduction instructions in the readme.

Emilios1995:
This fixes #3897.

~~I wasn't able to get the tests to pass on my machine, even on `main`, so I didn't consider whether to add tests for this change. 
Please let me know whether I should do something other than `yarn install`, `yarn build` and `yarn test` for them to work.~~

Still, I verified that the changes fix my issue by testing on a project that uses the custom id field. I tested both: Refetching, and loading more data on a pagination fragment.


packages/react-relay/relay-hooks/useRefetchableFragmentNode.js
```diff
@@ -516,10 +518,21 @@ if (__DEV__) {
     getInitialIDAndType(
       memoRefetchVariables: ?Variables,
       fragmentRefPathInResponse: $ReadOnlyArray<string | number>,
+      fragmentNode: ReaderFragment,
+      componentDisplayName: string,
```
josephsavona, line 522: nit: pass `identifierField` instead of recomputing it. i would also default to `id` if identifierField is null/undefined

josephsavona (changes requested):
Overall this looks good, some small nits. Could you please create at least one unit test (see existing useRefetchableFragment unit tests) to ensure that this feature doesn't break? Thanks!

Emilios1995:
> Overall this looks good, some small nits. Could you please create at least one unit test (see existing useRefetchableFragment unit tests) to ensure that this feature doesn't break? Thanks!

Thanks for the review @josephsavona! Unfortunately, after running the existing tests, I realized that my changes break some existing functionality related to refetching or paginating `@fetchable` fragments.

Specifically, when paginating or refetching a `@fetchable` fragment, the query that the compiler generates accepts a variable named `id`, even though the fragment's `identifierField` might be something else, like `fetch_id`, for example. 

This is unlike the compiler's behavior when setting the `nodeInterfaceField` setting: In that case, the query that the compiler generates accepts a variable with the same name as `identifierField`. This is the behavior I had in mind when making this PR, but it breaks the tests related to refetching `@fetchable` types.

I think we can go forward in three different directions:
1. Make the hooks always pass both variables to the query: `identifierField` and `"id"`. The former covers use cases where `nodeInterfaceField` is set, and the later everything else, including `@fetchable` fragments. It would look something like [this](https://github.com/Emilios1995/relay/commit/35df5c9bc10c42aaa4a6c5de1ba9466b3ba3ebaa).
2. Is there some way we can get the `nodeInterfaceField` setting at runtime? If so, the hook can use that value to know what variable to pass, rather than `identifierField` as in the current state of this PR.
3. Maybe #3638 was wrong in setting the query's variable name to be `nodeInterfaceField / identifierField`. Instead, it should have remained just `id` to be consistent with the behavior for `@fetchable` types.

IMO, option 1 is very pragmatic, but both 1 and 2 are a bit leaky since they force the hooks to be aware of the compiler's quirks. Option 3 sounds more correct to me. However, adopting it would be a breaking change since some users of `nodeInterfaceField` might already be going around this bug by manually passing the id variable to the refetch/pagination variables. (Though I'm not sure how much we should worry about a breaking change affecting something that was always a workaround anyway!)

What do you think?

josephsavona:
> is there some way we can get the nodeInterfaceField setting at runtime? 

It should already be available on the refetch metadata (?)

Emilios1995:
@josephsavona The scope of this PR increased a bit. Based on your comments, I decided to add another field to the refetchable fragments' metadata, since as per my previous comments, the `identifierField` metadata wasn't enough for the hooks to effectively determine what the query variable should be named. 

The new metadata field is pretty explicit so the hooks don't need to have any complex logic to figure it out. It's named `identifierQueryVariableName` and it is exactly what the name conveys.

I know you asked for a test in the hooks to make sure this doesn't break, but unfortunately that's not straightforward. Currently _all_ the GraphQL operations for the tests are being compiled by the relay-compiler as configured in `scripts/config.tests.json`. This configuration configures the `packages` folder as its source. 

This makes it impossible to create a new test suite whose GraphQL ops are compiled with a different compiler setting, which is what we would need to test with the `nodeInterfaceIdField` setting.

There are ways around that but all of them would have a pretty heavy footprint on how the codebase is structured. (For example, one solution could be to exclude the generated artifacts from source control, and generate them on demand before the tests are run. This would allow us to compile and run the test suite as normal, compile it again with the `nodeInterfaceIdField` set, and then do a second run of `jest` but with CLI args so that it only runs a subset of the test, namely, the ones that are relevant to the `nodeInterfaceIdField` such as the refetchable and pagination ones.)

I think that with this change, adding a test to the hooks becomes less important since now the compiler is responsible for explicitly letting the hooks know what the variable should be named. The hooks should just use the provided metadata field, so there isn't any special logic it has to perform in the case when `nodeInterfaceIdField` is set. 

If you insist on having a new test for the hooks, I would require instruction on how exactly to make that happen (given the need of a separate compiler config as I explained above.)

Emilios1995:
@josephsavona Have you had a chance to look at my latest changes?

kav:
Any chance of getting this or something similar merged @josephsavona?

captbaritone (commented):
Overall this looks good, and thanks for working on it! I left a few comments. Most are trivial, but there is one about allowing the variable name to be configured independently of the field name. Curious to hear your thoughts.

Apologies for the slow response on this PR.

packages/react-relay/relay-hooks/react-cache/useRefetchableFragmentInternal_REACT_CACHE.js
```diff
@@ -448,8 +455,8 @@ function useRefetchFunction<TQuery: OperationType>(
       // If the query needs an identifier value ('id' or similar) and one
       // was not explicitly provided, read it from the fragment data.
       if (
-        identifierField != null &&
-        !providedRefetchVariables.hasOwnProperty('id')
+        identifierQueryVariableName != null &&
```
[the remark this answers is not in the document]
Emilios1995: So, not quite. The warning below was already there before this PR. It was necessary because it doesn't only check for the existence of `identifierField`, rather, it makes sure we have an `identifierValue`, which is defined as `fragmentData[identifierField]`. 

Even if the compiler invariant holds and we _have_ a `identifierField` whenever there is an `identifierQueryVariableName`, `identifierValue` might be missing for a number of different reasons outside the control of the compiler. For example, there might be a malfunction in the Relay runtime or perhaps even in the users's GraphQL server.

Nevertheless, if we hit the warning we're definitely in a corrupted state and there's a case to be made for triggering an exception via an invariant, but I would you to confirm that this is indeed what we want since I'm not familiar with the team's heuristics for when to use a warning and when to go for an invariant.

Changing the compiler output as you're describing would still be nice. I'll see if I can get around to doing it.

morrys:
any news on this PR? 👍

Emilios1995:
> any news on this PR? 👍

I've been a bit busy lately, but this week I started implementing the changes that Jordan requested. They'll be ready soon!

Emilios1995:
Hey @captbaritone! I finally got around to implementing the changes you requested. 

Please take a look at the last three commits. The first two of them add support for configuring the variable name, as you suggested. (The default keeps the current beahvior.) 
The last commit implements the "extra point" change you suggested. Namely, changing the compiler output so that the query variable name and the identifier field live as non-nullable fields inside a new struct type. This change is propagated to the interfaces exposed from `relay-runtime`, helping consumers such as the flow code in `react-relay` avoid having to handle "impossible states" where one of the fields is present and the other is not.

packages/relay-runtime/util/getRefetchMetadata.js
```diff
@@ -58,14 +62,17 @@ function getRefetchMetadata(
...
-    identifierField == null || typeof identifierField === 'string',
-    'Relay: getRefetchMetadata(): Expected `identifierField` to be a string.',
-  );
+  const identifierInfo = refetchMetadata.identifierInfo;
+  if (identifierInfo) {
+    invariant(
+      identifierInfo.identifierField == null ||
+        typeof identifierInfo.identifierField === 'string',
+      'Relay: getRefetchMetadata(): Expected `identifierField` to be a string.',
+    );
```
captbaritone, line 71: Should we add a similar invariant for `identifierQueryVariableName`?

packages/relay-runtime/util/getRefetchMetadata.js
```diff
@@ -58,14 +62,17 @@ function getRefetchMetadata(
       'this is likely a bug in Relay.',
     componentDisplayName,
   );
-  const identifierField = refetchMetadata.identifierField;
-  invariant(
-    identifierField == null || typeof identifierField === 'string',
-    'Relay: getRefetchMetadata(): Expected `identifierField` to be a string.',
-  );
+  const identifierInfo = refetchMetadata.identifierInfo;
+  if (identifierInfo) {
```
captbaritone, line 66:
```suggestion
  if (identifierInfo != null) {
```

packages/react-relay/relay-hooks/useLoadMoreFunction.js
```diff
@@ -201,7 +207,7 @@ function useLoadMoreFunction<TVariables: Variables>(
 
       // If the query needs an identifier value ('id' or similar) and one
       // was not explicitly provided, read it from the fragment data.
-      if (identifierField != null) {
+      if (identifierInfo != null) {
```
captbaritone, line 210: I'm confused by the relationship between `identifierField` and `identifierInfo` in this scope. My sense right now is that they are actually both coming from the same metadata but that `identifierInfo` is extracted locally and `identifierField` is extracted by the parent function and passed in. Is that correct?

If not, maybe we could add a comment explaining the difference between the two? If not, maybe we could get rid of one of them and either always pass in a full `identifierInfo` object, or avoid the need for parent functions to pass it in, and instead always extract it locally?

captbaritone (changes requested):

captbaritone:
Thanks for your continued work on this @Emilios1995! Left another round of comments.

packages/relay-runtime/util/getRefetchMetadata.js
Emilios1995: Sure, it doesn't hurt and it's consistent with the rest of the function. Added!

packages/react-relay/relay-hooks/useLoadMoreFunction.js
Emilios1995: You're right, having both was redundant. I just changed it so that this function doesn't need parent functions to pass `identifierInfo`. Now we just extract it locally.

packages/relay-runtime/yarn.lock
```diff
@@ -0,0 +1,111 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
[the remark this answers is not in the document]
Emilios1995: Removed.

Emilios1995:
> Thanks for your continued work on this @Emilios1995! Left another round of comments.

Thank you @captbaritone! I just pushed some changes addressing your comments.

Emilios1995:
@captbaritone Hey! Any chance we could get this merged?

captbaritone:
This has landed. Folks should be able to try it out using the @main tag of the NPM packages. Note that you'll need to upgrade both the `relay-runtime` and `relay-compiler` packages.

Emilios1995:
> This has landed. Folks should be able to try it out using the @main tag of the NPM packages. Note that you'll need to upgrade both the `relay-runtime` and `relay-compiler` packages.


Thank you, @captbaritone!

Lalitha-Iyer:
Thank @Emilios1995 for this!  Works great.

Lalitha-Iyer:
One thing I had to figure out was adding `nodeInterfaceIdVariableName` in the schema config since we used a different variable name than "id".  Probably worth documenting it someplace.

kav:
It's worth noting at present it appears the default implementation of `getDataID` sill uses a hardcoded value of `id` rather than looking up the `nodeInterfaceIdField` value so needs to be provided as well.

`getDataID: (fieldValue) => fieldValue.nodeId`

https://github.com/facebook/relay/issues/4199#issue-1553756962

Emilios1995:
@Lalitha-Iyer That's a good point. We did document the new variable [here](https://github.com/facebook/relay/pull/4053/files#diff-ea3d70d0ba8d3a78d700b797f4b88c0a189ad82ff96ae63eaa6639edcf604745R76) but I can see how it would be better to have a comprehensive guide documenting all the customization options. 

Such a guide would also be a good place to remind people they might need to customize the `getDataID` too, as @kav highlighted.

@captbaritone Would you like there to be a new Guide on relay.dev called "Using a different field for global IDs" or something similar documenting everything?

PR #4054, opened 2022-08-29
[Docs] Remove optional initialQueryRef arg from  use-query-loader.md example 

damassi:
As per [this stack overflow question](https://stackoverflow.com/questions/68708300/how-to-obtain-the-initialqueryref-in-relay), the docs are confusing on first glance. It looks like the user _needs_ to provide an `initialQueryRef` argument and the docs reference a router, but thats quite a bit out of context, and the argument isn't needed. 

I imagine quite a few devs have looked at this example, couldn't figure out how to obtain the initialQueryRef and have given up. 

This removes the reference from the code example for clarity sake and a follow-up PR should add a fully working example where the initial query ref is passed.


rbalicki2:
Good call, thanks for this change!

For additional context, the behavior of this parameter is somewhat broken: if useQueryLoader is passed an initialQueryRef and the hook commits, it will dispose of the initial query ref on unmount. If it is passed one and doesn't commit, it won't. (The behavior is also probably sketchy if you call `loadQuery` or `dispose` before commit.)

But: this hook should dispose of the initial query ref, though, because it doesn't own that query ref. So, e.g. a parent useQueryLoader's query ref, which is being used as an initial query ref here, will now be disposed and in an invalid state.

A better pattern (in **all cases**) is to do `queryRef ?? someInitialQueryRef` wherever we use the query ref with a default. This also has the benefit of being more accurately typed.

PR #4057, opened 2022-09-02
Strip extensions on file import paths when writing out TypeScript generated code

### LINKED ISSUE #4056: Relay Resolver has module resolution problem in TypeScript
Same issue as https://github.com/facebook/relay/issues/3989, but it was not completely fixed.

The dynamic imports in the AST omit extensions, however the type imports still have the extensions.

It appears that 
https://github.com/facebook/relay/blob/639abe1b23ed2187500514c331bed3aebede2f02/compiler/crates/relay-typegen/src/typescript.rs#L97
and
https://github.com/facebook/relay/blob/639abe1b23ed2187500514c331bed3aebede2f02/compiler/crates/relay-typegen/src/typescript.rs#L101

Both also need to strip extensions, as typescript requires that extensions not be provided.

ernieturner:
Fixes #4056

In the typegen for Typescript when we're writing the pathname for module and type imports, this PR strips any trailing `.ts` that might exist on the path. This is required for TypeScript file imports which don't like the trailing file suffix. This was partially fixed in https://github.com/facebook/relay/pull/4002 but not completely. 

I'm new to this repo and it's been a while since I've written some Rust, so definitely let me know if I'm doing anything wrong here or if there's any code I've missed, thanks!

I've also filled out the CLA as well.

### DIFF
--- a/compiler/crates/relay-typegen/src/typescript.rs
+++ b/compiler/crates/relay-typegen/src/typescript.rs
@@ -95,10 +95,24 @@ impl Writer for TypeScriptPrinter {
     }
 
     fn write_import_module_default(&mut self, name: &str, from: &str) -> FmtResult {
-        writeln!(&mut self.result, "import {} from \"{}\";", name, from)
+        let from_without_extension = if from.ends_with(".ts") {
+            from.strip_suffix(".ts").unwrap_or(from)
+        } else {
+            from
+        };
+        writeln!(
+            &mut self.result,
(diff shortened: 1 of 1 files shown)


rbalicki2:
Looks good to me! Thank you for making this change. I'll go ahead and import and merge it.
For future reference, it looks like this may be fixed upstream at some point: https://github.com/microsoft/TypeScript/issues/37582

Issue #4082, opened 2022-10-06
Using @required on the `node` field of a connection type causes Relay compilation to fail


ernieturner:
When attempting to use the `@required` directive on the response from a Connection type, Relay compilation errors with `All references to a field must have matching @required declaration`. As a quick example

```
const { data } = usePaginationFragment(graphql`
    fragment RefetchableAndRequired on Query @refetchable(queryName: "PaginationQuery") {
      users(first: $count after: $cursor) @connection(key: "PaginationQuery_users") {
        edges {
          node @required(action: THROW) {
            id
          }
        }
      }
    }
  `,
  queryRef,
);
```

Running this will give an error in the compiler internals

```
[ERROR] Error in the project `xxx`: ✖︎ All references to a field must have matching @required declarations. The field 'node` is @required here

path/to/file.test.tsx:463:19
  462 │                 edges {
  463 │                   node @required(action: THROW) {
      │                   ^^^^
  464 │                     id

  ℹ︎ but not @required here
<generated>: <missing source>
```

The `<generated>: <missing source>` makes me think this is happening internally in Rust. I attempted to create a number of tests to reproduce this, but wasn't able to for some reason, so I'm not sure what is unique about this. I'll probably keep poking at tests to see if I can reproduce this, but figured I'd open this in the meantime.

captbaritone:
Sorry for the pain here. This is a limitation in Relay right now, and a very poor error message. The issue is that Relay automatically inserts some additional fields into connections, and these generated fields don't have any @required directives. This conflicts with @required's validation which enforces that all instances of a field within a selection have the same @required directive. 

However, since the fields are generated, we don't have a good place to point to. If someone wants to open a PR to add this summary to the docs that would be super helpful!

Ideally we could at least have a better error message here. I think that would require either the @required transform knowing about the fields that connections insert, (or perhaps something generic that handles all generated fields?), or going the other way, and having @connection transform encode some version of @required's validation rules.

ernieturner:
Ah ok, glad to know it isn't just me! So to clarify, what I'm seeing is that `@required` isn't supported on the `node` field, nor any of it's direct descendants. But grandchildren fields can be marked with the directive. That is

```graphql
edges {  # ✅
  node { # ❌
    id # ❌
    subThing { ❌
      name # ✅
    }
  }
}
```

Does that jive with your understanding? I'll write up a PR to add some of this content to the docs.

captbaritone:
[...]

PR #4084, opened 2022-10-10
Add more content to doc site about required directive

ernieturner:
More content for the `@required` directive page to add notes about

+ How the Relay compiler will error if you use a LOG/NONE action directly beneath a THROW action
+ Caveats with locality in that the Relay compiler will error if you aren't consistent with the directive on same field in a single fragment
+ Caveats with using the directive alongside the `@connection` directive. 

Written to partially address https://github.com/facebook/relay/issues/4082 to at least document the error message and what it means.

### DIFF
--- a/website/docs/guides/required-directive.md
+++ b/website/docs/guides/required-directive.md
@@ -12,7 +12,7 @@ keywords:
 
 import DocsRating from '@site/src/core/DocsRating';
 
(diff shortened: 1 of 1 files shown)


captbaritone:
Thanks @ernieturner this looks really well done ❤️

ernieturner:
@captbaritone This issue came up internally and I hadn't realized that this change to the docsite hasn't actually been merged. Could we get this merged to the external site?

PR #4106, opened 2022-10-26
Fix typo in load-query.md documentation

jmancherje:
I believe this line should read:

"... will be retained by the relay store, preventing the data from being garbage collected" instead of "... will be retained by the relay store, preventing ~~**it**~~ the data from being garbage collected".


captbaritone:
Thanks!

PR #4112, opened 2022-10-30
Fixed typescript generation for empty variables query

### LINKED ISSUE #4111: Bad typescript type generation for queries with empty variables
## Current behavior

When using `useLazyLoadQuery` with a Query that has no variables, in a Typescript project, the Relay compiler will output the following type for the variables:

```typescript
export type Component_Query$variables = {};
```

The problem with this type is that it allows the following Typescript code to be valid:

```typescript
const data = useLazyLoadQuery<Component_Query>(
    graphql`
      query Component_Query {
        ...
      }
    `,
    { fetchPolicy: 'store-or-network' },
  )
```

This is a simple mistake that happened in our codebase, where we forgot to pass an empty argument for `variables`, and instead passed the `options` with a `fetchKey`. This results in valid code that triggers no Typescript errors, however the `fetchKey` is passed as a variable instead of as an option so it does not work. Actually, any attribute can be passed into the variables without triggering any errors

### Expected behavior

The following type or equivalent should be generated

```typescript
export type ChatBubble_CurrentUserQuery$variables = { [K in any]: never };
```

This would allow only an empty object to be valid as variables, so the `useLazyLoadQuery` usage in the example above would be invalid unless fixed to pass an empty object as variables.

fazo96:
Fixes #4111 

I built the relay compiler from main with this change and tested it on our project, and it fixes the issue

### DIFF
--- a/compiler/crates/relay-typegen/src/typescript.rs
+++ b/compiler/crates/relay-typegen/src/typescript.rs
@@ -225,7 +225,7 @@ impl TypeScriptPrinter {
 
     fn write_object(&mut self, props: &[Prop]) -> FmtResult {
         if props.is_empty() {
-            write!(&mut self.result, "{{}}")?;
+            write!(&mut self.result, "{{ [K in any]: never }}")?;
             return Ok(());
         }
 
@@ -402,7 +402,7 @@ mod tests {
     fn exact_object() {
         assert_eq!(
             print_type(&AST::ExactObject(ExactObject::new(Vec::new()))),
-            r"{}".to_string()
+            r"{ [K in any]: never }".to_string()
         );
 
         assert_eq!(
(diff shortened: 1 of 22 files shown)


alunyov (changes requested):
Thank you for submitting this PR! Can you add a test that verifies the updates?

fazo96:
hi @alunyov, I saw that there were already tests checking for this but as a result of my changes they are failing. I have updated them to match my changes and they pass locally

CI should be green now as a result

murtlatif:
@fazo96 Thank you for submitting this PR, after encountering this issue it'd be awesome if we can get this in. Could we re-request review on this to move forward here since everything seems to look good?

Also thoughts on using `Record<string, never>` rather than `[K in any]: never`?

fazo96:
hey @murtlatif like you said, it looks like this has been resolved in #4404

Closing this PR

PR #4114, opened 2022-10-31
Fix import in Docs about commitMutation

jaroslav-kubicek:
Just small DOCs fix in https://relay.dev/docs/next/api-reference/commit-mutation/


rbalicki2:
Thanks for finding and fixing this!

PR #3972, opened 2022-06-11
[VSCode Extension] Support for multiple relay configs

tbezman:
## Goal

Support repositories with multiple Relay configs when using the VSCode Extension. 

### Changes

We had to subtly change the architecture of the extension to be more contextual around a given project. We still support the happy path of no configuration required if you have a relay config in your root.

- **Breaking Change**: `relay.pathToConfig` and `relay.rootDirectory` have been deprecated in favor of `relay.projects`, an array of `{ pathToConfig: string, rootDirectory: string }`. 
- Instead of a global `RelayExtensionContext`, we now have a `RelayExtensionContext` with `projects` inside, and a `RelayProjectExtensionContext`. The latter is used for project specific functions in the extension code.
- Logging looks a bit better with a prefixed `[Project Name] ` in the output so you know which project is causing that output. This is done by passing a contextual logger function around.
- The extension's commands execute across all of the projects. Warning / Error messages should only show once per command as not to create a jarring UX.

### Testing
- I tested this out on the https://github.com/relayjs/relay-examples repo (thanks @captbaritone for the suggestion).
- I have a PR over there with configuration for this next release ready to go once this lands (https://github.com/relayjs/relay-examples/pull/235)

### Demo

Here's a demo in the relay-examples repo, one extension, multiple projects, multiple LSPs running based on the configuration in https://github.com/relayjs/relay-examples/pull/235

https://user-images.githubusercontent.com/6754223/175756716-800673e4-9ba9-40ca-a3c3-af686d007cd2.mp4

PR #4127, opened 2022-11-15
Bump version of VSCode extension, update README

ernieturner:
Does a major bump of the VSCode extension and adds new configuration options to README. I did a major version bump because (from what I could tell) the old `config.rootDirectory` and `config. pathToConfig` aren't taken into account, so this would likely break for users upgrading.

### DIFF
--- a/vscode-extension/README.md
+++ b/vscode-extension/README.md
@@ -12,7 +12,7 @@ Search for "Relay GraphQL" in the VS Code extensions panel or install through [t
 
 ## Setup
 
-- Make sure you have a Relay config file somewhere in your project.
+- Make sure you have at least one Relay config file somewhere in your project.
   - We support standard config file formats (`.yml`, `.js`, `.json`), and the the `relay` field in your `package.json`
 - Make sure you have the `relay-compiler` installed in your project. The bare minimum for this extension is v13.
 - Make sure you are able to run the `relay-compiler` command from the command line. If `yarn relay-compiler` works, it's very likely that the extension will work. See more on [debugging below](#debugging).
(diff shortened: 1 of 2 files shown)


captbaritone (changes requested):

ernieturner:
@captbaritone Did some re-organizing and cleanup of the config section to hopefully make it more clear how each of the config options works depending on single vs multiple.

ernieturner:
@captbaritone We should be able to kick off https://github.com/facebook/relay/actions/workflows/vscode.yml now right?

captbaritone:
Sorry for the delay, I've been out of office. I just tried kicking it off and got a failure related to a version mismatch. Looking into it now.

https://github.com/facebook/relay/actions/runs/3831747693/jobs/6521246260

ernieturner:
No worries! I just pushed up [this PR](https://github.com/facebook/relay/pull/4166) to bump the vscode engine which might fix this?

captbaritone:
Looks like the publish failed for a new reason: https://github.com/facebook/relay/actions/runs/3832309035

I suspect it has something to do with caching node_modules? Going to look into if we can run without cache.

captbaritone:
I'm able to reproduce the issue locally:

```
cd vscode-extension
yarn install --frozen-lockfile --ignore-scripts
yarn vsce publish
```

It errors with the conflicts before it tries to publish and realizes it has no auth.

The error seems to come from here: https://github.com/microsoft/vscode-vsce/blob/8b49e9dfdf909ad3af2b9ec9c825f5b501f6d75e/src/package.ts#L1179

I added some logging and it seems like it's not actually a case sensitivity issue, but rather that the files are getting processed twice for some reason. (This aligns with the logging which shows the files having the same case).

I'm out of time to investigate today. @ernieturner  would you be able to take a look?

ernieturner:
@captbaritone I think I have this working, but in the middle of testing it out I realized that [the multi-config PR](https://github.com/facebook/relay/pull/3972) hasn't actually landed in this repo yet. It looks like those changes have been imported but hasn't merged. Can we get that fixed up first, then I'll put up a PR to fix the publishing step?

PR #4129, opened 2022-11-15
Add configuration option to keep both ID and query text in artifacts

xuorig:
This pull request is not complete, but I'd love to get your thoughts before pushing it further.

We are currently in the process of adopting PQs. However, we want to do so progressively.  An approach we're interested in is keeping both the original text and the `id` on the client, meaning we could dynamically switch between IDs and full query documents at runtime. In this PR I've added a config option that enables this behaviour.

It's a bit challenging to test given that the current compiler tests don't exercise the code path leading to the generation of an ID, so I may need to use another approach if going for this solution.

### DIFF
--- a/compiler/crates/relay-compiler/src/artifact_content/content.rs
+++ b/compiler/crates/relay-compiler/src/artifact_content/content.rs
@@ -178,11 +178,16 @@ pub fn generate_operation(
     fragment_locations: &FragmentLocations,
 ) -> Result<Vec<u8>, FmtError> {
     let mut request_parameters = build_request_params(normalization_operation);
-    if id_and_text_hash.is_some() {
+
+    if id_and_text_hash.is_some() && project_config.always_keep_operation_text {
+        request_parameters.id = id_and_text_hash;
+        request_parameters.text = text.clone();
+    } else if id_and_text_hash.is_some() {
         request_parameters.id = id_and_text_hash;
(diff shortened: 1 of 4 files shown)


xuorig:
Just pushed another approach to this problem that feels less brittle. Moving the logic of deciding which params to keep to the persister instead. This feels a bit better, the current logic was executed during serialization of the artifacts which required a more generic approach. This really feels more like a persister-level decision. Thoughts?

soneymathew:
This feature landed in relay as part of this PR
https://github.com/facebook/relay/pull/3917
Can this approach work for you?

xuorig:
@rbalicki2 apologies for the direct ping, but we would still appreciate a feature like this in Relay rather than using a custom fork. Anything we can do to push this PR forward (or this one https://github.com/facebook/relay/pull/3917)?

captbaritone:
Thanks for the PR @xuorig. I've decided to start with @tomgasson's PR, but as he pointed out, you are both on the same page. Would you mind doing a review of https://github.com/facebook/relay/pull/3917 to see if there's anything that you noticed while implementing this PR that we might have missed over there?

captbaritone:
I'm going to close this for now, but I'll be happy to reopen if https://github.com/facebook/relay/pull/3917 doesn't end up working out.

PR #4131, opened 2022-11-15
Fix LSP completion icon for fields

beaumontjonathan:
# Incorrect LSP completion icons for scalar fields

I noticed that the VSCode extension seemed to use weird icons for completion suggestions of _value_ and _text_ for items which are actually _field_s based on the VSCode IntelliSense [docs](https://code.visualstudio.com/docs/editor/intellisense#_types-of-completions). E.g., `String` types were showing up just as plaintext type. All non-struct/union/interface/object types should just be considered a field.

### Before
Screenshot 2022-11-15 at 22 20 32

### After
Screenshot 2022-11-15 at 22 46 01


rbalicki2:
Thanks!! This makes a lot more sense, thank you for fixing it

beaumontjonathan:
No worries! I'm just trying out the VSCode plugin for the first time & there are a few other things I'd like to fix up if I get the time

PR #4134, opened 2022-11-21
Feature: LSP Fragment completions

beaumontjonathan:
# LSP Fragment Completions

Adds support for `__typename`, fragment, and inline fragment completions for `interface`/`union`/`type`.

Completions are ordered using [`sortText`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion), with the following `precedence`:

1. Fields on the type, e.g., fields from `type`/`interface`
2. `__typename`
3. `...on T` inline fragments
4. `...F` fragments

The modified tests show the behaviour, but a summary here is:

## `interface`

- `__typename: String!` completion
- `...on T` completions for the `interface` type itself and all types which implement it
- `...F` completions for all fragments `F` on the `interface` or any `implements`' of it

### Example

Screenshot 2022-11-21 at 13 59 39

## `union`

- `__typename: String!` completion
- `...on T` completions for the `union` type itself and for each variant
- `...F` completions for all fragments `F` on the `union` or any variants of it

### Example

Screenshot 2022-11-21 at 13 59 56

## `type`

- `__typename: String!` completion
- `...F` completions for all fragments `F` on the `type` or on any `interface`s or `union`s it implements

### Example

Screenshot 2022-11-21 at 14 00 10

## Questions:

- [x] Does Relay have a preferred style for spacing of fragments? Which should be used on these completions? Should this be a user configurable option?
  - `...on T` vs `... on T`
  - `...F` vs `... F`
- [x] Should inline fragment completions exist for the type itself? E.g., when fragmenting on a `type User`, should `...on User` be suggested?
  - At the moment this MR adds this for `interface` & `union` but not `type` - should this be consistent?
- [x] Will this feature be annoying for users of the LSP in orgs with a large codebase & lots of fragments? Should either of these be user configurable?
- [x] Is the `sortText` robust enough? This has been implemented by prefixing a number to the start of the `label`; it seems to work for me locally ([code](https://github.com/facebook/relay/pull/4134/files#diff-04b7155ac238238b6434c2f32c30439229c0127075572f2d2f7519b574ff90c6R1018))


compiler/crates/relay-lsp/src/completion/mod.rs
```diff
@@ -710,6 +753,7 @@ fn resolve_completion_items_for_inline_fragment_type(
     type_: Type,
     schema: &SDLSchema,
     existing_inline_fragment: bool,
+    include_fragment: bool,
```
alunyov, line 756: nit: boolean arguments maybe confusing sometimes.

Maybe we should change this into the callback -> `format_fragment_completion_label`?

alunyov (changes requested):
This is super cool! 

Thank you for working on this, just one small comment on the boolean argument, but overall looks great.

alunyov:
> Should inline fragment completions exist for the type itself? E.g., when fragmenting on a type User, should ...on User be suggested?

I don't think we need this, I think we automatically inline these fragments anyways, so they are redundant. 

> Will this feature be annoying for users of the LSP in orgs with a large codebase & lots of fragments? Should either of these be user configurable?

I think it should be fine. As long as you can refine the list of suggestions base on what you've already typed. Can you? Will it filter out irrelevant fragment spreads?

> Is the sortText robust enough?

Hmm, seems like it should be fine.

beaumontjonathan:
Awesome, thanks for looking at this so quickly! :D

-----

> > Should inline fragment completions exist for the type itself? E.g., when fragmenting on a type User, should ...on User be suggested?
> 
> I don't think we need this, I think we automatically inline these fragments anyways, so they are redundant.

Cool, I can make a small change to filter out the base type for `interface` & `union` so it has the same behaviour as `type`, i.e., so the following doesn't happen

```graphql
fragment Test on MyInterfaceType {
  |
  # Completions:
  # ...on MyInterfaceType     <-- Remove this one
  # ...on ConcreteType
}
```

-----

> > Will this feature be annoying for users of the LSP in orgs with a large codebase & lots of fragments? Should either of these be user configurable?
> 
> I think it should be fine. As long as you can refine the list of suggestions base on what you've already typed. Can you? Will it filter out irrelevant fragment spreads?

This is the behaviour while typing - so long as it _exactly_ matches the completion label then it persists, but if you close the completion window & reopen it, then you don't get any completions (*except for when you're writing a fragment spread, i.e., `...Foo|` because that's handled explicitly [here](https://github.com/beaumontjonathan/relay/blob/693cc37f419786bb8817e9496142875599ac70f4/compiler/crates/relay-lsp/src/completion/mod.rs#L550))

This logic could definitely be improved, but it would mean adjusting the AST to allow inline fragments with no identifiers, i.e., allowing `...on |`.

https://user-images.githubusercontent.com/17055343/203138940-ae206454-a939-4821-94b9-9f86e6088729.mov

compiler/crates/relay-lsp/src/completion/mod.rs
beaumontjonathan: I completely agree about boolean arguments, I rewrote with a callback and it seemed more confusing

We already use a boolean argument in a similar situation for `resolve_completion_items_for_fragment_spread`. Here `existing_fragment_spread: bool` is used to describe whether the completion is being used to rename an existing spread.

I've renamed this function and argument for inline fragments to more closely align to the existing one for fragment spreads. Now the argument is `existing_fragment_spread`.

beaumontjonathan:
@alunyov I've made some changes to clean up the arguments a little, and in the process explored how completion works for existing inline fragments. I added more tests to document this & fixed the behaviour in my change.

There is still plenty of weirdness in `relay-lsp`, but I don't think any of these changes degrades it. There's definitely more I'd like to explore about allowing completion items which aren't just at the start of the line, e.g., a partially typed inline fragment which currently doesn't work.

```graphql
{
  ...on StartOfFragmentNam|
  # ^ Completions
  # - ...on StartOfFragmentName
  # - ...on StartOfFragmentNameOther
}
```

Let me know if you'd like anything else changed :smile:

alunyov (approved):

alunyov:
Thank you @beaumontjonathan for working on this. This looks great!

I'll test this internally, and I think we can ship this. 

A few ideas on how we can improve this further (but maybe as a follow-up PRs);

- I think currently, if you type `...` it doesn't show the list of fragments. I think it wasn't working before this PR, but since we start showing fragment spreads in the suggestion list, it would be cool if we can start showing all possible fragment spreads in the case when people type `...`.
- Apply lexicographical sorting to the suggestion list. We use this for error suggestions here: https://github.com/facebook/relay/blob/908fe115d618ff032e551caf9aaec0d544ec23d5/compiler/crates/schema/src/suggestion_list.rs#L65)
  - for field names: use the user input and show suggestions in the order that is closed in this lexicographical distance to already typed value.
  - for fragment spreads: use the current operation/fragment name as the sorting reference for the suggestion list. It is a heuristic, but you more likely to use similar fragments (something like UserProfile, UserProfilePicture, UserProfileDescription, etc... are more likely the correct fragment to suggest first, rather then something that's starts with "A" like ...ApplicationSettings, etc).

compiler/crates/relay-lsp/src/completion/mod.rs
```diff
@@ -734,24 +770,24 @@ fn resolve_completion_items_for_inline_fragment_type(
...
+                label: format!("...on {type_name}"),
                 kind: None,
                 detail: None,
                 documentation: None,
                 deprecated: None,
                 preselect: None,
                 sort_text: None,
                 filter_text: None,
-                insert_text: Some(format!("{} {{\n\t$1\n}}", label)),
+                insert_text: Some(format!("...on {type_name} {{\n\t$1\n}}")),
```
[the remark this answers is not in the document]
beaumontjonathan: Sounds good, done

beaumontjonathan:
> * I think currently, if you type `...` it doesn't show the list of fragments. I think it wasn't working before this PR, but since we start showing fragment spreads in the suggestion list, it would be cool if we can start showing all possible fragment spreads in the case when people type `...`.

This is reasonable, and definitely the desired behaviour.

I looked into this initially, but it would require a change to the parser - at the moment we won't recognize that it's a fragment unless if has a identifier after the `...` [code](https://github.com/beaumontjonathan/relay/blob/8dc927fb0f0d38d8c47622eaa91939577fbd5e4e/compiler/crates/graphql-syntax/src/parser.rs#L1279).

It's definitely possible to fix this, but would have a larger blast radius & I think it could come later.

> * Apply lexicographical sorting to the suggestion list. We use this for error suggestions here: https://github.com/facebook/relay/blob/908fe115d618ff032e551caf9aaec0d544ec23d5/compiler/crates/schema/src/suggestion_list.rs#L65

^ This is really interesting. For editor completions, I think if we get the above working, then we should get this for free from the editor. This is the current behaviour for field completions in VSCode

https://user-images.githubusercontent.com/17055343/203425992-16e6bf05-68cc-4fec-9016-4740fdf602c9.mov

alunyov:
Oh, one valid case that we need to fix here. We should not add `__typename` for the root types (Query, Mutation, Subscription).

We have a validation that prevents from adding these fields: https://github.com/facebook/relay/blob/15cc327065c6a03d19549234cd63d2c2727dbe50/compiler/crates/relay-transforms/src/validations/disallow_typename_on_root.rs#L45-L47

So we should not add this to the autocomplete list for these types.

PR #4135, opened 2022-11-22
Fix: Field alias completions

beaumontjonathan:
# Fix field alias completions

The parser currently refuses to parse documents with incomplete aliases, e.g.,

```graphql
query {
  user {
    aliasField: # Expected a non-variable identifier (e.g. 'x' or 'Foo')
  }
}
```

And also places the error dialog in the wrong location - on the 

## Before

Screenshot 2022-11-22 at 23 00 50

#### Before (error message visible)

Screenshot 2022-11-22 at 23 00 12

## After

Screenshot 2022-11-22 at 22 57 56


alunyov (changes requested):
Thank you for adding this change!

Can you add a test with the updated error message?

beaumontjonathan:
@alunyov Fair comment, I've added a test :smile:

PR #4147, opened 2022-12-08
Add a new package eslint-plugin-relay-internal

captbaritone:
Adds a new package that we can use to add lint rules that apply just to the Relay code base. It starts with `no-mixed-import-and-require` which should catch when we mix require and import, which is disallowed in WWW's Flow config and so breaks our sync.

Pushing now to see what problems crop up when imported into Phabricator. Will followup there with whatever work is needed to get this to work with internal linting.

## Test Plan

I've confirmed that if I break the lint rule's tests they are caught by our top-level (Yarn workspace) jest run.

Screenshot 2022-12-08 at 3 37 12 PM

---
I've also confirmed that the lint rule triggers both in the editor and at the CLI.

Screenshot 2022-12-08 at 3 39 01 PM

Screenshot 2022-12-08 at 3 39 33 PM

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,7 +14,7 @@ module.exports = {
   root: true,
   // TODO - migrate this onto @react-native-community/eslint-config
   extends: ['fbjs'],
-  plugins: ['jest', 'relay', 'react-hooks'],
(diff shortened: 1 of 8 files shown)


captbaritone:
Failing test is pre-existing.

Issue #4146, opened 2022-12-03
Network errors not caught by error boundary in React Native when using `useLazyLoadQuery`, `useFragment` and a `store-and/or-network` fetch policy


levibuzolic:
> *Note:* This _may_ be a bug with the React Native renderer, I'm still investigating to understand exactly where the difference in behaviour is coming from. However given some of the Relay hooks APIs appear to behave correctly int he same environment I'm starting an issue here until I can find out more.

We're observing unusual behaviour in React Native (both native and web) where network errors aren't being bubbled up by Relay but only with a specific combination of Relay hook APIs.

This issue happens when using `useLazyLoadQuery` with a fetch policy of `store-and-network` or `store-or-network` **and** the data requiring network is specified in a `useFragment`. If a network error is raised, rather than being caught by the closest error boundary the component will instead render to completion _without_ it's required data, violating the type safety contracts. In production environments this results in JS level errors as the rendered components try and access data that isn't present.

I've created a [minimal reproduction in a repo](https://github.com/levibuzolic/react-native-relay-network-error) containing a handful of test cases demonstrating the broken behaviour, along witht he expected working behaviour when using slightly different APIs.

I've included more detail about the working test cases int he README.md of that repo, but I'll outline the details of the reproduction below.

## ❌ [`LazyQueryWithError`](https://github.com/levibuzolic/react-native-relay-network-error/blob/main/src/LazyQueryWithError.tsx)

Primary example of the bug, the fetch is setup to throw a JS error to simulate a network error. (The behaviour is the same with real network errors)

**Expected behaviour:** The network error would bubble up and be caught by the parent error boundary and the component would never render to completion.

**Observed behaviour:** The network error is ignored and the component renders to completion. Typically in a real environment this would lead to JS errors as the rendered component tries to access data that isn't present.

Curiously if you trigger another render in the component (e.g. by pressing a button to set some state) the error will then be correctly handled, it's only on the initial render that the error is ignored.

![Kapture 2022-12-03 at 22 14 48](https://user-images.githubusercontent.com/721323/205438241-cb515379-64a8-4598-9d4f-c47d7fdbb66b.gif)

### Simplified example

```typescript
import * as React from 'react';
import {Text} from 'react-native';
import {graphql, useFragment, useLazyLoadQuery} from 'react-relay';
import type {Query} from './__generated__/Query.graphql';
import type {Query_query$key} from './__generated__/Query_query.graphql';

export default function Parent() {
[...]

PR #4157, opened 2022-12-23
[WIP] Unit test for https://github.com/facebook/relay/issues/4146

alunyov:
This is a unit-test based on the examples from https://github.com/facebook/relay/issues/4146

It looks like the component with `useFragment` is rendered when renderPolicy = partial, unstable_isConcurrent = false.


packages/react-relay/relay-hooks/__tests__/__snapshots__/useLazyLoadQuery-ErrorBoundary-test.js.snap
```diff
@@ -0,0 +1,35 @@
...
+<div>
+  Error
+  : 
+  Test
+</div>
+`;
+
+exports[`useLazyLoadQueryNode with ErrorBoundary with (renderPolicy = partial, unstable_isConcurrent = false) should throw an error when the query fails: [Initial Render] 1`] = `"Loading..."`;
+
+exports[`useLazyLoadQueryNode with ErrorBoundary with (renderPolicy = partial, unstable_isConcurrent = false) should throw an error when the query fails: [Render after request] 1`] = `<div />`;
```
alunyov, line 25: This is where we're rendering a component with `useFragment` with missing data, instead of re-throwing the network error.

PR #4159, opened 2022-12-28
Align @defer / @stream compiler assumptions to GraphQL spec

### LINKED ISSUE #4158: Align @defer and @stream with current draft specification
Relay doesn't define the definition of these directives itself, but the rust compiler makes some assumptions on the shape of these directives. These assumptions are not in line with the currently proposed definitions for these directives:

- [@defer spec definition](https://github.com/robrichard/graphql-spec/blob/incremental/spec/Section%203%20--%20Type%20System.md#defer)
- [@stream spec definition](https://github.com/robrichard/graphql-spec/blob/incremental/spec/Section%203%20--%20Type%20System.md#stream)

Namely the `@stream` directive argument being `initial_count` instead of `initialCount` and the nullability of the `if` and `label` arguments.

The specification is already in the Draft stage and the current discussion is (to my knowledge) no longer focused on the definition of these directives, but rather on the shape of payload used to deliver the incremental results. I don't think there will be any further iteration on the directives, so I would propose to align Relay's assumptions of those directives with the spec. This would allow people to play around with existing server-side implementations and provide further feedback on the incremental deliver portion of the spec.

tobias-tengler:
### @stream

- Rename `initial_count` argument to `initialCount`
- Allow `initial_count` argument to be not defined, since the official directive definition specifies it as nullable and with a default value of `0`
- Validate `initialCount` argument being a non-negative int, if it's defined
- Make `if` argument non-null

### @defer

- Make `if` argument non-null

## Open questions

- Should I also change the shape of `@stream_connection`? I haven't touched its signature yet, since it's not part of the GraphQL specification.
- Some of the changes in this PR are **breaking changes**. `@stream` / `@defer` are experimental features in OSS, so it _shouldn't_ be a problem to just ship these changes. To my knowledge, Meta should be using these features in production though. Is this a blocker for these breaking changes?

Closes #4158


alunyov (changes requested):
Thanks for looking into this! 

I think @rbalicki2 were also looking into this at some point (so maybe have more context).

Specifically for this PR - I don't think we will be able to land that internally. So we need to figure out the strategy on how to make this work.

Some high-level observations:

1) We need to be able to minimize the changes in this diff to reduce the risk of breaking things internally.

2) For the `@defer/@stream` we need to somehow express to the compiler what type of directive is expected. Maybe via config, I think we have schema_config section in the compiler config maybe put it there. We can add an explicit version - something like what type of arguments is expected, or even SDL for these directives, or maybe have something like enum DEFER_STREAM_MODE = CustomRelay | OSS

tobias-tengler:
I also just watched the GraphQL spec meetings of December and there seems to be once more discussion around the shape of the response. In particular the necessity of the `label` on the directives is being put to question. It's also proposed to no longer include the `path` in each incremental patch. So I would anticipate that Relay _might_ has to do some more changes to how `@defer / @stream` are handled in OSS. So having a config option might be the best thing to support Meta and OSS side-by-side.

Tiedye:
Any updates?

tobias-tengler:
I can have another go at this in the coming weeks. I'll try to add a config option and keep the default as the previous behavior so all the test changes and so on can be reverted again and we have a smaller diff to land

tobias-tengler:
Closing this in favor of https://github.com/facebook/relay/pull/4467

PR #4161, opened 2022-12-30
fix: syntax higlighting homepage in darkmode not working

aliifam:
before:
image
after:
image

Make the syntax highlighter in dark mode work properly and be easy to read

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -45,4 +45,4 @@ untyped-import
 untyped-type-import
 
 [version]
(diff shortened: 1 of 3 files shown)


alunyov (changes requested):
Thank for looking into this. 

Based on the screenshot, this PR does fix the text color, but I don't think it fixes the code-highlighting (it's all white text).

How does it look in the light-mode?

PR #4162, opened 2022-12-30
[VSCode] Provide autocompletion for compiler configuration

tobias-tengler:
When the VSCode extension is active, it will now autocomplete configuration options of the compiler, both in the `relay.config.json` and the `package.json`.

Single project

relay-single-project

Multi project

relay-multi-project


vscode-extension/src/json-schemas/relay-config-json.jsonc
```diff
@@ -0,0 +1,588 @@
...
+        "enable_relay_resolver_transform": {
+          "type": "boolean"
+        },
+        "use_named_imports_for_relay_resolvers": {
+          "type": "boolean",
+          "default": true
+        },
+        "relay_resolver_model_syntax_enabled": {
+          "type": "boolean"
+        },
```
tobias-tengler, line 498: Should feature flags be properly typed or should I just not type them as they are likely very transient?

tobias-tengler (commented):
I created the JSON schema by hand, since I thought the configuration doesn't change that often. Investing in automation felt like a waste of time...

captbaritone:
This is so rad! I'm a little nervous about this falling out of sync with our config, but given that our config is already poorly documented, having this to help people out seems very worthwhile. One thing to look into: We use serde to define our config serialization/deserialization. I wonder if there are any existing tools to auto generate json schema from serde definitions?

If that does not exist, I wonder if there's some way to test that this definition stays in sync with our existing config schema?

tobias-tengler:
@captbaritone I thought about that and researched some tools, but all of them would require some manual modification of the schema after it was generated. The Rust config is also interleaved with options that aren't publicly exposed and just used for internal state.
I assumed that the existing options weren't likely to be (often) changed or removed and it wouldn't be _that bad_ to maybe miss a new minor option. If they do get out of sync, I think we could think about automating it. But as it's just an assisting tool and it getting out of sync doesn't break anything, I think it's not worth the effort at the moment. I think it should suffice if someone knows about the existence of these files and can raise a comment during code review if they spot them not being updated alongside a configuration change....

One thing I was thinking about was that it would be nice to ship the JSON schema alongside the compiler, so that the autocompletion always reflects the set of options the compiler exposes and not just the latest version of those options. I'll play around with it a bit.

captbaritone:
Given that the LSP is:

1. Already watching files and capable of providing diagnostics
2. Knows which file is _actually_ the config file (not just based on filename matching)
3. Knows how to parse and report errors on the config format

It does seem like we could find a way to have our main LSP process report errors. In fact, I believe we try to parse the config on startup and currently just crash. Maybe it would be sufficient to find that code path and look into converting the serde parse error into a DiagnositcError.

I think that happens here: https://github.com/facebook/relay/blob/2ac84385c208840dce2170661e93fdf636b0b248/compiler/crates/relay-bin/src/main.rs#L312

Maybe ` start_language_server` could accept some kind of Result of the config, or even just the config path and then do custom error handling?

Related, ideally the LSP and compiler in watch mode could actively watch the config file for changes and either automatically restart, or at least prompt for a manual restart.

tobias-tengler:
@captbaritone I think this would be great, if the compiler can actually surface wrong configuration directly in the file. I still think the JSON schema has its place though. I mainly introduced it for _discovery_ purposes and assistance in writing that configuration. I think the LSP can take over after that and help me out when there are _logical_ errors in my configuration, like invalid paths. But honestly I think the error reporting is plenty sufficient already, it's mostly about discovery and ease of use.

captbaritone:
Good point. This provides autocomplete in addition to error reporting.

captbaritone:
> One thing I was thinking about was that it would be nice to ship the JSON schema alongside the compiler, so that the autocompletion always reflects the set of options the compiler exposes and not just the latest version of those options. I'll play around with it a bit.

Sounds good. Let me know when you're ready for this PR to get looked at again.

gulpfile.js
```diff
@@ -367,6 +367,20 @@ const relayCompiler = gulp.parallel(
...
+          '.',
+          'compiler',
+          'crates',
+          'relay-config',
+          'src',
+          'config-schema.json',
+        ),
+      )
+      .pipe(gulp.dest(path.join(DIST, 'relay-compiler')));
+  },
```
tobias-tengler, line 383: I have never used gulp, but I just copied the approach used for the `schema-extensions.graphql` above. However when I run `yarn gulp release` in the root directory, the output is:
image
I would expect both `schema-extensions.graphql` and `config-schema.json` to be in the root of `relay-compiler`, but they aren't. Am I missing something or is this expected when executing it locally? (I'm on Windows)

tobias-tengler (commented):
Okay so what I've done now is remove the JSON schema from the VS Code extension and placed it in the `relay-config` crate. Being colocated with the actual config should hopefully help with reminding other developers to also update the schema.
This schema file is then copied via gulp to the `relay-compiler` NPM package. The VS Code extension already has a mechanism to locate the compiler binary in the `node_modules`, so I'm just piggybacking off of that to also locate the `config-schema.json`. If it exists it will be used to provide auto-completion in Relay's JSON based configuration files.
With this the config auto-completion should always match the compiler version.

vscode-extension/src/providers/textDocumentContentProvider.ts
```diff
@@ -0,0 +1,49 @@
...
+  'package-json-relay-config-schema';
+
+export class RelayTextDocumentContentProvider
+  implements TextDocumentContentProvider
+{
+  private readonly jsonSchemaPath: string;
+
+  constructor(context: RelayExtensionContext) {
+    this.jsonSchemaPath = path.join(
+      context.relayBinaryExecutionOptions.binaryPath,
```
captbaritone, line 25: If the user changes their `binaryPath` in the config, do we correectly recompute this?

vscode-extension/src/providers/textDocumentContentProvider.ts
```diff
@@ -0,0 +1,49 @@
...
+  onDidChange?: Event<Uri> | undefined;
+
+  provideTextDocumentContent(uri: Uri): ProviderResult<string> {
+    if (uri.authority === PACKAGE_JSON_RELAY_CONFIG_SCHEMA_PATH) {
+      return `{"properties": { "relay": { "$ref": "${RelayTextDocumentContentProvider.scheme}://${RELAY_CONFIG_SCHEMA_PATH}", "description": "Relay.js configuration" }}}`;
+    }
+
+    if (uri.authority === RELAY_CONFIG_SCHEMA_PATH) {
+      try {
+        return fs.readFileSync(this.jsonSchemaPath, {encoding: 'utf-8'});
```
captbaritone, line 41: Oder versions of the compiler will not have this file and might error here. Should we check the compiler's version before expecting this to exist?

compiler/crates/relay-config/src/config-schema.json
```diff
@@ -0,0 +1,588 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
```
captbaritone, line 2: Do you think we could add a comment above each struct/enum that is included here informing folks that changes made to that struct need to be replicated here?

captbaritone: This is also making me think that for folks who describe their config using typescript, it would be nice to have typescript definitions as well. Are there tools to generate typescript from json-schema (or vise-versa?)

gulpfile.js
captbaritone: I'm afraid I'm not sure what's going on here. Maybe you can add some logging to investigate?

vscode-extension/src/providers/textDocumentContentProvider.ts
tobias-tengler: No, I should change this. I was thinking that when the path is changed the entire extension must reload anyways, but that's not true, since the path is just used to start the compiler once, if it's changed after that it doesn't affect the system.

compiler/crates/relay-config/src/config-schema.json
tobias-tengler: I've used a JSON Schema to TypeScript npm package before, but I'm sure it exists the other way around as well. 
So you're saying there should be TypeScript types for the config. 
Where should these live? `@types/relay-compiler`? I think an entirely different repository will be even harder to sync. I'm personally of the opinion that the types should be inclined into the Relay repository, so I could also just add it straight to the relay-compiler package. We could then have a script that created the JSON schema from these TypeScript definitions at build, so that only the TS types have to be kept in sync.

tobias-tengler: Okay I've added the TypeScript types and tested two packages to generate the JSON schema from the TypeScript types. In both cases the schema is losing expressiveness in the form `oneOf`. With the handwritten schema you could for example express that only "schema" or "schemaDir" is supported, which is now no longer possible. If we go the reverse and generate the Types from the JSON schema, the types are also weird in some places.
Is this loss of expressiveness okay in light of the easier maintainability? I guess it's still better to have some guidance, but it could also be confusing because both Single project and Multiproject options are available at the root, which would lead to the compiler panicking...

vscode-extension/src/providers/textDocumentContentProvider.ts
tobias-tengler: I've added a check whether or not the package version of `relay-compiler` is >=14.2, since I'm assuming that will be the version this feature might be released in.

compiler/crates/relay-config/src/config-schema.json
captbaritone: What's the reason for needing two scripts to generate the schema? I don't follow that bit.

I suspect the best fix for the oneOf issue would be to move to a config format that was less overloaded.

What do you personally think of the tradeoff of manual vs generated? It seems like the generated one is still much better than nothing, even if it's not optimal, and that it's better than having to maintain two different versions, but I haven't tried it out enough to know how badly the generated schema misses the mark.

tobias-tengler: It's just to showcase what two different packages would output. This is not final. The generated JSON file would also not be checked in normally and the build integration is still missing. I'll mark as draft again to avoid confusion.

I would prefer the generated, if it was a bit better.
The goal of this was to assist you in writing the config file. It does that, but the oneOf issue suggested invalid configuration options that might lead to errors, which is bad. I'll try to see if there is something I can do to improve the situation, but I don't have anymore time today, sadly...

vscode-extension/src/providers/register.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  const isSemverRangeSatisfied = semver.satisfies(
+    packageManifest.version,
+    '>=14.2',
+  );
+
+  const isPrerelease = semver.prerelease(packageManifest.version) != null;
+
+  if (!isSemverRangeSatisfied && !isPrerelease) {
```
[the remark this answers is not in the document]
tobias-tengler: I'm not sure I'm following... We have to read the packageJson file above to find the version to do the semver checks. If the version doesn't match we don't register the provider and it will never attempt to read the JSON schema.

PR #4165, opened 2023-01-03
New tutorial in the docs

davidmccabe:
New tutorial
Copied to the v14 docs so there are two copies of it.
Misc. small style changes in the stylesheets.


website/docs/getting-started/installation-and-setup.md
```diff
@@ -13,8 +13,27 @@ keywords:
...
 
-## Installation
+In many situations, the easiest way to install Relay is with the `create-relay-app` package written by Tobias Tengler. Contrary to what the name suggests, this package *installs* Relay on your existing app.
+
+It currently supports apps built on Next, Vite, and Create React App. If you aren't on one of those platforms, or if it doesn't work for you for some reason, proceed to the manual steps below.
+
+To run it, make sure you have a clean working directory and run:
+
+```
+npm create @tobiastengler/relay-app
```
alunyov, line 25: Do we want to start using `install-relay` here?

I saw we already have a repo for this: https://github.com/relayjs/install-relay - can we move the @tobias-tengler PR to that repo?

Also, I think we'll need to setup the publishing to this npm module: https://www.npmjs.com/package/install-relay

davidmccabe: Yes, it's waiting for Legal to stamp that repo to make it a public repo, at that point Tobias is going to submit the PR to that repo.

I think this is fine to ship as-is in the meantime.  I'll change it to our repo as soon as that is ready.

PR #4166, opened 2023-01-03
Bump vscode engine to match types

ernieturner:
Bumping the minimum VSCode engine in extension package.json to fix publishing step - https://github.com/facebook/relay/actions/runs/3831747693/jobs/6521246260

### DIFF
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -119,7 +119,7 @@
     "build-local": "vsce package"
   },
   "engines": {
-    "vscode": "^1.60.0"
+    "vscode": "^1.65.0"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


captbaritone:
Perfect! thanks

captbaritone:
I don't see any way to have vsce run a dry run publish which would catch this, but if such a thing were possible, it would be nice to add that to our CI so that this would get caught earlier.

ernieturner:
Ya, that's odd they don't have a `--dry-run` option there. I think it makes the most sense to just never touch the `engines` field and leave it at `1.60.0` from here on out.

captbaritone:
If only json supported comments 😞

PR #4178, opened 2023-01-07
Fix tutorial typos

alvarlagerlof:
Found some typos in the tutorial. The content seems to be defined in two places so I updated them both. Please let me know if only one should be updated.


davidmccabe:
Thank you!! And yes, there are two copies of everything, thanks for catching that.

PR #4185, opened 2023-01-12
Update connections documentation

kendepelchin:


rbalicki2:
Thanks for making this change! Can you go ahead and update the unversioned file, as well? Otherwise, looks great!

davidmccabe:
Thanks!!

davidmccabe:
@rbalicki2  There have been a few PRs that don't update the versioned file, so I'll just let them all land and then do another diff where I sync the two. (Very easy to do in batch, just a `cp` command)

kendepelchin:
> Thanks for making this change! Can you go ahead and update the unversioned file, as well? Otherwise, looks great!
> 
> 
> 
> 

Sorry didn't see this as fast as I wanted. I can always update it if needed

Issue #4092, opened 2022-10-14
Bug: Polling does not stop after a component is unmounted


sajinshrestha:
## Issue
I'm trying to switch from Apollo Client to Relay, but running into an issue with polling queries. Polling does not stop after the component is unmounted.

Example code:  
```
  const query = useLazyLoadQuery<TodoAppQuery.TodoAppQuery>(
    TodoAppQuery.default,
    {
      userId: "me"
    },
    {
      networkCacheConfig: {
        poll: 1_000
      }
    }
  );
```

## Expected behavior
Polling to stop after the component is unmounted.

Bug Repo - https://codesandbox.io/s/happy-hamilton-29f8r0?file=/src/PageTwo.tsx

```
"react": "^18.2.0",
"react-dom": "^18.2.0",
"react-relay": "^14.1.0",
"relay-runtime": "^14.1.0"
```

josephsavona:
This is definitely a bug, a possible cause is that in some cases we delay cleanup from hooks until after a timeout (in case another component is in the process of mounting and needs the same data). A workaround is to switch to `useQueryLoader()` and load the query in a useEffect.

merrywhether:
`useQueryLoader` also exhibits the same bug, but dropping down one more level to `loadQuery` + `useEffect` does work:

```
  const environment = useRelayEnvironment();

  useEffect(() => {
    const queryRef = loadQuery(
      environment,
      opts.subscription,
      opts.variables,
      { networkCacheConfig: { poll: POLL_MS },
    );
    return queryRef.dispose;
  }, [opts, environment]);
```

PR #4187, opened 2023-01-13
feat(bug): update-hook

admirsaheta:
Update to the useLazyLoadQuery hook 
-  Added useEffect and useRef inside to store the fetchObservable subscription directly to the object.
- Which then stores the subscription inside the useEffect hook 
- After which it unsubscribed from the subscription in the cleanup function

Reported issue #4092


alunyov (changes requested):
Thank you for submitting this PR!

Can you add a test case for these changes (a case that would fail without these changes)?

PR #4188, opened 2023-01-13
Update mutations-updates.md

rogerroelofs:
Fix Like Mutation. The mutation uses a variable named `doesLike` while the data field is named `doesViewerLike`.


davidmccabe:
Thanks!!

PR #4192, opened 2023-01-18
Can we migrate Relay from common js to es6 modules?

captbaritone:
Differential Revision: D42580242


captbaritone:
Just a trial to see if GitHub CI has any problems with this.

PR #4194, opened 2023-01-19
Fix unexpected bold texts

mizdra:
There was a case where an asterisk was intended to be printed but was treated as a special character for bold.

You can check the issue at the following URL.

- https://github.com/facebook/relay/blob/10e20cc7344cb9130308fb5e5b3bb6f5e95e3514/packages/relay-compiler/README.md
- image

### DIFF
--- a/packages/relay-compiler/README.md
+++ b/packages/relay-compiler/README.md
@@ -68,7 +68,7 @@ file sources, and "listen" to the file changes in the "watch" mode. If
   enabling this the babel plugin needs `artifactDirectory` to be set as well.
   [string]
 - `excludes` Directories to ignore under `src`. [array] [default:
-  ["**/node_modules/**", "**/__mocks__/**", "**/__generated__/**"]]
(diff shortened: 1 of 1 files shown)


packages/relay-compiler/README.md
```diff
@@ -68,7 +68,7 @@ file sources, and "listen" to the file changes in the "watch" mode. If
   enabling this the babel plugin needs `artifactDirectory` to be set as well.
   [string]
 - `excludes` Directories to ignore under `src`. [array] [default:
-  ["**/node_modules/**", "**/__mocks__/**", "**/__generated__/**"]]
+  ["\*\*/node_modules/\*\*", "\*\*/__mocks__/\*\*", "\*\*/__generated__/\*\*"]]
```
mizdra, line 71: `**/node_modules/**` should be correct.

https://github.com/facebook/relay/blob/10e20cc7344cb9130308fb5e5b3bb6f5e95e3514/compiler/crates/relay-compiler/src/config.rs#L593-L599

PR #4197, opened 2023-01-20
Small fix on tutorial (interfaces-polymorphism.md)

vbornand:
Small error on getting poster.joined. Must use data.joined due to data masking.

The code is correct in the source. Error only in the doc.


alunyov (changes requested):
Thank you for the contribution, and sorry with the late review.

Can you please update the source of this file and re-submit PR? https://github.com/facebook/relay/blob/e0fb6f896d235b19fc8957798c6eaf27b180108b/website/docs/tutorial/interfaces-polymorphism.md?plain=1

Thank you!

PR #4220, opened 2023-02-14
Fix typos

MattKleinsmith:


alunyov (changes requested):
Thank you for the fix! Can you also update the source of these versioned docs https://github.com/facebook/relay/blob/main/website/docs/tutorial/queries-1.md

PR #4225, opened 2023-02-18
If artifactDir doesn't exist, create it

### LINKED ISSUE #3782: Custom `artifactDirectory` has to exist and is not automatically created (like `__generate__`)
When specifying a custom `artifactDirectory` in the relay compiler config, ​the compiler requires this directory to exist. Otherwise you get:

```
[ERROR] Config `/home/.../frontend/relay.config.js` is invalid:
​- The `artifactDirectory` does not exist at `/home/.../frontend/src/query-types`.
```

This is particularly annoying when using git, as you cannot check empty folders into git. The typical "`.gitkeep` file in folder" trick does not work well, because the relay-compiler deletes all files from the artifact directory on each compile, always leading to a dirty working directory.

I think the best is to use the same behavior as for the default `__generated__` folder: create it, if it doesn't exist.

nrempel:
Closes #3782

This PR is in response to the comment https://github.com/facebook/relay/issues/3782#issuecomment-1029060702

With this change, the artifactDirectory is created instead of failing with `ArtifactDirectoryNotExistent`.

Note: I've removed the `ArtifactDirectoryNotExistent` case and used `unwrap` here instead since this case should be impossible with this change. If you prefer to keep the error variant, I'm happy to throw it back in.

### DIFF
--- a/compiler/crates/relay-compiler/src/config.rs
+++ b/compiler/crates/relay-compiler/src/config.rs
@@ -8,6 +8,7 @@
 use std::env::current_dir;
 use std::ffi::OsStr;
 use std::fmt;
+use std::fs::create_dir_all;
 use std::path::Path;
 use std::path::PathBuf;
 use std::sync::Arc;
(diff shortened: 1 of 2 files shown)


nrempel:
Hi @alunyov, pinging you regarding https://github.com/facebook/relay/issues/3782#issuecomment-1029060702

Do you think this is a worthy change? If not, happy to close this PR. Thanks!

PR #4227, opened 2023-02-21
Json schema

tbezman:


tobias-tengler:
Is your goal also to provide the JSON schema via the VS Code extension? I already explored this in #4162, but I got stuck on creating a really great schema that correctly utilizes oneOf to differentiate between multi and single project setups and didn't have much time recently to look into it.
Your approach looks much more integrated. Maybe you can re-use the editor parts I already built :)

tbezman:
@tobias-tengler My plan was to write some documentation alongside this and recommend people use this in their relay config since most editors support json schemas. I haven't throughly inspected the schema. You down to chat and talk through some of the problems you faced?

tobias-tengler:
Sure :) I started by modeling the config options in typescript, so that people using `relay.config.js` could also profit from the types. Then I tried to derive a JSON schema from that since I only have little experience with Rust and so TypeScript felt more natural to me. The problem I faced was that the TypeScript to JSON package I used produced anyOf instead of oneOf for unions like the Multi or Single project config. This lead to all options being available in the schema so it was hard to differentiate in the edtior where a property belonged to. Maybe that issue isn't present in the Rust converter you're using...

tbezman:
@tobias-tengler Wow, great insight! It looks like the package I'm using is having the same issue. I'm looking now to see what kind configuration they support to resolve this. Is it a fundamental issue or should we be able to work around this?

tobias-tengler:
I considered doing a find and replace in the schema to just replace all anyOf with oneOf, but it's extremely Jacky of course 🙈
I haven't found an option to toggle the behavior in my JS lib, hopefully you have more luck 🤞

tbezman:
@tobias-tengler Seems like the library we're using makes the decision for us based on some heuristics https://github.com/GREsau/schemars/blob/master/schemars_derive/src/schema_exprs.rs#L274

I manually changed the `anyOf` to `oneOf` and VSCode seems to be handling it fine. I don't like this but my god this is so much better than what we have today

tbezman:
Like, I wouldn't mind writing a short script to change the `anyOf` to `oneOf` after generation

tbezman:
@tobias-tengler Here's an example of what was outputted by this PR https://pastebin.com/nfxAcp5H

tbezman:
It also looks like we could generate a typescript file using something like https://www.npmjs.com/package/json-schema-to-typescript so people get autocomplete in `relay.config.json` and `relay.config.js`

tobias-tengler:
On my phone so I can't check thoroughly, but at a glance the "schema" (and maybe "src") property should be required in the single config. I think it's not, due to how the config is constructed internally, but as a developer I have to specify them or the compiler panics.

.github/workflows/docusaurus.yml
```diff
@@ -19,6 +19,22 @@ jobs:
...
+          override: true
+
+      - uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --features vendored --manifest-path compiler/Cargo.toml --bin relay-config-schema --release
+      
+      - name: Generate JSON Schema
+        run: |
+          ./compiler/target/release/relay-config-schema > website/static/relay-config-schema.json
```
tobias-tengler, line 35: We are doing this so people can reference the JSON schema via the Relay website?
In my old PR I opted into including the JSON schema in the relay-compiler npm package, so that the JSON schema would match the installed version of the compiler in the future.

scripts/fixJsonSchema.js
```diff
@@ -0,0 +1,16 @@
...
+// The Rust Crate `schemars` is not properly outputting `oneOf` so we'll manually edit it here
+
+const fs = require('fs');
+
+const schemaPath = './website/static/relay-config-schema.json';
+
+const fileContents = fs.readFileSync(schemaPath, 'utf-8');
+const parsedFileContents = JSON.parse(fileContents);
+
+parsedFileContents.oneOf = parsedFileContents.anyOf;
```
tobias-tengler, line 11: There should be some nested `anyOf`s that need to be covered as well if I remember correctly.

tobias-tengler (commented):
@tbezman Could you upload the JSON schema for me to review as well? I can't checkout and build the code rn.

.github/workflows/docusaurus.yml
tbezman: This makes way more sense and would be versioned appropriately. I'll make this change tonight.

scripts/fixJsonSchema.js
tbezman: Call me crazy, but I'm pretty sure all of the `anyOf`s should be converted to `oneOf`s?

tobias-tengler: Yea, I agree. I don’t think there’s any anyOf in there that can’t be replaced by a oneOf.

PR #4237, opened 2023-03-05
Add missing relay-runtime to installation script

MarceloPrado:
Self-explanatory, it seems the `relay-runtime` dependency is missing in the installation instructions.

After following the ["Installing in a Project"](https://relay.dev/docs/getting-started/installation-and-setup/) guide, I jumped to the [tutorial docs](https://relay.dev/docs/tutorial/queries-1/) and noticed I didn't have the runtime installed.


rbalicki2:
Thank you for catching and fixing this! Could you please modify the non-versioned markdown file as well? Otherwise, lgtm!

MarceloPrado:
@rbalicki2 thanks for the feedback, addressed!

Issue #3795, opened 2022-02-10
Document multi-project config options


kejistan:
There is some decent [documentation for the single-project config](https://github.com/facebook/relay/tree/main/packages/relay-compiler#configuration) options, but no mention of multi-project configs in any of the documentation. The [test project](https://github.com/facebook/relay/blob/d6ee691846bd806a113922a466dbb4d7b6b562fc/compiler/test-project/relay.json) uses a config in the multi-project format and [the source for the format](https://github.com/facebook/relay/blob/d6ee691846bd806a113922a466dbb4d7b6b562fc/compiler/crates/relay-compiler/src/config.rs#L523) is pretty easy to read, but it'd be nice to have this documented in the same place as the single-project format.

alunyov:
I think it is a good suggestion.

This section is something that should go into "Advanced Setup" or similar, as most of the projects do not need this config.

sibelius:
when should one use multi-project config?

if I have a monorepo with many relay different project should I have a single relay config with all projects? or should I have many relay configs?

alunyov:
@sibelius 

I think for multi-config you'll need to have these two requirements:

1) You need to build all relay projects in the mono-repo with a single command
2) Every sub-project in the mono-repo have different schema/extensions/JS/TS language/persist config, etc

hanford:
are multi project configurations still possible? I'm trying to configure one with the rust based compiler but see

```
Error("unknown field `projects`, expected one of `schema`, `src`, `artifactDirectory`, `include`, `extensions`, `excludes`, `schemaExtensions`, `noFutureProofEnums`, `language`, `customScalars`, `eagerEsModules`, `persistConfig`, `isDevVariableName`, `codegenCommand`, `jsModuleFormat`, `schemaConfig`, `moduleImportConfig`, `typegenPhase`, `featureFlags`", line: 0, column: 0).
```

orta:
As it can't hurt to have another working example, I switched to a monorepo and have this `relay.config.json`:

```json
{
  "root": ".",
  "sources": {
    "apps/puzzmo.com": "app",
    "apps/studio": "studio"
  },
  "projects": {
    "app": {
      "language": "typescript",
      "output": "apps/puzzmo.com/src/__generated__",
      "schema": "api-schema.graphql",
      "eagerEsModules": true
    },
    "studio": {
      "language": "typescript",
      "output": "apps/studio/src/__generated__",
      "schema": "api-schema.graphql",
      "eagerEsModules": true
[...]

PR #4238, opened 2023-03-07
Support multiple Relay project configurations in extension

ernieturner:


.nvmrc
```diff
@@ -0,0 +1 @@
+16
```
ernieturner, line 1: Let me know if there's a reason we don't want to do this. When I tried running `yarn` from the root of the repo on Node 18 I was getting

```
[15:30:49] 'bundleTask' errored after 118 ms
[15:30:49] Error: error:0308010C:digital envelope routines::unsupported
    at new Hash (node:internal/crypto/hash:71:19)
    at Object.createHash (node:crypto:133:10)
```

Pretty sure it's a webpack problem with the version we're on: https://github.com/webpack/webpack/issues/14532

package.json
```diff
@@ -20,9 +20,6 @@
     "test-dependencies": "node ./scripts/testDependencies.js",
     "typecheck": "flow check"
   },
-  "workspaces": [
-    "vscode-extension"
-  ],
```
ernieturner, line 25: Using yarn workspaces wrecks havoc with the VSCode extension publishing tool. It fails to find the right dependencies, duplicates deps, etc. I'm not sure if there was benefit we were getting from this, but removing it makes the bundle/publishing of the extension much easier.

vscode-extension/tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "lib": ["es2020"],
     "outDir": "out",
     "rootDir": "src",
-    "sourceMap": true,
+    "sourceMap": false,
```
ernieturner, line 8: No need to generate sourcemap files within the compiled extension output

vscode-extension/package.json
```diff
@@ -115,8 +145,11 @@
     "typecheck": "tsc --noEmit",
     "prettier-check": "prettier -c .",
     "lint": "eslint --max-warnings 0 .",
-    "vscode:prepublish": "tsc",
-    "build-local": "vsce package"
+    "vscode:prepublish": "rm -f tsconfig.tsbuildinfo && rm -rf out && yarn run esbuild-base -- --minify",
+    "build-local": "vsce package",
+    "esbuild-base": "esbuild ./src/extension.ts --bundle --outfile=out/extension.js --external:vscode --format=cjs --platform=node",
+    "esbuild": "npm run esbuild-base -- --sourcemap",
+    "esbuild-watch": "npm run esbuild-base -- --sourcemap --watch"
```
ernieturner, line 152: Pretty much just copied this from the VSCode docs https://code.visualstudio.com/api/working-with-extensions/bundling-extension#using-esbuild

ernieturner:
I don't really understand the unit test failures at this point since my changes shouldn't have touched any of that. Maybe the cargo.lock file needs to be updated, not sure.

vscode-extension/src/compiler.ts
```diff
@@ -6,52 +6,50 @@
...
-    cwd: context.relayBinaryExecutionOptions.rootPath,
+    cwd: context.project.binaryExecutionOptions.rootPath,
   });
 
   terminal.sendText(
-    `${context.relayBinaryExecutionOptions.binaryPath} ${args.join(' ')}`,
+    `${context.project.binaryExecutionOptions.binaryPath} ${args.join(' ')}`,
   );
 
-  terminal.show();
```
ernieturner, line 34: This disables the terminal auto-showing when switching to a workspace that starts the compiler by default. I don't see much reason to auto-show this when switching workspaces and personally find it a bit annoying

noghartt:
Any update on that PR? What do we need to finish and merge this PR?

cc @captbaritone @tobias-tengler

sho-pb:
Do you plan to release this Pull Request?

I added it to relay.packages and settings.json as documented, but it did not work correctly and upon investigation I noticed that this PR has not yet been merged.

ldeveber:
> Do you plan to release this Pull Request?
> 
> I added relay.projects to settings.json as documented, but it did not work correctly and upon investigation I noticed that this PR has not yet been merged.

I just came looking to find out why that wasn't working, since it was documented as a feature. Do we know when this would be released?

Screenshot 2024-02-13 at 5 10 07 PM

ohardy-agrisight:
Same for me, can we merge this MR and do a release ? Necessary for mono repo :)

net:
What is everyone doing as a workaround to handle monorepos with multiple relay configs?

noghartt:
> What is everyone doing as a workaround to handle monorepos with multiple relay configs?

You can pass projects into your `relay.config.js` like this:

```ts
module.exports = {
  root: ".",
  sources: {
    // put here the source of each project, the root dir of each package of your monorepo
  },
  projects: {
    project1: {
      schema: "...",
      language: "typescript",
    },
    // ...
  },
}
```

shadiramadan:
The workaround doesn't support `artifactDirectory` unfortunately.

ArturGraefenstein:
Here is what works for monorepo and the VSCode extension:  [https://github.com/facebook/relay/issues/3795#issuecomment-2789225473](url)

PR #4254, opened 2023-03-21
Extract events into named types

captbaritone:
Being able to refine log events to named types in tests would be very helpful, so I've(?) refactored the union of log events to be a union of named types.

This refactor brought to you by ChatGPT. 

Prompt: 

> Can you refactor this Flow code to extract each log event type into its own named type, with the LogEvent type being a union of those types?

Took a few tries because the output was longer than it could emit as one response:

> Can you do the last ones, starting with StorePublishLogEvent (and the resulting union of all named types)


captbaritone:
CI failure is unrelated. Started breaking with https://github.com/facebook/relay/commit/41ace35007f65783c76d762ec09aac5e325695ba. Looking into that separately.

Issue #4257, opened 2023-03-24
install missing field handler for `node` by default or add to docs


robertf224:
Hi!  My company adopted Relay recently and it has been a joy to use 😄 

Just ran into some behavior I found unexpected.  I was building a controlled user dropdown with props like this:
```typescript
interface UserDropdownProps {
  fragment: UserFragment$key;
  onSelect: (id: string) => void;
}
```

I'm doing (1) a query inside of the component to search/render users (2) triggering the onSelect callback when one is selected (3) fetching the user from outside the component after selection (4) then rendering it inside the component via the fragment prop.  I was expecting the user to be cached during (3) so that (4) wouldn't cause my dropdown to suspend, but I realized it was suspending because I didn't have a missing field handler for the `node` field, realized my mistake after finding [this comment](https://github.com/facebook/relay/issues/3219#issuecomment-713172012).  In hindsight this makes some sense, but I think I sort of reasonably expected that `node` queries would automatically hit the cache given it's part of the spec, and the missing field handler [docs](https://relay.dev/docs/guided-tour/reusing-cached-data/filling-in-missing-data/) seem to suggest you're supposed to just be supplying missing field handlers for non-`node` fields: "However, Relay doesn't have this knowledge by default, so we need to configure it to encode the knowledge that a node(id: 4) "is the same as" user(id: 4)."

Any reason we shouldn't just install a `node` missing field handler by default?  Or otherwise any opposition to my adding a note to this effect to the docs?  Thanks!

sibelius:
can you add to the docs?

robertf224:
docs PR merged, closing

PR #4259, opened 2023-03-24
update docs for missing field handlers

robertf224:
in response to https://github.com/facebook/relay/issues/4257


rbalicki2:
Thanks for catching and fixing this!! Once you sign the CLA, I'll import and get it landed.

PR #4265, opened 2023-03-28
WIP

gearnode:


captbaritone:
@gearnode Seems reasonable. I think you also need to include this artifact down below in the `main-release` step. Otherwise it won't be included in the published package.

PR #4267, opened 2023-03-29
Fix grammar error in the fragments page of docs

h7y:
> The avoid these and many other problems, we can move the data requirements for the Story component into Story.tsx.

The -> To


rbalicki2:
Thanks for noticing and fixing this!

PR #4275, opened 2023-04-07
Create stand-alone PR to fix VSCode extension build

ernieturner:
This ports over a subset of the changes from [this PR](https://github.com/facebook/relay/pull/4238) that just fix some build issues with the VSCode extension. This works to build the extension locally, which should in theory also work with the Github Action publish step, but I can't fully verify that since it requires the marketplace PAT to run. However, we **should not** do a publish after this PR merges because the current tagged version is 2.0.0 and we should wait for the other PR to land before we publish, especially since we're not going to publish 2.0.0 and instead are planning on publishing version 1.3.0.


ernieturner:
@captbaritone Here's the PR to just fix the build changes for the extension.

captbaritone:
Thank you!

Issue #4045, opened 2022-08-14
[website]: Elaborate why checking in generated files is recommended or remove the recommendation


jakobvase:
Hi Relay!
On the website, there are two places where you recommend adding the relay generated files to source control:

* https://relay.dev/docs/getting-started/step-by-step-guide/#41-configure-relay-compiler
* https://relay.dev/docs/glossary/#compiler

I've been digging around a lot, and I can't find an explanation anywhere. What are the benefits of checking in the generated files to source control? The best answer I've got is an old discussion on the discord:

> charklewis — 08/09/2021
Has anyone had any situations where uploading the generated graphql files to git is required? Or do you generate it as required (ie locally, in CI/CD etc) 

> Gabriel Nordeborn — 08/09/2021
I typically commit them to source control, but I think what people do differs. In general, if you're not too annoyed by a bunch of extra files, it's quite nice to keep them in source control, so you can spot changes between versions etc

> lenny — 08/09/2021
one thing to keep in mind with this approach is merge conflict resolution, as the artifacts tend to cause conflict if people are touching the same files or even fragments that affect the same queries. Wonder how folks deal with this? 

> Gabriel Nordeborn — 08/09/2021
Yeah that's a good point. One way of dealing with it is having a CI action that makes sure that the right output is generated, which guards for that at least some

But if that's the reason for the recommendation, I don't think it's good enough to warrant a recommendation in the official docs?

PR #4284, opened 2023-04-14
Remove recommendation to commit generated files to source control

amonger:
This recommendation seems to be anecdotal, and doesn't have any reason to back it up.

It would be good to remove this as a recommendation (even suggesting to gitignore?) Or providing a reason why we should hold onto these files.

Related PR: https://github.com/facebook/relay/issues/4045

### DIFF
--- a/website/docs/getting-started/step-by-step-guide.md
+++ b/website/docs/getting-started/step-by-step-guide.md
@@ -205,7 +205,7 @@ cd your-app-name
 yarn start
 ```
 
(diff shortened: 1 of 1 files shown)


captbaritone:
This is intentional. Without checking in these files other tools that operate on your code in CI, or people who want to edit non-Relay code will have a broken experience unless they independently run Relay.

I'd be happy to accept a PR that adds this context.

PR #4293, opened 2023-04-22
Replace invalid variable characters for import name

JingLi1998:
Closes: #4294 

When using provided variables, with ESM, the Relay compiler incorrectly assumes that the module path can also be reused for the default import name.

This is not valid Javascript as variables cannot use '.' or '-' characters.

To fix, simply replace '.' and '-' with underscores.

Before:
```
import include-can-edit-providers-three.relayprovider from './../include-can-edit-providers-three.relayprovider';
import include-can-edit-providers.relayprovider from './../include-can-edit-providers.relayprovider';
const providedVariablesDefinition: ProvidedVariablesType = {
  "__relay_internal__pv__includecaneditprovidersrelayprovider": include-can-edit-providers-three.relayprovider,
  "__relay_internal__pv__includecaneditprovidersthreerelayprovider": include-can-edit-providers.relayprovider
};
```

After:
```
import include_can_edit_providers_two_relayprovider from './../include-can-edit-providers-two.relayprovider';
import include_can_edit_providers_relayprovider from './../include-can-edit-providers.relayprovider';
const providedVariablesDefinition: ProvidedVariablesType = {
  "__relay_internal__pv__includecaneditprovidersrelayprovider": include_can_edit_providers_relayprovider,
  "__relay_internal__pv__includecaneditproviderstworelayprovider": include_can_edit_providers_two_relayprovider
};
```

I'm unfamiliar with Rust so am unsure if there is a better approach.

### DIFF
--- a/compiler/crates/relay-codegen/src/build_ast.rs
+++ b/compiler/crates/relay-codegen/src/build_ast.rs
@@ -1860,12 +1860,14 @@ impl<'schema, 'builder, 'config> CodegenBuilder<'schema, 'builder, 'config> {
                             provider.module_path().to_str().unwrap().intern(),
                         )
                     };
+                
+                let variable_name = (provider.original_variable_name.to_string() + "_provider").intern();
 
                 Some(ObjectEntry {
                     key: def.name.item.0,
                     value: Primitive::JSModuleDependency(JSModuleDependency {
                         path: provider_module,
-                        import_name: ModuleImportName::Default(provider.module_name),
+                        import_name: ModuleImportName::Default(variable_name),
                     }),
                 })
             })


captbaritone:
Thanks for looking at this! The approach does not seem robust though, since there may be other characters that are valid for filenames for but not for JS variables. What about using `provider.original_variable_name.0`, perhaps with a prefix to avoid name collisions. I believe that's already a GraphQL variable, which I believe means its also a valid JS variable name.

JingLi1998:
@captbaritone thank you for the suggestion it makes a lot of sense. I've updated the implementation to use the existing variable and added a prefix of `__relay_internal__pv__`.

compiler/crates/relay-codegen/src/build_ast.rs
```diff
@@ -1860,12 +1860,14 @@ impl<'schema, 'builder, 'config> CodegenBuilder<'schema, 'builder, 'config> {
                             provider.module_path().to_str().unwrap().intern(),
                         )
                     };
+                
+                let variable_name = ["__relay_internal__pv__", &provider.original_variable_name.to_string()].join("").intern();
```
captbaritone, line 1864: Can we replace the slice+join with a simpler concatenation using the `+` operator?

Also, can we go with `<variableName>_provider`. A suffix instead of a prefix and avoid the cryptic "pv" dimminution.

JingLi1998: @captbaritone Thanks for the feedback I've updated it to reflect the changes you requested. I had assumed that the `__relay_internal__pv` was the preferred way to prefix variables since I saw it used elsewhere in the codebase 😅

PR #4303, opened 2023-05-04
Update printer.rs

sibelius:
enable tsx on relay resolvers

### DIFF
--- a/compiler/crates/relay-codegen/src/printer.rs
+++ b/compiler/crates/relay-codegen/src/printer.rs
@@ -661,7 +661,7 @@ fn get_module_path(js_module_format: JsModuleFormat, key: StringKey) -> Cow<'sta
             let extension = path.extension();
 
             if let Some(extension) = extension {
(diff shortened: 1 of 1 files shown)


compiler/crates/relay-codegen/src/printer.rs
```diff
@@ -661,7 +661,7 @@ fn get_module_path(js_module_format: JsModuleFormat, key: StringKey) -> Cow<'sta
             let extension = path.extension();
 
             if let Some(extension) = extension {
-                if extension == "ts" || extension == "js" {
+                if extension == "ts" || extension == "tsx" || extension === "js" {
```
ubmit, line 664: Shouldn't it be `extension == "js"` instead?

```suggestion
                if extension == "ts" || extension == "tsx" || extension == "js" {
```

captbaritone:
Would you mind doing a quick search in the compiler to see if there are other places we make this same, or similar errors?

PR #4306, opened 2023-05-09
docs: Fix typo in `getting-started/installation-and-setup.md`

alan910127:
Fix a typo in the relay config file,
should be `excludes` instead of `exclude`.


captbaritone:
Thanks!

Issue #3856, opened 2022-04-03
babel-plugin-relay and @babel/plugin-transform-modules-commonjs generate undefined variable references with eagerEsModules enabled


kejistan:
I'm not sure yet whether this is a bug in the relay plugin or a bug in the babel plugin, but the end result is that the `graphql` tags are with a sequence expression and there is an additional import added to the top of the file. Unfortunately at some point that import is renamed, but only some of the references are updated. I've put together a [minimal repro here](https://github.com/kejistan/relay-ts-example).

The result of babel on the minimal file are in the repo's README, but I'm adding an annotated version below:

```js
"use strict";

// Added by babel-plugin-relay. initially named _testQuery, but later renamed to _testQuery2
var _testQuery2 = _interopRequireDefault(require("./__generated__/testQuery.graphql"));

var _reactRelay = require("react-relay");

function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }

// The result of expanding the graphql tag, I've trimmed the details to make the issue more obvious,
// The second element in this sequence expression should also be _testQuery2, but is still named
// _testQuery.
const query = (_testQuery2.default.hash && /* trimmed */, _testQuery);
```

The expectation is that all references to `_testQuery` and `_testQuery2` should be the same.

artola:
@alunyov I found the same problem, when `eagerEsModules` is enabled the generated code is not correct, see the Babel compiled code a similar example of @kejistan shown above. In this case, a Jest test with this Babel config:

```javascript
module.exports = {
  presets: [
    '@babel/preset-env',
    ['@babel/preset-react', {runtime: 'automatic'}],
  ],
  plugins: ['babel-plugin-relay'],
};
```

Note: `@babel/preset-env` calls `@babel/plugin-transform-modules-commonjs`.

With eagerEsModules = true ... throws `ReferenceError: _trialQuery is not defined`

```txt
"use strict";

Object.defineProperty(exports, "__esModule", {
  value: true
});
exports["default"] = Trial;

var _trialQuery2 = _interopRequireDefault(require("./__generated__/trialQuery.graphql"));

var _reactRelay = require("react-relay");

var _jsxRuntime = require("react/jsx-runtime");

function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { "default": obj }; }

function Trial() {
[...]

PR #4309, opened 2023-05-10
Avoid reusing AST nodes

eoin:
Reusing AST nodes can result in subtle bugs, for example [this issue](https://github.com/facebook/relay/issues/3856) is caused by AST node reuse breaking [memoization](https://github.com/babel/babel/blob/main/packages/babel-helper-module-transforms/src/rewrite-live-references.ts#L275-L276) in the `@babel/plugin-transform-modules-commonjs` plugin.

The first commit here introduces [@babel/helper-check-duplicated-nodes](https://github.com/babel/babel/tree/main/packages/babel-helper-check-duplicate-nodes) which breaks a number of tests. The second commit fixes those tests.

Fixes https://github.com/facebook/relay/issues/3856.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
   "dependencies": {
     "@babel/core": "^7.18.6",
     "@babel/generator": "^7.18.6",
+    "@babel/helper-check-duplicate-nodes": "^7.18.6",
     "@babel/parser": "^7.18.6",
     "@babel/plugin-proposal-nullish-coalescing-operator": "^7.0.0",
     "@babel/plugin-proposal-optional-catch-binding": "^7.0.0",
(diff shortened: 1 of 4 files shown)


captbaritone:
Awesome! Thanks for working on this. Would it be possible to add a regression test for https://github.com/facebook/relay/issues/3856. It should be as simple as adding a file to https://github.com/facebook/relay/tree/main/packages/babel-plugin-relay/__tests__/fixtures

captbaritone:
Also, [babel-check-duplicated-nodes](https://www.npmjs.com/package/babel-check-duplicated-nodes) looks pretty simple and pretty stable. Could we just vendor that one function instead of taking on the additional dependency?

eoin:
@captbaritone thanks for the review. I hadn't spotted that there's [@babel/helper-check-duplicate-nodes](https://github.com/babel/babel/tree/main/packages/babel-helper-check-duplicate-nodes), which appears to have changed more over time, has test coverage, etc... I'm inclined to depend on that, instead of vendoring. What do you think?

Regarding a regression test; the fixture + snapshot approach is tricky, in that for the issue to manifest we need to add something to the transform chain that relies on AST node uniqueness. That's doable, but before I go down that path, is it worth it?

captbaritone:
Makes sense. I'll import this as is. Thanks again!

PR #4312, opened 2023-05-10
Upgrade rustix to prevent build error

captbaritone:
Currently the compiler build is failing in GitHub CI with the following error:

```
error[E0425]: cannot find function `makedev` in module `c`
  --> /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/rustix-0.37.3/src/backend/libc/fs/makedev.rs:44:17
   |
44 |     unsafe { c::makedev(maj as i32, min as i32) }
   |                 ^^^^^^^ not found in `c`
   |
help: consider importing this function
   |
2  | use crate::fs::makedev;
   |
help: if you import `makedev`, refer to it directly
   |
44 -     unsafe { c::makedev(maj as i32, min as i32) }
44 +     unsafe { makedev(maj as i32, min as i32) }
   |
```

Update done with:

```
cargo update -p rustix@0.37.3 --precise 0.37.19
```

I'm not entirely sure what in this upgrade is resolving the error, but it does seem to resolve it locally. Will check CI on this PR to confirm it fixes it in CI as well.

### DIFF
--- a/compiler/Cargo.lock
+++ b/compiler/Cargo.lock
@@ -95,9 +95,9 @@ dependencies = [
 
 [[package]]
 name = "bitflags"
-version = "1.2.1"
+version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cf1de2fe8c75bc145a2f577add951f8134889b4795d47466a54a5c846d691693"
+checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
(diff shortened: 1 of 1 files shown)


anmonteiro:
0.37.3 -> 0.37.4 also works so I suspect it was the libc upgrade in https://github.com/bytecodealliance/rustix/compare/v0.37.3...v0.37.4

anmonteiro:
Another alternative is `cargo update`, which also seems to work and dedupes a few of the packages, including Rustix.

PR #4313, opened 2023-05-11
Create shared mock for scheduler

captbaritone:
The version of scheduler we get internally is always the latest. However, on GitHub we get the version is a transitive dependency from `react-test-renderer`.

Some methods in the unstable_mock have been renamed between these two versions. This mock file provides a centralized place to reconcile those differences so that the same tests can work both internally and on GitHub.

I also tried manually upgrading `react-test-renderer` to the `@experimental` tag in the hopes of getting GitHub to use the same version we have internally, and then transitively inherit the more recent version of `scheduler`. Unfortunately that created many test failures for reasons I don't currently fully understand.

### DIFF
--- a/packages/react-relay/__tests__/mockScheduler.js
+++ b/packages/react-relay/__tests__/mockScheduler.js
@@ -12,9 +12,8 @@
 const invariant = require('invariant');
 const Scheduler = require('scheduler/unstable_mock');
 
-// The version of scheduler we get internally is always the latest. However, on
-// GitHub we get the version is a transitive dependency from
-// `react-test-renderer`.
+// The version of scheduler we get internally is always the latest. However, the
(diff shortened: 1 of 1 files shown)


packages/react-relay/__tests__/mockScheduler.js
```diff
@@ -0,0 +1,37 @@
...
+ * @format
+ * @oncall relay
+ */
+
+const invariant = require('invariant');
+const Scheduler = require('scheduler/unstable_mock');
+
+// The version of scheduler we get internally is always the latest. However, on
+// GitHub we get the version is a transitive dependency from
+// `react-test-renderer`.
```
eoin, line 17:
```suggestion
// The version of scheduler we get internally is always the latest. However, the
// version we get on GitHub is a transitive dependency from `react-test-renderer`.
```

PR #4317, opened 2023-05-17
Use same field name for pagination that compiler generates

dwwoelfel:
It looks like the compiler [generates the refetchable query](https://github.com/facebook/relay/blob/4b1b59eb06dae90ba3c6f9e5034ac7845fee0cf2/compiler/crates/relay-transforms/src/refetchable_fragment/fetchable_query_generator.rs#L94) using the `nodeInterfaceIdField` as the variable name, but the runtime always uses "id" as the variable name.

This should fix `usePaginationFragment` for users that set a custom `nodeInterfaceIdField` in their `schemaConfig`.

### DIFF
--- a/packages/react-relay/relay-hooks/useLoadMoreFunction.js
+++ b/packages/react-relay/relay-hooks/useLoadMoreFunction.js
@@ -214,7 +214,7 @@ function useLoadMoreFunction<TVariables: Variables>(
             identifierValue,
           );
         }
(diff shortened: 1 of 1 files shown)


captbaritone:
I think this is a duplicate of https://github.com/facebook/relay/pull/4053?

captbaritone:
Closing for now. Feel free to ping here if this is different than https://github.com/facebook/relay/pull/4053.

PR #4337, opened 2023-06-07
update fbjs to address ReDOS from ua-parser-js

cross19xx:
This PR bumps `fbjs` from `3.0.4` to `3.0.5`. `3.0.5` contains a fix for the [Regular Expression Denial of Service (ReDoS)](https://security.snyk.io/vuln/SNYK-JS-UAPARSERJS-3244450) vulnerability found in `ua-parser-js` and resolved [here](https://github.com/facebook/fbjs/pull/507)

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "eslint-plugin-react-hooks": "4.6.0",
     "eslint-plugin-relay": "1.8.3",
     "eslint-plugin-relay-internal": "link:./packages/eslint-plugin-relay-internal",
-    "fbjs": "^3.0.2",
+    "fbjs": "^3.0.5",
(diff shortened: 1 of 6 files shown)


captbaritone:
Looks like our version fbjs needs to stay in sync with the version used in the Meta monolith. We only use fbjs on trusted inputs, so I don't believe we are concerned with the ReDOS possibility.

cross19xx:
> Looks like our version fbjs needs to stay in sync with the version used in the Meta monolith. We only use fbjs on trusted inputs, so I don't believe we are concerned with the ReDOS possibility.

@captbaritone  But, let's not forget this is an open-sourced project, and others will be concerned about the ReDOS possibility

PR #4339, opened 2023-06-10
Normalize separators in paths used in Watchman dirname expressions

leops:
It seems like Watchman only supports using forward slashes as paths separators in `dirname` expressions (at least). This can be an issue on Windows if any of the paths specified in the Relay configuration file (such as `src` or `schemaExtensions`) contains a platform-specific path separator (backward slash) as the `dirname` expressions generated by the Relay compiler to find and watch sources files will silently fail and return no match.

It might be possible to fix this at the Watchman level somehow but it seems like it would be a heavy, possibly breaking change, so I came up with a simpler solution specifically for the Relay Compiler that ensures all path separators are normalized to forward slashes before sending the path to Watchman.

I've tried to keep the impact of normalization low by avoiding string reallocations if the path doesn't contain any backward slash, and the new function is covered by a few checks in a new unit test.

### DIFF
--- a/compiler/crates/relay-compiler/src/file_source/watchman_query_builder.rs
+++ b/compiler/crates/relay-compiler/src/file_source/watchman_query_builder.rs
@@ -5,6 +5,8 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+use std::borrow::Cow;
+use std::path::Path;
 use std::path::PathBuf;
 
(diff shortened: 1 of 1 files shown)


captbaritone:
Hey @leops thanks for looking into this. Could you open an issue (or find an existing one) on the Watchman repo? Would love to rule out fixing this at the proper level of abstraction before we resort to trying to solve it here.

PR #4347, opened 2023-06-16
Enable __token field based on compiler schema config

### LINKED ISSUE #3941: Unknown field "__token" when using @fetchable
https://relay.dev/docs/next/guided-tour/refetching/refetching-fragments-with-different-data/ mentions the ability to use @fetchable for refetchable fragments rather than needing to implement the Node type. I followed a guide (https://dev.to/marais/fetchable-with-relay-1meb) to doing this, but ran into an issue with "__token" not being defined for my @fetchable type. I tried adding a custom handler for this directive on my backend (apollo server), but it wouldn't let me override a field prefixed with "__" (internal field).

Is better support for this on the way, or is this just deprecated / unsupported externally? It would be much nicer if I could avoid bloating my `Query.node` field with every single type

edvinerikson:
This fixes #3941 so that fetchable directive is usable in OSS. 

# Why is this needed?

fields starting with `__` is reserved in the GraphQL spec and since this field isn't part of the spec any attempt of using the field on a server that is spec compliant will result in an error. 

# What changed?

* Added a new opt-in `enableTokenField`  config option.
* Conditionally include the `__token` field in refetchable fragment selections based on the config option.

# How has this been tested?

One could argue we should add a new test asserting the old behavior, I wasn't too sure how to go about adding a special fixture though.

The tests now assert that `__token` is not being generated. 

cc @captbaritone

### DIFF
--- a/compiler/crates/relay-compiler/tests/compile_relay_artifacts/fixtures/fragment-on-non-node-fetchable-type.expected
+++ b/compiler/crates/relay-compiler/tests/compile_relay_artifacts/fixtures/fragment-on-non-node-fetchable-type.expected
@@ -154,13 +154,6 @@ fragment fragmentOnNonNodeFetchableType_ProfilePicture on User {
             "alias": null,
             "args": null,
             "kind": "ScalarField",
-            "name": "__token",
-            "storageKey": null
-          },
-          {
-            "alias": null,
-            "args": null,
-            "kind": "ScalarField",
             "name": "id",
             "storageKey": null
           }
@@ -204,7 +197,6 @@ fragment fragmentOnNonNodeFetchableType_RefetchableFragment on NonNodeStory {
     id
   }
   fetch_id
-  __token
 }
 
 
@@ -292,13 +284,6 @@ fragment fragmentOnNonNodeFetchableType_RefetchableFragment on NonNodeStory {
       "args": null,
       "kind": "ScalarField",
       "name": "fetch_id",
-      "storageKey": null
-    },
-    {
-      "alias": null,
-      "args": null,
-      "kind": "ScalarField",
-      "name": "__token",
       "storageKey": null
     }
   ],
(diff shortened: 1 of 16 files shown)


captbaritone:
Thanks for the quick PR! A few style suggestions. Other than that, my main concern is that we get this tested.

Let's ensure we have tests that cover both cases. You can change https://github.com/facebook/relay/blob/d5d1c07d7a44856f82c9a652697e9662a091d4c6/compiler/crates/relay-transforms/tests/refetchable_fragment/mod.rs#LL18C60-L18C67 to build a custom schema config and then use a pattern [like this](https://github.com/facebook/relay/blob/0bf353a3c9f023adcebdae69e70595f45409cad1/compiler/crates/relay-docblock/tests/parse/mod.rs#L64-L66) to make it conditional based on a comment in the fixture file.

Then please add tests which ensure we cover both cases.

Finally, does this introduce a possibility that a type would have neither a an id field _or_ a token field? Do we have validation anywhere to catch that?

edvinerikson:
I'll look into adding tests for this using a flag like that, thanks for the pointers!

edvinerikson:
>Finally, does this introduce a possibility that a type would have neither a an id field or a token field? Do we have validation anywhere to catch that?

We still fetch the `id` field if it exists. And we always fetch the fetchable id. As long as the type you add `fetchable` to has an `id` field we should be OK.But I suppose we could add validation for the scenario where `tokenFieldEnabled: false` and the type does not have the `id` field available.

edmundgentle:
Is there a reason this PR hasn't been accepted yet? I'm running into this problem and without editing the generated files, I can't see another way around this until this PR is merged?

captbaritone:
Apologies for loosing track of this. Let me import/rebase this and play around with it a bit.

captbaritone:
Import failed due to this PR being old and our internal tool getting stuck when trying to rebase. I'm going to take these commits, rebase them myself and open a new PR to import. The author info on the commit should remain

captbaritone:
Rebased here: https://github.com/facebook/relay/pull/4889 Going to close this PR. We can continue the conversation on that PR.

Issue #3965, opened 2022-06-10
Typesafe Updaters with TypeScript language in compiler config cause compiler to panic


captbaritone:
Ideally we could catch this and emit a diagnostic pointing to the use of the typesafe updater.

```
╭─[00:00]: captbaritone at moffo in ~/projects/relay-resolvers-todomvc on branch*
╰─➤ yarn relay
yarn run v1.22.18
warning ../package.json: No license field
$ relay-compiler
[INFO] querying files to compile...
[INFO] [default] compiling...
thread 'tokio-runtime-worker' panicked at 'Getters and setters with different types are not implemented in typescript. See https://github.com/microsoft/TypeScript/issues/43662', crates/relay-typegen/src/typescript.rs:240:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[ERROR] A thread that the Relay compiler spun up did not shut down gracefully: panic
[ERROR] Compilation failed.
[ERROR] Unable to run relay compiler. Error details:
A thread that the Relay compiler spun up did not shut down gracefully: panic
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
╭─[00:00]: captbaritone at moffo in ~/projects/relay-resolvers-todomvc on branch*
╰─➤
```

captbaritone:
cc @rbalicki2

TJ09:
The listed blocker, https://github.com/microsoft/TypeScript/issues/43662, is resolved as of TypeScript 5.1 (released in June, we're now on 5.2). I'm not sure if Relay has a process for requiring a specific minimum version of TS, but is it now theoretically possible to remove the compiler restriction and allow typesafe updaters on TypeScript?

tobias-tengler:
I've already implemented it here: https://github.com/facebook/relay/pull/4370
Just needs to be merged...

tobias-tengler:
It's all ready to go from my side for a couple months now :) Someone on the Relay team like @captbaritone or @alunyov needs to review and merge it...

tobias-tengler:
This can now be closed!

PR #4370, opened 2023-07-09
[relay-compiler] Type-safe updaters for TypeScript

### LINKED ISSUE #4212: New tutorial uses mutation APIs not available for TypeScript users
## Reporting issues with Relay

This is more of a "maybe we should provide a link elsewhere" but you cannot use `optimisticUpdater` in a TypeScript project as it triggers a compiler panic:

```
$ relay-compiler
[INFO] querying files to compile...
[INFO] [default] compiling...
thread 'tokio-runtime-worker' panicked at 'Getters and setters with different types are not implemented in typescript. See https://github.com/microsoft/TypeScript/issues/43662', crates/relay-typegen/src/typescript.rs:249:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[ERROR] A thread that the Relay compiler spun up did not shut down gracefully: panic
[ERROR] Compilation failed.
[ERROR] Unable to run relay compiler. Error details: 
A thread that the Relay compiler spun up did not shut down gracefully: panic
error Command failed with exit code 1.
```

Not strictly a problem ( because you can read the thread and see someone from this team saying "hey this thing is blocking relay") but it does mean folks outside of FB can't [follow the mutations section](https://relay.dev/docs/tutorial/mutations-updates/#improving-the-ux-with-an-optimistic-updater)

Maybe just a link to the reference to optimisticResponse part of [`useMutation`](https://relay.dev/docs/api-reference/use-mutation/#return-value)'s reference?

tobias-tengler:
TypeScript [supports specifying different getter and setter types](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-1.html#unrelated-types-for-getters-and-setters) since 5.1, so I've updated the type generation logic around `@updatable` and `@assignable` to  work with TypeScript.

I've also updated the notice on the website and made the documentation around type-safe updaters available externally (if this is not wanted, just revert 33612d24d8d28c5116623031eaf26ac3e475e7e0).

Companion PR for `@types/relay-runtime`: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66013 (Merged)

Closes #4212 #3965


website/docs/guided-tour/updating-data/imperatively-modifying-linked-fields.md
```diff
@@ -21,7 +21,7 @@ import {OssOnly, FbInternalOnly} from 'docusaurus-plugin-internaldocs-fb/interna
 
 :::caution
 
-Because in TypeScript, [getters and setters cannot have different types](https://github.com/microsoft/TypeScript/issues/43662), and the generated types of getters and setters is not the same, `readUpdatableQuery` is currently unusable with TypeScript. `readUpdatableFragment` is usable, as long as the updatable fragment contains only scalar fields.
+`readUpdatableQuery` and `readUpdatableFragment` require TypeScript 5.1 or later.
```
captbaritone, line 24: What happens today if you try to use TypeSafe updaters with an earlier version of TypeScript? Syntax error?

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/updatable-fragment-spread-and-regular-spread.expected
```diff
@@ -0,0 +1,48 @@
...
+	firstName
+}
+==================================== OUTPUT ===================================
+import { FragmentRefs } from "relay-runtime";
+export type updatableFragmentSpreadAndRegularSpreadQuery$variables = {};
+export type updatableFragmentSpreadAndRegularSpreadQuery$data = {
+  readonly me: {
+    readonly $updatableFragmentSpreads: FragmentRefs<"updatableFragmentSpreadAndRegularSpread_updatable_user">;
+    readonly " $fragmentSpreads": FragmentRefs<"updatableFragmentSpreadAndRegularSpread_user">;
+  } | null;
```
captbaritone, line 23: In the Flow version this is object `?` instead of `| null`. I believe `?` also includes `undefined`. In general, values read out of the store can always be `null` (server said they were null) or `undefined` (value was missing from the store).

I wonder if this is an existing imperfection in the TypeScript types?

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/updatable-fragment-spread-and-regular-spread.expected
```diff
@@ -0,0 +1,48 @@
...
+};
+export type updatableFragmentSpreadAndRegularSpreadQuery = {
+  response: updatableFragmentSpreadAndRegularSpreadQuery$data;
+  variables: updatableFragmentSpreadAndRegularSpreadQuery$variables;
+};
+-------------------------------------------------------------------------------
+import { FragmentRefs } from "relay-runtime";
+export type updatableFragmentSpreadAndRegularSpread_updatable_user$data = {
+  firstName: string | null;
```
captbaritone, line 32: See above. Should this also include `undefined`, as it does in Flow?

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/updatable-fragment-spread-multiple.expected
```diff
@@ -0,0 +1,47 @@
...
+-------------------------------------------------------------------------------
+import { FragmentRefs } from "relay-runtime";
+export type updatableFragmentSpreadAndRegularSpread_2_updatable_user$data = {
+  firstName: string | null;
+  readonly " $fragmentType": "updatableFragmentSpreadAndRegularSpread_2_updatable_user";
+};
+export type updatableFragmentSpreadAndRegularSpread_2_updatable_user$key = {
+  readonly " $data"?: updatableFragmentSpreadAndRegularSpread_2_updatable_user$data;
+  readonly $updatableFragmentSpreads: FragmentRefs<"updatableFragmentSpreadAndRegularSpread_2_updatable_user">;
+};
```
captbaritone, line 37: In Flow this object, and  `updatableFragmentSpreadAndRegularSpread_2_updatable_user$data` above is inexact (`...`). I don't have a good sense of why, but presumably that means we need some TypeScript equivalent? Or maybe it can be removed from the Flow types?

https://flow.org/en/docs/types/objects/#exact-and-inexact-object-types

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/updatable-fragment-spread-multiple.graphql
```diff
@@ -0,0 +1,14 @@
+query updatableFragmentSpreadAndRegularSpreadQuery {
+	me {
+		...updatableFragmentSpreadAndRegularSpread_updatable_user
```
captbaritone, line 3: Nit: Two spaces not tabs.

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/updatable-operation-plural-field-no-spreads.expected
```diff
@@ -0,0 +1,18 @@
...
+    id
+  }
+}
+==================================== OUTPUT ===================================
+export type UpdatableQuery$variables = {};
+export type UpdatableQuery$data = {
+  get nodes(): ReadonlyArray<{
+    readonly id: string;
+  } | null> | null;
+  set nodes(value: []);
```
captbaritone, line 13: Flow has an explicit `void` return value here. Should we have that (`undefined`) here as well?

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/updatable-operation-plural-field-with-spreads.expected
```diff
@@ -0,0 +1,27 @@
...
+}
+==================================== OUTPUT ===================================
+import { FragmentRefs } from "relay-runtime";
+export type UpdatableQuery$variables = {};
+export type UpdatableQuery$data = {
+  get nodes(): ReadonlyArray<{} | null> | null;
+  set nodes(value: ReadonlyArray<{
+    readonly __typename: "User";
+    readonly __id: string;
+    readonly " $fragmentSpreads": "Updatable_user";
```
captbaritone, line 19: See above about `inexact` objects.

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/updatable-operation-assignable-fragment-plural.expected
```diff
@@ -0,0 +1,40 @@
...
+import { FragmentRefs } from "relay-runtime";
+export type UpdatableQuery$variables = {};
+export type UpdatableQuery$data = {
+  get node(): {
+    readonly __typename: "User";
+    get parents(): ReadonlyArray<{}>;
+    set parents(value: ReadonlyArray<{
+      readonly __typename: "User";
+      readonly __id: string;
+      readonly " $fragmentSpreads": "Assignable_user";
```
captbaritone, line 26: Should this be wrapped in a `FragmentRefs` type?

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/updatable-operation-assignable-fragment.expected
```diff
@@ -0,0 +1,27 @@
...
+}
+==================================== OUTPUT ===================================
+import { FragmentRefs } from "relay-runtime";
+export type UpdatableQuery$variables = {};
+export type UpdatableQuery$data = {
+  get viewer(): {} | null;
+  set viewer(value: {
+    readonly __typename: "Viewer";
+    readonly __id: string;
+    readonly " $fragmentSpreads": "Assignable_viewer";
```
captbaritone, line 19: Should this be wrapped in a FragmentRefs type?

captbaritone (changes requested):
Thanks for working on this! I've left some comments, but overall I think this should be possible to land.

compiler/Cargo.lock
```diff
@@ -52,13 +52,13 @@ checksum = "9b34d609dfbaf33d6889b2b7106d3ca345eacad44200913df5ba02bfd31d2ba9"
 
 [[package]]
 name = "async-trait"
-version = "0.1.58"
+version = "0.1.71"
```
captbaritone, line 55: Could we omit this `Cargo.lock` change from the PR?

tobias-tengler: Sure, I just included it, since the pipeline would fail otherwise.

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/updatable-operation-plural-field-no-spreads.expected
tobias-tengler: I don't think it would make a difference, but I could be wrong.

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/updatable-fragment-spread-and-regular-spread.expected
tobias-tengler: If the store can also return undefined, then I'd say the types should definitely reflect that. I've added the necessary change: a0f14566813e66dfb3e8ac36a90c59dada1789c6 Might be too big for this PR though, so maybe a follow-up?

website/docs/guided-tour/updating-data/imperatively-modifying-linked-fields.md
tobias-tengler: Actually nothing, if you don't remove the `@ts-nocheck` in the generated files 😅 so I removed the warning...

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/updatable-operation-assignable-fragment.expected
tobias-tengler: I think this might've been a bug, please check if my proposed fix is the correct one: c8d3d10dc0e90741eed1672adbab2a8a1a1a82bf

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/updatable-fragment-spread-multiple.expected
tobias-tengler: How would such an inexact type be typed in Typescript, would it just be a union of
```ts
type inexact = original_obj & { [k: string]: unknown };
```

Might need more input from some TypeScript wizard before I add this...

tobias-tengler:
@captbaritone I addressed everything that I could, please give it another look 🙏

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/updatable-fragment-spread-multiple.expected
captbaritone: Maybe we can first start with understanding why its inexact in Flow? What breaks if we make it inexact? Is this specific to updatable fragments, or are all fragment keys inexact in Flow?

Maybe the semantics of TypeScript are different and we don't actually need them to be inexact in TypeScript.

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/updatable-operation-assignable-fragment.expected
captbaritone: That looks right to me.

captbaritone:
@tobias-tengler Thanks for the quick turn around. Left a few more comments.

Have you tried running this with a real TypeScript project? Does it generate valid types that appear to work with a minimal example of using typesafe updaters? Would love to have validation that this works end to end.

tobias-tengler:
@captbaritone I've built an example [here](https://github.com/tobias-tengler/relay-typesafe-ts-updaters).

While testing I noticed that the `babel-plugin-relay` doesn't behave correctly.
When having an `@assignable` fragment, the generated artifact will only export a `validate` function, not an artifact. `babel-plugin-relay` doesn't yet know about this, so it tries to replace the graphql tag with an import - which fails of course, since there's nothing to import.

I've fixed this in cd7ae1e658e796121a430027ccbfb2e64736e76a. Do you think that is fine? Meta probably uses another transform, if this hasn't led to issues for you yet, right?

If we want to go forward like this, I can also submit a PR for Next.js to update their transform. `vite-plugin-relay` should be covered by the update to `babel-plugin-relay`.

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/updatable-fragment-spread-multiple.expected
tobias-tengler: In other places inexact objects are used for fragments, if their MaskStatus is unmasked. I know too little about how fragments work to infer why it would be important to have the same characteristics when updating a fragment... 

I tried tracing the git log for an explanation of why an InexactObject was being used, but haven't found anything. @rbalicki2 you seem to have added the code, do you perhaps remember why the InexactObject was being used?

captbaritone:
Thanks for taking the time to validate this end to end!

> the generated artifact will only export a validate function, not an importable node. babel-plugin-relay doesn't yet know about this

My understanding of the Babel transform is that it simply replaces the `graphql` tagged template literal with an import of the generated artifact for the fragment. The transform does not make any assumptions about the exports of that module. It sounds like the compiler is correctly creating the generated artifact for the `@assignable` fragment. 

Maybe that assumption is not actually correct? Or perhaps only correct when using `require` instead of `import`, or relative paths rather than Haste style global module names for imports?

Could you share what error you are seeing as well as what the transform is generating before you apply your fix?

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/updatable-operation-plural-field-with-spreads.graphql
```diff
@@ -0,0 +1,9 @@
+query UpdatableQuery @updatable {
+  nodes(ids: [4]) {
+    ...Updatable_user
+  }
+}
+
+fragment Updatable_user on User @assignable {
+  __typename
+}
```
captbaritone, line 9: Internal lint rules will reject the diff if trailing line endings are not included :(

Could you add one here?

compiler/crates/relay-typegen/src/visit.rs
```diff
@@ -1566,8 +1573,8 @@ fn make_prop(
                                          });
                                          let assignable_fragment_spread_ref= Prop::KeyValuePair(KeyValuePairProp {
```
captbaritone, line 1574: Unrelated: Strange that `cargo fmt` is not adding a space between `assignable_fragment_spread_ref` and `=`...

tobias-tengler:
The problem I’m seeing is that normally the compiler creates and default exports a node from the artifact: https://github.com/tobias-tengler/relay-typesafe-ts-updaters/blob/main/src/__generated__/AppAssignableQuery.graphql.ts

For assignable fragments it doesn’t: https://github.com/tobias-tengler/relay-typesafe-ts-updaters/blob/main/src/__generated__/testAssignableUser.graphql.ts
Which should be correct since the assignable fragment acts just as a marker from my understanding, we’re never actually passing it to a relay-runtime API.
With `require` the plugin probably works as is, since `require` doesn’t demand any exports. I think this is only a problem with `eagerEsModules`.

i think there are two options:
1. Make the plugin aware of this case and remove graphql tag in the case of an assignable fragment (what I implemented)
2. Default export an empty object or another dummy value from the assignable fragment artefact to satisfy the esm import

compiler/crates/relay-typegen/src/visit.rs
tobias-tengler: Was `cargo fmt` not run in a while? If I run it locally I get ~150 file changes. But I think my Rust toolchain is just setup incorrectly...

tobias-tengler:
I've updated the example application I've built to introduce the breaking changes related to the assignable fragment: https://github.com/tobias-tengler/relay-typesafe-ts-updaters/commit/7e699d7ac8d884f13d1eba66e89455b06b405d07

The error I get in the browser:

> Uncaught SyntaxError: The requested module '/src/__generated__/AppAssignableUser.graphql.ts' does not provide an export named 'default' (at App.tsx:11:30)

The generated source looks like this:

```js
import _AppAssignableUser from "/src/__generated__/AppAssignableUser.graphql.ts";
// ...
_AppAssignableUser.hash && _AppAssignableUser.hash !== "a1488fcf5af224c81a9df5a6d12e1b14" && console.error("The definition of 'AppAssignableUser' appears to have changed. Run `relay-compiler` to update the generated files to receive the expected data."),
_AppAssignableUser;
```

Which is incorrect, there is no default export in `AppAssignableUser.graphql.ts` and also no hash.
I just checked and it seems like they are also not created when using flow.

tobias-tengler:
Bump @captbaritone 😇

captbaritone:
Sorry I haven't been able to follow up here, I've started looking into the hash issue and haven't yet found a satisfactory answer. Open questions I'd like to answer:

1. Is there a reason we don't can't have a hash exported from the assignable fragment artifact?
2. Why don't we see this same error internally? Reading the Babel transform code (which is the same as we use internally) it looks to me like the hash check is performed in dev for Flow/Haste import/require. Still need to understand why that's not the case.

I'll try to followup next week, but in the mean time you can dig into these if you want to help unblock.

The ideal solution is that the assignable fragment artifact exports a hash, since that would avoid the throw issue and also enable better feedback in dev when you forget to run the compiler.

tobias-tengler:
@captbaritone I've reverted my changes to `babel-plugin-relay` and instead added the hash export to the assignable fragments. 

This is what a generated artifact looks like now: https://github.com/tobias-tengler/relay-typesafe-ts-updaters/commit/150f54a33c92aa4d604a49665be178e7bccc0f7f

tobias-tengler:
@captbaritone Did you have time this week to look further into this?

I think for your second point it's clear what's happening:
- The assignable fragment artefact doesn't have a default export
- `require` always gives you and empty object
- The transform replaces the graphql fragment directive + string with `_AppAssignableUser.hash && _AppAssignableUser.hash !== "something" && console.error(...)`
- `_AppAssignableUser` is `const _AppAssignableUser = require(...)` so it will be the following object`{ validate: func() }`
- `_AppAssignableUser.hash` therefore evaluates to `false`, no error is logged and everything seems correct

tobias-tengler:
@captbaritone sorry for the ping, I'm sure you're very busy, but I don't want this to stall. Could you take a look at my above comments or point out another member of the Relay Team that could do so?

captbaritone:
Thanks for the ping and sorry for the delay. I'll look in the morning!

compiler/crates/relay-compiler/src/artifact_content/content.rs
```diff
@@ -718,12 +720,43 @@ fn generate_assignable_fragment(
...
+        content_sections.push(ContentSection::Generic(section));
+    }
+    // -- End Fragment Node Hash Section --
+
+    // -- Begin Fragment Node Export Section --
+    let mut section = GenericSection::default();
+    write_export_generated_node(&project_config.typegen_config, &mut section, "node", None)?;
+    content_sections.push(ContentSection::Generic(section));
+    // -- End Fragment Node Export Section --
+
```
captbaritone, line 753: Amazing! Love to see this foundational improvement to Relay (will help both Flow and TS users)

Would you mind splitting this addition of the hash into its own PR?

captbaritone (changes requested):
This is great! Can we split the addition of hashes into its own PR? After that I can import.

tobias-tengler:
@captbaritone I moved the addition of the node to the following PR: https://github.com/facebook/relay/pull/4409

tobias-tengler:
@captbaritone bump 🙈

michaelmcneilnet:
Will this be released soon? This is a very valuable fix for us.

tobias-tengler:
I think the team is thinking about cutting a 15.1.0 release soon, see https://github.com/facebook/relay/issues/4440
Should probably be a few week out. if you can't wait you can always use the `relay-compiler@main` tag.

Issue #4379, opened 2023-07-18
Nullable types in TypeScript are incorrectly typed


tobias-tengler:
The Relay store can emit `null` **and** `undefined` for fields. In flow this is correctly typed via `?`, but in TypeScript the fields type is just a union of the raw type and `null`. The `undefined` case isn't covered.

PR #4380, opened 2023-07-18
[relay-compiler] Fix type generation of nullable types in TypeScript

tobias-tengler:
Closes: #4379

## Release Notes

Previously, the relay-compiler would generate a union of the actual type and `null` (`T | null`) for nullable fields when targeting TypeScript. This was not correct, since the Relay store can also return `undefined` in the case of missing data. This version now produces a union of the actual type, `null` and `undefined` (`T | null | undefined`). 
Consequently you now also have to check both the `null` and `undefined` case, before accessing a field's value:

```ts
const data = useLazyLoadQuery(/* ... */);

// Notice the `!=` instead of `!==` to check against null and undefined
if (data.nullableField != null)
{
  // Safely access data.nullableField
}
```

Since this is a pretty big change, we're offering the `typescriptExcludeUndefinedFromNullableUnion` feature flag in the `relay-compiler` config to keep the old type generation behavior for now:

```json
{
  "language": "typescript",
  "typescriptExcludeUndefinedFromNullableUnion": true
}
```

Just be aware that we will remove this feature flag in a future release and that you might run into an unexpected `undefined`, if you're not checking for `undefined` before accessing a field - especially if it's a client-only field.


captbaritone:
Thanks!! 

My guess is that this is going to be potentially difficult for folks to upgrade to. To help support their migration, would you add a [compiler feature flag](https://github.com/facebook/relay/blob/51fee1513e2b0619752aea57484d11a5f77eddc2/compiler/crates/common/src/feature_flags.rs#L22) that gates this behavior, with the correct behavior being the default? We can plan to have at least one release where this change is opt-in before we remove the flag.

tobias-tengler:
@captbaritone should it be a feature flag or can the switch also be on the [TypegenConfig](https://github.com/facebook/relay/blob/51fee1513e2b0619752aea57484d11a5f77eddc2/compiler/crates/relay-config/src/typegen_config.rs#L69)? This is already being passed to the TypescriptPrinter, so the change would be less disruptive to make.

captbaritone:
Yeah, that's fine. Could you also do a writeup about this change that we can include in the next release notes? Ideally something that covers:

1. Values can be `undefined` in the case of missing data.
2. This has always been true, the types were wrong
3. We recommend performing null checks using `== null` to cover both null and undefined
4. An example of setting the feature flag/typegen config in a config file to temporarily opt out
5. This feature flag/option is temporary and will be dropped in a future version. Its intended for migration purposes only.

tobias-tengler:
Where should I do the writeup?

captbaritone:
> Where should I do the writeup?

Wish I had a better answer to this. For now, maybe just a comment here on the PR or in a linked Google Doc (or similar)

tobias-tengler:
@captbaritone Sorry for taking some time to get back to this - I was on a small vacation.
I've added the release notes to the PR description.

While looking over the code again, I noticed some other type generation issues regarding client schema extensions:
1. Non-Null client schema extensions are typed as `type`, while they should be `type | undefined` in both Flow and TypeScript
2. Client schema extensions are included as optional on the `rawResponse`. Why are the client-only fields even included in the `rawResponse`?

I tried to tackle 1, but I lack the understanding of what `conditional` is in the code-base and am afraid to change something there. Could you maybe take a look at this?

ernieturner:
@tobias-tengler @captbaritone I was trying to upgrade a project of mine to Relay 15 and came across this addition. I'm not quite ready to tackle this change and update all my code, so I attempted to use the `typescript_exclude_undefined_from_nullable_union` flag to disable this, but I don't think it's actually possible to enable this flag for single project configurations. Or am I missing something? 

In the [config.rs](https://github.com/facebook/relay/blob/main/compiler/crates/relay-compiler/src/config.rs) file the [`ConfigFileProject`](https://github.com/facebook/relay/blob/main/compiler/crates/relay-compiler/src/config.rs#L949) (which is used in the Relay config when configuring multiple projects in a single config) flattens out the `typegen_config: TypegenConfig` which is where this new config was added, but that doesn't appear to exist on the [`SingleProjectConfigFile`](https://github.com/facebook/relay/blob/main/compiler/crates/relay-compiler/src/config.rs#L662) which is what I'm using (I think). 

Is this just an oversight and theres something I'm missing or do we need to update the `SingleProjectConfigFile` to let people set all of the various typegen flags?

tobias-tengler:
@ernieturner Thanks for bringing this up! This was indeed a small oversight on my part 😅 Should be fixed with #4482

tobias-tengler:
@gajus The reason I created this PR was this comment by one of the maintainers: https://github.com/facebook/relay/pull/4370#discussion_r1265754991

You don't have to duplicate all your if conditionals, just use `==` instead of `===`, since `undefined` and `null` share the same semantic value, even though their types are different.

I think an example of where this change could help you catch bugs, is someone deleting a record from the store. This could result in your linked fields or arrays suddently containing an `undefined` value.

In theory this could also happen for non-nullable types, so in hindsight this should've maybe been solved in the store except for the type generation.

Maybe @captbaritone could provide some additional insights.

gajus:
> You don't have to duplicate all your if conditionals, just use == instead of ===, since undefined and null share the same semantic value, even though their types are different

That's inconvenient because we enforce strict checks across our codebase. So it is going to be either:

```diff
  );

  if (profile === null) return null;
+  if (profile === undefined) return null;

  return (
    <>
      <Helmet>
        <title>{`Opportunities offered by ${profile.firstName} ${profile.lastName}`}</title>
      </Helmet>
```

or

```diff
  );
-  if (profile === null) return null;
+  // eslint-disable-next-line no-eq-null, eqeqeq
+  if (profile == null) return null;

  return (
    <>
      <Helmet>
        <title>{`Opportunities offered by ${profile.firstName} ${profile.lastName}`}</title>
      </Helmet>
```

Neither of which is a nice change.

gajus:
@captbaritone any chance of reconsidering this?

gajus:
@kassens @alunyov @SamChou19815 is it possible to get some follow up here?

alunyov:
So, I think the rationale for this change is well described by @tobias-tengler - in Relay, null and undefined are different things. If the type system can help reflect the actual runtime behavior, it is a good change from the correctness perspective.

However, from a practical point of view, I agree that this change may break a lot of existing codebases, where, for various reasons, the handling of null/undefined values may be different. We also had to disable this flag for one of our internal projects that was using TypeScript. It is likely we won’t be updating that project to a different null handling in the near future, so the feature flag will probably stay.

@gajus, am I understanding correctly that your proposal is to revert this change (type updates) and handle this at runtime by always returning null values and never undefined?

gajus:
That's correct. I'd propose to revert this. There is extremely little value for this to be exposed to the client. You could make it exposed behind a feature flag (the inverse of the current behavior), but even then it is very hard to see how this would benefit someone even if they are starting the project from the ground up.

captbaritone:
I think there may be some confusion. The change here makes the type script generated types match the long-standing behavior of the runtime (and the generated Flow types). undefined is (and has been) a possible value in these locations forever. The types are just being updated to reflect this truth. This change does not introduce new undefined values, but just corrects the types to correctly represent the fact that these values can, in fact, be undefined.

gajus:
I see. In which case, I would want to have an option to just return `null` for all the cases that can return `null | undefined`. We were quite happy all this time assuming that these are the only possible values.

We are currently working on a wrapper to do this for us, but I think that keeping the current behavior as the default will severely cripple new user adoption of Relay.

ernieturner:
Doesn't the `typescriptExcludeUndefinedFromNullableUnion: true` flag in the Relay config give you that option to keep the return type as only nullable and not `null | undefined`? You shouldn't need to create your own wrapper for this as that flag should just revert the behavior back to Relay < 16. When we upgraded to Relay 16 we just enabled that flag and no other changes were necessary.

gajus:
The release notes explicitly state that the feature flag is going to be removed in the future, so we need to proactively plan for it.

And also, the feature flag does not cover all the cases, e.g.

Screenshot 2023-11-28 at 10 18 21 AM

Even though `"typescriptExcludeUndefinedFromNullableUnion": true`, the above code still includes `undefined` as a possible return value.

ernieturner:
Given this discussion it seems like it's reasonable to reverse course on the plan to remove the feature flag and just keep it around forever. I can't speak to the feasibility of that, but it seems like something that people should be allowed to opt into forever if they want. 

Regarding the use case where you're still getting `undefined` in the union, that seems odd as we haven't had that problem come up once we enabled that feature flag. Maybe there's something different with how that fragment is defined? Not sure what might be happening there.

captbaritone:
> And also, the feature flag does not cover all the cases, e.g.

Interesting. Did version 16 return only null there? If so, can you open a separate issue?

captbaritone:
I'm open to leaving the feature flag indefinitely. We are starting to look at alternate ways to handle missing data more explicitly, perhaps piggybacking on our [explicit error handling efforts](https://github.com/facebook/relay/issues/4416). So, there's a possibility that we'll have a fundamental solution which can allow us to avoid returning `undefined` in this positions (since they will turn into errors) at which point we'll be able to update the types to be `T | null` while still being accurate.

captbaritone:
I'm not particularly concerned about the current default interfering with new adoptions of Relay. Handling both null and undefined as a single concept with `== null` should be acceptable for most users.

captbaritone:
I've updated the release notes to clarify our thinking here: https://github.com/facebook/relay/releases/tag/v16.0.0

alex-statsig:
> > And also, the feature flag does not cover all the cases, e.g.
> 
> Interesting. Did version 16 return only null there? If so, can you open a separate issue?

I'm not sure if this issue got opened, but I realized during some upgrades that `@types/react-relay` types `useFragment` to return `KeyTypeData<TKey> | null | undefined` in recent versions

PR #4381, opened 2023-07-18
[relay-compiler] Fix LocalPersister repersist

tobias-tengler:
Currently the `--repersist` with the LocalPersister only adds new operations without cleaning up old operations. My expectation of a repersist would be to persist the currently active operations and getting rid of the outdated operations.

## Changes

- Do not panic if LocalPersister file doesn't exist yet and just print a warning
- Do not initialize LocalPersister when repersisting as to remove outdated operations

### DIFF
--- a/compiler/crates/relay-bin/src/main.rs
+++ b/compiler/crates/relay-bin/src/main.rs
@@ -269,16 +269,17 @@ async fn handle_compiler_command(command: CompileCommand) -> Result<(), Error> {
         config.artifact_writer = Box::new(ArtifactValidationWriter::default());
     }
 
-    config.create_operation_persister = Some(Box::new(|project_config| {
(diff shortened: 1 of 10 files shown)


captbaritone:
Looks like this is in a draft state, so I'll wait to look at this in detail. Feel free to ping me when you want eyes on this.

captbaritone:
I think that cleaning up old operations and then deploying would potentially break all users who currently have your site open in a browser and then attempt to navigate since the query id they have might have been cleaned up. For production queries I think persisted queries need to be ~append only with some eventual cleanup depending upon how old of a client you expect to have. For native apps (React Native in Relay's case) that might be months or even years.

Maybe we need a separate flag for this cleanup action? Or maybe you have a different expectation of `--repersist` than I do?

PR #4404, opened 2023-08-08
[relay-compiler] Correctly type empty variables in TypeScript

tobias-tengler:
Previously the compiler produced `{}` for empty objects in TypeScript. This is for example the case for the `...$variables` type, if no variables are present.

The problem is that `{}` does not mean "empty object" in TypeScript. It means "any value except null and undefined".
Because of this you're able to input **anything** besides those two values as variables for an operation without variables.

I opted to produce the more appropriate `Record<PropertyKey, never>`, which effectively represents an object without properties.


captbaritone:
Thanks!

PR #4405, opened 2023-08-09
Fix typo in testing-relay-with-preloaded-queries.md

Uk1288:
Update `testing-relay-with-preloaded-queries.md`


captbaritone:
Thanks for working on this fix. Could you add this to the non-versioned version of this document as well? https://github.com/facebook/relay/blob/main/website/docs/guides/required-directive.md

If you want, you could also back-port these fixes to previous versions as well.

PR #4421, opened 2023-08-27
[relay-lsp] Automatic renaming

tobias-tengler:
This PR adds support for renaming fragments, operations and Relay Resolver fields via the LSP.

relay-lsp-file-rename

relay-lsp-resolver-rename

## Changes

- Support renaming operations via OperationDefinitionName
- Support renaming fragments via FragmentDefinitionName and FragmentSpreadName
- Rename all fragment references on fragment rename
- Support renaming Relay resolver fields via FieldName (disabled for now, due to another bug)
- Rename all Relay Resolver field references on Relay Resolver field rename (disabled for now, due to another bug)
- Rename fragments and operations within renamed files

### DIFF
--- a/compiler/crates/relay-lsp/src/docblock_resolution_info.rs
+++ b/compiler/crates/relay-lsp/src/docblock_resolution_info.rs
@@ -8,13 +8,14 @@
 use common::Span;
 use graphql_ir::reexport::StringKey;
 use graphql_ir::FragmentDefinitionName;
+use graphql_syntax::Identifier;
 use relay_docblock::DocblockIr;
 use relay_docblock::On;
 
 pub enum DocblockResolutionInfo {
     Type(StringKey),
     RootFragment(FragmentDefinitionName),
-    FieldName(StringKey),
+    FieldName(Identifier),
     Deprecated,
 }
 
@@ -44,9 +45,7 @@ pub fn create_docblock_resolution_info(
             }
 
             if resolver_ir.field.name.span.contains(position_span) {
-                return Some(DocblockResolutionInfo::FieldName(
-                    resolver_ir.field.name.value,
-                ));
+                return Some(DocblockResolutionInfo::FieldName(resolver_ir.field.name));
             }
 
             if let Some(output_type) = &resolver_ir.output_type {
@@ -87,6 +86,10 @@ pub fn create_docblock_resolution_info(
                 if root_fragment.location.contains(position_span) {
                     return Some(DocblockResolutionInfo::RootFragment(root_fragment.item));
                 }
+            }
+
+            if resolver_ir.field.name.span.contains(position_span) {
+                return Some(DocblockResolutionInfo::FieldName(resolver_ir.field.name));
             }
 
             // @deprecated key
--- a/compiler/crates/relay-lsp/src/references/mod.rs
+++ b/compiler/crates/relay-lsp/src/references/mod.rs
@@ -80,10 +80,7 @@ fn get_references_response(
                         On::Type(type_) => type_.value.item,
                         On::Interface(interface) => interface.value.item,
                     },
-                    DocblockIr::TerseRelayResolver(_) => {
-                        // TODO: Implement support for terse relay resolvers.
-                        return Err(LSPRuntimeError::ExpectedError);
-                    }
+                    DocblockIr::TerseRelayResolver(terse_resolver) => terse_resolver.type_.item,
                     DocblockIr::StrongObjectResolver(_) => {
                         // TODO: Implement support for strong object.
                         return Err(LSPRuntimeError::ExpectedError);
@@ -94,7 +91,7 @@ fn get_references_response(
                     }
                 };
 
-                let references = find_field_locations(program, field_name, type_name)
+                let references = find_field_locations(program, field_name.value, type_name)
                     .ok_or(LSPRuntimeError::ExpectedError)?
                     .into_iter()
                     .map(|location| transform_relay_location_to_lsp_location(root_dir, location))
--- a/compiler/crates/relay-lsp/src/rename/mod.rs
+++ b/compiler/crates/relay-lsp/src/rename/mod.rs
@@ -1,0 +1,702 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+//! Utilities for providing the rename feature
+
+use std::collections::HashMap;
+use std::path::PathBuf;
+
+use common::Location as IRLocation;
+use common::NamedItem;
+use common::Span;
+use graphql_ir::FragmentDefinition;
+use graphql_ir::FragmentSpread;
+use graphql_ir::OperationDefinitionName;
+use graphql_ir::Program;
+use graphql_ir::Visitor;
+use graphql_syntax::ExecutableDefinition;
+use intern::string_key::Intern;
+use intern::string_key::StringKey;
+use lsp_types::request::PrepareRenameRequest;
+use lsp_types::request::Rename;
+use lsp_types::request::Request;
+use lsp_types::PrepareRenameResponse;
+use lsp_types::TextEdit;
+use lsp_types::Url;
+use lsp_types::WorkspaceEdit;
+use rayon::prelude::IntoParallelRefIterator;
+use rayon::prelude::ParallelIterator;
+use relay_docblock::DocblockIr;
+use relay_docblock::On;
+use resolution_path::ArgumentParent;
+use resolution_path::ArgumentPath;
+use resolution_path::DirectiveParent;
+use resolution_path::DirectivePath;
+use resolution_path::IdentParent;
+use resolution_path::IdentPath;
+use resolution_path::ResolutionPath;
+use resolution_path::ResolvePosition;
+use resolution_path::VariableIdentifierParent;
+use resolution_path::VariableIdentifierPath;
+
+use crate::docblock_resolution_info::create_docblock_resolution_info;
+use crate::docblock_resolution_info::DocblockResolutionInfo;
+use crate::find_field_usages::find_field_locations;
+use crate::location::transform_relay_location_to_lsp_location;
+use crate::Feature;
+use crate::GlobalState;
+use crate::LSPRuntimeError;
+use crate::LSPRuntimeResult;
+use lazy_static::lazy_static;
+
+lazy_static! {
+    static ref ARGUMENTS_DIRECTIVE: StringKey = "arguments".intern();
+    static ref ARGUMENTDEFINITIONS_DIRECTIVE: StringKey = "argumentDefinitions".intern();
+}
+
+/// Resolve a [`Rename`] request to workspace edits
+pub fn on_rename(
+    state: &impl GlobalState,
+    params: <Rename as Request>::Params,
+) -> LSPRuntimeResult<<Rename as Request>::Result> {
+    let uri = &params.text_document_position.text_document.uri;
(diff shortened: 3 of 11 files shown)


tobias-tengler:
Unrelated, but I find the Rust setup very weird. Maybe it's because I'm a Rust beginner, but it seems like this project should have a `rust-toolchain.toml` file specifying a singular toolchain to use. At the moment there are two different toolchains being used for building and formatting the project which is kind of cumbersome for local development... Without looking at the Github Action I also wouldn't have been able to find the correct versions to install.

captbaritone:
This is so exciting! Thank you for working on this. I see the PR is in draft state, but please let me know when you'd like eyes on it.

For resolvers, I can see the following rename refactors being useful:

1. Rename field, which renames the resolver and all instances of that field on all documents
2. Rename a type, same but for a type (resolvers which define types are not yet documented in open source)

captbaritone:
Regarding the toolchain issue, its likely that we have some disfunction here since we have a different system to manage this stuff internally. Do you want to try opening a PR that adds the `rust-toolchain.toml` file and aligns us on a single version? For my context, where are the two places where we define different versions for formatting and compiling?

tobias-tengler:
The two versions are here: https://github.com/facebook/relay/blob/main/.github/workflows/ci.yml

I think this is the case, since the rustfmt.toml file is referencing formatting options that are only available in the nightly version!?

compiler/crates/relay-lsp/src/find_field_usages/find_field_locations.rs
```diff
@@ -19,6 +19,7 @@ pub(crate) fn find_field_locations(
     type_name: StringKey,
 ) -> Option<Vec<IRLocation>> {
     let type_ = program.schema.get_type(type_name)?;
+    // todo: this doesn't find usages of renamed RelayResolver fields
```
tobias-tengler, line 22: The renaming of a Relay Resolver field only works once. After the rename there aren't any more references returned from this FieldUsageFinder.

I believe this is because the rename of the field happens without user interaction, so there are some caches and so that aren't updated as they usually would. We could probably trigger the ingestion of changes from within the rename code, but I don't know how at the moment.
Maybe one of you can quickly help out?

compiler/crates/relay-lsp/src/rename/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+            };
+
+            index += 1;
+        }
+    }
+
+    Ok(Some(WorkspaceEdit {
+        changes: Some(rename_changes),
+        ..Default::default()
+    }))
```
tobias-tengler, line 281: I'm really unhappy with this code, but I'm just hitting the ceiling of my Rust knowledge. I tried to make it more functional by using iterators, but I struggled with `collect`, ignore errors while iterating and in general how to avoid the many `.unwrap()` calls.
I would appreciate input from some Rust veterans 🦀 Also feel free to just make the changes yourself in the code!

tobias-tengler (commented):
@captbaritone @alunyov this PR would be ready for review, with some open problems you might be able to help me with :)

As proposed by @captbaritone I added support for renaming Relay Resolver fields (I've left out the types for now, since I know too little about them).

compiler/crates/relay-lsp/src/rename/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+                        root_dir,
+                    );
+
+                    let location = common::Location::new(source_location_key, docblock_field.span);
+                    let lsp_location =
+                        transform_relay_location_to_lsp_location(root_dir, location)?;
+
+                    merge_text_edit(&mut changes, lsp_location, &params.new_name);
+
+                    // todo: rename JS function
```
tobias-tengler, line 126: I think it would be really cool, if we could also rename the JS function of the Relay Resolver field, but I think the location of the function name isn't yet tracked by the LSP. I could probably hack together some lookahead after the DocBlock to find the function definition, but I don't think it would be very stable 🤪

compiler/crates/relay-lsp/src/rename/mod.rs
martindisch: I've already made a couple of small improvements locally today, but will have to rebase since you changed a fair amount now. I'll walk you through it tomorrow.

orta:
Such a cool feature!

tobias-tengler:
@captbaritone @alunyov The problem of only being able to rename Relay Resolvers once seems to have a larger underlying issue that also affects finding references of Relay Resolvers.

After some investigation, the problem occurs since the `FieldUsageFinder` is fed the schema of the "program". This schema seems to be stale or at least the `fields` collection inside of it isn't updated as it should, when the Relay Resolver field name changes. This causes lookups of the `FieldID` on that schema to always resolve the old field name and thus were not able to find references after a rename of the field.
I resolved this by getting the schema from the GlobalState as you can see in b46799bc661fcbd13a554530ac42a2fedcf1fe37. This seems to get the fresh schema that is built in response to file changes. 
With this the lookup of references works. However if you're changing the field name at the usage site to something that doesn't exist, it is now still found as a reference. I suspect another issue, where field usages aren't cleaned up properly.
This problem seems to go a bit deeper than I'd like, so I would appreciate if one of you could maybe take a look. I'm sure you're able to diagnose what could be the cause faster than me...

compiler/crates/relay-lsp/src/rename/mod.rs
captbaritone: I think we'll have to leave that out for now. Eventually I would love Relay to understand the surrounding JavaScript (and types!) and to be able to infer a lot of what is currently duplicated between the docblock and the actual code (similar to what I'm exploring with [Grats](https://grats.capt.dev/)). However, that is a larger project.

compiler/crates/relay-lsp/src/rename/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub struct FragmentFinder {
+    fragment_locations: Vec<IRLocation>,
+    fragment_name: StringKey,
+}
+
+impl FragmentFinder {
+    pub fn get_fragment_usages(program: &Program, name: StringKey) -> Vec<IRLocation> {
```
captbaritone, line 343: Would it be possible to reuse https://github.com/facebook/relay/blob/main/compiler/crates/relay-lsp/src/references/mod.rs#L119?

compiler/crates/relay-lsp/src/references/mod.rs
```diff
@@ -80,10 +83,7 @@ fn get_references_response(
                         On::Type(type_) => type_.value.item,
                         On::Interface(interface) => interface.value.item,
                     },
-                    DocblockIr::TerseRelayResolver(_) => {
-                        // TODO: Implement support for terse relay resolvers.
-                        return Err(LSPRuntimeError::ExpectedError);
-                    }
+                    DocblockIr::TerseRelayResolver(terse_resolver) => terse_resolver.type_.item,
```
captbaritone, line 86: ❤️

captbaritone, line 86: ❤️

compiler/crates/relay-lsp/src/docblock_resolution_info.rs
```diff
@@ -89,6 +88,10 @@ pub fn create_docblock_resolution_info(
                 }
             }
 
+            if resolver_ir.field.name.span.contains(position_span) {
+                return Some(DocblockResolutionInfo::FieldName(resolver_ir.field.name));
```
captbaritone, line 92: ❤️

compiler/crates/relay-lsp/src/find_field_usages/find_field_locations.rs
```diff
@@ -5,21 +5,25 @@
...
 use graphql_ir::Visitor;
 use intern::string_key::StringKey;
+use schema::SDLSchema;
 use schema::Schema;
 
 use crate::find_field_usages::FieldUsageFinder;
 
 pub(crate) fn find_field_locations(
     program: &Program,
+    schema: &Arc<SDLSchema>,
```
captbaritone, line 21: Given your comment, it sounds like the presence of the schema here is to work around a larger bug/issue of the program.schema being stale. Could we add a comment here clarifying that? That way if/when we resolve that issue we can switch back to the simpler approach of using `program.schema`.

captbaritone:
This is so great! I love all the helpers we're adding here, as well as the little extra  "go to definition" fixes you added along the way. I'm also excited about other possible renames we could add from here! Aside: I'd love to add LSP support for `.grapql` schema files as well! That would open a lot of options!

Finally, I'm appreciative of the identification of the stale schema issue. I suspect that was subtly breaking other click-to-definition features relating to resolvers. Would you mind filing a GitHub issue for us to investigate? You could include a link to that issue in code comments in the places where you are having to work-around it.

One larger thing to include in your test plan is how this feature works with multi-project configs where one project uses another project as its "base" project. This setup allows one project to extend another project. For example, you might have an internal admin tool that extends your public facing app. The internal tool can reference fragments/schema used in the public facing app, but not vice versa.

Would be good to validate that this works correctly when renames are performed in a parent project and in a child project.

I left a few comments, but they are mostly superficial. Looking forward to getting this merged!

captbaritone (changes requested):

compiler/crates/relay-lsp/src/find_field_usages/find_field_locations.rs
tobias-tengler: I have actually reverted the change and disabled Relay Resolver field renaming for now - it was still buggy. I've raised an issue to track it: https://github.com/facebook/relay/issues/4447

compiler/crates/relay-lsp/src/rename/mod.rs
```diff
@@ -0,0 +1,413 @@
...
+    field_name: StringKey,
+    type_name: StringKey,
+    new_field_name: &str,
+    program: &Program,
+    root_dir: &PathBuf,
+) -> LSPRuntimeResult<HashMap<Url, Vec<TextEdit>>> {
+    let mut changes = HashMap::<Url, Vec<TextEdit>>::new();
+
+    for location in find_field_locations(program, field_name, type_name).unwrap_or_else(|| vec![]) {
+        let lsp_location = transform_relay_location_to_lsp_location(root_dir, location)?;
```
captbaritone, line 300: Internally, we may have a huge number of items here and below in `rename_fragment`. I wonder if it would improve performance to tackle these in parallel (par_iter).

tobias-tengler:
@captbaritone I addressed all your feedback and also tried to test with this multi-project base setup you mentioned. I managed to setup a multi-project config (which had way too many gotchas - will create some issues later), but things started to fall apart as soon as I added the base option and actually tried to reference fragments from the base project.
The compiler would just panic with the following message
> thread 'tokio-runtime-worker' panicked at 'Expected location for fragment BaseProjectFragment.', crates/relay-typegen/src/write.rs:495:29

If you can supply me with a working multi-project base setup I would be happy to test it again. 
Is there anything you expect to not work? If I understand the base projects correctly, there shouldn't be anything that could go wrong. Maybe the fragment renames don't propagate across project boundaries, but this should also affect other features then, since I'm discovering them via the Visitor.

captbaritone:
Hey! I haven't forgotten about this, just been very busy. Hope to get to this next week. In the mean time, do you have a repro for that panic? Would you mind filing an issue?

tobias-tengler:
@captbaritone I filed an issue with a reproduction in https://github.com/facebook/relay/issues/4471.

tobias-tengler:
@captbaritone any updates?
If you feel uneasy about merging this, I could also move the automatic renaming part behind a feature flag :)

tobias-tengler:
@captbaritone or @alunyov could you review this please? This would be such an awesome addition!

PR #4428, opened 2023-08-30
[relay] Show a helpful error if a resolver returns an interface with no implementors

monicatang:
**Context:**

Without this we generate a broken Flow type.

We will eventually support Relay Resolvers that return interfaces. Today they are not permitted in the common case. However, once we do allow them, we must ensure that the schema includes at least once concrete type that implements that interface. Without that property, we will generate invalid Flow types.

**This change**

Adds additional validation for implementors of client-defined interfaces if they are being used as a return type for RelayResolvers and tests. Also updates an existing test to have it only test the existing validation for blocking interfaces as a return type.

**Test plan**
`./scripts/update-fixtures.sh`
`cargo test`

specifically:
`cargo test -p relay-transforms --test client_edges_test`
`cargo test -p relay-compiler --test relay_compiler_compile_relay_artifacts_test relay_resolver_edge_to_interface_with_no_implementors`
`cargo test -p relay-compiler --test relay_compiler_compile_relay_artifacts_test relay_resolver_edge_to_interface_with_child_interface_and_no_implementors`


compiler/crates/relay-transforms/tests/client_edges/fixtures/client-edge-to-client-interface.invalid.graphql
```diff
@@ -15,6 +15,10 @@ interface ClientOnlyInterface implements Node {
   id: ID!
 }
 
+type BestFriend implements ClientOnlyInterface {
```
captbaritone, line 18:
```suggestion
# Add a concrete type so that we don't trigger an unrelated compiler error.
type BestFriend implements ClientOnlyInterface {
```

compiler/crates/relay-compiler/tests/compile_relay_artifacts/fixtures/relay-resolver-edge-to-interface-with-no-implementors.expected
```diff
@@ -0,0 +1,38 @@
...
+      │   ^^^^^^^^^^^^^^
+    4 │     name
+
+
+✖︎ No types implement the client interface SomeInterface. For a client interface to be used as a @RelayResolver @outputType, at least one Object type must implement the interface.
+
+  <generated>:2:35
+    1 │ # expected-to-throw
+    2 │ query relayResolverEdgeToInterfaceWithNoImplementorsQuery {
+      │                                   ^^^^^^^^^^^^^
```
captbaritone, line 37: This underline being in the wrong place is not a problem with your diff. Our test util is not fully roust in this case and does not print location info correctly.

compiler/crates/relay-compiler/tests/compile_relay_artifacts/fixtures/relay-resolver-edge-to-interface-with-no-implementors.expected
```diff
@@ -0,0 +1,38 @@
...
+✖︎ Client Edges that reference client-defined interface types are not currently supported in Relay.
+
+  relay-resolver-edge-to-interface-with-no-implementors.graphql:3:3
+    2 │ query relayResolverEdgeToInterfaceWithNoImplementorsQuery {
+    3 │   resolver_field {
+      │   ^^^^^^^^^^^^^^
+    4 │     name
+
+
+✖︎ No types implement the client interface SomeInterface. For a client interface to be used as a @RelayResolver @outputType, at least one Object type must implement the interface.
```
captbaritone, line 32: Can we update the error message to be more generic?
```suggestion
✖︎ No types implement the client interface SomeInterface. Interfaces returned by a @RelayResolver @outputType must have at least one concrete implementation.
```

compiler/crates/relay-transforms/src/client_edges.rs
```diff
@@ -363,7 +363,18 @@ impl<'program, 'sc> ClientEdgesTransform<'program, 'sc> {
             }
 
             match edge_to_type {
-                Type::Interface(_) => {
+                Type::Interface(interface_id) => {
+                    let interface = schema.interface(interface_id);
+                    let implementing_objects =
+                        interface.recursively_implementing_objects(Arc::as_ref(schema));
```
captbaritone, line 369: Can we add a test case which covers the case where the interface is implemented by another interface, but there are still no concrete implementations? Should still error.

captbaritone:
Awesome! Left a few comments, but this looks spot on so far.

captbaritone (changes requested):

compiler/crates/relay-transforms/src/client_edges.rs
monicatang: added a test case called `relay-resolver-edge-to-interface-with-child-interface-and-no-implementors`!

compiler/crates/relay-compiler/tests/compile_relay_artifacts/fixtures/relay-resolver-edge-to-interface-with-no-implementors.expected
monicatang: Right, it should be underlining `resolver_field` here. Just wondering, where does that get determined?

compiler/crates/relay-transforms/src/errors.rs
```diff
@@ -168,7 +168,7 @@ pub enum ValidationMessage {
     },
 
     #[error(
-        "No types implement the client interface {interface_name}. For a client interface to be used as a @RelayResolver @outputType, at least one Object type must implement the interface."
+        "No types implement the client interface {interface_name}. Interfaces returned by a @RelayResolver @outputType must have at least one concrete implementation."
```
captbaritone, line 171: This validation now fires even for non-@outputType resolvers
```suggestion
        "No types implement the client interface {interface_name}. Interfaces returned by a @RelayResolver must have at least one concrete implementation."
```

compiler/crates/relay-compiler/tests/compile_relay_artifacts/fixtures/relay-resolver-edge-to-interface-with-no-implementors.expected
captbaritone: Here: https://github.com/facebook/relay/blob/74bb8abd92d13460fb90be290c65719cba74c62e/compiler/crates/relay-compiler/tests/compile_relay_artifacts/mod.rs#L206

The problem is that the fixture file actually models two different files, and so the error location should actually get mapped to an offset based on if its an error in the schema portion of the file or the fragment portion.

PR #4433, opened 2023-09-03
Use same toolchain for build and lint

martindisch:
While building this codebase, I noticed several stumbling blocks that might confuse and discourage first-time contributors coming from other Rust projects, which I would like to help address in the near future.

The first one I'll bring up is that we're using different toolchains for formatting (nightly) and building (stable). I assume nightly is used for formatting/linting because some nice rules such as `group_imports` are unfortunately still unstable. The current setup means that:

- if people use a stable toolchain like they're used to from most projects, their IDE might apply formatting that doesn't respect these nightly-only rules, which they only notice once CI fails and might have trouble figuring out
- if they use nightly they won't have this problem, but will have to deal with the usual inconveniences of working with nightly builds, such as compiler panics and other unstable strangeness

As much as I like `group_imports = "StdExternalCrate"`, I'm not sure if it's worth the toolchain hassle. So I opened this PR to show what it would look like with stable-only and get your opinion. It doesn't actually change much in the existing files because the default behavior for the rules in question is mostly preserving existing formatting, the biggest downside I see is that we're no longer able to enforce the order of the imports for new files in CI. If consistency is a priority here, that would have to be maintained through the regular code review process. But in exchange we'd get a more streamlined development setup.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -136,7 +136,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly-2022-06-26
+          toolchain: 1.72.0
           override: true
           components: rustfmt
       - name: "rustfmt"
--- a/compiler/crates/common/src/diagnostic_check.rs
+++ b/compiler/crates/common/src/diagnostic_check.rs
@@ -124,11 +124,9 @@ mod escalate_tests {
 
         escalate(DiagnosticSeverity::WARNING, &mut diagnostics);
 
-        assert!(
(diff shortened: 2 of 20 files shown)


captbaritone:
The challenge here is that internally this code is managed by a global Rust setup used across a larger mono-repo. Relay needs to conform to those norms and share those versions. Those config files/scripts live outside Relay's code base, so we can't include them here. Let me see if I can find this source of truth internally and find out:

1. Which version we should be using
2. If we can find some way to ensure they stay in sync

captbaritone:
@alunyov I think you found another fix for this right?

PR #4434, opened 2023-09-04
Enable goto def in LSP to goto the actual schema definition

bigfootjon:
I basically just took @captbaritone 's POC and made it configurable based on our conversation [on Threads](https://www.threads.net/@janjonzen/post/CwwfXJ7LlL_)

This should be reasonably generalized to support different use-cases, but I'm open to rewriting all of this.

Test plan:

```
cd compiler
cargo build --release --bin relay
cd ../vscode-extension
yarn build-local
```

In a private project, I configured the new option and overrode the relay binary location:

```diff
+ "relay.pathToExtraDataProviderScript": "../scripts/graphql_lookup.sh",
+ "relay.pathToRelay": ".../relay/compiler/target/release/relay"
```

I manually installed the `relay-2.1.0.vsix` file in my VSCode and reloaded

I opened a file with a Relay GraphQL query in it and ctrl-clicked (goto def on my machine) and it went to the correct file!

### DIFF
--- a/compiler/crates/relay-bin/src/main.rs
+++ b/compiler/crates/relay-bin/src/main.rs
@@ -111,6 +111,11 @@ struct LspCommand {
     /// Verbosity level
     #[clap(long, arg_enum, default_value = "quiet-with-errors")]
     output: OutputKind,
+
+    /// Script to be called to lookup the actual definition of a GraphQL entity for
+    /// implementation-first GraphQL schemas.
+    #[clap(long)]
+    extra_data_provider_script: Option<String>,
 }
 
 #[derive(clap::Subcommand)]
@@ -309,11 +314,15 @@ async fn handle_compiler_command(command: CompileCommand) -> Result<(), Error> {
     Ok(())
 }
 
-struct ExtraDataProvider {}
+struct ExtraDataProvider {
+    script: String,
+}
 
 impl ExtraDataProvider {
-    pub fn new() -> ExtraDataProvider {
-        ExtraDataProvider {}
+    pub fn new(script: String) -> ExtraDataProvider {
+        ExtraDataProvider {
+            script
+        }
     }
 }
 
@@ -324,7 +333,7 @@ impl LSPExtraDataProvider for ExtraDataProvider {
 
     fn resolve_field_definition(
         &self,
-        _project_name: String,
+        project_name: String,
         parent_type: String,
         field_info: Option<FieldSchemaInfo>,
     ) -> Result<Option<FieldDefinitionSourceInfo>, String> {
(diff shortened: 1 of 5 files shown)


vscode-extension/README.md
```diff
@@ -70,6 +70,13 @@ Path to a Relay config relative to the `rootDirectory`. Without this, the compil
 
 An array of project configuration in the form `{name: string, rootDirectory: string, pathToConfig: string}`. If omitted, it is assumed your workspace uses a single Relay config and the compiler will search for your config file. But you can also use this configuration if your Relay config is in a nested directory. This configuration must be used if your workspace has multiple Relay projects, each with their own config file.
 
+#### `relay.pathToExtraDataProviderScript` (default: `null`)
+
+Path to a script to look up the actual definition for a GraphQL entity for implementation-first GraphQL schemas. This script will be called for "goto definition" requests to the LSP instead of opening the schema.
+The script will be called with 2 arguments. The first will be the relay project name, the second will be either "Type" or "Type.field" (a type or the field of a type, repectively).
```
bigfootjon, line 76: Maybe we should future-proof this by calling with some command like "lookup default Foo.bar" instead of hardcoding this as a lookup action? I say this because I see there's a "stats" thing in the ExtraDataProvider trait which might warrant this sort of future proofing

compiler/crates/relay-bin/src/main.rs
```diff
@@ -306,13 +314,69 @@ async fn handle_compiler_command(command: CompileCommand) -> Result<(), Error> {
...
+            .output()
+            .expect("Failed to run extra data provider script");
+
+        let result = String::from_utf8(result.stdout).expect("Failed to parse output");
+
+        // Parse file_path:line_number:column_number
+        let result = result.trim();
+        let result = result.split(':').collect::<Vec<_>>();
+        if result.len() != 3 {
+            return Ok(None);
```
captbaritone, line 356: This function returns a result. Can we return an Err here?

compiler/crates/relay-bin/src/main.rs
```diff
@@ -306,13 +314,69 @@ async fn handle_compiler_command(command: CompileCommand) -> Result<(), Error> {
...
+    ) -> Result<Option<FieldDefinitionSourceInfo>, String> {
+        let entity_name = match field_info {
+            Some(field_info) => format!("{}.{}", parent_type, field_info.name),
+            None => parent_type,
+        };
+        let result = Command::new(&self.script)
+            .arg(project_name)
+            .arg(entity_name)
+            .output()
+            .expect("Failed to run extra data provider script");
```
captbaritone, line 348: Can we return an Err here instead of crashing the LSP?

vscode-extension/README.md
captbaritone: I'd be fine making this a "locate" command and exposing a separate option for a command to support the query info.

captbaritone: Should this option clarify that you need a new-ish version of relay-compiler for this to work?

captbaritone (changes requested):
Awesome!! Thank you for working on this.

captbaritone:
With v2.1.0 of the VSCode extension now published (thanks @alex-statsig ), I have an example repo showing this feature working with Grats: https://github.com/captbaritone/grats-relay-example/commit/64a5ed4cc928d71807ee5895c348dbbc09f956cc

It's a bit of a bummer that I need a separate script. I wonder if it would be nicer to allow the config option to be a bash command, rathe than just a path to an executable. But in that world, it's a bit unclear how the project and entity name get passed.

Maybe you could pass the LSP an option like: 

```bash
relay lsp --locateCommand="pnpm run grats locate %entity"
```

Or something?

alex-statsig:
Another thought on this - it would be great to have a separate "locateCommand" setting for "Go to implementations"

I used this setting to allow easier navigation to our backend resolvers (which is amazing), but now theres no easy way to get to the schema (currently manually defined, ideally will be code-first eventually).

Perhaps another way to support that would be to allow returning two locations, or more conveniently an option to still include the schema.graphql definition that relay finds rather than replacing it.

PR #4455, opened 2023-09-29
[vscode] Make the Relay Compiler Terminal Transient

ianobermiller:
If you enable `relay.autoStartCompiler`, the extension will open a compiler in a new Terminal on every VS Code reload (e.g. Developer: Reload Window). This can easily lead to many copies of the compiler running.

Instead, we can mark the terminal with `isTransient`, and it won't survive reloads. While this means reloads must restart the compiler, it is very fast and this avoid many "Relay Compiler" terminals.

Test Plan:

- Debug on a project with `"relay.autoStartCompiler": true`
- Reload the window
- Note there is only one "Relay Compiler" terminal

### DIFF
--- a/vscode-extension/src/compiler.ts
+++ b/vscode-extension/src/compiler.ts
@@ -25,6 +25,7 @@ export function createAndStartCompiler(context: RelayExtensionContext) {
   const terminal = window.createTerminal({
     name: 'Relay Compiler',
     cwd: context.relayBinaryExecutionOptions.rootPath,
+    isTransient: true,
   });
 
   terminal.sendText(


Lalitha-Iyer:
cc @tbezman @captbaritone is this something you could help review and merge? Thanks!

captbaritone:
Awesome! Thanks for tracking this down.

tobias-tengler:
Just wanted to add my two cents: The compiler startup time depends on the size of your application. For us it takes 20-30 seconds. I would image it takes even longer in larger code bases. Would be cool if there was a solution that didn't require the compiler to restart.

ianobermiller:
@tobias-tengler The only thing I can think of would be if we could somehow query the relay compiler process, but that seems quite complicated. Considering that a VS Code reload already restarts things like ESLint and TypeScript, this brings nice parity, but that is a good point that startup is not always fast.

captbaritone:
I think this change is better than the current state, and restarting is probably not too common. Let's start here.

PR #4460, opened 2023-10-03
Make all compiler fixture tests async

captbaritone:
This change makes all of our Rust fixture tests async. This is a pre-requisite to supporting integration fixture tests which run the entire compiler (including disk IO) as a fixture test.


alunyov:
This is landed as 650dfe46f0571cb693f35e5ae2c35d1be759810c

PR #4467, opened 2023-10-06
Allow configurable `@defer`/`@stream` names

robrichard:
This is not an attempt to get Relay to conform to the current defer/stream spec draft, but rather just a small step in that direction by allowing configurable directive & argument names. 

Currently Relay requires the argument `initial_count` on `@stream`, but the GraphQL spec uses camelCase for all names.

This PR updates Relay compiler to follow the same pattern used by `ConnectionInterface`, introducing `DeferStreamInterface` to allow configuring the names and arguments of the defer & stream directive.

The first commit adds the configuration updating all touch points to reference the configuration. 

The second commit changes the defaults to camelCase, to match the defaults used in `ConnectionInterface`. This makes the diff substantially larger as many tests and fixtures need to be updated. I'm happy to remove this commit from this PR if desired.

### DIFF
--- a/compiler/crates/common/src/named_item.rs
+++ b/compiler/crates/common/src/named_item.rs
@@ -63,7 +63,18 @@ impl FromStr for DirectiveName {
 }
 
 impl_lookup!(DirectiveName);
-#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd, Hash)]
+#[derive(
+    Clone,
+    Copy,
+    Debug,
+    Deserialize,
+    Eq,
+    PartialEq,
+    Ord,
+    PartialOrd,
+    Hash,
+    Serialize
+)]
 pub struct ArgumentName(pub StringKey);
 
 impl fmt::Display for ArgumentName {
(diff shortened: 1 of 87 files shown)


captbaritone:
Looks like the "CI / Compiler output check" CI step is failing. Can update https://github.com/facebook/relay/blob/main/scripts/config.tests.json to opt back into the old names?

robrichard:
@captbaritone all checks are passing now. It was failing because `relay-extensions.graphql` is statically added to schemas to add the `@stream_connection` directive. 

I updated `connection_constants` to decouple the `@stream_connection` argument names from the `@stream` argument names. So with these updates `@stream_connection` will always use the current snake_case names, but they will be transformed to the `@defer` and `@stream` argument names passed in through the config.

captbaritone:
@robrichard Can you add a followup PR (I'll try to merge this one as is) to make the panics in this file into diagnostics? 

https://github.com/facebook/relay/blob/6fe9d760a63494824e77c9b1a0097f87b95c8e24/compiler/crates/relay-transforms/src/defer_stream/directives.rs#L32-L66

These can now happen as a result of not specifying the the right config for your schema and no user error should result in a panic from the compiler.

PR #4470, opened 2023-10-07
Build compiler playground as part of GitHub CI (take 2)

captbaritone:
We reverted this in https://github.com/facebook/relay/commit/6911daa825956038ba3b12286eda1d5776f31992 because it was failing in CI. Can't repro locally, so trying again on GitHub to see what the error looks like.


captbaritone:
JS Lint failure is unrelated

PR #4473, opened 2023-10-07
[relay-lsp] Ensure documents are synced before calculating completions

### LINKED ISSUE #4472: [relay-lsp] Suggestions from wrong context
Last week, while working on a rather large file, I noticed that the LSP would often not suggest any fields or fields from a different level in the selection set.

The issue is easily reproduced (more so in larger files) if you accept the completion item for a field returning an object type. 

In response to that the LSP will insert a snippet like the following:
```graphql
newObjectField {
    $cursor
}
```
Afterwards it triggers the `editor.action.triggerSuggest` command to reopen the completion menu at the new cursor position.

What's happening is that the `textdocument/didchange` and `textdocument/completion` commands from the client are not necessarily sent in that order and if they are, the LSP might still be processing the document update while completion items are already being calculated at the new cursor position. And of course, if the LSP is computing completion items for the new location, while still working with the old document, weird things can happen like no suggetions because the position is outside the GraphQL document or the position being in a different selection set.

I think there are two things at play here that both need to be addressed in order to correct the behavior:
1. Relay should wait to calculate completion items until all sync operations are completed - this can be easily done and I will send a PR for it in a follow-up.
2. There seems to be a bug in the LSP client (VS Code in this case) where `textdocument/didchange` and `textdocument/completion` are not sent in this order if a completion item triggers `editor.action.triggerSuggest` - I will raise an issue for this in their repository.

tobias-tengler:
Fixes #4472

### DIFF
--- a/compiler/crates/relay-lsp/src/server/mod.rs
+++ b/compiler/crates/relay-lsp/src/server/mod.rs
@@ -152,7 +152,7 @@ where
     );
 
     let task_processor = LSPTaskProcessor;
-    let task_queue = TaskQueue::new(Arc::new(task_processor));
+    let mut task_queue = TaskQueue::new(Arc::new(task_processor));
     let task_scheduler = task_queue.get_scheduler();
 
     config.artifact_writer = Box::new(NoopArtifactWriter);
@@ -190,6 +190,12 @@ fn next_task(
     }
 }
 
+static NOTIFCATIONS_MUTATING_LSP_STATE: [&str; 3] = [
+    "textDocument/didOpen",
+    "textDocument/didChange",
+    "textDocument/didClose",
+];
+
 struct LSPTaskProcessor;
 
 impl<TPerfLogger: PerfLogger + 'static, TSchemaDocumentation: SchemaDocumentation + 'static>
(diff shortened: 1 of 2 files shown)


captbaritone:
Interesting! Looks reasonable, but are you able to validate this change on its own, or do we need the VSCode fix as well?

tobias-tengler:
I would say it catches 40% of the race conditions between `textdocument/didchange` and `textdocument/completion`. It definitely also needs a fix or input on a strategy to use from the VSCode team or LSP clients in general.

tobias-tengler:
It seems like all the issues were being caused by the TaskQueue essentially fire-and-forgetting the processing of tasks. I opted to execute all the notification tasks concerning the document content, e.g. `textdocument/didChange`, serially i.e. blocking so that no other tasks can be processed while the content of a text document is being updated. All features relying on the current state of the text document should now no longer run into race conditions.

I also verified that this fixes the problems in my PR adding folding features: #4480

tobias-tengler:
While this already works really reliably, I'd like to extend the solution with something a little bit more sophisticated to also wait for ongoing `textdocument/*` requests to finish, before processing notifications that alter the document state serially. 

Will need to smarten up on Rust synchronization mechanisms first though 😅

tobias-tengler:
@captbaritone @alunyov this would be ready to review :)

I've also made a quick recording to showcase what exactly this solves (it might also be the source of other subtle LSP bugs):

|Before|After|
|---|---|
| before-completions <br/> Completions are shown from wrong context, because the completions request finishes faster than the documentChange notification that precedes it | after-completions <br/> Completions are shown in correct context, as notifications that modify the LSP state are now executed in isolation (serially) and other requests stay in the queue until the notification has been processed |

captbaritone:
Landed as https://github.com/facebook/relay/commit/2b502883c730aefd9ff37eb6c6074e63731537bc

captbaritone:
Thanks again for this excellent PR and your patience!

PR #4478, opened 2023-10-12
Fixed example code in infinite scroll section of the tutorial

rainandbare:
The tutorial seemed to have originally broken out a `NewsfeedContents` component to house the `NewsfeedContentsFragment` but then it was decided that that was too much work/too confusing for the tutorials purposes. ✅ However some of the old code and references were left in the instructions making that section hard to follow. 

Also added here are the return types on `useLazyLoadQuery`, `useFragment` and `usePaginationFragment` because they are a little hard to find and having the correct types makes the tutorial much smoother and easier to follow.

### DIFF
--- a/website/docs/tutorial/connections-pagination.md
+++ b/website/docs/tutorial/connections-pagination.md
@@ -267,7 +267,7 @@ function StoryCommentsSection({story}) {
         />
       )}
       // change-line
-      {isPending && <CommentsLoadingSpinner />}
+      {isPending && <Small Spinner />}
     </>
   );
 }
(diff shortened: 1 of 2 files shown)


website/docs/tutorial/connections-pagination.md
```diff
@@ -267,7 +267,7 @@ function StoryCommentsSection({story}) {
         />
       )}
       // change-line
-      {isPending && <CommentsLoadingSpinner />}
```
rainandbare, line 270: This component doesn't exist in the example repository

PR #4480, opened 2023-10-12
[relay-lsp] Add folding of linked fields & inline fragments

tobias-tengler:
relay-lsp-folding


tobias-tengler:
Depends on #4473 for consistent updating of the folding ranges.

PR #4481, opened 2023-10-12
Always use / as path separators in generated artifacts

captbaritone:
Differential Revision: D50243120

### DIFF
--- a/compiler/crates/relay-config/src/project_config.rs
+++ b/compiler/crates/relay-config/src/project_config.rs
@@ -8,6 +8,7 @@
 use std::fmt;
 use std::path::Path;
 use std::path::PathBuf;
+use std::path::MAIN_SEPARATOR;
 use std::str::FromStr;
 use std::sync::Arc;
 use std::usize;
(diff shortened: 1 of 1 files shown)


alloy:
While this fixes the path to always have slashes for imports of artefacts, it's still broken for imports of resolver modules. I have [a fix up for this which shows you where the issue lies](https://github.com/facebook/relay/commit/eb69a7eb7af183da25aab778aae86a79afba0c61), but as it's my first foray into Rust it's using a crate to do the work, so your utility should be used instead.

captbaritone:
@alloy Yeah, I suspect there are a few places where this is broken which we'll need to track down. 

Maybe you could start with PR including a fixture test which triggers the bug? You can add the test here: https://github.com/facebook/relay/tree/main/compiler/crates/relay-compiler/tests/relay_compiler_integration/fixtures

Some tips on how to rebuild fixture tests can be found here: https://github.com/facebook/relay/blob/main/.github/CONTRIBUTING.md

captbaritone:
This landed.

PR #4499, opened 2023-10-20
Diagnostic when additional args are passed to defer & stream directives

robrichard:
Followup from https://github.com/facebook/relay/pull/4467#issuecomment-1765078461

### DIFF
--- a/compiler/crates/relay-codegen/src/build_ast.rs
+++ b/compiler/crates/relay-codegen/src/build_ast.rs
@@ -1269,7 +1269,8 @@ impl<'schema, 'builder, 'config> CodegenBuilder<'schema, 'builder, 'config> {
         let DeferDirective { if_arg, label_arg } = DeferDirective::from(
             defer,
             &self.project_config.schema_config.defer_stream_interface,
-        );
+        )
+        .unwrap();
         let if_variable_name = if_arg.and_then(|arg| match &arg.value.item {
             // `true` is the default, remove as the AST is typed just as a variable name string
             // `false` constant values should've been transformed away in skip_unreachable_node
@@ -1321,7 +1322,8 @@ impl<'schema, 'builder, 'config> CodegenBuilder<'schema, 'builder, 'config> {
                 } = StreamDirective::from(
                     stream,
                     &self.project_config.schema_config.defer_stream_interface,
-                );
+                )
+                .unwrap();
                 let if_variable_name = if_arg.and_then(|arg| match &arg.value.item {
                     // `true` is the default, remove as the AST is typed just as a variable name string
                     // `false` constant values should've been transformed away in skip_unreachable_node
--- a/compiler/crates/relay-transforms/src/defer_stream/directives.rs
+++ b/compiler/crates/relay-transforms/src/defer_stream/directives.rs
@@ -5,9 +5,12 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+use common::Diagnostic;
 use graphql_ir::Argument;
 use graphql_ir::Directive;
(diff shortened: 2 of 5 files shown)


compiler/crates/relay-transforms/src/defer_stream/directives.rs
```diff
@@ -47,7 +58,10 @@ impl<'a> StreamDirective<'a> {
     /// Extracts the arguments from the given directive assumed to be a @stream
     /// directive.
     /// Panics on any unexpected arguments.
-    pub fn from(directive: &'a Directive, defer_stream_interface: &DeferStreamInterface) -> Self {
+    pub fn from(
+        directive: &'a Directive,
+        defer_stream_interface: &DeferStreamInterface,
+    ) -> Result<Self, Diagnostic> {
```
captbaritone, line 64: I think we have `DiagnosticResult` (singular) for this.

compiler/crates/relay-transforms/src/defer_stream/mod.rs
```diff
@@ -444,4 +444,10 @@ enum ValidationMessage {
         arg_name: ArgumentName,
         directive_name: DirectiveName,
     },
+
+    #[error("Unexpected argument to @defer: {arg_name}")]
```
captbaritone, line 448:
```suggestion
    #[error("Unexpected argument to @defer: '{arg_name}'.")]
```

compiler/crates/relay-codegen/src/build_ast.rs
```diff
@@ -1269,7 +1269,8 @@ impl<'schema, 'builder, 'config> CodegenBuilder<'schema, 'builder, 'config> {
         let DeferDirective { if_arg, label_arg } = DeferDirective::from(
             defer,
             &self.project_config.schema_config.defer_stream_interface,
-        );
+        )
+        .unwrap();
```
robrichard, line 1273: @captbaritone I was able to capture the Diagnostic errors in the other places `DeferDirective`/`StreamDirective` are used, but I'm not sure how to handle it here. This change doesn't stop the panics because I think this will panic before it gets to the transforms where the errors are handled. 

It doesn't look like this file captures errors and I think it might be a big change to introduce that, but my Rust experience is limited so let me know how you think we should handle this?

compiler/crates/relay-transforms/src/defer_stream/mod.rs
```diff
@@ -444,4 +444,10 @@ enum ValidationMessage {
...
+    UnexpectedArgumentDefer { arg_name: ArgumentName },
+
+    #[error("Unexpected argument to @stream: {arg_name}")]
```
captbaritone, line 451:
```suggestion
    #[error("Unexpected argument to @stream: '{arg_name}'.")]
```

compiler/crates/relay-transforms/src/defer_stream/directives.rs
```diff
@@ -19,7 +22,10 @@ impl<'a> DeferDirective<'a> {
     /// Extracts the arguments from the given directive assumed to be a @defer
     /// directive.
     /// Panics on any unexpected arguments.
-    pub fn from(directive: &'a Directive, defer_stream_interface: &DeferStreamInterface) -> Self {
+    pub fn from(
+        directive: &'a Directive,
+        defer_stream_interface: &DeferStreamInterface,
+    ) -> Result<Self, Diagnostic> {
```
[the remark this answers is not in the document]
robrichard: I only saw `DiagnosticsResult` (plural) here: https://github.com/facebook/relay/blob/main/compiler/crates/common/src/diagnostic.rs#L21

PR #4507, opened 2023-10-26
[relay-lsp] Incremental syncing

tobias-tengler:
Currently the the entire content of a document is sent to the LSP on every change. This is wasteful and also clutters up logs, if you're trying to debug the LSP logs.

This PR switches the LSP to incremental syncing, which means the client is sending the full document content to the LSP on open (`textdocument/didopen`), but only incremental patches as the document changes (`textdocument/didchange`).
`lsp-textdocument` is used to hold the open text documents in memory and apply the incremental patches.


alunyov:
Thank you for adding this. Let me import this and test this internally, we also have few other extensions that are using our LSP crates (specifically GlobalState) so I might need to update and test these too.

PR #4509, opened 2023-10-27
[relay-lsp] Calculate LSP locations from cached JavaScriptSourceFeatures

tobias-tengler:
Currently, when transforming a location to an LSP location, JavaScriptSourceFeatures are extracted from the corresponding file and then used to calculate the LSP location.
The problem is that what's on disk doesn't necessarily match the current document state. If you have unsaved changes in your editor, locations will be calculated on the older (persisted-to-disk) file content.

This PR tries to get the JavaScriptSourceFeatures from the synced sources first (which represent the current document state) and if it can't, falls back to reading from disk.

### DIFF
--- a/compiler/crates/relay-lsp/src/find_field_usages.rs
+++ b/compiler/crates/relay-lsp/src/find_field_usages.rs
@@ -31,7 +31,6 @@ use schema::Type;
 use serde::Deserialize;
 use serde::Serialize;
 
-use crate::location::transform_relay_location_to_lsp_location;
 use crate::server::GlobalState;
 use crate::LSPRuntimeError;
 use crate::LSPRuntimeResult;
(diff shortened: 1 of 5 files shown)


tobias-tengler:
@captbaritone I'm working on some refactorings and it would be great to get this one in beforehand.

captbaritone:
Thanks so much for this. Sorry I've let it languish. There is still one wrinkle, which is that for most of these LSP features the location information is computed from the `program` IRs which I believe are derived from the on-disk representation of the file, not from the potentially unsaved state.

I guess if you have unsaved changes and we report a location in that file that we've derived from on-disk state we're going to show. you the wrong location anyway so this is no worse? Or put another way, there's no way at that point that we can derive a correct location so it's not a problem if its still wrong.

tobias-tengler:
Yes, that's true. I like the approach you took in your inlay hint PR with having both options available and making their behavior explicit through naming. I assume your PR is more likely to land, so I'll just leave this open for now and will close it once yours is merged :)

PR #4512, opened 2023-11-03
Example with missing validation error for a required field argument on the resolver field

alunyov:
The variable $scale is not defined on the query.


compiler/crates/relay-compiler/tests/compile_relay_artifacts/fixtures/relay-resolver-with-args-fragment-spread.expected
```diff
@@ -0,0 +1,225 @@
...
+  profile_picture(scale: $scale)
+}
+
+query relayResolverWithArgsFragmentSpread_Query {
+  ...relayResolverWithArgsFragmentSpread_Fragment
+}
+
+fragment relayResolverWithArgsFragmentSpread_Fragment on Query {
+  pop_star {
+    pop_star_name(scale: $scale)
```
alunyov, line 13: The variable $scale is not defined on the `relayResolverWithArgsFragmentSpread_Query`. We should throw a validation error in this case.

Issue #4465, opened 2023-10-06
Export persisted queries to file


tomgasson:
When using persisted queries but working in a FE-only repo, we end up without the query text available for our mock graphql setup.

I saw [this recent commit: Add a subcommand to dump definitions to file](https://github.com/facebook/relay/commit/1ae14fe737013ac41a105a23a84ce187930fab15).

We're [doing a very similar thing in our fork](https://github.com/atlassian-forks/relay/commit/322a3fb05b478b041651a56d54a1337ed60be669)

This is a gentle request to externalise that subcommand.

We used the query id, but it would also be fine to use the name since there's no risk of versioning conflation in this situation.

PR #4515, opened 2023-11-05
[relay-compiler] @preloadable Support

tobias-tengler:
This is a continuation of #4110

Open TODOs:

- [ ] Value of exported `node` should be formatted using `JSONPrinter`
- [ ] Compiler should throw if `@preloadable` is being used without persisted queries, because [`loadQuery` will throw at runtime otherwise](https://github.com/facebook/relay/blob/10ca82f96f0d763a223ab4d0b14cc5ea2e6b838c/packages/react-relay/relay-hooks/loadQuery.js#L305-L309)
- [ ] Move `PreloadableConcreteRequest<T>` from `@types/react-relay` to `@types/relay-runtime`
- [ ] Add tests


compiler/crates/relay-compiler/src/config.rs
```diff
@@ -404,7 +407,34 @@ Example file:
...
+                                id_and_text_hash,
+                                artifact.artifact_source_keys.clone(),
+                                artifact.source_file,
+                            ))
+                        }
+                        _ => None,
+                    })
+                    .filter_map(|artifact| artifact)
+                    .collect()
+            })),
```
alunyov, line 437: For the code organization purposes, I would move it to an individual module/function with something like `default_generate_extra_artifacts_fn`

compiler/crates/relay-config/src/project_config.rs
```diff
@@ -383,6 +383,21 @@ impl ProjectConfig {
         self.create_path_for_artifact(source_file, filename)
     }
 
+    pub fn path_for_language_specific_artifact(
```
alunyov, line 386: why do we need a language specific `path_for_artifact`?

alunyov (commented):
Thank you for working on this!

compiler/crates/relay-config/src/project_config.rs
tobias-tengler: `path_for_artifact` adds `.graphql.{js|ts}` as an extension. In the previous PR and from what I've read, the $parameters files have been named `<OperationName>$parameters.{js|ts}`, without the `.graphql`.

If desired I can of course re-use the existing method, but the artifacts would be named differently then 😊

tobias-tengler:
@alunyov this would be ready to go from my side.

I'd like to add tests, but it seems like there is no existing test setup for `generate_extra_artifacts`. Is there something that could be pulled into OSS from the internal Relay version or what would be the best way to go about writing tests in your opinion?

tobias-tengler:
@alunyov would you mind taking another look at this? It would be incredibly valuable for us to have this feature :)

alunyov:
@tobias-tengler sorry for the delay. I'll try to get to this next week

compiler/crates/relay-compiler/src/build_project/generate_artifacts.rs
```diff
@@ -228,6 +229,30 @@ fn generate_normalization_artifact(
...
+pub fn generate_preloadable_query_parameters_artifact(
+    project_config: &ProjectConfig,
+    normalization: &Arc<OperationDefinition>,
+    id_and_text_hash: &Option<QueryID>,
+    source_keys: Vec<ArtifactSourceKey>,
+    source_file: SourceLocationKey,
+) -> Artifact {
+    let query_id = id_and_text_hash
+        .clone()
+        .expect("Expected operation artifact to have an `id`. Ensure a `persistConfig` is setup for the current project.");
```
alunyov, line 241: Theoretically, you don't need query persisting for @prelodable query to work. You can just fetch a query with the text. Yes, it will probably increase the size of that artifact, but still it won't include the normalization part of the query, so it's a code-size win compared to the full normalization artifact.

compiler/crates/relay-config/src/project_config.rs
alunyov: Can we also use this method to generate path `for *.graphql.js` artifacts?

alunyov: So here:

https://github.com/facebook/relay/pull/4515/files#diff-c4a1061d9a7a07501974396d27729b0a9b4bac18dd4729ee2982bbb6ff412a42R379-R385

We would call to `self.path_for_language_specific_artifact(format!("{}.graphql.ts", artifact_name)))`

alunyov (changes requested):
Overall, a very clean PR! Thank you for adding this.

Can you add a test for this? I think we have a way to add integration tests now: https://github.com/facebook/relay/tree/main/compiler/crates/relay-compiler/tests/relay_compiler_integration

compiler/crates/relay-compiler/src/build_project/generate_artifacts.rs
tobias-tengler: The code generated for `PreloadableQueryRegistry` requires `params.id` at the moment: https://github.com/facebook/relay/blob/54586eccb033cbdd0896bb1d5d53727a523e3032/compiler/crates/relay-compiler/src/artifact_content/content.rs#L347-L350
and `loadQuery` will throw at runtime, if there s no `id`: https://github.com/facebook/relay/blob/54586eccb033cbdd0896bb1d5d53727a523e3032/packages/react-relay/relay-hooks/loadQuery.js#L304-L309

alunyov: Ah, ok... Maybe the right thing to do is to remove these invariants. But for this PR I think it make sense to keep it as is (with only persisted query id). Thanks!

tobias-tengler:
@alunyov I tried to add a test for this behavior, but there's something going wrong. You can find the failing test here: https://github.com/facebook/relay/pull/4515/commits/0d15d61ddb7db33ea9483fb4b7227f535f3214ff
Running the compiler against the exact same example outside of a test works. In the test I receive a `None` for `id_and_text_hash` here: https://github.com/facebook/relay/blob/0d15d61ddb7db33ea9483fb4b7227f535f3214ff/compiler/crates/relay-compiler/src/build_project/generate_extra_artifacts.rs#L46

Any idea what could be the cause?

compiler/crates/relay-compiler/tests/relay_compiler_integration/fixtures/javascript_preloadable_query.expected
```diff
@@ -0,0 +1,106 @@
...
+    "name": "fooQuery",
+    "operationKind": "query",
+    "text": null
+  }
+};
+})();
+
+node.hash = "21bf4f020aaeb6ce67d04911a13d42a3";
+
+require('relay-runtime').PreloadableQueryRegistry.set(node.params.id, node);
```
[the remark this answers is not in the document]
alunyov: It looks like this flag is missed here:

https://www.internalfb.com/code/fbsource/[c61d25cea342fb77f15e9ff159f21cb429a2e93c]/fbcode/relay/oss/crates/relay-compiler/src/artifact_content/content.rs?lines=342-349

tobias-tengler:
@alunyov Added an integration test for each language option and fixed the `eagerEsModules` handling. Could you kindly give it another look? :)

alunyov (approved):
This looks great!

alunyov:
@tobias-tengler lets fix these rust formatting issues, and I'll import this internally.

tobias-tengler:
@alunyov fixed :) It's weird that my editor doesn't format correctly. I'm using VS Code, rust-analyzer and I'm working from within the `compiler` directory. Normally the editor should format this correctly or am I missing something? I often had to fix formatting by hand with this project...

tobias-tengler:
Do you have an idea why this fails? https://github.com/facebook/relay/actions/runs/7452983745/job/20277398152?pr=4515

alunyov:
> Do you have an idea why this fails? https://github.com/facebook/relay/actions/runs/7452983745/job/20277398152?pr=4515

Did you run `https://github.com/facebook/relay/blob/main/scripts/update-fixtures.sh` to generate the fixture tests?

alunyov:
> @alunyov fixed :) It's weird that my editor doesn't format correctly. I'm using VS Code, rust-analyzer and I'm working from within the `compiler` directory. Normally the editor should format this correctly or am I missing something? I often had to fix formatting by hand with this project... The formatting job in the CI is using a (different) nightly version of Rust than the build. I added a rustup override for 1.72.0 for the compiler as that's what's being used to build the compiler in CI. With the nightly version from the lint job I can't build the project unfortunately.

Sorry about that... We have a weird internal rust configuration, that we somehow is trying to replicate in OSS. 

cc @captbaritone - were there any discussion with internal Rust team on how to properly setup OSS rust projects that needs to be synced internally?

tobias-tengler:
Oh f*** me 😂 I totally wrote them by hand, whoops.

compiler/crates/relay-compiler/tests/relay_compiler_integration/fixtures/preloadable_query_flow.expected
```diff
@@ -0,0 +1,129 @@
...
+  "language": "flow",
+  "schema": "./schema.graphql",
+  "eagerEsModules": true,
+  "persistConfig": { 
+    "file": "./operations.json" 
+  }
+}
+
+//- operations.json
+{}
```
alunyov, line 19: @tobias-tengler somehow the query text is missing in the operations - this is maybe related to the way the integration test prints these files...

tobias-tengler: This is the input. The local persister throws if the file doesn't already exists. This is just here to satisfy the persister.
Screenshot 2024-01-08 at 22 23 20

compiler/crates/relay-compiler/src/artifact_content/content.rs
```diff
@@ -39,6 +39,115 @@ use super::content_section::GenericSection;
...
+    query_id: &QueryID,
+) -> Result<Vec<u8>, FmtError> {
+    let mut request_parameters = build_request_params(normalization_operation);
+    let cloned_query_id = Some(query_id.clone());
+    request_parameters.id = &cloned_query_id;
+
+    let mut content_sections = ContentSections::default();
+
+    // -- Begin Docblock Section --
+    let v = match query_id {
```
tyao1, line 57: nit: Prefer explicit variable names
```suggestion
    let extra_annotations = match query_id {
```

compiler/crates/relay-compiler/src/config.rs
```diff
@@ -408,7 +409,7 @@ Example file:
             header: config_file.header,
             codegen_command: config_file.codegen_command,
             load_saved_state_file: None,
-            generate_extra_artifacts: None,
+            generate_extra_artifacts: Some(Box::new(default_generate_extra_artifacts_fn)),
```
tyao1, line 412: My concern is there could be existing commands at Meta that shouldn't generate extra artifacts. Could this be configured just for the OSS binary?

compiler/crates/relay-config/src/project_config.rs
```diff
@@ -373,17 +373,32 @@ impl ProjectConfig {
...
+    }
+
+    pub fn path_for_language_specific_artifact(
+        &self,
+        source_file: SourceLocationKey,
+        artifact_file_name: String,
+    ) -> PathBuf {
+        let filename = match &self.typegen_config.language {
+            TypegenLanguage::Flow | TypegenLanguage::JavaScript => {
+                format!("{}.js", artifact_file_name)
```
tyao1, line 396: I might be missing something, does this change regular artifact paths from `.graphql.js` to `.js`?

tobias-tengler: Nope. This method is passed `<name>.GraphQL` as `artifact_file_name` by `artifact_path_for_definition`. So the old artifact names remain.

compiler/crates/relay-compiler/src/config.rs
tobias-tengler: I'm not sure how I could configure this just for the OSS binary. Are there any existing examples for such a case?

tomgasson: At Atlassian we've customised our relay-bin to [output persisted mocks](https://github.com/facebook/relay/issues/4465). 


Easy to tweak so that we chain these calls since we're already customising that - but since this is a new feature, potentially it can be a different config (or, feature flagged rather than a config) and done with an extra `artifacts.extend` just before the `generate_extra_artifacts_fn` runs [here](https://github.com/facebook/relay/blob/9f135b8f92a750c7bc79c08bda0c149539ac0bc8/compiler/crates/relay-compiler/src/build_project/mod.rs#L360-L370)?

alunyov: Good point @tyao1 

@tobias-tengler I think you can move that config to `relay-bin`:

```
diff --git a/compiler/crates/relay-bin/src/main.rs b/compiler/crates/relay-bin/src/main.rs
index f832244070dd..51bab9e93822 100644
--- a/compiler/crates/relay-bin/src/main.rs
+++ b/compiler/crates/relay-bin/src/main.rs
@@ -18,6 +18,7 @@ use intern::string_key::Intern;
 use log::error;
 use log::info;
 use relay_compiler::build_project::artifact_writer::ArtifactValidationWriter;
+use relay_compiler::build_project::generate_extra_artifacts::default_generate_extra_artifacts_fn;
 use relay_compiler::compiler::Compiler;
 use relay_compiler::config::Config;
 use relay_compiler::errors::Error as CompilerError;
@@ -296,6 +297,8 @@ async fn handle_compiler_command(command: CompileCommand) -> Result<(), Error> {
         );
     }
 
+    config.generate_extra_artifacts = Some(Box::new(default_generate_extra_artifacts_fn));
+
     let compiler = Compiler::new(Arc::new(config), Arc::new(ConsoleLogger));
 
     if command.watch {
diff --git a/compiler/crates/relay-compiler/src/config.rs b/compiler/crates/relay-compiler/src/config.rs
index b42255bb86a4..f5861eeaf6a7 100644
--- a/compiler/crates/relay-compiler/src/config.rs
+++ b/compiler/crates/relay-compiler/src/config.rs
@@ -57,7 +57,6 @@ use watchman_client::pdu::ScmAwareClockData;
 
 use crate::build_project::artifact_writer::ArtifactFileWriter;
 use crate::build_project::artifact_writer::ArtifactWriter;
-use crate::build_project::generate_extra_artifacts::default_generate_extra_artifacts_fn;
 use crate::build_project::generate_extra_artifacts::GenerateExtraArtifactsFn;
 use crate::build_project::get_artifacts_file_hash_map::GetArtifactsFileHashMapFn;
 use crate::build_project::AdditionalValidations;
@@ -409,7 +408,7 @@ Example file:
             header: config_file.header,
             codegen_command: config_file.codegen_command,
             load_saved_state_file: None,
-            generate_extra_artifacts: Some(Box::new(default_generate_extra_artifacts_fn)),
+            generate_extra_artifacts: None,
             generate_virtual_id_file_name: None,
             get_artifacts_file_hash_map: None,
             saved_state_config: config_file.saved_state_config,
@@ -576,7 +575,11 @@ impl fmt::Debug for Config {
         } = self;
 
+                "Some(Fn)"
+            } else {
+                "None"
+            }
         }
 
         f.debug_struct("Config")
```

tobias-tengler: Moved it to relay-bin :)

alunyov:
@tobias-tengler you'll also need to update integration test setup, to include the default extra generator to the test config

alex-statsig:
Can / should this support mutations?

I've noticed mutations often add up to a large chunk of bundle sizes (sometimes due to including `...EntirePage` to ensure data is all refetched). It seems like the AST shouldn't be needed immediately for mutations, and committing the mutation with appropriate variables could be decoupled (especially with persisted queries removing the need for the entire query text in the bundle).

It's not quite the same flow as preloading queries, but a similar idea I think to prioritize the part needed to issue the request (and defer the part to process the response). It seems especially applicable since mutations are not used more often than they are used.

PR #4540, opened 2023-11-30
FIX(Graphql): Add support for input with default values

eMerzh:
Hello here 👋 

We have a few inputs with default required fields that have default values like here :

```graphql
input StoryPinInput {
  id: ID!
  isGlobal: Boolean! = false
}
```

but the TS generated for this look like 
```ts
export type StoryPinInput = {
  id: string;
  isGlobal: boolean;
};
```
which is not ideal ... 

so i marked the field as optional if it has a default :)

### DIFF
--- a/compiler/crates/relay-typegen/tests/generate_typescript/fixtures/default-input.expected
+++ b/compiler/crates/relay-typegen/tests/generate_typescript/fixtures/default-input.expected
@@ -1,0 +1,27 @@
+==================================== INPUT ====================================
+mutation feedbackUnLikeMutation($input: FeedbackUnLikeInput) {
+  feedbackUnLike(input: $input) {
+    feedback {
+      id
+    }
+  }
+}
+==================================== OUTPUT ===================================
+export type FeedbackUnLikeInput = {
+  feedbackId?: string | null | undefined;
+  silent?: boolean;
+};
+export type feedbackUnLikeMutation$variables = {
+  input?: FeedbackUnLikeInput | null | undefined;
+};
+export type feedbackUnLikeMutation$data = {
+  readonly feedbackUnLike: {
+    readonly feedback: {
+      readonly id: string;
+    } | null | undefined;
+  } | null | undefined;
+};
+export type feedbackUnLikeMutation = {
+  response: feedbackUnLikeMutation$data;
+  variables: feedbackUnLikeMutation$variables;
+};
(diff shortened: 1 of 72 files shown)


eMerzh:
@captbaritone sorry for the ping 🥹 .... is there a way to bump those PRs before they go down the black hole of 2nd page ? 

if there's something we can do to help ? 🤷

captbaritone:
Sorry for the delayed response. Can you add a fixture test here? https://github.com/facebook/relay/tree/main/compiler/crates/relay-typegen/tests/generate_typescript/fixtures I think the test schema has an example you can trigger here: https://github.com/facebook/relay/blob/54586eccb033cbdd0896bb1d5d53727a523e3032/packages/relay-test-utils-internal/testschema.graphql#L677

Maybe you could also ensure there'a fixture test that validates what we do for Flow in this case?

Info on updating fixture tests can be found here: https://github.com/facebook/relay/blob/main/.github/CONTRIBUTING.md#pull-requests

Defaults in GraphQL are confusing, but I believe we should should ensure this generates `{ isGlobal?: boolean }` and not `{ isGlobal: boolean | null | undefined }` since I believe with required arguments with defaults you are permitted to omit the argument but not pass an explicit null.

eMerzh:
awesome! ❤️  .... i'll take a look at that

eMerzh:
here... i hope it's better now :) 

tried to not add input but i prefer avoid touching too much other tests to add this

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/default-input.graphql
```diff
@@ -0,0 +1,7 @@
+mutation feedbackUnLikeMutation($input: FeedbackUnLikeInput) {
+  feedbackUnLike(input: $input) {
+    feedback {
+      id
+    }
+  }
+}
```
captbaritone, line 7: Would you mind adding this same fixture to the Flow tests?

https://github.com/facebook/relay/tree/main/compiler/crates/relay-typegen/tests/generate_flow/fixtures

captbaritone (changes requested):

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/default-input.expected
```diff
@@ -0,0 +1,27 @@
...
+  feedbackUnLike(input: $input) {
+    feedback {
+      id
+    }
+  }
+}
+==================================== OUTPUT ===================================
+export type FeedbackUnLikeInput = {
+  feedbackId?: string | null | undefined;
+  silent?: boolean;
```
captbaritone, line 12: nice!

compiler/crates/relay-typegen/tests/generate_typescript/fixtures/default-input.graphql
eMerzh: arg, my bad... here they are :)

eMerzh:
awesome ❤️  thanks for the review

PR #4544, opened 2023-12-07
fix path on Windows

### LINKED ISSUE #3272: Can't resolve graphl file when using Webpack 5 (and nextjs)

artola:
- closes #3272

@sibelius, here is a PR addressing this long-standing issue. This is a follow-up to your [comment](https://github.com/facebook/relay/issues/3272#issuecomment-1836003327).


sibelius (commented):
can you add a test?

artola:
> can you add a test?

A test has been added to ensure that paths are properly handled. On both POSIX and non-POSIX platforms, it should always emit forward slashes.

PR #4547, opened 2023-12-10
[Rough Draft] Document Relay Resolvers

captbaritone:
For the last few years we've been experimenting with expanding [Relay Resolvers](https://relay.dev/docs/guides/relay-resolvers/) to allow them to return client state that can change over time. We are now close to releasing this new capability as a stable part of Relay, and so these docs my initial stab at documenting their full feature set. **I expect these docs will go through a few revisions before we land, so they still have a few "TODO"s scattered around in the docs (and in this PR).**

**Warning**

These docs document the API we **expect** to ship. Today it still requires enabling both compiler and runtime feature flags, and using a special instance of the Relay Store. Additionally, some of the types and values what will be exported as top level exports of the `relay-runtime` module are not yet exported. **Following these docs today is not expected to work** Thanks for your patience!

## TODO

List of work items still to be done for these docs:

- [ ] Ensure we expose the `LiveState` type from `relay-runtime` root
- [ ] Decide on canonical name for Live Resolvers ("live data?", "live resolvers?", "live fields?")
  - [ ] How does this align with "derived fields"
- [ ] Unify around name for "client schema". Client resolver schema? Client state schema?
  - [ ] This is confusing because we already have client schema extensions
- [ ] What about type exports for TypeScript?
  - [ ] LiveState
  - [ ] SuspenseSentinel
- [ ] Expose our live resolver batching method directly on Relay environment
- [ ] I think that we currently allow Resolver types to be @deprecated, but in documenting this, I realized that we the spec does not actually allow this.


website/blog/2023-03-30-relay-15.mdx
```diff
@@ -70,7 +70,7 @@ Flow users will now get types inferred from `graphql` literals with more Relay A
 
 ### Relay Resolver improvements
 
-A significant portion of our development effort since our last release has gone into improving [Relay Resolvers](https://relay.dev/docs/guides/relay-resolvers) (a mechanism for exposing derived data in the graph). It is worth noting that Relay Resolvers are still experimental and API changes might occur in the future.
+A significant portion of our development effort since our last release has gone into improving [Relay Resolvers](https://relay.dev/api-reference/relay-resolvers/introduction/) (a mechanism for exposing derived data in the graph). It is worth noting that Relay Resolvers are still experimental and API changes might occur in the future.
```
monicatang, line 73: Is the last sentence still applicable after we release them as stable? Maybe we can remove the "Relay Resolvers are still experimental" bit

website/docs/api-reference/relay-resolvers/defining-fields.md
```diff
@@ -0,0 +1,41 @@
...
+title: "Defining Fields"
+slug: /api-reference/relay-resolvers/defining-fields/
+description: How to define fields for your client state schema using Relay Resolvers
+---
+
+Defining fields on a client type is as simple as defining a resolver function which accepts an instance of your model type [TODO link] as its first argument and returns the field value. Note that the exported function name must match the field name.
+
+## Syntax
+
+When defining a field string after `@RelayResolver` is a GraphQL `TypeName` followed by a dot followed by the field
```
monicatang, line 12: Nit - Suggestion for clarity (IMO)
```suggestion

Relay resolvers are marked via docblocks above a resolver function.
`@RelayResolver` is the tag to indicate the start of any relay resolver definition.
To define a field, add a GraphQL `TypeName` followed by a dot followed by the field
```

website/docs/api-reference/relay-resolvers/limitations.md
```diff
@@ -0,0 +1,32 @@
+---
+id: limitations
+title: "Limitations"
+slug: /api-reference/relay-resolvers/limitations/
+description: Limitations of Relay Resolvers
+---
+
+Relay Resolvers are do have some limitations. Here we will collect a list of known limitations and provide alternatives where possible.
```
monicatang, line 8:
```suggestion
Here are some known limitations of RelayResolvers and their alternatives, if any.
```

website/docs/guides/relay-resolvers/live-fields.md
```diff
@@ -0,0 +1,119 @@
...
+import type { LiveState } from 'relay-runtime';
+
+/**
+ * @RelayResolver Query.counter: Int
+ * @live
+ */
+export function counter(): LiveState<number> {
+  return {
+    read: () => store.getState().counter,
+    subscribe: (cb) => {
```
TJ09, line 26: A more descriptive name (like "callback") or a typehint would do wonders here. `cb` is a bit opaque.

website/blog/2023-03-30-relay-15.mdx
captbaritone: I think it's okay for a blog post to be a point in time artifact. That said, maybe I can add and an "update" section.

website/docs/guides/relay-resolvers/live-fields.md
captbaritone: Good point. Thanks for pointing that out. Will fix.

website/docs/guides/relay-resolvers/defining-types.md
```diff
@@ -0,0 +1,49 @@
...
+### Defining a “weak” type
+
+If your type does not have unique identifiers, you can define it as “weak” by adding the `@weak` docblock tag. Weak types are defined by a docblock followed by an exported type definition matching the types name. Resolvers that define edges to weak types will need to return a fully populated model object[TODO LINK] matching this type.
+
+```tsx
+/**
+ * @RelayResolver ProfilePicture
+ * @weak
+ */
+export type ProfilePicture = { url: string, height: number, width: numbe };
```
Blitz2145, line 44: Small typo here should be `width: number`

website/docs/guides/relay-resolvers/defining-types.md
```diff
@@ -0,0 +1,49 @@
...
+
+:::info
+Each client state GraphQL type is backed by a JavaScript object type which these docs will refer to as its "model type". Resolvers "on" this type will be passed an instance of this type as their first argument.
+:::
+
+Resolver types are defined using the `@RelayResolver` tag followed by the name of the type you are defining. By default Relay assumes your client types are “strong”, meaning each instance has a unique ID. This property allows Relay to apply a number of optimizations, such as memoizing resolver computation.
+
+### Defining a “strong” type
+
+Strong types ar defined by a docblock followed by an exported function whose name matches the type's name, and which accepts an ID as its first argument and returns an instance of the type’s model [TODO link]. Resolvers that define edges to this type will simply need to return the ID of the object, rather than deriving the model themselves.
```
monicatang, line 18: is it right that the model resolver can only accept one argument (the id argument)? Seems to be the case when I tried it. IIUC, instead of "first" argument we can say "only"?

website/docs/guides/relay-resolvers/defining-fields.md
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * @RelayResolver User.name: String
+ */
+export function name(user: UserModel): string {
+  return user.name;
+}
+```
+
+:::note
+Relay will takes care of efficiently recomputing resolvers when any of their inputs (in this case the model instance) change, so you don’t need to worry about memoizing your resolver function.
```
christiansany, line 32:
```suggestion
Relay will take care of efficiently recomputing resolvers when any of their inputs (in this case the model instance) change, so you don’t need to worry about memoizing your resolver function.
```

website/docs/api-reference/relay-resolvers/dockblock-format.md
```diff
@@ -0,0 +1,153 @@
...
+ * @weak
+ */
+export type ProfilePicture = {
+  url: string;
+  width: number;
+  height: number;
+} 
+```
+
+See the [Weak Types](../../guides/relay-resolvers/defining-types.md#Defining a “weak” type) guide for more information including how to define an edge to a weak type.
```
tyao1, line 119: The link is somehow broken

PR #4550, opened 2023-12-12
fix: correct typo `interaction`

JeffGuKang:
`intecation` => `interaction`


alunyov:
Thank you for the fix, can you also update the source of this page?

captbaritone:
@JeffGuKang The source of truth for this content is here: https://github.com/facebook/relay/blob/main/website/docs/tutorial/queries-2.md

PR #4565, opened 2024-01-06
Add fuzz-harness for graphql-syntax

silvergasp:


silvergasp:
See #4566 for more context.

Issue #4037, opened 2022-08-06
Improve docs for relay.config.js with Typescript support from JSDoc


EloB:
It would be good if this section:
https://relay.dev/docs/getting-started/installation-and-setup/#compiler-configuration

Had the same documenation like Next.js:
https://nextjs.org/docs/basic-features/typescript#type-checking-nextconfigjs

So the editor can understand the shape of the config.
https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#type

PR #4569, opened 2024-01-07
feat(compiler): add type declaration with relay config file types

noghartt:
resolve https://github.com/facebook/relay/issues/4037


noghartt:
> I'm a little hesitant to merge this given that I don't think it's fully complete. For example I don't see feature flags or mutli-project config. Would you be interested in taking it the rest of the way and capturing the full config? If so, I think this would be a great addition.

I'd love to take the rest of the config and documentate it.

But yes, it isn't fully complete because I didn't find some good documentations related to the rest of the config. I saw that the `relay-config` crate has other fields, but I couldn't find anything that documentate these configurations, any idea where I can find it?

What I follow was the [README](https://github.com/facebook/relay/tree/main/packages/relay-compiler) from the `react-compiler` package, but I think that is a little outdated, right?

tobias-tengler:
If it's just about adding the TypeScript declarations, I think you can work of the declaration file I derived from the config last year: https://github.com/facebook/relay/blob/d873d5b8927b4a5bb6ca2c4b9a4c504ea0723b6b/packages/relay-compiler/index.d.ts
I would also like to continue my effort regarding the VSCode integration for config completion in JSON files: https://github.com/facebook/relay/pull/4162 Now that I have some more experience with the compiler, maybe I have more luck than back then :D

sibelius (commented):
cool

noghartt:
cc @captbaritone @tobias-tengler 

I improved the types and wrote more JSDocs based on the `config.rs` as mentioned before. Improve the doc too, what do you think?

packages/relay-compiler/index.d.ts
```diff
@@ -0,0 +1,327 @@
...
+  /**
+   * Let users enable the Relay Resolvers.
+   * 
+   * @see https://relay.dev/docs/guides/relay-resolvers/
+   */
+  enable_relay_resolver_transform?: boolean;
+  /** Enable deprecated `@outputType` on Relay Resolvers. */
+  relay_resolver_enable_output_type?: RelayConfigFeatureFlag;
+  /** Enable returning interfaces from Relay Resolvers without `@outputType` */
+  relay_resolver_enable_interface_output_type: RelayConfigFeatureFlag;
```
captbaritone, line 297: I think all of these are optional?

packages/relay-compiler/index.d.ts
```diff
@@ -0,0 +1,327 @@
...
+  /** Root directory relative of this configuration. */
+  root?: string;
+  /** A header to be prepend to each generated artifact. */
+  header?: string[];
+  /**
+   * A mapping directory from directory paths (relative to the root) to a project.
+   * If a path is a subdirectory of another path, the more specific path wins.
+   */
+  sources: Record<string, string>;
+  /** Command name that for relay compiler. */
```
captbaritone, line 80: Not sure about this comment.

packages/relay-compiler/index.d.ts
```diff
@@ -0,0 +1,327 @@
...
+  persist?: RelayConfigPersitConfigLocal;
+  /**
+   * Restrict @relay_test_operation to directories matching this regex
+   */
+  testPathRegex?: string;
+  /** Generates a `// @relayVariables name1 name2` header in generated operation files. */
+  variableNamesComment?: boolean;
+  featureFlags?: RelayConfigFeatureFlags;
+  /** A placeholder for allowing extra information in the config file */
+  extra: unknown;
```
captbaritone, line 203: Presumably this should be optional?

captbaritone (changes requested):
So exciting! This is going to be such a good improvement. A few requests:

1. Could you do another pass to ensure no properties which are actually optional are marked as non-optional?
2. In the Rust code, can you add a comment at the top of each struct that is part of this config with a note that any changes made to the struct should also be made to these types?

captbaritone:
Another point: We don't have authoritative docs on this config. As a baseline we should link out to this definition from our docs as a "Here's where you can go to find out what options are available in the config". As a stretch, I wonder if it would be possible to use something like https://www.npmjs.com/package/docusaurus-plugin-typedoc to generate an actual docs page (or pages) from this?

Maybe we can start with the former and then (optionally) explore the later in a separate PR.

packages/relay-compiler/index.d.ts
tobias-tengler: This should probably just be `[configKey: string]: any`?

noghartt: I pushed a fix into this tht turns it into an optional field.

> This should probably just be `[configKey: string]: any`?

In case, I think that if we should type it, we should type accordingly to the `serde::Value` struct, that can be any JSON-friendly value. What do you think?

noghartt:
> Another point: We don't have authoritative docs on this config. As a baseline we should link out to this definition from our docs as a "Here's where you can go to find out what options are available in the config". As a stretch, I wonder if it would be possible to use something like https://www.npmjs.com/package/docusaurus-plugin-typedoc to generate an actual docs page (or pages) from this?
> 
> Maybe we can start with the former and then (optionally) explore the later in a separate PR.

Would be really cool if we could have a docs specifically for these configs, explaining which things each field are doing. I think that we can derive some docs from this config, like one about feature flags and what are the features related.

packages/relay-compiler/index.d.ts
```diff
@@ -0,0 +1,332 @@
...
+   */
+  isDevVariableName?: string;
+  /** 
+   * Formatting style for generated files.
+   * 
+   * @default "commonjs"
+   */
+  jsModuleFormat?: 'commonjs' | 'haste';
+  /** Options for configuring the output of compiler diagnostics. */
+  diagnosticReportConfig: RelayConfigDiagnosticReport;
```
captbaritone, line 61: I think we need to understand defaults a little better, since I know a config without this value (for example, but others as well) is accepted by the compiler.

Ah, looking a bit more closely, it looks like the struct itself is annotated with `default`. Can you check each of the structs and see how that impacts these types?

https://github.com/facebook/relay/blob/ece1fd953b987d08ab1833d5c1f63dee7728a4ab/compiler/crates/relay-compiler/src/config.rs#L666

website/docs/getting-started/installation-and-setup.md
```diff
@@ -74,7 +75,7 @@ module.exports = {
 ```
 
 This configuration also can be specified in `"relay"` section of the `package.json` file.
-For more details, and configuration options see: [Relay Compiler Configuration](https://github.com/facebook/relay/tree/main/packages/relay-compiler)
+For more details, and configuration options see: [Relay Compiler Configuration](https://github.com/facebook/relay/tree/main/packages/relay-compiler/index.d.ts)
```
captbaritone, line 78: Love this! Could you also add a similar note to https://github.com/facebook/relay/blob/main/packages/relay-compiler/README.md#configuration?

captbaritone (changes requested):

website/docs/getting-started/installation-and-setup.md
noghartt: Added! I updated the `README` with a note admonition containing the same info.

packages/relay-compiler/index.d.ts
noghartt: I checked all the structs and update some JSDocs with new `@default` values. I think that everything is OK now, what do you think?

packages/relay-compiler/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+   * 
+   * @default
+   * ```js
+   * { 
+   *   criticalLevel: "error",
+   * }
+   * ```
+   */
+  diagnosticReportConfig?: RelayConfigDiagnosticReportConfig;
+  featureFlags: RelayFeatureFlags;
```
captbaritone, line 71: This is optional.

packages/relay-compiler/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+  /** 
+   * Keep the previous compiler behavior by outputting an union of the raw type and null, 
+   * and not the **correct** behavior of an union with the raw type, null and undefined.
+   * 
+   * @default false
+   */
+  typescript_exclude_undefined_from_nullable_union: boolean;
```
captbaritone, line 78: This is optional

captbaritone (changes requested):

packages/relay-compiler/index.d.ts
noghartt: Fixed!

packages/relay-compiler/index.d.ts
noghartt: Fixed!

captbaritone:
I've imported this, but it will race with https://github.com/facebook/relay/pull/4585 which changes the config types. I'll try to fix it up internally before I land.

captbaritone:
Hey. After importing this, I'm seeing a bunch more errors. How have you been validating this addition? Could you find a way to ensure it's working end to end on a given config file? Maybe include a screenshot of what errors look like in VSCode?

packages/relay-compiler/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+   * @default "__DEV__"
+   */
+  isDevVariableName?: string;
+}
+
+type RelayConfigMultiProjectItemSchemaConfig =
+  | { 
+      /** The path of your schema file. */
+      schema: string 
+    };
```
captbaritone, line 127: I don't think a semicolon is valid here.

packages/relay-compiler/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+   * 
+   * @default
+   * ```js
+   * { 
+   *   criticalLevel: "error",
+   * }
+   * ```
+   */
+  diagnosticReportConfig?: RelayConfigDiagnosticReportConfig;
+  featureFlags?: RelayFeatureFlags;
```
captbaritone, line 71: This type is not defined.

packages/relay-compiler/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+  enumModuleSuffix?: string;
+  /**
+   * (Flow only) When set, generated input types will have the listed fields optional even if the schema defines them as required.
+   */
+  optionalInputFields?: string[];
+  /**
+   * (TypeScript only) Whether to use the `import type` syntax introduced in Typescript version 3.8. This will prevent warnings from `importsNotUsedAsValues`.
+   */
+  useImportTypeSyntax?: boolean;
+  customScalarTypes?: CustomScalarTypes;
```
captbaritone, line 158: This type is not defined

packages/relay-compiler/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+   */
+  flowTypegen?: {
+    /**
+     * This option controls whether or not a catch-all entry is added to enum type definitions values that may be added 
+     * in the future. Enabling this means you will have to update your application whenever the GraphQL server schema adds 
+     * new enum values to prevent it from breaking.
+     * 
+     * @default false
+     */
+    noFutureProofEnums?: NoFutureProofEnums;
```
captbaritone, line 174: This type is not defined

packages/relay-compiler/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+  /** 
+   * For now, this also disallows fragments with variable definitions. 
+   * This also makes `@module` to opt in using `@no_inline` internally
+   * 
+   * **NOTE** that the presence of a fragment in this list only controls 
+   * whether a fragment is *allowed* to use `@no_inline`: whether the fragment is 
+   * inlined or not depends on whether it actually uses that directive.
+   */
+  no_inline?: RelayConfigFeatureFlag;
+  enable_3d_branch_arg_generation?: bool;
```
captbaritone, line 346: `bool` is not a valid TypeScript type.

packages/relay-compiler/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+  text_artifacts?: RelayConfigFeatureFlag;
+  skip_printing_nulls?: RelayConfigFeatureFlag;
+  /** Enable support for the experimental `@alias` directive on fragment spreads. */
+  enable_fragment_aliases?: RelayConfigFeatureFlag;
+  /** Print queries in compact form. */
+  compact_query_text?: RelayConfigFeatureFlag;
+  /** Create normalization nodes for client edges to client objects. */
+  emit_normalization_nodes_for_client_edges?: boolean;
+  /** Fully build the normalization AST for Resolvers. */
+  enable_resolver_normalization_ast?: bool;
```
captbaritone, line 357: `bool` is not a valid TypeScript type.

captbaritone (changes requested):

packages/relay-compiler/index.d.ts
noghartt: I think that this isn't a problem, but I removed it to follow the pattern

packages/relay-compiler/index.d.ts
noghartt: Fixed

packages/relay-compiler/index.d.ts
noghartt: I defined this type, fixed

packages/relay-compiler/index.d.ts
noghartt: Oh, nice catch, fixed

packages/relay-compiler/index.d.ts
noghartt: Fixed

captbaritone:
@noghartt Can you please add a test plan that demonstrates this working end to end? For example, a screenshot of it reporting errors on an invalid config file in VSCode? (and one of it not erroring on a valid config file)?

noghartt:
> @noghartt Can you please add a test plan that demonstrates this working end to end? For example, a screenshot of it reporting errors on an invalid config file in VSCode? (and one of it not erroring on a valid config file)?

Of course. But in case, VSCode and LSP does not infer type checking over type infered objects that comes from JSDocs annotation, so it'll won't trigger a visual error for the final user, only if they have a `relay.config.ts` and let the `tsserver` does it for us.

But in case, I took some screenshots that shows this type working, I'll let also a test plan containing how I test it:

## Test plan

- [ ] build the packages using gulp with the command `yarn gulp mainrelease`
- [ ] create a new `relay.config.js` file
- [ ] add type annotations targeting the exported `RelayConfig` type from `./dist` build package

the boilerplate 

```js
// relay.config.js

/** @type {import('./dist/relay-compiler').RelayConfig} */
module.exports = {
  // ...
}
```

## Screenshots

starting single project
image

multi project config
image

feature flags
image

---

As an example, even the `next.config.js` aren't being typechecked by the LSP.

image

compiler/crates/common/src/feature_flags.rs
```diff
@@ -17,6 +17,7 @@ use serde::Serialize;
 
 use crate::Rollout;
 
+/// **NOTE**: Every change on this struct should be reflected into the `relay-compiler/index.d.ts` file.
```
alunyov, line 20: I wonder if we can make use of something like this: https://github.com/Aleph-Alpha/ts-rs here?

We would generate these typescript definitions based on the actual rust types?

noghartt: I did some POCs with [specta](https://docs.rs/specta/latest/specta/) yesterday, it seems to works well. I think that we can bring it to these structs too.

They have just a limitation around some JSDocs comments, like the `@default` and others, but it's working.

alunyov: Nice, @noghartt! I think this would be a much better solution to always keep these definitions in sync.

captbaritone: @noghartt Just checking in here. Were you able to make any progress with Spectra? Do you think it can offer a viable automated solution?

noghartt: > @noghartt Just checking in here. Were you able to make any progress with Spectra? Do you think it can offer a viable automated solution?

I think that we can bring an automated solution yes. But as I commented, specta doesn't seem to fit all the types for us, as an example, it doesn't fill the JSDocs related fields, like the `@default`. I couldn't find a workaround for that scenario.

Except that, it seems to works well. Do you think that we can bring this improvement in this PR or maybe moving to another?

noghartt:
Well done by @captbaritone on this PR https://github.com/facebook/relay/pull/4723. We can follow this improvement and bring some type emission following the ideas of `JsonSchema`!

Issue #4567, opened 2024-01-06
Cargo doc fails to generate documentation


silvergasp:
When familiarising myself with the relay compiler codebase I ran;
```
cd compiler
cargo doc --open
```

There are a slew of errors that come up with URLs int the documentation that aren't formatted correctly in docstrings.

silvergasp:
I've got a partial fix in #4568, but I haven't added this the CI yet to prevent future regressions.

captbaritone:
https://github.com/facebook/relay/pull/4568 has merged. Would you like to add a test to CI?

silvergasp:
> #4568 has merged. Would you like to add a test to CI?

Done in #4580

silvergasp:
Scratch that WIP in #4580

silvergasp:
Now it's done

PR #4580, opened 2024-01-10
Add cargo doc to CI

silvergasp:
Closes: #4567


silvergasp:
Lint failure doesn't seem to be related to this change.

.github/workflows/ci.yml
```diff
@@ -100,6 +100,25 @@ jobs:
       - name: "Run tests"
         run: cargo test --manifest-path=compiler/Cargo.toml --locked ${{ matrix.target.features && '--features' }} ${{ matrix.target.features }}
 
+  
+  build-docs:
+    name: Build docs (${{ matrix.target.os }})
```
captbaritone, line 105: This name could be confusing since it kinda sounds like this is building our documentation website. Maybe instead call it "Build Cargo Docs" or something?

For bonus points, maybe we could figure out how to build this as part of the actual website build and include it as a part of the website!

PR #4581, opened 2024-01-10
[relay-lsp] Rename refactoring

tobias-tengler:
This is another attempt at https://github.com/facebook/relay/pull/4421. I've removed the automatic renaming part to make the feature purely opt-in (can still be made automatic later, maybe also behind a flag).

## Changes
- Adds support for the following rename operations
  - Renaming operations
  - Renaming fragments (both from definition and spread)
  - Rename fragment arguments (both from definition and usage)
  - Rename operation variables (both from definition and usage)
    - Global variables aren't yet supported. I plan to add support for this in a later PR alongside better LSP support for global variables in general


captbaritone:
Thanks for continuing to work on this!

1. Can "Fix "goto references" for TerseRelayResolver" be split into a separate PR?
2. How could we add unit tests for this code? Happy to hop on a VC if you'd like and we can brainstorm

compiler/crates/relay-lsp/tests/rename/fixtures/rename_fragment_argument_from_definition.expected
```diff
@@ -0,0 +1,40 @@
...
+  me {
+    ...UserFragment @arguments(RENAMED: $include_firstname)
+  }
+}
+
+fragment UserFragment on User @argumentDefinitions(RENAMED: { type: "Boolean!" }) {
+  name @skip(if: $return_firstname)
+  ... @defer(if: $RENAMED) {
+    og_name: name
+  }
```
tobias-tengler, line 32: Any idea why the rename works for the argument on the custom directive but not for `@skip`? Is it treated differently by the Visitor?

tobias-tengler (commented):
@captbaritone I've added fixture tests similar to the ones that were already present for other LSP features in bb101280533e30ef89cf94f0e1bb474c8f8dffe8 :)

compiler/crates/relay-lsp/tests/rename/fixtures/rename_fragment_argument_from_definition.expected
tobias-tengler: Nvm, I can get a handle on ot through visit_variable, but it feels not intuitive.

tobias-tengler:
@captbaritone Dammit, missed the release window again 😅 Would still love to get this in though!

compiler/crates/relay-lsp/src/rename/mod.rs
```diff
@@ -0,0 +1,638 @@
...
+}
+
+impl FragmentFinder {
+    pub fn get_fragment_usages(program: &Program, name: StringKey) -> Vec<IRLocation> {
+        let mut finder = FragmentFinder {
+            fragment_locations: vec![],
+            fragment_name: name,
+        };
+        finder.visit_program(program);
+        finder.fragment_locations
```
captbaritone, line 452: Fragment names can also be referenced in Relay resolver docblocks in the `@rootFragment` tag. Would be cool if we could handle those as well.

captbaritone (commented):
We're trying to get into a more regular release cadence, so hopefully we can merge this soon and cut another release including it!

Overall these changes look really sound, and the tests are particularly great. Thanks for adding them!

Some thoughts:
1. Variable renaming is intentionally local to the operation, but variables in GraphQL are global. Every fragment that is transitively spread into the operation has access to it. This gets extra gross because a fragment can be spread into multiple queries, meaning it forms a coupling between the variable names in one query with the variable names in the other. I suspect this means that the approach you take here (where it's intentionally local) is the best one, but I wanted to make sure that was an intentional decision.
2. I'm curious how this works with project inheritance. In Relay with multi-project config, you can define a project as extending another project. The child project has visibility into the parent project. If I rename a fragment defined in a parent project, does it also rename all instanced where that fragment is referenced from a child project?
3. "Make into local variable" would be an incredible refactor. To be able to take a variable in a fragment that is currently referencing a global query variable, define as a fragment argument, and then update all fragment spreads to pass the argument explicitly.
4. Have you been able to validate these changes end to end in VSCode? If so, could you include a video showing that so we can make it part of the test plan?

compiler/crates/relay-lsp/src/rename/mod.rs
tobias-tengler: I've been looking into this and I think I'll have to introduce tracking of those `@rootFragment` references (with location information) within the `Program`. Currently the `@rootFragment` name is only tracked as argument on the internal relay resolver field directive. I don't see how I could include location information there, without weirdly serializing it to GraphQL primitives.
If you have a nice idea on how to tackle this - let me know :)

tobias-tengler:
@captbaritone I've updated this to also support renaming of proper fragment arguments (variable definitions) now that #4651 is merged.

Would love to see this merged. It pains me that this is dragging around since last August.

compiler/fixture_dirs.txt
```diff
@@ -36,7 +36,7 @@ crates/relay-docblock/tests/parse
 crates/relay-docblock/tests/to_schema
 crates/relay-lsp/tests/find_field_usages
 crates/relay-lsp/tests/hover
-crates/relay-schema-generation/tests/docblock
+crates/relay-lsp/tests/rename
```
captbaritone, line 39: Looks like a bad rebase?

tobias-tengler: Oops, yeah. One sec

compiler/crates/relay-lsp/src/rename/mod.rs
```diff
@@ -0,0 +1,616 @@
...
+                    fragment_name: fragment_definition.name.value,
+                    argument_name: variable_name,
+                }
+            } else {
+                // The variable not being defined in the fragment definition
+                // doens't mean it's not defined at all. There's also the
+                // possibility of it being a global operation variable.
+                // For now we just rename the identifier without modifying
+                // any other parts.
+                RenameKind::Identifier
```
captbaritone, line 346: This seems like it might be more confusing than helpful. If I click a global variable and select rename, and it only renames the thing I was clicking, I would probably be confused since I was expecting the global variable to get renamed. Should we just skip offering the refactor in this case?

tobias-tengler: Yes, I agree. I'll remove it.

compiler/crates/relay-lsp/src/rename/mod.rs
```diff
@@ -0,0 +1,616 @@
...
+                });
+        }
+
+        self.default_visit_fragment(fragment);
+
+        self.current_scope.fragment_name = None;
+    }
+
+    // For some reason the argument on @skip isn't visited,
+    // but the "argument variable" is visited through this...
```
captbaritone, line 546: I think conditionals are transformed in the IR into their own node type. So @skip and @include get transformed into their or IR node

tobias-tengler: Good to know. Would you recommend a different approach that what I've taken here?

tobias-tengler: Removed :)

compiler/crates/relay-lsp/src/rename/mod.rs
tobias-tengler: Removed :)

compiler/crates/relay-lsp/src/rename/mod.rs
captbaritone: I think this is fine.

captbaritone:
I've been able to get this working internally, which is rad!

However, I noticed one case which does not seem to work correctly:

```typescript
const UserProfile = graphql`
  fragment RelayReaderTestCreatesFragmentPointersWithVariableArgumentsUserProfile on User
  @argumentDefinitions(size: {type: "[Int]"}) {
    id
    ...RelayReaderTestCreatesFragmentPointersWithVariableArgumentsUserProfilePicture
      @arguments(size: $size)
  }
`;
```

When I try to rename the `$size` variable it only renames the definition, not it's value in the `@arguments` directive.

tobias-tengler:
I'm able to reproduce this with a test. Working on a fix :)

tobias-tengler:
@captbaritone Fixed!

captbaritone:
FYI the LSP crate has some other internal users which I will need to update to work with these changes. It's on my todo list for next week.

tobias-tengler:
@captbaritone This would also be awesome to have in today's release 🤞

tobias-tengler:
@captbaritone Were you able to address the internal users of the LSP crate? Is there anything I can do to get this merged?

tobias-tengler:
@captbaritone The one year anniversary of me first raising a PR for this is coming up, so would you do me the honors and get this through? 😇
If you can give me some hints what exactly is blocking internally, I could maybe come up with a solution that's easier to land as well!

captbaritone:
Sorry for the egregious delay. Still need to get our GitHub CI building reliably in order to get `@main` builds publishing to NPM. Currently it's hitting issues running out of disk space.

Issue #3538, opened 2021-07-09
TypeScript: Property 'invalidateStore' does not exist on type 'RecordSourceProxy'


lukebrandonfarrell:
When using `commitLocalUpdate` to invalidate the store the TS compiler throws and error: 

```
Property 'invalidateStore' does not exist on type 'RecordSourceProxy'
```

You can reproduce by writing the following in a TypeScript project:

```
commitLocalUpdate(relayEnvironment, (store : RecordSourceProxy) => {
      store.invalidateStore(); // Property 'invalidateStore' does not exist on type 'RecordSourceProxy'
 });
```

I believe that `RecordSourceProxy` needs to be updated in `updater: StoreUpdater` to `RecordSourceSelectorProxy`.

If someone points me in the right direction on where to make this change, I can make a PR.

maxprilutskiy:
A year after, the issue is still there. It's seems to be possible to invalidate store after mutation, but it isn't when using `commitLocalUpdate`
<img width="711" alt="Screenshot 2023-03-14 at 22 55 10" src="https://user-images.githubusercontent.com/5614659/225149850-9750d8bb-0a73-416f-b459-557994ee8d64.png">

vxm5091:
Would be super helpful to have this implemented or for there to be another documented alternative to invalidating the store outside of a mutation or subscription.

nossila:
This is already fixed by the typescript implementation here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/relay-runtime/lib/mutations/commitLocalUpdate.d.ts

And @vxm5091 you should be able to do something like this to invalidating the store:
```typescript
import { useRelayEnvironment } from 'react-relay/hooks'
import { commitLocalUpdate } from 'react-relay';

export const useInvalidateRelayStore = () => {
  const environment = useRelayEnvironment();
  return () => {
    commitLocalUpdate(environment, store => {
      store.invalidateStore()
    });
  };
}
```

I think we can close this issue in my opinion.

PR #4584, opened 2024-01-12
Fix commitLocalUpdate type

nossila:
The store received by `commitLocalUpdate` function is actually an instance of `RecordSourceSelectorProxy` and not `RecordSourceProxy` like we can see [here](https://github.com/facebook/relay/blob/d37a5aa/packages/relay-runtime/store/RelayPublishQueue.js#L414).

This fixed issue #3538

### DIFF
--- a/packages/relay-runtime/mutations/commitLocalUpdate.js
+++ b/packages/relay-runtime/mutations/commitLocalUpdate.js
@@ -11,11 +11,11 @@
 
 'use strict';
 
-import type {IEnvironment, StoreUpdater} from '../store/RelayStoreTypes';
+import type {IEnvironment, SelectorStoreUpdater} from '../store/RelayStoreTypes';
 
 function commitLocalUpdate(
   environment: IEnvironment,
-  updater: StoreUpdater,
+  updater: SelectorStoreUpdater,
 ): void {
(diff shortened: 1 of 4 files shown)


noghartt (commented):
cool

tobias-tengler:
You've changed the Flow type annotations. If you require the change in TypeScript as suggested by the title you need to make the change in the [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/relay-runtime/index.d.ts) repository for the relay-runtime package.

nossila:
@tobias-tengler thanks for the heads up!! I updated the implementation to follow the approach in [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/relay-runtime/lib/mutations/commitLocalUpdate.d.ts)

nossila:
@tobias-tengler you are right, I pointed the link to the wrong line. I tried fixing the `flow typecheck` but no success, so I'm closing this PR because I don't use flow daily and this issue doesn't really affect me since I'm using typescript.

For for typescript it was enough to use `SelectorStoreUpdater` on `commitLocalUpdate` but for flow don't seems to be enough.

PR #4588, opened 2024-01-19
Report errors if location provider fails

captbaritone:
Setting up the location provider in the most recent version of the VSCode extension is nearly impossible because if you don't do it perfectly (define the script using the right relative path, know which CWD the script will run in, use the right argument, parse it correctly, return the location in the correct format...) it simply fails silently with no hint as to what went worng.

Adding this logging will at least enable folks to debug what's happening in the Relay LSP Output panel, if they set the log level correctly.

The other option would be to simply return an error if there is a location provider provided and it errors trying to derive location info. I'm not sure how fallible we except these to be (I happen to know our Hack version has known issues where certain fields/types can't really get resolved).

Test Plan:

https://github.com/facebook/relay/assets/162735/c2e2c806-e6f1-4141-a450-a34774d766e6


orta:
Thanks, came to see if something like this existing in the source code!

PR #4596, opened 2024-01-23
fix(doc): update config name following update

eMerzh:
since https://github.com/facebook/relay/commit/4e68155d732b the custom scalar config has changed,
so let's update the doc


tobias-tengler:
Maybe it would also be helpful to update the Release notes with the new name in camelCase `customScalarTypes`. In my PR I only mentioned the Rust variable names, but for the developer those fields are `customScalars` and `customScalarTypes` and not `custom_scalars` and `custom_scalar_types`.

alunyov:
> @alunyov Maybe it would also be helpful to update the Release notes with the new name in camelCase `customScalarTypes`. In my PR I only mentioned the Rust variable names, but for the developer those fields are `customScalars` and `customScalarTypes` and not `custom_scalars` and `custom_scalar_types`.

Done!

PR #4607, opened 2024-01-31
Update queries-2.md

tpuric:
There was a typo. hovecard -> hovercard


noghartt (commented):
Good.

I think that you could bring this fix for all other versions of the documentation too.

captbaritone (changes requested):
The source of truth for this text is here: https://github.com/facebook/relay/blob/main/website/docs/tutorial/queries-2.md

The change here (in the most recent version) is helpful too, so you can leave that. You can also go back and make this change to previous versions if you want, but that's fully optional.

tpuric:
I'll collate this into a separate PR, thanks for the feedback.

PR #4608, opened 2024-01-31
chore(doc): improve relay compiler doc for persist

eMerzh:
because this config was undocumented just found it in the release note


eMerzh:
i'm also thinking about removing the ! exists if persist file does not exists instead creating it, what do you think?

noghartt:
Maybe we can improve on the type too?

https://github.com/facebook/relay/pull/4569

PR #4621, opened 2024-02-20
Validate edgeTypeName argument points to an existing type

tobias-tengler:
Currently you can input any string for the `edgeTypeName` argument on the `@prependNode` and `@appendNode` directives. 

This change enforces that the value you supply points to an existing object type within the schema.

### DIFF
--- a/compiler/crates/common/src/feature_flags.rs
+++ b/compiler/crates/common/src/feature_flags.rs
@@ -123,6 +123,10 @@ pub struct FeatureFlags {
     /// in @refetchable transform
     #[serde(default)]
     pub prefer_fetchable_in_refetch_queries: bool,
+
+    /// Disable validation of the `edgeTypeName` argument on `@prependNode` and `@appendNode`.
+    #[serde(default)]
+    pub disable_edge_type_name_validation_on_declerative_connection_directives: FeatureFlag,
 }
 
 fn default_as_true() -> bool {
--- a/compiler/crates/relay-transforms/src/declarative_connection.rs
+++ b/compiler/crates/relay-transforms/src/declarative_connection.rs
@@ -23,6 +23,8 @@ use graphql_ir::Transformer;
 use intern::string_key::Intern;
 use intern::string_key::StringKey;
 use lazy_static::lazy_static;
+use schema::suggestion_list::did_you_mean;
+use schema::suggestion_list::GraphQLSuggestions;
 use schema::SDLSchema;
 use schema::Schema;
 use schema::Type;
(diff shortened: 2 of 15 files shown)


compiler/crates/relay-transforms/src/declarative_connection.rs
```diff
@@ -273,6 +273,30 @@ impl Transformer for DeclarativeConnectionMutationTransform<'_> {
                     Some(connections_arg) => {
                         let edge_typename_arg = node_directive.arguments.named(*EDGE_TYPENAME_ARG);
                         if let Some(edge_typename_arg) = edge_typename_arg {
+                            if let Some(edge_typename_value) = edge_typename_arg
+                                .value
+                                .item
+                                .get_constant()
+                                .and_then(|c| c.get_string_literal())
```
captbaritone, line 280: Out of curiosity, what happens if you try to use a variable here? Does that trigger some other reasonable error message?

captbaritone:
So good! Thank you!

compiler/crates/relay-transforms/src/declarative_connection.rs
tobias-tengler: It doesn't error. Is using a variable considered an error? I've added a test case that demonstrates that it currently doesn't fail.

captbaritone:
Still working on landing this. It looks like it's triggering a number of errors internally. Trying to figure out if this is because we're catching real errors (and if so, how we can resolve them), or if there's something else going on here.

tobias-tengler:
Is this still blocked internally?

captbaritone:
Sorry, haven't been able to revisit. Adding to my todo list for next week.

captbaritone:
So we have many infractions in our codebaese today, and I don't see an obvious way to "fix" these issues at scale. Maybe we need a mechanism to selectively ignore this warning for existing infractions? Perhaps we could leverage our feature flag infra here?

Also, would be cool to suggest types with similar names in this error, I think we have examples of that elsewhere in the codebase.

compiler/crates/common/src/feature_flags.rs
```diff
@@ -108,6 +108,10 @@ pub struct FeatureFlags {
     /// Allow non-nullable return types from resolvers.
     #[serde(default)]
     pub allow_resolver_non_nullable_return_type: FeatureFlag,
+
+    /// Disable validation of the `edgeTypeName` argument on `@prependNode` and `@appendNode`.
+    #[serde(default)]
+    pub disable_edge_type_name_validation_on_declerative_connection_directives: bool,
```
captbaritone, line 114: Could we use `FeatureFlag` here instead and then use [`is_enabled_for`](https://github.com/facebook/relay/blob/fdd270b61bb1101e4bd0056bfa83add0b420d257/compiler/crates/common/src/feature_flags.rs#L131) on the typename passed? That would allow us to implement a ratchet internally where we could improve things over time.

tobias-tengler: Done :)

tobias-tengler:
@captbaritone I fixed the conflicts :)

tobias-tengler:
@captbaritone why was this reverted? Internal reasons?

captbaritone:
@tobias-tengler Yeah, internal reasons. Just some internal miscommunication with getting the feature flag set. I have an internal diff open to re-land.

PR #4629, opened 2024-02-26
[relay-lsp] Add go to definition support for type defined in variable definition

tobias-tengler:


captbaritone:
Nice! I wish we had tests for go-to-definition. I wonder if the tests we have for hover could be adapted...

tobias-tengler:
It's on my todo list 😅 I want to add tests in a separate PR, before   proceeding with the go to definition support within schema files...

PR #4631, opened 2024-02-27
[relay-transforms] Infer @refetchable queryName from fragment name

tobias-tengler:
This makes the `queryName` argument on the `@refetchable` directive nullable and if no value is specified, generates a query name in the format of `<FRAGMENT_NAME>RefetchQuery` for the refetch query.

The motivation is that I've never encountered a case where I or other developers at my current job wished to specify a name that doesn't follow this pattern. We've even made this format a requirement.

I think this change reduces boilerplate and still gives you the option to specify a name manually _if_ you wish to do so. The change is also backwards compatible.


captbaritone:
This idea has been raised before, and I believe the rational for the current behavior was that having the explicit query name in the fragment makes it easier to debug because the name of the query is always explicit. If you want to look for refetch queries, you can see its name right in the code.

That said, I agree it's annoying. What do you think about offering an LSP autofix if the name is omitted?

tobias-tengler:
That makes sense. I like the idea of the autofix (there are lots of other opportunities as well). I'll try to come up with something over the weekend 😅

PR #4632, opened 2024-02-28
fix: fix requestCache conflicts when have multiple ReactRelayQueryRenderer   

hannanstd:
Sometimes we have to have two or more `ReactRelayQueryRenderer` with different environments at the same time.
So the problem is that `requestCache` is used as a shared variable to hold both environments's caches. 
On the other hand, as you know each environment can have its `configName` while we creating it.
It makes sense that the `requestCacheKey` consists of this `configName` to isolate each environment from the others.
This PR does that. [see the added Unit Tests](https://github.com/facebook/relay/pull/4632/files#diff-63e24882a3c84e88a7fdfbe17e069ec8545880927653411eee5cfd0f9cdc2555R1213). 
Also, it can solve some issues that were reported before regarding the SSR.


captbaritone:
At this point we are not investing in improving the older higher-order-components APIs and are instead focused on the hooks-based APIs.

Do the hooks apis (`useLazyLoadQuery`, etc) allow you to get the behavior you want?

hannanstd:
@captbaritone 
Hello, Thank you for your answer. but it is not investing in improving. this problem/bug exists in the relay and many legacy projects use it.
This change would not cause any issues or break changes. it just adds the environment `configName` to the `cachekey` and can be reviewed quickly but it helps others.

PR #4636, opened 2024-03-01
Use DefaultValue for default value of InputValueDefinition

tobias-tengler:


tobias-tengler:
@captbaritone Next PR on the road to SchemaDocument resolution-path support :)

compiler/crates/schema/src/in_memory/mod.rs
```diff
@@ -1675,7 +1675,11 @@ impl InMemorySchema {
                     Ok(Argument {
                         name: ArgumentName(arg_def.name.value),
                         type_: self.build_input_object_reference(&arg_def.type_)?,
-                        default_value: arg_def.default_value.clone(),
+                        default_value: if let Some(default_value) = &arg_def.default_value {
+                            Some(default_value.value.clone())
+                        } else {
+                            None
+                        },
```
captbaritone, line 1682: Here (and the similar place in relay docblock) could you use `default_value.map(...)`? Might be a little cleaner.

compiler/crates/graphql-syntax/src/node/type_system.rs
```diff
@@ -515,7 +515,7 @@ impl fmt::Display for DirectiveLocation {
 pub struct InputValueDefinition {
     pub name: Identifier,
     pub type_: TypeAnnotation,
-    pub default_value: Option<ConstantValue>,
+    pub default_value: Option<DefaultValue>,
```
captbaritone, line 518: Is there more context on why it's useful to have a DefaultValue here?

If we want to do this, I think we should move where `DefaultValue` is defined, perhaps into `compiler/crates/graphql-syntax/src/node/constant_value.rs`? Right now it's defined with the executable definition AST constructs, so I don't think it makes sense to share it between both types of documents without moving it.

captbaritone (changes requested):

compiler/crates/graphql-syntax/src/node/type_system.rs
tobias-tengler: The `ConstantValue` doesn't have span information, so we can't determine whether we are on the default value of a `InputValueDefinition` in the path resolution. The `DefaultValue` has a span definition and made sense to me here, as it's the same concept as a default value for a `VariableDefinition`.

captbaritone: Cool. What do you think about moving the definition into constant_value.rs?

PR #4639, opened 2024-03-04
Implement SchemaDocument path resolution

tobias-tengler:


tobias-tengler:
@captbaritone This would be ready to review :)

captbaritone:
@tobias-tengler Sorry for the slow action here. There's an internal package which makes heavy use of these structs and I got a little stuck trying to figure out how to migrate that internal crate. Let me take another look...

PR #4647, opened 2024-03-12
Input fields with defaults shouldn't be required

obmarg:
I added a new non-null input field with a default to a schema I work on, and relay started throwing an error for that field:

```
[ERROR] Error: ✖︎ Missing required fields '["kind"]' of type 'AccessTokenCreateInput'

  app/routes/_dashboard/_personal-account/settings/-access-tokens/token-form.tsx:41:30
   40 │   ) {
   41 │     accessTokenCreate(input: { name: $name, accountId: $accountId }) {
      │                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   42 │       ... on AccessTokenCreateSuccess {
```

This seems to go against [the spec which says][1]:

> An input field is required if it has a non-null type and does not have
> a default value. Otherwise, the input object field is optional.

This PR updates the validation code to consider fields with default values as optional, in line with the spec.

I've added some tests for this, though I'm not familiar enough with relay to know whether I'm actually covering both of the cases I've fixed.  Please let me know if there's improvements to be made.

[1]: http://spec.graphql.org/October2021/#sec-Input-Object-Required-Fields

### DIFF
--- a/compiler/crates/graphql-ir/src/build.rs
+++ b/compiler/crates/graphql-ir/src/build.rs
@@ -1569,7 +1569,7 @@ impl<'schema, 'signatures, 'options> Builder<'schema, 'signatures, 'options> {
         let mut required_fields = type_definition
             .fields
             .iter()
-            .filter(|x| x.type_.is_non_null())
+            .filter(|x| x.type_.is_non_null() && x.default_value.is_none())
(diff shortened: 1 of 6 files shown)


captbaritone:
Thanks!

PR #4652, opened 2024-03-15
Activate LSP for schema and schema extension files

tobias-tengler:
This is in preparation for adding LSP support for schema and schema extension files.

### DIFF
--- a/compiler/crates/relay-lsp/src/server/lsp_state.rs
+++ b/compiler/crates/relay-lsp/src/server/lsp_state.rs
@@ -37,6 +37,7 @@ use lsp_types::Url;
 use relay_compiler::config::Config;
 use relay_compiler::get_parser_features;
 use relay_compiler::FileCategorizer;
+use relay_compiler::FileGroup;
 use relay_compiler::ProjectName;
 use relay_docblock::parse_docblock_ast;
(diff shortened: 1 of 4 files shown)


captbaritone:
Awesome! I'm curious how/if you were able to verify these changes. This LSP stuff can be hard to test (especially when they require changes to the VSCode extension as well)

tobias-tengler:
Yes, I was able to verify it. I started the extension development host and launched my lsp playground repository, which is configured to point to the locally built relay binary. I then observed the extension displayed the `Relay: ready` message, when I opened the configured `schema.graphql` file.

compiler/crates/relay-lsp/src/server/lsp_state.rs
```diff
@@ -521,12 +530,34 @@ impl<TPerfLogger: PerfLogger + 'static, TSchemaDocumentation: SchemaDocumentatio
             js_server.process_js_source(uri, text);
         }
```
captbaritone, line 531: Should this block live in the FileGroup::Source match case, since it's js-specific?

captbaritone (changes requested):
Looks good! Just one qustion

tobias-tengler:
Addressed :)

Issue #4654, opened 2024-03-20
Unhandled JS Exception: Cannot read property 'providedVariables' of undefined


vestrelatlassian:
Hello, my team is attempting to integrate Relay into a React Native app we are building. Everything's working great until we try using `usePaginationFragment`...

Loading the preloaded query works perfectly fine. Data is fetched from our Relay-compliant GraphQL server and displayed in the app. The error occurs when calling `loadNext`...

|                                                                                                 |                                                                                                                                 |
|-------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------|
| ![wtf](https://github.com/facebook/relay/assets/164123801/1ae8cbbf-91f6-475d-a1fe-3d2783a3c5e0) | ![Screenshot 2024-03-19 at 5 33 13 PM](https://github.com/facebook/relay/assets/164123801/eca919b3-6be1-4c76-a4a9-e3252190e653) |

I have spent an entire day looking through the official Relay docs and examples and googling around for answers. No luck. Any ideas?

---------

Here's the code I'm playing around with. It's not production code so don't judge lol.

```tsx
export const PopularFeedPaginationQuery = graphql`
  query PopularFeedPaginationQuery {
    ...PopularFeedPaginationFragment
  }
`;

export const PopularFeedPaginationFragment = graphql`
  fragment PopularFeedPaginationFragment on Query
  @argumentDefinitions(first: { type: "Int", defaultValue: 2 }, after: { type: "String" })
  @refetchable(queryName: "PopularFeedPaginationFragment") {
    popularFeed(first: $first, after: $after)
      @connection(key: "PopularFeedPaginationFragment_popularFeed") {
      edges {
        node {
          content {
            id
            title
          }
        }
      }
    }
  }
`;

export function PaginatedPopularFeed({
  queryRef
} : {
  queryRef: PreloadedQuery<PopularFeedPaginationQueryType>;
}): ReactElement {
  const queryData = usePreloadedQuery<PopularFeedPaginationQueryType>(PopularFeedPaginationQuery, queryRef);

  const {
    data, hasNext, loadNext
  } = usePaginationFragment<PopularFeedPaginationQueryType, PopularFeedPaginationFragment$key>(
    PopularFeedPaginationFragment, queryData
  );

  function loadNextPage() {
    loadNext(2);
  }

  const contentList = data.popularFeed.edges
    ?.flatMap(edge => (edge?.node?.content ? [edge.node.content] : [])) ?? [];

[...]

PR #4663, opened 2024-04-04
Prevent name collision between derived refetchable query and fragment name

captbaritone:
As identified in #4654 it's possible to use an existing fragment name as the `queryName` argument of `@refetchable` and end up with a collision. The result is that we end up importing the fragment's artifact when we expect to import the queries. It's possible we even overwrite that same path with two artifacts.

~One thing I noticed is that for regular queries we require that the name end with `Query` but it looks like we don't for queries generated with `queryName`. Perhaps we need to enforce the same convention with these queries?~

This PR now includes additional validations to prevent this error.


compiler/crates/relay-compiler/tests/compile_relay_artifacts/fixtures/connection-name-matches-fragment.invalid.expected
```diff
@@ -0,0 +1,630 @@
...
+          "cursor": "after"
+        },
+        "backward": null,
+        "path": [
+          "node",
+          "comments"
+        ]
+      },
+      "fragmentPathInResult": [],
+      "operation": require('connectionNameMatchesFragmentFragment.graphql')
```
captbaritone, line 512: This is the problem. We meant to import the generated operation's artifact, but we accidentally got the fragment's.

PR #4670, opened 2024-04-05
Add SchemaDocument find references LSP support

tobias-tengler:


tobias-tengler:
@captbaritone This would now also be ready.
If you cut a release today, please don't forget to also release the VS Code extensions, so the LSP is activated for `.graphql` files. I've already bumped the version in https://github.com/facebook/relay/commit/964e078395a8b5ef2d86eea33d524cda375e3572, so you just need to run the action: https://github.com/facebook/relay/actions/workflows/vscode.yml

PR #4671, opened 2024-04-09
Add Techjobs.be as a user

evaetj:
We've been happy relay users for the last year. Pull to add us to the user list :)


philipheinser:
@evaetj For what product are you using relay, your website and app does not seam to use it?

evaetj:
We have been it in our internal back-end tools (to support us in automating things) and will in the near future also be using it in our public website

PR #4680, opened 2024-04-21
Add code action to create missing operation variable

tobias-tengler:
Depends on #4679

This extends the `ExpectedOperationVariableToBeDefined` error to include the variable name and type in the diagnostic data. Using this data we create a code action that appends the missing variable with the correct type to the operation variables.

### DIFF
--- a/compiler/crates/graphql-ir/src/errors.rs
+++ b/compiler/crates/graphql-ir/src/errors.rs
@@ -149,9 +149,6 @@ pub enum ValidationMessage {
         next_type: String,
     },
 
-    #[error("Expected variable `${0}` to be defined on the operation")]
-    ExpectedOperationVariableToBeDefined(VariableName),
-
     #[error(
         "Expected argument definition to have an input type (scalar, enum, or input object), found type '{0}'"
     )]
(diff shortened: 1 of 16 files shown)


compiler/crates/relay-lsp/src/code_action.rs
```diff
@@ -244,53 +282,3 @@ fn create_code_action(
         ..Default::default()
     })
 }
-
-#[cfg(test)]
-mod tests {
```
tobias-tengler, line 249: TODO: Re-add this test and tests for the new functionality

compiler/crates/graphql-ir/src/errors.rs
```diff
@@ -539,6 +536,13 @@ pub enum ValidationMessage {
     ResolverInMutation,
 }
 
+#[derive(Clone, Debug)]
+pub struct ValidationDiagnosticCode;
+
+impl ValidationDiagnosticCode {
+    pub const EXPECTED_OPERATION_VARIABLE_TO_BE_DEFINED: i32 = 1;
```
tobias-tengler, line 543: I'm not a fan of this being just an integer. Since the diagnostic code is also shown in the LSP client, ideally it would be something like `relay001`. But having a string here will likely mean that we can no longer pattern match. Maybe there's something like a macro we can utilize on the errors enum to define a diagnostic code there and we can easily reference it and it gets properly formatted when sent to the client?

captbaritone:
Looks like this is still in draft state. Let me know if/when it's ready for review.

tobias-tengler:
I want to take a different approach. Closing for now.

PR #4681, opened 2024-04-21
Add code action to create missing fragment argument

tobias-tengler:
Depends on #4680

This renames the `GlobalVariables` error to `ExpectedOperationVariableToBeDefined` and emits the error for each problematic variable identifier, instead of a single error for multiple infringements. Also extends the error to include the variable name and type in the diagnostic data. Using this data we create a code action that appends the missing argument with the correct type to the fragment arguments (only using `@argumentDefinitions` for now).
No diagnostic is emitted for orphan fragments that aren't referenced in an operation. I plan on supporting this case and add a refactor for global variables in a later PR.

### DIFF
--- a/compiler/crates/common/src/diagnostic.rs
+++ b/compiler/crates/common/src/diagnostic.rs
@@ -12,6 +12,7 @@ use std::fmt::Write;
 
 use lsp_types::DiagnosticSeverity;
 use lsp_types::DiagnosticTag;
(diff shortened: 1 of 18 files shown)


captbaritone:
Looks like this is still in draft state. Let me know if/when it's ready for review.

tobias-tengler:
I want to take a different approach. Closing for now.

Issue #4226, opened 2023-02-20
codegen hash values are different


kmin-283:
### fragment
```
fragment SellingProductListFragmentContainer_store_representActiveProducts on Store
      @argumentDefinitions(
        count: { type: "Int", defaultValue: 10 }
        cursor: { type: "ID" }
        filter: { type: "ProductFilter", defaultValue: { statuses: [ACTIVE], representStatus: ACTIVE } }
      )
      @refetchable(queryName: "SellingProductListFragmentContatinerRepresentActiveProducts") {
        representActiveProducts: products(first: $count, after: $cursor, filter: $filter)
          @connection(key: "SellingProductListFragmentContainer_store_representActiveProducts") {
          edges {
            node {
              _id
            }
          }
        }
      }
```
When the relay-compiler command is executed, generated the hash value is

<img width="437" alt="Screenshot 2023-02-20 at 4 44 26 PM" src="https://user-images.githubusercontent.com/63279334/220043493-4a09ed5d-1ac8-4c63-866d-1bf691f5398f.png">

but when i execute same fragment in this code it return different hash value
```
const { parse, print } = require('graphql')
const crypto = require('node:crypto')

const gql = String.raw
const text = gql`
  fragment SellingProductListFragmentContainer_store_representActiveProducts on Store
  @argumentDefinitions(
    count: { type: "Int", defaultValue: 10 }
    cursor: { type: "ID" }
    filter: { type: "ProductFilter", defaultValue: { statuses: [ACTIVE], representStatus: ACTIVE } }
  )
  @refetchable(queryName: "SellingProductListFragmentContatinerRepresentActiveProducts") {
    representActiveProducts: products(first: $count, after: $cursor, filter: $filter)
      @connection(key: "SellingProductListFragmentContainer_store_representActiveProducts") {
      edges {
        node {
          _id
        }
      }
    }
  }
`

const ast = parse(text)
const definition = ast.definitions[0]
const printed = print(definition)
const hash = crypto.createHash('md5').update(printed, 'utf8').digest('hex')

console.log(hash)
``` 
the hash value is
```
296c46ae6772194c8404c631edebccbf
```

I put the same input into the md5 hash function, but why are the results different?

cometkim:
Hey @kmin-283, I recently received a report of the same issue from https://github.com/cometkim/vite-plugin-relay-lite/issues/53

And found https://github.com/graphql/graphql-js/pull/2797 from graphql-js' v15.4.0 release.

[...]

PR #4692, opened 2024-05-04
inline graphql-js' printer

cometkim:
See #3628 and #4226


captbaritone:
This seems sensible to me. Feel free to ping me when this diff is ready for review (I see it's still marked as Draft).

captbaritone:
I wonder if there's a sensible way to add some tests which assert that the compiler and this code arrive at the same hash?

cometkim:
That's is easy if we can run the printer standalone. Or, it would be ideal to publish to NPM and directly use its WASM or NAPI distribution

cometkim:
Workaround: Commit md5 hashes for all the fixtures in the graphql-text-printer crate test suite. Then we can compare it when running the babel plugin tests.

cometkim:
@captbaritone I added idempotency testing using compiler-side fixtures. Fixed 4 failures related to spaces between multiple definitions and it all passes.

packages/babel-plugin-relay/printGraphQL.js
```diff
@@ -0,0 +1,184 @@
...
+ */
+
+const { visit } = require('graphql');
+
+/**
+ * Converts an AST into a string, using one set of reasonable
+ * formatting rules.
+ */
+function print(ast: any): string {
+  return visit(ast, { leave: printDocASTReducer });
```
captbaritone, line 21: Something weird here. By version 16.8.1 which we are supposed to be using, this had changed and there wasn't a top-level `leave` here anymore, each child visitor had a `leave` key.

I think there's some weird Yarn issues, becuase if I run `yarn` in this directory, I get a local `node_modules` directory with `graphql@16.8.1` which causes this to fail with many failures that look like this:

```
  ● printGraphQL › tests printer idempotence: conditions.expected

    TypeError: visitFn.call is not a function

      19 |  */
      20 | function print(ast: any): string {
    > 21 |   return visit(ast, { leave: printDocASTReducer });
         |                             ^
      22 | }
      23 |
      24 | const printDocASTReducer: any = {

      at Object.visit (packages/babel-plugin-relay/node_modules/graphql/language/visitor.js:197:15)
      at print (packages/babel-plugin-relay/printGraphQL.js:21:29)
      at packages/babel-plugin-relay/__tests__/printGraphQL-test.js:27:38
```

I think our test runner does not perform that yarn install for some reason. I'm honestly not even sure why this package has its own yarn.lock.

captbaritone (changes requested):
I'm not able to land this internally due to the issue with yarn.lock (see my inline comment).

Do you mind looking into it? Maybe you can do some git blaming to figure out why we even have that yarn.lock in the first place?

cometkim:
I think it makes more sense for each package to have a lockfile since there is no workspaces setup, or gitignoring. `yarn install` is needed to run the test suite for `react-relay`, but it doesn't seem to be running anywhere?

Anyway, I'd suggest clearing it and activating the workspaces, it would improve your CI cache hits a lot.

captbaritone:
@cometkim Yeah, I could have sworn we had a workspace setup. Here's where we do the global yarn install. I guess we'll need to either setup the workspace or manually run install for this package.

https://github.com/facebook/relay/blob/main/.github/workflows/ci.yml#L45

cometkim:
setup here? or another pr?

captbaritone:
Either way. Here might be easier since we can use CI to validate that it worked. Whatever you prefer.

cometkim:
https://github.com/facebook/relay/pull/4711

I'm not sure it could actually boost the CI, let's see.

captbaritone:
Checking back here. Are we stuck on this Yarn issue?

cometkim:
Not really. It is a result of prioritizing another task according to what you mentioned:

> Maybe you can do some git blaming to figure out why we even have that yarn.lock in the first place?

PR #4693, opened 2024-05-07
Type useLazyLoadQuery with hook syntax

alexmckenley:
Differential Revision: D57076552


captbaritone:
Looks like this landed. Not sure why the PR didn't close itself.

PR #4696, opened 2024-05-13
Improve website build+deploy

zpao:
I got distracted looking at something else for Relay and noticed the repo itself is several GB. This is almost certainly primarily from the `gh-pages` branch. It currently gets rebuilt on every commit to `main` and since the content of the pages include the last modified by & time using that commit data, every single generated page gets rebuilt as well. For example, https://github.com/facebook/relay/commit/79a357573cfc5fc2ce8de15ba133630f3edac8d1 updated 4672 files. That build job took almost 30 minutes (https://github.com/facebook/relay/actions/runs/9067652200)

So I did a few things
1. Removed the modified by stuff. It's not useful and honestly potentially misleading.
2. Updated actions to latest versions. This alone shaved off a bunch of time
3. Persisted the docusaurus cache.

Warm build (typical) time is now 6 minutes. https://github.com/zpao/relay/actions/runs/9052335508

### DIFF
--- a/.github/workflows/docusaurus.yml
+++ b/.github/workflows/docusaurus.yml
@@ -42,6 +42,22 @@ jobs:
         with:
           persist-credentials: false
 
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+          cache: 'yarn'
+          cache-dependency-path: website/yarn.lock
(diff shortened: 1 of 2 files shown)


captbaritone:
This is awesome. Thank you!

PR #4697, opened 2024-05-13
Add code action to rename fragment

tobias-tengler:
At the moment there's a code action that suggests correct operation names, if you hover over an operation name or the operation name is associated with a diagnostic.
This PR adds the same thing for fragment definition names.


tobias-tengler:
In the future, once https://github.com/facebook/relay/pull/4581 is merged (I'm still hopeful), i'd like to extend the code action to perform an actual rename for the fragment. 
Really looking forward to both of these PRs landing. It's incredibly frustrating to rename components / fragments currently.

Issue #4000, opened 2022-07-05
Provided variables and Live Resolver modules should allow injected dependencies


leoasis:
Looking at [how Provided variables modules should be defined](https://relay.dev/docs/next/api-reference/graphql-and-directives/#provided-variables), and [how Live Resolvers modules are defined](https://github.com/captbaritone/redux-to-relay-with-live-resolvers-example/blob/step-1/src/relay/resolvers/RootAllTodosResolver.js), the way it is currently expected to access external services (like the Redux store, or a module to ask for feature flags), is by using singletons at the module level.

This is not compatible with Node.js environments (when doing SSR for example), where the modules are loaded once and multiple requests are handled sharing the same module state.

It would be ideal if the signature for provided variables modules and live resolvers allowed to get access to a "context" object that could be injected in the Relay Environment, so that we could put these services to be accessed by the modules, instead of accessing them as singletons. Something like this:

```
const store = ...
const featureFlags = ...
const environment = new Environment({
  moduleContext: { store, featureFlags } // name TBD
});

return <ReduxProvider store={store}>
  <RelayEnvironment environment={environment}>
    <App />
  </RelayEnvironment>
</ReduxProvider>
```

And they would be consumed like this:

```
// a Provided Variable
export default {
  get(moduleContext): boolean {
    return moduleContext.featureFlags.check('todo_should_include_timestamp');
  },
};
```

```
// a Live Resolver
export default function RootAllTodosResolver(moduleContext) {
  return selectLiveState(moduleContext.store, (state) => {
    return state.todos.map((todo) => todo.id);
  });
}
```

Would this be something that would be acceptable for contribution? Thank you in advance for reading this!

captbaritone:
Thanks for writing this up! Overall it looks right to me. Some concerns to track:

1. We need to convey the fact that these context values are not expected to change overtime. There will not be a mechanism to have dynamic values in this context (unless you are tracking them yourself like with a live resolver).
2. There's not really any way for these values to be typesafe. We can't statically know which environment a given resolver/variable provider will be used with. This is the same limitation that React Context has, so I think it's acceptable, but it's worth considering. How will we type these context values?
3. How will we thread this context value though?

I think a good next step would be a lightweight proposal for how we can achieve #3. This could be a rough PR, or even a document.

Markionium:
Hiya @captbaritone and @leoasis,

I took an attempt at implementing the approach discussed above in this draft #4704.

[...]

PR #4704, opened 2024-06-04
Add context to relay resolvers

Markionium:
This is a draft PR to implement an approach mentioned in #4000.

I've currently put the `resolverContext` on the `Store` instead of the `Environment`.  It seems that the `Store` might be a better place than the `Environment`?

If defined on the environment the context would need to go from the environment into the store down to the reader? Since a store could be used by different environments if the resolverContext would be on the environment that might cause more issues than it being on the store? 

The following is an example of how this would be used. 

One would provide the context as follows.

```ts
const store = new LiveResolverStore(source, {
    resolverContext: {
      counter: createObservableCounter(),
      world: "World!"
    },
  });
  ```

Then the resolvers could use the context similar to `readFragment`

```ts
/**
 * @RelayResolver Query.hello_context: String
 *
 * Say `Hello, ${world}!`
 */
import {resolverContext} from '../../ResolverFragments';

function hello_context(): string {
  const world = resolverContext<{world: string}>().world;
  return `Hello, ${world}!`;
}
```

```ts
/**
 * @RelayResolver Query.hello_world_with_context: String
 * @live
 *
 * Say `Hello ${world}!`
 */
function hello_world_with_context(): LiveState<string> {
  const dependency = resolverContext<{greeting: string}>();

  return {
    read() {
      return `Hello ${dependency.greeting}!`;
    },

    subscribe(callback) {
      return () => {
        // no-op
      };
    },
  };
}
```


captbaritone:
Hey @Markionium thanks for picking this up!

I agree that attaching the context to the store is the right choice.

However, rather than accessing the context via a magic function, my preference would be that the context object get passed in as the third argument to the resolver function. This would keep alignment with a traditional `graphql-js` style resolver function.

```ts
/**
 * @RelayResolver User.bestFriend: User
 */
export function myResolver(user, _, ctx) {
    return ctx.userById(user.bestFriendId);
}
```

In terms of type safety, I think we can allow users to provide a module/export name in the compiler config for the project. We can then include that type in our generated type assertions (note that these are not yet built for TypeScript, but it should be pretty easy to add if you are interested in another project).

You still need to ensure your compiler config and the store you create actually match, but that's something you can get right once and in one place so probably an okay end state.

Beyond that we are also exploring a more light-weight way to define resolvers using an approach similar to Grats. In that world it could be even simpler, and work similar to how it works in Grats (see [the docs here](https://grats.capt.dev/docs/docblock-tags/context)).

captbaritone:
Connecting the dots. @alloy and I have been discussing this a bit more on this Gist: https://gist.github.com/alloy/f0c9c90ff7a28f3b17850021488979d5

captbaritone:
Looks like this is getting close, but still in draft. Give me a ping when this is ready for review! Excited to see this!

drewatk:
Hey @captbaritone! We're ready for your review here. @Markionium can you please un-mark as draft?

compiler/crates/relay-typegen/src/visit.rs
```diff
@@ -588,6 +625,35 @@ fn import_relay_resolver_function_type(
...
+                Some(LiveResolverContextType {
+                    name: context_import.name.clone().intern(),
+                    import_path: context_import.package.clone().intern(),
+                })
+            }
+            None => None,
+        }
+    } else {
+        None
+    };
```
Markionium, line 655: Since only the `import_path` is different we could do something like this (Note I didn't run this)

I think it doesn't matter too much but it might make it a bit clearer.

```suggestion
    let context_import = if resolver_metadata.live {
        match &typegen_context
            .project_config
            .typegen_config
            .live_resolver_context_type
        {
            Some(context_type) => {
                let import_path = match context_type {
                    LiveResolverContextTypeInput::Path(context_import) => {
                        typegen_context.project_config.js_module_import_identifier(
                            &typegen_context.project_config.artifact_path_for_definition(
                                typegen_context.definition_source_location,
                            ),
                            &PathBuf::from(&context_import.path),
                        )
                    }
                    LiveResolverContextTypeInput::Package(context_import) => {
                        context_import.package.clone().intern()
                    }
                };

                Some(LiveResolverContextType {
                    name: context_type.name.clone().intern(),
                    import_path,
                })
            },
            None => None,
        }
    } else {
        None
    };
```

compiler/crates/relay-typegen/src/visit.rs
```diff
@@ -473,7 +477,13 @@ fn get_resolver_arguments(
     encountered_enums: &mut EncounteredEnums,
     custom_scalars: &mut std::collections::HashSet<(StringKey, PathBuf)>,
     schema_field: &Field,
+    context_import: Option<LiveResolverContextType>,
 ) -> Vec<KeyValuePairProp> {
+    let void_type = match typegen_context.project_config.typegen_config.language {
+        TypegenLanguage::Flow | TypegenLanguage::JavaScript => AST::RawType(intern!("void")),
+        TypegenLanguage::TypeScript => AST::RawType(intern!("undefined")),
+    };
+
```
Markionium, line 486: In javascript, we probably don't want to show any types at all? Or we could have TS docblock types, but in that case perhaps it would make more sense to align JS with TS?

drewatk: Good catch. This might be caught later in the writer for JS also just by never writing any types but I'm not sure about that.

packages/react-relay/__tests__/LiveResolvers-test.js
```diff
@@ -1711,3 +1711,136 @@ test('provided variables and resolvers', () => {
...
+    }
+  `;
+
+  const operation = createOperationDescriptor(FooQuery, {});
+  const environment = new RelayModernEnvironment({
+    network: RelayNetwork.create(jest.fn()),
+    store: new LiveResolverStore(RelayRecordSource.create(), {
+      gcReleaseBufferSize: 0,
+      liveResolverContext: {
+        greeting: {myHello: 'Hello Allemaal!'},
```
alloy, line 1728: @Markionium I zie you 😄

packages/relay-runtime/store/RelayReader.js
```diff
@@ -1415,17 +1424,29 @@ function getResolverValue(
...
   let resolverResult = null;
   let resolverError = null;
+
+  const shouldPassContextToLiveResolver = field.kind === 'RelayLiveResolver';
   try {
     const resolverFunctionArgs = [];
     if (field.fragment != null) {
       resolverFunctionArgs.push(fragmentKey);
+    } else {
+      // Set first argument to `null` in case we have resolver context
```
alloy, line 1434: Nit: The code is pushing `undefined`, not `null. What gives?

packages/relay-runtime/store/__tests__/resolvers/HelloWorldResolverWithContext.js
```diff
@@ -0,0 +1,43 @@
...
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict-local
+ * @format
+ * @oncall relay
+ */
+
+'use strict';
+
+import type {LiveResolverContextType} from '../../../mutations/__tests__/LiveResolverContextType';
```
alloy, line 14: Nit: For clarity sake, I think it would be good to rename this type to self-document that it's a test fixture representing the host's context typing, not a generic Relay provided context typing.

scripts/config.tests.json
```diff
@@ -57,7 +57,8 @@
         "relay_resolver_enable_interface_output_type": { "kind": "enabled" }
       },
       "language": "flow",
-      "experimentalEmitSemanticNullabilityTypes": true
+      "experimentalEmitSemanticNullabilityTypes": true,
+      "liveResolverContextType": { "name": "LiveResolverContextType", "path": "packages/relay-runtime/mutations/__tests__/LiveResolverContextType" }
```
alloy, line 61: Do we need another project that exercises the `package` module specifier variant?

yarn-error.log
```diff
@@ -0,0 +1,7442 @@
+Arguments:
```
alloy, line 1: Whoopsie, accidental check-in

compiler/crates/relay-typegen/src/write.rs
```diff
@@ -178,7 +178,10 @@ pub(crate) fn write_operation_type_exports_section(
     write_import_actor_change_point(actor_change_status, writer)?;
     runtime_imports.write_runtime_imports(writer)?;
     write_fragment_imports(typegen_context, None, encountered_fragments, writer)?;
-    write_relay_resolver_imports(imported_resolvers, writer)?;
+    // TODO: Add proper support for Resolver type generation in typescript
```
alloy, line 181: This is follow-up work for us, right? Can you link a WI/issue?

compiler/crates/relay-typegen/src/write.rs
```diff
@@ -449,7 +452,11 @@ pub(crate) fn write_fragment_type_exports_section(
     write_custom_scalar_imports(custom_scalars, writer)?;
 
     runtime_imports.write_runtime_imports(writer)?;
-    write_relay_resolver_imports(imported_resolvers, writer)?;
+
+    // TODO: Add proper support for Resolver type generation in typescript
```
alloy, line 456: Ditto

packages/react-relay/__tests__/__generated__/LiveResolversTestCounterContextQuery.graphql.js
```diff
@@ -0,0 +1,104 @@
...
+
+'use strict';
+
+/*::
+import type { ClientRequest, ClientQuery } from 'relay-runtime';
+import type { LiveState } from "relay-runtime";
+import {counter_context as queryCounterContextResolverType} from "../../../relay-runtime/store/__tests__/resolvers/LiveCounterContextResolver.js";
+import type { LiveResolverContextType } from "../../../relay-runtime/mutations/__tests__/LiveResolverContextType";
+// Type assertion validating that `queryCounterContextResolverType` resolver is correctly implemented.
+// A type error here indicates that the type signature of the resolver module is incorrect.
```
alloy, line 25: For those folks that check the artefacts into SCM, I wonder if this will take up too much space vs the benefit of the inline documentation?

packages/relay-runtime/store/RelayReader.js
drewatk: I forgot to change the comment, but I changed it to `undefined` to keep consistent with the `: undefined` we will push for the args value

scripts/config.tests.json
drewatk: Possibly? As far as I could find there was only one project config for all tests

packages/react-relay/__tests__/__generated__/LiveResolversTestCounterContextQuery.graphql.js
drewatk: FWIW, the assertion and these two comment lines predate our changes.

compiler/crates/relay-typegen/src/visit.rs
drewatk: I don't think this work because in the `Some(LiveResolverContextType {` I get an error `no field `name` on type `&LiveResolverContextTypeInput` because at that point the type isn't specific enough to know that both types share the `name` field

compiler/crates/relay-typegen/src/visit.rs
captbaritone: Would be good to add a test which confirms what happens here. Could be an integration test or a typegen test

compiler/crates/relay-typegen/src/visit.rs
```diff
@@ -529,6 +539,13 @@ fn get_resolver_arguments(
                 optional: false,
             });
         }
+    } else if resolver_metadata.live {
+        resolver_arguments.push(KeyValuePairProp {
+            key: "key".intern(),
+            value: AST::RawType("void".intern()),
```
captbaritone, line 545: Preferable to use the `intern!("void")` macro here and above. That creates the string once statically rather than needing to intern it each time this branch is run.

compiler/crates/relay-typegen/src/visit.rs
```diff
@@ -529,6 +539,13 @@ fn get_resolver_arguments(
                 optional: false,
             });
         }
+    } else if resolver_metadata.live {
```
captbaritone, line 542: I think it's possible to have a resolver that:

1. Does not define a rootFragment
2. Is not `@live`
3. Reads arguments

In this case, we probably still need some placeholder parent object/key argument.

So maybe we just do `else` here instead of `else if resolver_metadata.live`.

Either way, let's make sure we have a test case for that combination.

compiler/crates/relay-typegen/src/visit.rs
```diff
@@ -546,14 +563,34 @@ fn get_resolver_arguments(
...
     }
+
     if !args.is_empty() {
         resolver_arguments.push(KeyValuePairProp {
             key: "args".intern(),
             value: AST::ExactObject(ExactObject::new(args)),
             read_only: true,
             optional: false,
         });
+    } else if resolver_metadata.live {
```
captbaritone, line 574: Same as above, I think this probably applied even to non-live resolvers?

compiler/crates/relay-typegen/src/visit.rs
```diff
@@ -546,14 +563,34 @@ fn get_resolver_arguments(
...
+        resolver_arguments.push(KeyValuePairProp {
+            key: "args".intern(),
+            value: void_type,
+            read_only: true,
+            optional: false,
+        });
     }
+
+    if resolver_metadata.live {
```
captbaritone, line 583: Same as above, I think this probably applied even to non-live resolvers?

compiler/crates/relay-typegen/src/visit.rs
```diff
@@ -588,6 +625,35 @@ fn import_relay_resolver_function_type(
         &PathBuf::from(resolver_metadata.import_path.lookup()),
     );
 
+    let context_import = if resolver_metadata.live {
```
captbaritone, line 628: Even non-live resolvers should be able to read from context I think. I could imagine a purely derived resolver which wants to look into context to see something like the user's configured language or if they are in dark mode.

compiler/crates/relay-typegen/src/visit.rs
```diff
@@ -588,6 +625,35 @@ fn import_relay_resolver_function_type(
...
+        match &typegen_context
```
captbaritone, line 629: Optional: Might be cleaner to model this match as a `.map` since the Some case always returns a Some and the None case always returns None.

packages/relay-runtime/store/RelayReader.js
```diff
@@ -1415,17 +1424,29 @@ function getResolverValue(
 
   let resolverResult = null;
   let resolverError = null;
+
+  const shouldPassContextToLiveResolver = field.kind === 'RelayLiveResolver';
```
captbaritone, line 1428: As stated above, I think we can unconditionally pass context to resolvers. Does not need to be just live resolvers.

packages/relay-runtime/store/RelayStoreTypes.js
```diff
@@ -1329,3 +1329,12 @@ export type LiveState<+T> = {
...
 };
+
+/**
+ * Context that will be provided to live resolvers if
+ * `liveResolverContext` is set on the Relay Store.
+ * This context will be passed as the third argument to the live resolver
+ */
+export type LiveResolverContext = {
+  [key: string]: mixed,
+};
```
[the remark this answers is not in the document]
captbaritone: Agreed. I think context should be able to be whatever value the user wants to provide. I don't think Relay needs to have an opinion. A common case might be a Redux store.

compiler/crates/relay-config/src/typegen_config.rs
```diff
@@ -58,6 +58,25 @@ pub struct CustomScalarTypeImport {
...
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(untagged)]
+pub enum LiveResolverContextTypeInput {
+    Path(LiveResolverContextTypeInputPath),
+    Package(LiveResolverContextTypeInputPackage),
+}
+
+#[derive(Default, Debug, Serialize, Deserialize, Clone)]
+pub struct LiveResolverContextTypeInputPath {
+    pub name: String,
```
captbaritone, line 70: Can we use `StringKey` here to avoid having to intern the string each time we use it?

compiler/crates/relay-config/src/typegen_config.rs
```diff
@@ -58,6 +58,25 @@ pub struct CustomScalarTypeImport {
...
+    pub path: PathBuf,
+}
+
+#[derive(Default, Debug, Serialize, Deserialize, Clone)]
+pub struct LiveResolverContextTypeInputPackage {
+    pub name: String,
+    pub package: String,
```
captbaritone, line 77: Can we use `StringKey` in these to places as well?

captbaritone (changes requested):
This is looking so good! Amazing work so far and very close I think. Main bit of feedback (spread across the PR) is that I think _all_ resolvers should get context, not just `@live` resolvers.

compiler/crates/relay-typegen/src/write.rs
drewatk: #4772 Added

compiler/crates/relay-typegen/src/write.rs
drewatk: #4772 Added

drewatk:
@captbaritone @Markionium @alloy Ready for another look here. Addressed all comments and added some fixture tests as well. All rust tests are passing on my local, seems to be failing on `ubuntu` but not `mac` in CI.

drewatk:
@captbaritone Gentle Ping :)

compiler/crates/relay-typegen/src/typegen_state.rs
```diff
@@ -136,6 +136,13 @@ pub(crate) struct ImportedResolver {
     pub resolver_name: ImportedResolverName,
     pub resolver_type: AST,
     pub import_path: StringKey,
+    pub context_import: Option<LiveResolverContextType>,
```
captbaritone, line 139:
```suggestion
    pub context_import: Option<ResolverContextType>,
```

compiler/crates/relay-typegen/src/typegen_state.rs
```diff
@@ -136,6 +136,13 @@ pub(crate) struct ImportedResolver {
...
+}
+
+#[derive(Clone, Copy)]
+pub(crate) struct LiveResolverContextType {
```
captbaritone, line 143:
```suggestion
pub(crate) struct ResolverContextType {
```

packages/relay-runtime/store/RelayStoreSubscriptions.js
```diff
@@ -164,7 +172,12 @@ class RelayStoreSubscriptions implements StoreSubscriptions {
     }
     let nextSnapshot: Snapshot =
       hasOverlappingUpdates || !backup
-        ? RelayReader.read(source, snapshot.selector, this._resolverCache)
+        ? RelayReader.read(
+            source,
+            snapshot.selector,
+            this._resolverCache,
+            this._resolverContext || undefined,
```
captbaritone, line 179:
```suggestion
            this._resolverContext || undefined,
```
```suggestion
            this._resolverContext,
```

packages/relay-runtime/store/RelayStoreTypes.js
```diff
@@ -1378,3 +1378,10 @@ export type LiveState<+T> = {
    */
   subscribe(cb: () => void): () => void,
 };
+
+/**
+ * Context that will be provided to live resolvers if
+ * `resolverContext` is set on the Relay Store.
+ * This context will be passed as the third argument to the live resolver
+ */
+export type ResolverContext = Object;
```
captbaritone, line 1387:
```suggestion
export type ResolverContext = mixed;
```

I don't think we need to be opinionated at all as to what your context is.

website/docs/guides/relay-resolvers/context.md
```diff
@@ -0,0 +1,120 @@
...
+    subscribe: (callback) => {
+      return context.store.subscribe(callback);
+    },
+  };
+}
+```
+
+  </TabItem>
+
+  <TabItem value="Flow">
```
captbaritone, line 67: Context won't work with the Flow syntax until we do some additional work. Can you replace these Flow tabs with a note saying it's not supported for now?

website/docs/guides/relay-resolvers/context.md
```diff
@@ -0,0 +1,120 @@
...
+id: context
+title: "Context"
+slug: /guides/relay-resolvers/context
+description: Context in Relay Resolvers
+---
+import {FbInternalOnly, fbContent} from 'docusaurus-plugin-internaldocs-fb/internal';
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+In order to pass a service, or other values to be shared with all resolvers, the `LiveResolverStore` provides a means of passing context. This gets passed to the third argument of all resolvers (live and non-live).
```
captbaritone, line 11: Maybe we could call out that this is analogous to the context argument used on the server which usually holds things like the database connection: https://graphql.org/learn/execution/#root-fields--resolvers

website/docs/guides/relay-resolvers/limitations.md
```diff
@@ -7,9 +7,9 @@ description: Limitations of Relay Resolvers
 
 Relay Resolvers are do have some limitations. Here we will collect a list of known limitations and provide alternatives where possible.
 
-## No context or info arguments
+## No info argument
 
-In a full GraphQL implementation, resolvers would have access to a `context` and `info` argument. These arguments are not available in Relay Resolvers today. Supporting context is something we would like to do in the future, but have not yet implemented.
+In a full GraphQL implementation, resolvers would have access to an `info` argument. This argument is not available in Relay Resolvers today.
```
captbaritone, line 12: Yes!!

compiler/crates/relay-config/src/typegen_config.rs
```diff
@@ -129,6 +148,10 @@ pub struct TypegenConfig {
     /// A map from GraphQL error name to import path, example:
     /// {"name:: "MyErrorName", "path": "../src/MyError"}
     pub custom_error_type: Option<CustomTypeImport>,
+
+    // Indicates the type to import and use as the context for live resolvers.
```
captbaritone, line 152: If you use `///` here instead of `//` it will get picked up as part of the description in the json schema we generate. Can use that here and add similar comments to the `ResolverContextTypeInput` type and its variants?

captbaritone (changes requested):
Everything looks good! Just a few minor nits/fixups and we can land this!

captbaritone:
Rust test failures on CI are unrelated. We have an ongoing issue where it is running out of disk space.

website/docs/guides/relay-resolvers/limitations.md
drewatk: This was a fun change :)

compiler/crates/relay-compiler/relay-compiler-config-schema.json
```diff
@@ -1468,6 +1468,48 @@
           "default": false,
           "type": "boolean"
         },
+        "resolverContextType": {
```
captbaritone, line 1471: If you rerun the Rust tests with the flag to update snapshots this should update to include the description

compiler/crates/relay-config/src/typegen_config.rs
```diff
@@ -60,6 +60,25 @@ pub struct CustomTypeImport {
...
+pub enum ResolverContextTypeInput {
+    Path(ResolverContextTypeInputPath),
+    Package(ResolverContextTypeInputPackage),
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
+pub struct ResolverContextTypeInputPath {
+    pub name: StringKey,
+    pub path: PathBuf,
+}
```
captbaritone, line 74:
```suggestion
/// Specifies how Relay can import the Resolver context type from a path
#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
pub struct ResolverContextTypeInputPath {
    /// The name under which the type is exported from the module
    pub name: StringKey,
    /// The path to the module
    pub path: PathBuf,
}
```

Can we also clarify what the path is relative to?

compiler/crates/relay-config/src/typegen_config.rs
```diff
@@ -60,6 +60,25 @@ pub struct CustomTypeImport {
...
+
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
+pub struct ResolverContextTypeInputPackage {
+    pub name: StringKey,
+    pub package: StringKey,
+}
```
captbaritone, line 80:
```suggestion
/// Specifies how Relay can import the Resolver context type from a named package
#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
pub struct ResolverContextTypeInputPackage {
    /// The name under which the type is exported from the package
    pub name: StringKey,
    /// The name of the package
    pub package: StringKey,
}
```

compiler/crates/relay-typegen/src/visit.rs
```diff
@@ -525,12 +535,19 @@ fn get_resolver_arguments(
...
-                key: "rootKey".intern(),
+                key: intern!("rootKey"),
                 value: AST::RawType(format!("{fragment_name}$key").intern()),
                 read_only: false,
                 optional: false,
             });
         }
+    } else {
+        resolver_arguments.push(KeyValuePairProp {
+            key: "key".intern(),
```
captbaritone, line 546:
```suggestion
            key: intern!("key".intern),
```

captbaritone (changes requested):

captbaritone (approved):

captbaritone:
When trying to land this internally I found that it was going to be very difficult to rollout this change safely if we change how resolvers that don't read a parent object or rootFragment get called. So, I've updated the diff a bit internally to go back to the previous behavior where the first argument is omitted if there is no rootFragment and the resolver is not on a model type. I've updated the docs to match.

PR #4709, opened 2024-06-07
Normalize paths in diagnostics CLI output

captbaritone:
This should fix the current breakage of our windows tests. One thing I'm not sure about is if this will cause any problems with things like clickable file paths in the terminal on Windows.


captbaritone:
cc @alloy in case you know if replacing slashes in paths written to the CLI will break things for Windows users.

PR #4710, opened 2024-06-08
Support inlay hints

captbaritone:
First stab at supporting inlay hints

## Fragment Aliases

https://github.com/facebook/relay/assets/162735/8e0aff81-172b-4393-8c68-427eec3c3df3

## Field Args

Screenshot 2024-06-07 at 4 50 08 PM

### DIFF
--- a/compiler/crates/relay-lsp/src/inlay_hints.rs
+++ b/compiler/crates/relay-lsp/src/inlay_hints.rs
@@ -34,7 +34,7 @@ pub fn on_inlay_hint_request(
     let uri = params.text_document.uri.clone();
     let root_dir = state.root_dir();
 
-    if !is_file_uri_in_dir(&root_dir, &uri) {
+    if !is_file_uri_in_dir(root_dir, &uri) {
         return Err(LSPRuntimeError::ExpectedError);
     }
 
--- a/compiler/crates/relay-lsp/src/inlay_hints.rs
+++ b/compiler/crates/relay-lsp/src/inlay_hints.rs
@@ -1,0 +1,184 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+use std::path::PathBuf;
+
+use common::Location;
+use common::SourceLocationKey;
+use common::Span;
+use graphql_ir::Argument;
+use graphql_ir::FragmentDefinition;
+use graphql_ir::FragmentSpread;
+use graphql_ir::InlineFragment;
+use graphql_ir::OperationDefinition;
+use graphql_ir::Program;
+use graphql_ir::Visitor;
+use intern::string_key::Intern;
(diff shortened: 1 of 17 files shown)


compiler/crates/relay-lsp/src/inlay_hints.rs
```diff
@@ -0,0 +1,184 @@
...
+
+impl Visitor for InlayHintVisitor<'_> {
+    const NAME: &'static str = "InlayHintVisitor";
+
+    const VISIT_ARGUMENTS: bool = false;
+
+    const VISIT_DIRECTIVES: bool = false;
+
+    fn visit_operation(&mut self, operation: &OperationDefinition) {
+        if self.location_is_in_file(&operation.name.location) {
```
captbaritone, line 143: Not great that we have to iterate over every fragment/operation here...

compiler/crates/relay-lsp/src/inlay_hints.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    let file_path = absolute_file_path.strip_prefix(&root_dir).map_err(|_e| {
+        LSPRuntimeError::UnexpectedError(format!(
+            "Failed to strip prefix {:?} from {:?}",
+            root_dir, absolute_file_path
+        ))
+    })?;
+
+    let project_name = state.extract_project_name_from_url(&uri)?;
+    let program = state.get_program(&project_name)?;
```
captbaritone, line 59: Using this program's IR as the source of truth here will not work. This is based on what's on disk and not what's currently open in the editor. It will not reflect unsaved changes.

https://github.com/facebook/relay/assets/162735/66b21355-1c91-44be-ba26-4f52ba645694

compiler/crates/relay-lsp/src/utils.rs
```diff
@@ -62,7 +62,7 @@ pub fn extract_executable_definitions_from_text_document(
         .flat_map(|graphql_source| {
             let document = parse_executable_with_error_recovery_and_parser_features(
                 &graphql_source.text_source().text,
-                SourceLocationKey::standalone(text_document_uri.as_ref()),
+                SourceLocationKey::embedded(text_document_uri.path(), 1),
```
captbaritone, line 65: Hack here. Need to use the real index.

compiler/crates/relay-lsp/src/inlay_hints.rs
tobias-tengler: This should do the trick: https://github.com/facebook/relay/pull/4509

captbaritone: Yes! I ended up discovering that the hard way. I needed that change as well as figuring out how to derive the IR from the cached changed file. I did a hacky version of your change here, but let me see if I can do it via rebase and (finally) get your pr merged.

captbaritone:
This landed in https://github.com/facebook/relay/commit/dd9d3a03e969d1088573efa3bec4ac1555f4acd0 cc @tobias-tengler

Issue #4504, opened 2023-10-24
Docs: search behaves weirdly depending on the URL state


MarceloPrado:
It's quite annoying that Docusaurus search seems scoped to the current URL. As you can see from the repro, searching for `Imperatively` in the Relay home and in the "Next" version yields different results. This makes it harder to learn the framework, as finding the API/guide I need is often cumbersome.

I understand the team wants to "gate" access to certain docs depending on whether they are available or not. But to use my repro as an example, as a Relay user I already have access to the imperative APIs described in the [guide](https://relay.dev/docs/next/guided-tour/updating-data/imperatively-modifying-store-data/#internaldocs-banner), and hence, I expect them to be available in the major I'm at (v15).

Repro:

https://github.com/facebook/relay/assets/8047841/00d66e55-63b7-4b9c-99bf-4e9bd35dda5a

captbaritone:
Thanks for flagging. I was just looking into why search seems broken. Let me see what I can figure out here.

JUSTIVE:
other than the issue above, the whole search feature is broken. I've never had a successful search experience on the docs page.

MarceloPrado:
Fixed by #4715

PR #4715, opened 2024-06-14
Update Algolia app & key

zpao:
The app in use seems to be lost and doesn't appear to be configured correctly. It only seems to have crawler results for the "next" version, so contextual search in particular doesn't work when browsing other versions.


zpao:
Showing it working locally in multiple versions:

vNext:
image

v16:
image

MarceloPrado:
Fixes #4504

PR #4719, opened 2024-06-18
Report schema parsing errors for synced schema sources

tobias-tengler:
This adds support for diagnostics while typing (without saving) within schema documents.

Currently it's only reporting diagnostics for parsing errors, but it can be extended with actual validation of schema semantics in the future.

### DIFF
--- a/compiler/crates/common/src/text_source.rs
+++ b/compiler/crates/common/src/text_source.rs
@@ -113,6 +113,13 @@ impl TextSource {
                 character += 1;
             }
         }
+
+        if start_position != lsp_types::Position::default()
+            && end_position == lsp_types::Position::default()
+        {
+            end_position = lsp_types::Position::new(line as u32, character as u32);
+        }
+
         lsp_types::Range::new(start_position, end_position)
     }
 }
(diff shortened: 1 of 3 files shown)


compiler/crates/common/src/text_source.rs
```diff
@@ -113,6 +113,13 @@ impl TextSource {
                 character += 1;
             }
         }
+
+        if start_position != lsp_types::Position::default()
+            && end_position == lsp_types::Position::default()
+        {
+            end_position = lsp_types::Position::new(line as u32, character as u32);
+        }
```
tobias-tengler, line 121: There was a bug here in that if the diagnostic pointed to the last character you would always get a span where the end would point to 0,0. Now if you have a start, but no end, it uses the last character as the span's end.

compiler/crates/relay-lsp/src/server/lsp_state.rs
```diff
@@ -325,6 +325,45 @@ impl<TPerfLogger: PerfLogger + 'static, TSchemaDocumentation: SchemaDocumentatio
...
+        let mut diagnostics = vec![];
+        let text_source = schema_source.text_source();
+        let result = graphql_syntax::parse_schema_document(&text_source.text, source_location_key);
+
+        match result {
+            // In the future you could run additional validations based on the new schema document here
+            Ok(_) => (),
+            Err(raw_diagnostics) => {
+                diagnostics.extend(raw_diagnostics.iter().map(|diagnostic| {
+                    // TODO: The very last character is problematic for some reason
```
tobias-tengler, line 354:
```suggestion
```

PR #4723, opened 2024-06-23
[WIP] Generate JSON schema for compiler config

captbaritone:
One huge usability issue with Relay today is that the compiler config is wholely undocumented. To understand what options are available you either need to _try_ an option and try to decipher the sometime cryptic validation errors, or read the Rust code.

This PR attempts to start to close that gap. Our Rust structs and comments _are_ a good source of truth for this info, so ideally we still use that as our documentation/validation. So, here we derive [Json Schema](https://json-schema.org/) from our Rust structs/enums/types/serde tags. 

In VSCode this allows for nice autocomplete, hover tool tips, and yellow squiggles if you make an error (much nicer than a cryptic validation error when you run the compiler).

Some next steps this could enable:

1. Enable this editor experience directly if you have the Relay VSCode extension installed via the [JSON Schema contribution point](https://code.visualstudio.com/api/references/contribution-points#contributes.jsonValidation)
2. Derive a docs page about the [Docusaurus JSON Schema plugin](https://jy95.github.io/docusaurus-json-schema-plugin/

https://github.com/facebook/relay/assets/162735/7f5e807b-069c-4344-9e9c-506f3369b042

## Next Steps

There are a few unanswered things here that we'll need to consider:

1. How do we publish this schema such that users can find it and ensure that they use the correct schema for the version of the compiler they are using?
    1. Especially tricky if we want to enable it automatically in VSCode
    2. Maybe we could have the compiler validate that you've added the right value for `$schema`?
3. Can we get the docs plugin to work? (I think it might require Docusaurus 3)
4. Are there bugs? There are some serde attributes and little deserialization tricks which we might not have modeled correctly here. Difficult to know if we've captured them all.
5. Add more descriptions to our structs. Some fields are undocumented. This would provide us more incentive to get those struct/field/variant `///` comments very helpful and up to date.

## Acknowledgment

This PR aims to be a more maintainable and reliable alternative to #4569. Thanks for @noghartt for pushing the discussion of how we might fill this gap.


tobias-tengler:
Love the resumed work on this! I updated my old attempt to your version in https://github.com/facebook/relay/pull/4724
I think it would be easiest and best for adoption to publish the JSON schema as part of the relay-compiler npm package. The VS Code extension and other tooling could pick it up from there.

captbaritone:
> I updated my old attempt to your version in https://github.com/facebook/relay/pull/4724

Oh this is great! I wasn't sure if there was a way for the VSCode extension to locate the Json Schema file in node_modules. Looking forward to looking more closely at that PR.

Also, my sincere apologies for not acknowledging previous attempts at this. I had forgotten about your PR and also @tbezman! I've updated the description to reference both.

compiler/crates/relay-compiler/src/config.rs
```diff
@@ -696,15 +702,20 @@ struct MultiProjectConfigFile {
     feature_flags: FeatureFlags,
 
     /// Watchman saved state config.
+    #[schemars(skip)] // TODO: Fix me
```
captbaritone, line 705: This one's tricky because `ScmAwareClockData` and its serialization/deserialization is defined in its own crate which we don't own. We probably need to define our own wrapper around `ScmAwareClockData` that replicates its fields as suggested in https://graham.cool/schemars/examples/5-remote_derive/

gulpfile.js
```diff
@@ -360,6 +360,13 @@ const relayCompiler = gulp.parallel(
       })
       .pipe(gulp.dest(path.join(DIST, 'relay-compiler')));
   },
+  function copyCompilerConfigSchema() {
+    return gulp
+      .src(['relay-compiler-config-schema.json'], {
```
captbaritone, line 365: @tobias-tengler This should put the config schema in the root of the `relay-compiler` npm module.

Screenshot 2024-06-23 at 8 41 47 PM

compiler/crates/relay-compiler/tests/relay_config_schema_json_test.rs
```diff
@@ -0,0 +1,21 @@
...
+    let generator = gen::SchemaGenerator::from(settings);
+    let schema = generator.into_root_schema_for::<ConfigFile>();
+    let actual = serde_json::to_string_pretty(&schema).unwrap();
+    let expected = include_str!("../../../relay-compiler-config-schema.json");
+    let source_file_path = file!();
+
+    let expected_file_path = workspace_root()
+        .join(source_file_path)
+        .with_file_name("../../../relay-compiler-config-schema.json");
+    assert_file_contains(&actual, expected_file_path, expected)
```
captbaritone, line 20: This reuses some of the logic from our fixture files:

1. It will fail and show a diff if the current file does not match
2. Setting `UPDATE_SNAPSHOTS=1` will cause it to update the file

orta:
> How do we publish this schema such that users can find it and ensure that they use the correct schema for the version of the compiler they are using? 

Is pretty easy to solve, send a PR to https://github.com/SchemaStore/schemastore/blob/master/src/api/json/catalog.json with a link to a version you host on the relay website

captbaritone:
@orta Neat! How does version resolution work? If we add multiple versions there, who choses which one to use?

orta:
It basically doesn't - I think?

Might be able to do it with more control if you used the vscode extension to set the schema instead probably

captbaritone:
Yeah, that's what I think we'll do. Tobias already has a PR open.

PR #4724, opened 2024-06-23
Add editor support for JSON config schema

tobias-tengler:


vscode-extension/src/providers/textDocumentContentProvider.ts
```diff
@@ -0,0 +1,38 @@
...
+  provideTextDocumentContent(uri: Uri): ProviderResult<string> {
+    if (uri.authority === PACKAGE_JSON_RELAY_CONFIG_SCHEMA_PATH) {
+      return `{"properties": { "relay": { "$ref": "${RelayTextDocumentContentProvider.scheme}://${RELAY_CONFIG_SCHEMA_PATH}", "description": "Relay.js configuration" }}}`;
+    }
+
+    if (uri.authority === RELAY_CONFIG_SCHEMA_PATH) {
+      try {
+        return fs.readFileSync(this.configJsonSchemaPath, {encoding: 'utf-8'});
+      } catch {
+        return EMPTY_JSON_SCHEMA;
```
captbaritone, line 32: I think this is undefined?

vscode-extension/package.json
```diff
@@ -112,7 +112,17 @@
           "description": "Whether or not we should automatically start the Relay Compiler in watch mode when you open a project."
         }
       }
-    }
+    },
+    "jsonValidation": [
+      {
+        "fileMatch": "relay.config.json",
+        "url": "relay://relay-config-schema"
```
captbaritone, line 119: TIL You can define your own url schemes. This is perfect. I wasn't sure how we were going to be able to locate the user's local copy of the Json Schema, but this looks like a great approach.

captbaritone:
I just had another idea. Since the VSCode extension already knows how to invoke the compiler binary, and the compiler binary can construct the json schema, what if we added a separate compiler command to return the schema text on stdout? It seems like that might be simpler. It would also work better for us internally since we don't access the compiler via an NPM module but a separate binary deployment which would be hard to update to also sync a sibling file.

vscode-extension/src/providers/textDocumentContentProvider.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  static readonly scheme: string = 'relay';
+
+  cachedJsonSchema: string | undefined;
+
+  onDidChange?: Event<Uri> | undefined;
+
+  provideTextDocumentContent(uri: Uri): ProviderResult<string> {
+    if (uri.authority === PACKAGE_JSON_RELAY_CONFIG_SCHEMA_PATH) {
+      return `{"properties": { "relay": { "$ref": "${RelayTextDocumentContentProvider.scheme}://${RELAY_CONFIG_SCHEMA_PATH}", "description": "Relay.js configuration" }}}`;
```
captbaritone, line 28: This little recursion trick is so clever. Nice.

vscode-extension/src/providers/textDocumentContentProvider.ts
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    if (uri.authority === RELAY_CONFIG_SCHEMA_PATH) {
+      this.cachedJsonSchema ||= this.loadConfigJsonSchema();
```
captbaritone, line 32: If the user changes their relay binary path in the config file, does this get invalidated correctly?

tobias-tengler: No, but you have to restart the extension anyways because the binary information is only loaded once at startup atm.

vscode-extension/src/providers/register.ts
```diff
@@ -0,0 +1,32 @@
...
+import {workspace} from 'vscode';
+import {RelayExtensionContext} from '../context';
+import {RelayTextDocumentContentProvider} from './textDocumentContentProvider';
+
+export function registerProviders(context: RelayExtensionContext) {
+  const {binaryVersion} = context.relayBinaryExecutionOptions;
+
+  if (binaryVersion) {
+    const hasConfigJsonSchemaCommand =
+      semver.satisfies(binaryVersion, '>=18.0') ||
```
tobias-tengler, line 18: This probably needs to be updated depending on what the release schedule will be like. Just picked the next major for now.

tobias-tengler:
The Prettier and ESLint rules are conflicting... Nice

vscode-extension/src/providers/register.ts
captbaritone: Since it will be in the next release, can we just say `>17.0.0`?

vscode-extension/.eslintrc.js
```diff
@@ -25,6 +25,7 @@ module.exports = {
         'operator-linebreak': 'off',
         '@typescript-eslint/indent': 'off',
         '@typescript-eslint/object-curly-spacing': 'off',
+        '@typescript-eslint/brace-style': 'off',
```
captbaritone, line 28: Good call.

vscode-extension/src/providers/register.ts
captbaritone: Sorry, one more realization. What happens if you use a version of the compiler from the @main NPM tag? Will that work or not? Would be cool to be able to test this end to end before the next release by installing the new version of the extension and installing relay-compiler@main (note that @main releases are currently stuck due to broken CI)

tobias-tengler:
Verified that it would attempt to load the JSON schema for a prerelease compiler (but it wasn't working of course since there isn't a released version with the `config-json-schema` command yet).
Should be good to go :)

PR #4737, opened 2024-07-12
Fix render and dispose checks in useQueryLoader tests

monicatang:
Tests on `renderCount` would be testing React internals, which is unadvised since the tests may fail depending on the React version, and are probably ok to remove. But, these tests are testing Relay behavior -- whether queries are disposed upon re-render -- so I kept the React `renderCount` checks to clarify the expected render behavior that is triggering these query disposals.

I updated the `renderCount` to the correct number for React 19.


captbaritone:
Closing since the diff internally was abandoned.

PR #4740, opened 2024-07-13
Add Inlay Hints for Fragment arguments

tobias-tengler:
CleanShot 2024-07-13 at 16 31 43@2x

### DIFF
--- a/compiler/crates/relay-lsp/src/inlay_hints.rs
+++ b/compiler/crates/relay-lsp/src/inlay_hints.rs
@@ -8,8 +8,10 @@
 use common::Location;
 use common::Span;
 use graphql_ir::Argument;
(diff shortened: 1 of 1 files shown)


captbaritone:
I think there's an interesting question here of how we should use inlay hints. Is there a rational to insert them here and not next to output fields as well? This info is available on hover in both cases. Curious to hear the rational here.

Right now we are using inlay hints for `@alias` and that feels justified because it lets us show what Relay is inferring, and also shows it similar to the syntax we would ideally use.

The next place I was imagining using inlay hints was for fragment arguments defined using the directive syntax. Could we use inlay hints to show fragment arguments using the new syntax even if you define them using the directive syntax?

PR #4741, opened 2024-07-13
Fix some documentation links in errors or hover tooltips

tobias-tengler:
I noticed that some of the linked pages 404, are pointing to pages that aren't part of the actual documentation anymore or are needlessly pointing to the "next" version of the documentation.

### DIFF
--- a/compiler/crates/relay-schema/src/relay-extensions.graphql
+++ b/compiler/crates/relay-schema/src/relay-extensions.graphql
@@ -176,7 +176,7 @@ declare how null values should be handled at runtime. You can think of it as
 saying "if this field is ever null, its parent field is invalid and should be
 null".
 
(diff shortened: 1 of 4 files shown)


compiler/crates/relay-schema/src/relay-extensions.graphql
```diff
@@ -176,7 +176,7 @@ declare how null values should be handled at runtime. You can think of it as
 saying "if this field is ever null, its parent field is invalid and should be
 null".
 
-[Read More](https://www.internalfb.com/intern/staticdocs/relay/docs/guides/required-directive/) (FB only)
+[Read More](https://relay.dev/docs/guides/required-directive/)
```
tobias-tengler, line 179: Feel free to undo, if there is still FB specific information on this page, but considering that it's documented and used in OSS I thought it should use the link from the official docs.

Issue #4745, opened 2024-07-16
Remove @ts-nocheck from generated Typescript files


drewatk:
In order to ensure type safety, generated typescript files should not be exported with `@ts-nocheck`. This is currently present in both generated types for resolvers as well as generated fragments which are referenced from components. In order to create type assertions within this file, we will need to begin type-checking these files.

Related: #4742
This is part of the requirement for the resolver context work mentioned in #4000 and #4704

PR #4753, opened 2024-07-31
Enable checking Typescript files behind a compiler flag

drewatk:
Fix for #4745 

Behind a compiler flag, `typescript_check_generated_files`, removes the `@ts-nocheck` flag from being output within typescript generated files by the Relay Compiler. 

This is not safe to enable yet, but will allow for identifying Typescript issues in the generated code by enabling this flag to find those issues.

Notes:
- Added a new compiler flag (not a feature flag, can change to be a feature flag if that is more appropriate)
- Updated Tests


compiler/crates/relay-config/src/typegen_config.rs
```diff
@@ -114,6 +114,10 @@ pub struct TypegenConfig {
     #[serde(default)]
     pub typescript_exclude_undefined_from_nullable_union: bool,
 
+    // todo: should this go here or in a feature flag? Above flag is discussed here: https://github.com/facebook/relay/pull/4380
```
captbaritone, line 117: Feature flag is probably slightly preferable if it's feasible to thread it through without too much trouble. If it's very noisy to thread it through, this is fine.

compiler/crates/relay-compiler/tests/relay_compiler_integration/fixtures/preloadable_query_typescript.expected
```diff
@@ -53,19 +53,18 @@ export default node;
...
  */
 
 /* tslint:disable */
 /* eslint-disable */
-// @ts-nocheck
 
 // @relayRequestID ae6874c86ce5db2df8d6b253a6a0ec13
 
-import { ConcreteRequest, Query } from 'relay-runtime';
+import { ConcreteRequest } from 'relay-runtime';
```
captbaritone, line 67: I wonder if these fixture changes are actually out of date and come from https://github.com/facebook/relay/pull/4754?

compiler/crates/relay-compiler/tests/relay_compiler_integration/fixtures/preloadable_query_typescript.input
```diff
@@ -11,7 +11,8 @@ graphql`
   "eagerEsModules": true,
   "persistConfig": { 
     "file": "./operations.json" 
-  }
+  },
+  "typescriptCheckGeneratedFiles": true
```
captbaritone, line 15: I don't see any generated files changing as a result of this. Maybe the fixtures are out of date?

captbaritone (changes requested):
Thanks for working on this!

compiler/crates/relay-compiler/tests/relay_compiler_integration/fixtures/preloadable_query_typescript.expected
drewatk: Yep, will remove

compiler/crates/relay-compiler/tests/relay_compiler_integration/fixtures/preloadable_query_typescript.input
drewatk: Updated, but @captbaritone This begs the question, how much of the tests should have this flag on? Or should we leave it off in the tests for the time being?

captbaritone: Good question. Maybe just a add a single new test which turns it on. If these integration tests got typechecked (that wold be cool! Especially if/when we pull the library types into this repo!) then I'd stay turn it on for all of them, but without that, all we need to assert is that if you turn on the flag the suppression gets removed, and that's a separate assertion from any pre-existing test, so we can have that as its own test file.

Let me know if you need help figuring out how to add new fixtures.

compiler/crates/relay-config/src/typegen_config.rs
drewatk: Changed to feature flag!

compiler/Cargo.lock
```diff
@@ -429,7 +429,6 @@ dependencies = [
  "graphql-ir",
  "graphql-syntax",
  "intern",
- "relay-config",
```
drewatk, line 432: this seems wrong, should this change be reverted? Might have been an artifact of the other branch I was working on

compiler/crates/relay-compiler/tests/relay_compiler_integration/fixtures/preloadable_query_typescript.input
drewatk: Also, I agree that it would be good to start TS typechecking as a validation step within this repo. That way we can prevent regressions to the generated types. As it stands currently we have to check the output in our own repo first.

compiler/Cargo.lock
captbaritone: Oh yeah, our cargo.lock is out of date. I'll import https://github.com/facebook/relay/pull/4747 which should fix it. in the mean time you can revert or leave this, either way is fine.

compiler/crates/relay-compiler/tests/relay_compiler_integration/fixtures/preloadable_query_typescript.input
captbaritone: Yeah, for Flow we have our own codebase as a massive test case. For TypeScript we'll need something to echo that setup. Running typescript on integration tests could be a great way to achieve that pretty cheaply.

We have rust utilities that materialize these fixture files into multiple files in a directory and then run the compiler on them. Maybe that test could also run the TypeScript type check on them? Maybe a little brittle to shell out to TypeScript from within the Rust tests, but probably worth it.

Will require a little fiddling to get it to work both internally and in GitHub CI and ensure we keep the same version(s?) of TypeScript, but I bet we can figure it out.

Here's where we run the integration tests: https://github.com/facebook/relay/blob/main/compiler/crates/relay-compiler/tests/relay_compiler_integration.rs

cc @rbalicki2 regarding upstreaming TypeScript types and how we might add tests which assert their correctness.

drewatk: Added fixture test

compiler/Cargo.lock
captbaritone: This has landed on main now. Should be able to rebase.

compiler/crates/common/src/feature_flags.rs
```diff
@@ -124,6 +124,10 @@ pub struct FeatureFlags {
     /// Disable validation of the `edgeTypeName` argument on `@prependNode` and `@appendNode`.
     #[serde(default)]
     pub disable_edge_type_name_validation_on_declerative_connection_directives: FeatureFlag,
+
+    // Removes the @ts-nocheck from generated Typescript Files
```
captbaritone, line 128: Use `///` (three slashes) here to have Rust pick this up as a documentation comment. Should cause it to get included in the compiler-config-schema.

Also can we invert this naming? I think we want to end with checking ts files being the default, and defaulting flags to `false` is generally more intuitive. So, maybe we add this flag as `typescript_disable_checking_generated_files`?

captbaritone (changes requested):

compiler/crates/common/src/feature_flags.rs
drewatk: Then should we update all test snapshots to remove the nocheck? 

As it stands today, we'd break many TS implementations if we release with defaulting to checking TS.

Another question - do we prefer this type to be `bool` or `FeatureFlag`?

captbaritone: If it would break many TS implementations today, then maybe we could _start_ with it defaulting to true and then switch it to defaulting to false once we think we are in a place where it would work for most people?

I think `bool` is probably okay, but `FeatureFlag` would let you do a setup where you opt-out on a per-[something] basis by specifying a list of strings in the config. Could be fragment/query/mutation names, or some other pattern. Can be useful if you think people will need to do an incremental adoption.

drewatk: > If it would break many TS implementations today, then maybe we could start with it defaulting to true and then switch it to defaulting to false once we think we are in a place where it would work for most people?

This sounds good to me! I haven't found any examples in the existing where we default to true, how would I add that?

compiler/crates/common/src/feature_flags.rs
```diff
@@ -124,8 +124,44 @@ pub struct FeatureFlags {
...
+            disable_resolver_reader_ast: false,
+            enable_fragment_argument_transform: false,
+            allow_resolver_non_nullable_return_type: Default::default(),
+            disable_schema_validation: false,
+            disallow_required_on_non_null_fields: false,
+            prefer_fetchable_in_refetch_queries: false,
+            disable_edge_type_name_validation_on_declerative_connection_directives: Default::default(),
+            typescript_disable_checking_generated_files: default_true(),
+        }
+    }
```
captbaritone, line 161: I think it's okay if we skip this for now and rely on the derived default. I think it's okay to let all places that use default feature flags from `Default::default()` in Rust instead of via serde (just tests I think?) use `false` as the default here.

I assume that's what you started with. Was there some reason it was problematic to let those tests default to false?

drewatk:
Hey @captbaritone, let me know what you think about this approach and if there are any blockers to merging this

captbaritone:
Yikes. Sorry we lost track of this. Do you mind rebasing? I can merge.

PR #4754, opened 2024-07-31
Skip importing export types in TS

drewatk:
A fix for a typescript generation bug discovered from the work in #4753, with efforts to enable checking for all TS generated files #4745 

This fixes a class of Typescript errors in generated typescript files with TS checking enabled:
```
/src/__generated__/mutation.graphql.ts(11,27): error TS2305: Module '"relay-runtime"' has no exported member 'Mutation'.
/src/__generated__/mutation.graphql.ts(11,27): error TS6133: 'Mutation' is declared but its value is never read.
```

The root cause is the exported type is both not exported by the TS types, and not used due to typescript using the `export default` syntax.
Where flow would generate the following export:
```
module.exports = ((node/*: any*/)/*: Mutation<
  validateMutationTest9ChangeNameIncludeMutation$variables,
  validateMutationTest9ChangeNameIncludeMutation$data,
>*/);
```
Typescript would generate the following, not using the `Mutation` type:
```
export default node;
```

This fix removes the extraneous import type when the language is set as typescript.


compiler/crates/relay-compiler/src/build_project/artifact_generated_types.rs
```diff
@@ -27,8 +30,9 @@ impl ArtifactGeneratedTypes {
         operation: &OperationDefinition,
         skip_types: bool,
         is_client_only: bool,
+        project_config: &ProjectConfig,
```
captbaritone, line 33: Based on the principle of least knowledge should we maybe _just_ pass the language here instead of the full project_config?

compiler/crates/relay-compiler/src/build_project/artifact_generated_types.rs
```diff
@@ -27,8 +30,9 @@ impl ArtifactGeneratedTypes {
...
     ) -> Self {
-        if skip_types {
+        if skip_types || project_config.typegen_config.language == TypegenLanguage::TypeScript {
```
captbaritone, line 35: Not related to your change here, but I'm kinda bummed how this works. Having the set of imports defined far away from the code that uses them is pretty brittle. Ideally we could have a setup where the code that references an imported type also adds that type to a set, and then when we go to print the imports we use a sorted version of that set.

Does not need to be in scope for this PR, but would be a nice improvement for someone to do at some point.

captbaritone (changes requested):
Looks very sensible! Just one tiny nit

compiler/crates/relay-compiler/src/build_project/artifact_generated_types.rs
drewatk: Fixed

captbaritone:
Looks good! Happy to import/merge once the CLA is signed.

captbaritone:
@drewatk Thanks for the ping. Merged. We are still working on some issues with our CI which means we are not currently cutting our `@main` NPM releases `relay-compiler@main`. Were you depending upon using those releases to validate this change at Microsoft? If so, let me know and I can try to find a way to get a release here.

alloy:
@drewatk We should really consider adding some tests for the TS typings that get emitted.

PR #4766, opened 2024-08-11
Add @__RelayTypegenTypeOverride to use in CustomTransform to replace TS/JS type

Markionium:
This PR introduces and internal directive (`@__RelayTypegenTypeOverride`)  to use in a Typgen `CustomTransform` [(basic example)](https://github.com/facebook/relay/blob/0ed29d161fd9ea30346d18b9e96e6939a7e57ee1/compiler/crates/relay-transforms/src/typegen_type_override_directive.rs#L50-L79) 
It allows the `CustomTransform` to specify a type, path and generic arguments to be used in the JS/TS type instead of the default type.

I attempted to solve this with just the `CustomTransform` but didn't manage. I think the current approach should be generic enough for others to benefit from.

Example use case:
  
I want to specify localized strings through GraphQL and serve them using LiveResolvers. Some of these will have placeholders

```graphql
# schema.graphql
type ViewData {
  greetingLabel: String @localize(placeholders: ["name"]) # Value of this could be `Hello {name}`
}
```

Based on the above I want the compiler to generate a more strict type that I can use to check if this required placeholders for the string are also provided.

```ts
// __generated__/Fragment.ts
import { LocalizedString } from "@i18n/localization";
import { FragmentRefs } from "relay-runtime";
export type foo$data = {
  readonly greetingLabel: LocalizedString<"name"> | null | undefined; // <-- Notice the custom type  here
  readonly " $fragmentType": "foo";
};
export type foo$key = {
  readonly " $data"?: foo$data;
  readonly " $fragmentSpreads": FragmentRefs<"foo">;
};
```

Using this custom type I can make sure we "inject" the placeholder into the string by having a strongly typed function that ensures the correct placeholders are passed for each of the `LocalizedString`s.

```ts
const greeting = format(data.greetingLabel, { name: "Mark" }); // Typescript will require `name` to be provided.
```


Markionium:
@captbaritone would appreciate any thoughts you have on this. I can add more context/detail on my use case too if you need :)

PR #4767, opened 2024-08-11
Skip check for source control integration if noSourceControl: true is set in config.

Markionium:
This PR adds the option to add `noSourceControl: true` to the relay config which will skip the `git status` and `sl --version` commands when creating the configuration.

Those checks are run to help the compiler add/remove files from source control for you. However in cases where the generated files are not checked into source control this isn't helpful, especially when `git status` is "slow".

I noticed a significant decrease in performance between 16 and 17 when `git status` was introduced.

Another issue we have internally in Microsoft is that we use [BuildXL](https://github.com/microsoft/BuildXL) in some of our repos, which distributes building of our packages. However it will not allow reads to undeclared dependencies, a `git status` check will break that "rule" as it read across the whole repo. 

In the repo I work on we do not check in our artefacts and thus do not need source control integration.


tyao1:
For making it a config, you could add the config here:
https://github.com/facebook/relay/blob/main/compiler/crates/relay-compiler/src/config.rs

Markionium:
@tyao1 Updated the PR, tested to make sure it works for both scenarios (Single and Multi project config files). I'm not sure if there any tests for this source control integration that I could have updated/added to?

Markionium:
@tyao1 should be good to go now!

captbaritone:
Looks good!

PR #4770, opened 2024-08-13
Document current support for experimental semantic nullability

captbaritone:


website/docs/guides/semantic-nullability.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Proposed Solution
+
+[Semantic Nullability](https://github.com/graphql/graphql-wg/discussions/1410) is an early GraphQL spec proposal that aims to decouple error handling and nullability in the GraphQL spec to enable maximum resiliency while still exposing the "semantic nullability", (the nullability of the actual resolver function/method on the server) of the field to the client.
+
+The proposal works by allowing the schema to specify a new type of nullability of "null only on error". If a client sees this type, _and_ the client has some scheme for handling field errors out-of-band, it may treat the field that is exposed to user code as non-nullable.
```
benjie, line 24: Recommend using a different word versus `scheme` (because it's very close to `schema`)  - perhaps `strategy`?

```suggestion
The proposal works by allowing the schema to specify a new type of nullability of "null only on error". If a client sees this type, _and_ the client has some strategy for handling field errors out-of-band, it may treat the field that is exposed to user code as non-nullable.
```

website/docs/guides/semantic-nullability.md
```diff
@@ -0,0 +1,107 @@
...
+
+```js
+import { RelayFeatureFlags } from "relay-runtime";
+// @ts-ignore DefinitelyTyped is missing this
+RelayFeatureFlags.ENABLE_FIELD_ERROR_HANDLING = true;
+```
+
+Once enabled, you can add `@throwOnFieldError` to your fragments to indicate that the client should throw an error if any field errors are encountered when the fragment is read.
+
+In the below example, Relay's generated TypeScript of Flow types for `user.name` will be non-nullable.
```
benjie, line 62:
```suggestion
In the below example, Relay's generated TypeScript or Flow types for `user.name` will be non-nullable.
```

benjie (commented):
:+1:

website/docs/guides/semantic-nullability.md
```diff
@@ -0,0 +1,107 @@
...
+:::
+
+## Motivation
+
+One of GraphQL's strengths is its field-granular error handling which can dramatically improve response resiliency. However, today that error handling depends upon field nullability, which is the reason it is a [recommended best practice](https://graphql.org/learn/best-practices/#nullability) to default all fields to being nullable. This creates a trade-off where __enabling maximum resiliency means client developers must manually handle all possible permutations of field nullability__ within their components. [`@required`](./required-directive.md) can help a bit, but is ultimately a very blunt tool.
+
+## Proposed Solution
+
+[Semantic Nullability](https://github.com/graphql/graphql-wg/discussions/1410) is an early GraphQL spec proposal that aims to decouple error handling and nullability in the GraphQL spec to enable maximum resiliency while still exposing the "semantic nullability", (the nullability of the actual resolver function/method on the server) of the field to the client.
```
captbaritone, line 22: @benjie I wonder what the canonical link should be here. Lee's post is not the most up to date. Should this be your discussion post https://github.com/graphql/nullability-wg/discussions/58 maybe?

benjie: This is probably the best link right now:

https://github.com/graphql/graphql.github.io/blob/nullability-post/src/pages/blog/2024-08-14-exploring-true-nullability.mdx#our-latest-proposal

I need to either update what's being proposed in the issue you linked, or alternatively propose a slightly different alternative (more like what's in this blog post above).

Essentially what I'm proposing now is that users (people writing the web frontends/mobile apps/etc) would only ever consume a traditional-looking schema with traditional behaviors, but that schema may be different to the one that the server exposes because their client of choice (Relay, Apollo, URQL, etc) has factored in both how it handles nullability (e.g. "throw on read" would allow treating the semantically non-nullable types as if they were strictly non-nullable) and any other client-specific details such as local schema extensions.

benjie: To be clear: it's still essentially the same solution technically, it's just being presented differently - e.g. more carefully defining the audiences for each mode.

PR #4775, opened 2024-08-20
Fix raw response with selections on abstract type

tobias-tengler:


tobias-tengler:
@captbaritone Thoughts on this?

captbaritone:
This looks more correct to me at first glance! Thanks for working on this. I likely won't have time to look into it more closely until after GraphQL conf. One thing we'll have to look out for is if this will cause type errors with existing places people are using raw response types and how we'll manage that rollout if it's the case.

Let me see if anyone else on the team can take a look before then.

tobias-tengler:
@captbaritone This will certainly lead to type errors for existing projects. I've added a feature flag to enable an opt-out.

tobias-tengler:
@captbaritone Could we land this?

compiler/crates/relay-typegen/src/visit.rs
```diff
@@ -1707,14 +1770,47 @@ pub(crate) fn raw_response_selections_to_babel(
         }
     }
 
-    if !base_fields.is_empty() {
+    if !base_fields.is_empty() || !by_abstract_type.is_empty() {
+        let mut base_fields_map = selections_to_map(base_fields.clone().into_iter(), false);
+
+        if !by_abstract_type.is_empty() {
```
captbaritone, line 1776: Could we get a comment explaining the case we are handling here?

tobias-tengler:
@captbaritone Thanks for the review! 
One more question: Do you need a feature flag to turn of this addition at Meta? I just added one because I thought it's the kind of change you guys would want to have a flag for, but if you don't, I'll gladly get rid of it 😆

captbaritone:
> Do you need a feature flag to turn of this addition at Meta?

I suspect we will.

PR #4778, opened 2024-08-29
Add copy operation command

tobias-tengler:
I often find myself in a situation where I want to quickly run a Relay query outside of the configured Relay environment, i.e. a GraphQL IDE or another tool. When not using persisted operations, I need to go to the artifact, copy out the value of the `text` property and get rid of all the `\n` line breaks in the text. For persisted operations it's less tedious, I just grab the `id` from the artifact and look up the operation in the operation store, but it's still more cumbersome than it has to be.

This implements a new `relay/printOperation` request in the LSP that returns the transformed operation text of the operation under the cursor or the first operation inside the document. The VS Code Extension is also updated with a `relay.copyOperation` command that invokes that LSP request and copies the operation text to the clipboard.

Example of the command in action

### DIFF
--- a/compiler/crates/relay-lsp/src/graphql_tools.rs
+++ b/compiler/crates/relay-lsp/src/graphql_tools.rs
@@ -79,7 +79,7 @@ impl Request for GraphQLExecuteQuery {
 /// This function will return the program that contains only operation
 /// and all referenced fragments.
 /// We can use it to print the full query text
-fn get_operation_only_program(
+pub fn get_operation_only_program(
     operation: Arc<OperationDefinition>,
     fragments: Vec<Arc<FragmentDefinition>>,
     program: &Program,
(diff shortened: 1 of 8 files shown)


tobias-tengler:
I can no longer build locally and neither the CI can. Seems like your internal Rust version no longer matches the one defined in the CI. It's failing since you've merged https://github.com/facebook/relay/commit/176f96fd637e3879efc2dad0b7867b6703fe023c

zth:
Really nice! Did you consider making it a code action instead?

tobias-tengler:
At least for me the extension command seemed faster, as I don't first have to locate the operation inside the document if there's only one.
Where would you see the benefits of a code action?

tobias-tengler:
@captbaritone Had to update the minimum compiler version check to 18.

PR #4779, opened 2024-08-29
Update Rust toolchain to 1.76.0 in CI

tobias-tengler:
Since https://github.com/facebook/relay/commit/176f96fd637e3879efc2dad0b7867b6703fe023c the compiler can no longer be built in the GitHub CI. 
It seems there's a mismatch between Facebook's internal Rust toolchain version and the one used in OSS (1.72.0).
Through manual testing I found out that the compiler can be built using the 1.76.0 Rust toolchain.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,7 +95,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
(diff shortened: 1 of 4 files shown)


tobias-tengler:
Perfect. The `Cargo.lock` file is also out of date.

captbaritone:
Thank you!

PR #4780, opened 2024-09-03
Add relay-config library

robertf224:
Adding (or rather [reviving](https://www.npmjs.com/package/relay-config)) a `relay-config` library for loading/saving Relay configs.  Will port `babel-plugin-relay` to use this but hitting "module not found" errors in tests.


captbaritone:
Would love a more complete understanding of what problem(s) we're trying to solve here.

Here's what I'm inferring: The babel plugin currently needs to be able to see the Relay config file in order to apply the correct transformation. This requires:

1. Locating the same config file that the compiler finds
2. Parsing the config file in the same way the compiler does
3. Validate the config file in the same way the compiler does
4. Being able to have type-safe interactions with reading the config file once it's been validated

Am I correct in inferring that this is the set of problems we are trying to solve here?

package.json
```diff
@@ -37,6 +37,7 @@
     "@babel/traverse": "^7.14.0",
     "@babel/types": "^7.0.0",
     "@jest/create-cache-key-function": "^29.7.0",
+    "@rushstack/node-core-library": "^5.7.0",
```
captbaritone, line 40: I'm hesitant to bring in this dependency for a set of things that look like pretty thin wrappers around node apis. Do you think it could be avoided? What's the cost of not including it?

package.json
```diff
@@ -47,6 +48,8 @@
     "del": "6.0.0",
     "eslint": "^8.57.0",
     "eslint-config-fbjs": "4.0.0",
+    "mock-fs": "^5.2.0",
+    "json-schema-to-typescript": "^13.0.0",
```
captbaritone, line 52: Do we actually use this to generate typescript types, or just for json-schema runtime validation?

robertf224:
@captbaritone yeah that's about right.  For context I'm working on a nice integration layer right now between Relay and Next.js, but the same idea would apply to any tools building on top of Relay.  The two places I'm using what amounts to `loadConfig` is:
- To automatically populate the [Relay component of the Next.js config](https://nextjs.org/docs/architecture/nextjs-compiler#relay) (which sets up the SWC transform)
- I'm providing a `pull-schema` cli command that loads environment variables using [`@next/env`](https://github.com/vercel/next.js/tree/canary/packages/next-env), executes an introspection query using your Relay environment, and saves the schema to the path specified in the `schema` field of your Relay config

Providing a type-safe/standard library for loading Relay's config file will avoid bugs/hacks in tools building on top of Relay,  e.g. see this cli that just hardcodes [`relay.config.json`](https://github.com/tobias-tengler/create-relay-app/blob/79e3722a3b60e442940dfe4f0b720b7cee3536b7/src/misc/ProjectContext.ts#L40) and then does [a bunch of non-type-safe loading and manipulation on top](https://github.com/tobias-tengler/create-relay-app/blob/79e3722a3b60e442940dfe4f0b720b7cee3536b7/src/tasks/ConfigureRelayCompilerTask.ts#L27)

package.json
robertf224: using in a gulp task in the build to automatically generate from the schema file so we don't need to manually keep the two in sync

package.json
robertf224: using it for validating the loaded config against the json schema so we would otherwise need some dependency for doing that (e.g. `ajv` which this library uses under the hood).  also this dep should only be used in dev tools not anything at runtime

scripts/testDependencies.js
```diff
@@ -83,6 +83,7 @@ function testPackageDependencies(topLevelPackagePath, packagePath) {
     'relay-compiler',
     'relay-runtime',
     'react-relay',
+    'relay-config',
```
robertf224, line 86: not sure if this should be in here

robertf224:
btw @captbaritone this is where I have this logic duplicated in the integration library I wrote, would love to move upstream!  https://github.com/robertf224/next-relay/tree/main/packages/next-relay/src/config/relay-config

captbaritone:
Thanks for your work on this! It makes sense to me. A few thoughts:

1. What do you think about moving this directly into the `relay-compiler` NPM package? The compiler is really the source of truth for the shape of the config and where it looks for the config, so the two packages will will always need to stay in sync anyway.
2. Right now the babel transform also implements this search. If we're going to provide this logic in JS can we refactor the babel transform to import/use the new externally provided logic?

robertf224:
how would you want to structure the exports from `relay-compiler`?  looks like right now we basically just export the path to the binary as a string (https://github.com/facebook/relay/blob/main/packages/relay-compiler/index.js) so maybe would want to export a subpath or something to avoid a breaking change

robertf224:
Hey @captbaritone, got around to refreshing this:
- Removed `@rushstack/node-core-library` dep and switched to using `ajv` directly
- Tried moving to relay-compiler per your idea but seemed like it was gonna be messy (wasn't sure how to expose exports without breaking change) so just left as-is
- Ported babel-plugin-relay to use this library

Tests in other packages failing not clear why to me, but lmk what you think

PR #4784, opened 2024-09-06
Make workspace root lazy static to try to avoid cwd race condition

captbaritone:
Trying to get more info on the failure I see here: https://github.com/facebook/relay/actions/runs/10742885102/job/29796930703


captbaritone:
I think it’s a flaky situation due to integration tests running in that process changing the CWD in async tasks in a way that is unpredictable

captbaritone:
This landed as https://github.com/facebook/relay/commit/e1f2177c0cddff932fd04c90ccca068768b07dd8

Issue #4790, opened 2024-09-09
[Relay 18] Relay Resolver return type not getting derived from resolver function


waynezhang1995:
# Creating issues for Relay

## Questions regarding how to use Relay and/or GraphQL
We are experimenting with Relay 18 and have noticed that the return type of a Relay resolver with RelayResolverValue is no longer being derived in the generated TypeScript type definition

For example:
Given the following relay resolver implementation
```ts
type ContactInfo = {
  address: string;
  phone: string;
};

/**
 * @RelayResolver PersonalDetails.contactInfo: RelayResolverValue
 */
export function contactInfo(personalDetailsRef: RelayResolversPostalCodeFragment$key): ContactInfo {
  const data = readFragment(
      graphql`
        fragment RelayResolversPostalCodeFragment on PersonalDetails {
          address
          phone
        }
      `,
      personalDetailsRef,
    );

  // a very dummy example
  return {...data}
}
```

In the generated relay artifact that references the `contactInfo`

```
import { ConcreteRequest, Query } from 'relay-runtime';
import { FragmentRefs } from "relay-runtime";

// This is missing in Relay 18
// import { contactInfo as personalDetailsContactInfoResolverType } from "../RelayResolvers";

export type RelayResolversComplexRequiredThrowQuery$variables = Record<PropertyKey, never>;
export type RelayResolversComplexRequiredThrowQuery$data = {
  readonly viewer: {
    readonly userProperties: {
      readonly personalDetails: {
        readonly contactInfo: ReturnType<typeof personalDetailsContactInfoResolverType> | null;
      } | null;
    };
  };
};
```
Relay used to derive the return type `import { contactInfo as personalDetailsContactInfoResolverType } from "../RelayResolvers";`, but this is now missing in Relay 18

Appreciate any insights. Thanks

We want to keep signal strong in the GitHub issue tracker – to make sure that it remains the best place to track issues that affect the development of Relay.

If you have a question on how to use Relay, please get in touch with community members through one of the channels listed [here](https://relay.dev/help) instead.

## Reporting issues with Relay

We will be using GitHub Issues for our public bugs. We will keep a close eye on this and try to make it clear when we have an internal fix in progress. Before filing a new issue, make sure an issue for your problem doesn't already exist.

The best way to get your bug fixed is to provide a reduced test case. To make reproduction simple for you, and set-up simple for Relay's maintainers, you can use Glitch:
https://glitch.com/edit/#!/remix/relay-starter-kit

[...]

PR #4791, opened 2024-09-11
Adds back the missing resolver type import in Typescript files

Markionium:
When trying to improve Typescript types we made the wrong assumption that the resolver type import was always unused in Typescript files. 
We missed the fact that it is used in the scenario where `RelayResolverType` is provided as the return value of a resolver. (See issue #4790)

The TODO's remain as there is still work to do to add Context support to the resolvers.


captbaritone:
Hey. I'm hitting an internal infra issue importing this PR. I think there's some cases where if a PR needs to get rebased on import it can time out. Sorry to ask this, but do you mind reopening this as a new PR?

Markionium:
@captbaritone do we still need this though? Didn't you do almost exactly the same here?

https://github.com/facebook/relay/commit/4782743b8b0f2bc7a0cc133b92628e92e0ef57ac

captbaritone:
Ah, right. I last track of that. Should we just close this PR?

Markionium:
Yep, closed!

PR #4793, opened 2024-09-12
Replace story.image for story.thumbnail

theodorekouzelis:
The Story component expects the property to be named "thumbnail" https://github.com/relayjs/relay-examples/blob/main/newsfeed/src/components/Story.tsx#L24


captbaritone:
Thanks!

Issue #4755, opened 2024-07-31
Add Typescript validation to pipelines to prevent regressions


drewatk:
We'll need a way to validate Typescript code to ensure that we do not regress the generated code as we are moving towards typechecking checking generated TS files.

              Yeah, for Flow we have our own codebase as a massive test case. For TypeScript we'll need something to echo that setup. Running typescript on integration tests could be a great way to achieve that pretty cheaply.

We have rust utilities that materialize these fixture files into multiple files in a directory and then run the compiler on them. Maybe that test could also run the TypeScript type check on them? Maybe a little brittle to shell out to TypeScript from within the Rust tests, but probably worth it.

Will require a little fiddling to get it to work both internally and in GitHub CI and ensure we keep the same version(s?) of TypeScript, but I bet we can figure it out.

Here's where we run the integration tests: https://github.com/facebook/relay/blob/main/compiler/crates/relay-compiler/tests/relay_compiler_integration.rs

cc @rbalicki2 regarding upstreaming TypeScript types and how we might add tests which assert their correctness.

_Originally posted by @captbaritone in https://github.com/facebook/relay/pull/4753#discussion_r1698864973_

captbaritone:
Thanks for pulling this up into a top-level issue!

PR #4797, opened 2024-09-18
Add satisfies type assertion for typescript

### LINKED ISSUE #4772: Typescript Validation for Resolver Types
Relay Resolvers currently generate a type for Flow which typechecks resolver types using a type cast:

```js
import {User as userRelayModelInstanceResolverType} from "UserTypeResolvers";
// Type assertion validating that `userRelayModelInstanceResolverType` resolver is correctly implemented.
// A type error here indicates that the type signature of the resolver module is incorrect.
(userRelayModelInstanceResolverType: (
  id: User__id$data['id'],
  args: void,
) => LiveState<mixed>);
```

In typescript today, only the `import` statement is generated, with no implementation yet for the type assertion statement. 

The assertion should be added to validate that the imported function matches the expected type. This can leverage the `satisfies` [keyword added in Typescript 4.9](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html).

```ts
import {User as userRelayModelInstanceResolverType} from "UserTypeResolvers";
// Type assertion validating that `userRelayModelInstanceResolverType` resolver is correctly implemented.
// A type error here indicates that the type signature of the resolver module is incorrect.
(userRelayModelInstanceResolverType satisfies (
  id: User__id$data['id'],
  args: void,
) => LiveState<mixed>);
```

This will depend on #4745 is fixed and enabled by default to enable typechecking in generated files. It should also include the new context type which is being implemented in #4704. I will add references to this issue in #4704 where applicable.

It will also require a minimum Typescript version of TS 4.9 or higher for consumers, which can be checked as we implement #4755. We may have to have add a feature flag to disable this for consumers of the library which do not meet this minimum standard.

@captbaritone @alloy @Markionium.

drewatk:
Fixes #4772 

Adds a satisfies assertion to the live resolver import:

```
import {User as userRelayModelInstanceResolverType} from "UserTypeResolvers";
// Type assertion validating that `userRelayModelInstanceResolverType` resolver is correctly implemented.
// A type error here indicates that the type signature of the resolver module is incorrect.
(userRelayModelInstanceResolverType satisfies (
  id: User__id$data['id'],
  args: void,
) => LiveState<mixed>);
```

It will also require a minimum Typescript version of TS 4.9 or higher for consumers, which can be checked as we implement https://github.com/facebook/relay/issues/4755. We may have to have add a feature flag to disable this for consumers of the library which do not meet this minimum standard.


captbaritone:
Nice!!

PR #4821, opened 2024-10-15
Fix type generate of Result type

### LINKED ISSUE #4819: Are the Result<T, Error> type definitions wrong?
I wanted to play around with `@catch` and noticed that the Flow / TypeScript types and the compiler generated errors passed to the Result type don't make much sense.

The compiler will generate the following code:

```flow
export type FooQuery$data = {|
  +me: Result<?{|
    +firstName: ?string,
    +lastName: ?string,
  |}, $ReadOnlyArray<mixed>>,
|};
```

so the error type is `$ReadOnlyArray<mixed>`.

The type definitions in relay-runtime define the errors field on the `ErrorResult` as a `$ReadOnlyArray<Error>`:
https://github.com/facebook/relay/blob/e357063c26d0aaf0a5027ba591138ad172cd103f/packages/relay-runtime/experimental.js#L46C2-L55C67

So you'd end up with the `ErrorResult.errors` being `Array<Array<Error>>`. This doesn't seem right to me?

tobias-tengler:
Fixes #4819

### DIFF
--- a/compiler/crates/relay-typegen/src/visit.rs
+++ b/compiler/crates/relay-typegen/src/visit.rs
@@ -1805,7 +1805,7 @@ fn make_result_type(typegen_context: &'_ TypegenContext<'_>, value: AST) -> AST 
 
     AST::GenericType {
         outer: *RESULT_TYPE_NAME,
-        inner: vec![value, AST::ReadOnlyArray(Box::new(error_type))],
+        inner: vec![value, error_type],
     }
 }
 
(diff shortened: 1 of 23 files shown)


tobias-tengler:
Good ol' Cargo.lock acting up again. How refreshing 😅
Guess this needs to be merged first again: #4817

tobias-tengler:
Dammit @itamark was faster 😁

PR #4831, opened 2024-10-25
Document live value lifetime

captbaritone:


subtleGradient:
Maybe also add a note to add a `queryCacheExpirationTime` option to `new RelayStore` so that the garbage man will come on on time and sweep away the trash

PR #4833, opened 2024-10-25
Fix publishing npm packages

zpao:
This probably blames back to #4816 but I haven't completely proven that.

Essentially, this looks likely that something changed in npm. `npm-package-arg` thinks that `dist/foo` is a reference to a github repo, not a local path. Explicitly making that a relative path (`./dist/foo`) ensures it doesn't.

I haven't proven this will fix the issue but running `require('npm-package-arg')('dist/foo')` returns an object indicating it's treating it as a git repo. `require('npm-package-arg')('./dist/foo')` treats it like the directory it is.

My assumption is that this treatment like a git repo is that npm is deciding to go look it up and validate that the `main` tag exists. That might be fine but it's looking for `dist/foo` which doesn't exist and then that results in the git failure we're seeing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -276,7 +276,7 @@ jobs:
         if: github.ref == 'refs/heads/main' || github.ref_type == 'tag' && startsWith(github.ref_name, 'v')
         run: |
           for pkg in dist/*; do
-            npm publish "$pkg" ${TAG}
+            npm publish "./$pkg" ${TAG}
           done
         env:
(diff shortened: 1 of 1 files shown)


zpao:
Alright. I'm confident on the blame and reasoning. You were using node 14, which used npm 6. https://github.com/npm/cli/commit/a1f9be8a7f9b7a3a813fc3e5e705bc982470b0e2 made the switch to using pacote, which is where the actual bit to hit git happens.

PR #4839, opened 2024-11-02
Add __id to LSP completions

tobias-tengler:
Two of our developers reached out, because they weren't sure if they're using `__id` in the right places, since the LSP wouldn't suggest it.

I've extended the LSP completions to include the `__id` field to clear up those confusions.

To my knowledge the `__id` field should be a valid selection on any complex type. If there are special cases, please let me know and I'll update the implementation for the suggestion accordingly.

### DIFF
--- a/compiler/crates/relay-lsp/src/completion.rs
+++ b/compiler/crates/relay-lsp/src/completion.rs
@@ -47,6 +47,8 @@ use lsp_types::Documentation;
 use lsp_types::InsertTextFormat;
 use lsp_types::MarkupContent;
 use lsp_types::MarkupKind;
+use schema::field_descriptions::CLIENT_ID_DESCRIPTION;
+use schema::field_descriptions::TYPENAME_DESCRIPTION;
 use schema::Argument as SchemaArgument;
 use schema::Directive as SchemaDirective;
 use schema::InputObject;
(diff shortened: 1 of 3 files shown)


compiler/crates/relay-lsp/src/completion.rs
```diff
@@ -868,13 +870,31 @@ fn completion_items_for_request(
     }
 }
 
-fn resolve_completion_items_typename(type_: Type, schema: &SDLSchema) -> Vec<CompletionItem> {
+fn resolve_special_completion_items(type_: Type, schema: &SDLSchema) -> Vec<CompletionItem> {
+    let __id_item = CompletionItem {
+        label: "__id".to_owned(),
```
captbaritone, line 875: I wonder if this should be exposed by the SDLSchema? A little nervous having this fieldname and type hard coded in multiple modules.

captbaritone (commented):
Good catch. Overall looks good. One question before I import.

compiler/crates/relay-lsp/src/completion.rs
captbaritone: @tobias-tengler Thoughts on this? Would love to merge this improvement.

PR #4841, opened 2024-11-05
Docs: Fix fragments tutorial thumbnail typo

danielstocks:
Hi, I just ran through the tutorial and think I found a small typo.

In the [example code](https://github.com/relayjs/relay-examples.git) referenced in the tutorial the field is called `thumbnail` and not `image`. Fixed tutorial example to reflect this.

### DIFF
--- a/website/versioned_docs/version-v18.0.0/tutorial/fragments-1.md
+++ b/website/versioned_docs/version-v18.0.0/tutorial/fragments-1.md
@@ -25,7 +25,7 @@ const NewsfeedQuery = graphql`
           url
         }
       }
-      image {
+      thumbnail {
         url
       }
     }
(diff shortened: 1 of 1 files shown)


captbaritone:
Thanks!

You've updated only the v18 page here. Here is the source of truth: https://github.com/facebook/relay/blob/3eb627d7dcce8131890e4c25cc54aaa235040dd6/website/docs/tutorial/fragments-1.md#L4

(You can back port to v18, 18.1 etc if you want, but the important part is to update the source of truth doc

danielstocks:
@captbaritone Thanks, yeah I saw now it was already [fixed](https://github.com/facebook/relay/pull/4793/files) in the source of truth. Didn't realise since I was following the tutorial for v18 which is the current "default".

I'll leave it up to you if you want to merge or close this :)

Issue #4531, opened 2023-11-20
several errors possibly


lycha0206:
in https://relay.dev/docs/tutorial/connections-pagination/#improving-the-loading-experience-with-usetransition
there is no CommentsLoadingSpinner
I had to use SmallSpinner

but even so, the spinner is flashing for a split second and is gone (like 0.01s). For this one, I am not sure if it is some mistake I made or is it your code.

And then at https://relay.dev/docs/tutorial/connections-pagination/#step-5--call-usepaginationfragment

why is the function Newsfeed() all of a sudden becomes

NewsfeedContents({viewer}) {

and why adding `{viewer}` when down below in Step 6 it is immediately `{query}`

also the line

const {data, loadNext} = usePaginationFragment(NewsfeedFragment, viewer);

the NewsfeedFragment should be NewsfeedContentsFragment

And then in Step 6, what is the `{query}`?

as in `function NewsfeedContents({query}) {`
there is no query passed in.

I had to change the code to the following to work:

```tsx
export default function Newsfeed() {

  const queryData = useLazyLoadQuery<NewsfeedQueryType>(NewsfeedQuery, {});

  const { data, loadNext, hasNext, isLoadingNext } = usePaginationFragment(
    NewsfeedContentsFragment,
    queryData
  );

  function onEndReached() {
    loadNext(3);
  }

  const storyEdges = data.viewer.newsfeedStories.edges;
```

Then the page works, but the `viewer` in `  const storyEdges = data.viewer.newsfeedStories.edges;
` is underlined red. The message is: Property 'viewer' does not exist on type 'unknown'.

Also `storyEdge` in `{storyEdges.map((storyEdge) => (` is also underlined red.
The message is: Parameter 'storyEdge' implicitly has an 'any' type.

walliby:
I also cannot get the StoryCommentsSection to suspend with useTransition. Although looks like it does very briefly, before the network call finishes. There is an extra state update for some reason which changes isPending to false. Maybe related to one of these?

https://github.com/facebook/relay/issues/3082
https://github.com/facebook/relay/issues/4526

walliby:
@lycha0206 looking at the api doc for usePaginationFragment, you can return `isLoadingNext` instead of trying to `useTransition`. Here is how I got mine to work...
```
export default function StoryCommentsSection({ story }: Props) {
  const { data, loadNext, isLoadingNext } = usePaginationFragment(
    StoryCommentsSectionFragment,
    story
  );

[...]

Issue #4526, opened 2023-11-16
`refetch({}, "network-only")` suspends, even when inside `startTransition`


Zizico2:
## Reporting issues with Relay

`refetch({}, "network-only")` always causes React to suspend, even when inside `startTransition`. This is annoying, specially when rendering lists with `usePaginationFragment`, where you would want to show the existing items, while re-fetching the new ones, and not jump to a suspend boundary.  Furthermore, I think the difference in behavior for different `FetchPolicy`s is not that intuitive for a beginner. And `loadNext`, for example, will never suspend. I think everything that makes a network request should suspend, to avoid these confusions, and the user should be able to opt out of it with `startTransition`. This was the approach Apollo took (https://github.com/apollographql/apollo-client/issues/10676). 

I realize this might be too dramatic of a change, but at least we should be able to use `startTransition` to prevent `refetch` from suspending.

richardwu:
@Zizico2 did you ever figure out a fix for this? 

Seems like `refetch` ignores `startTransition` regardless of the `fetchPolicy`.

I'm on 16.2.0.

steveluscher:
Was `useTransition` introduced before React 18? You may be importing an API compatible version of `useTransition` from {somewhere} that's actually a noop.

steveluscher:
https://cdn.jsdelivr.net/npm/react@16.2.0/cjs/react.development.js

`useTransition` – not there.

richardwu:
Sorry I meant 16.2.0 react-relay. I'm on 18.2.0 react.

alexandernanberg:
I having this issue as well, e.g. pull down to refresh in React Native. Has anyone found a workaround? Using `fetchPolicy` `store-and-network` doesn't work well either since there is no indicator that content is being fetched

levibuzolic:
@alexandernanberg are you running in the new architecture? Transitions are a feature of concurrent rendering which is only available in React Native with the new architecture.

alexandernanberg:
Indeed seems to be an issue isolated to React Native, even though I'm on `0.76` and the new arch. Set up a similar env with web/nextjs and React 19, then it works as expected

lbxa:
Having the same issue as @alexandernanberg @Zizico2. Not being able to opt out of Suspense destroys the simple UX of pull to refresh lists with refetchable fragments in React Native. I'm running relay@18 on the new Arch.

yuliswe:
This is still an issue with react-native 0.76.5
[...]

Issue #3082, opened 2020-05-20
usePaginationFragment's refetch triggers twice when using useTransition 


Sicria:
I'm using the following `experimental` versions when running encountering this issue.

```
"react": "0.0.0-experimental-f42431abe"
"react-dom": "0.0.0-experimental-f42431abe"
"react-relay": "0.0.0-experimental-8cc94ddc"
```

I've been following the reference and seem to be hitting a strange bug. https://relay.dev/docs/en/experimental/api-reference#usepaginationfragment

I'm loading a component which is supplied with preloaded data, I'm then displaying that data in a list and allowing a user to change the `orderStatus` as a filter for that data.

I'm using the `usePaginationFragment` hook and the `refetch` method from that. All of that seems to work as intended but does suspend the component which is expected.

Applying a `useTransition` hook around the `refetch` causes some strange behaviour.

It seems the first change works as intended, but subsequent changes causes the `refetch` to trigger again and call the `onComplete` method.

I've tried commenting out the `startTransition` and it works fine but suspends, enabling it again causes `refetch` to trigger twice.

```
const queryData = usePreloadedQuery(QUERY, ordersQuery);
const {
  data: { orders },
  refetch,
  loadNext,
  hasNext,
  isLoadingNext,
} = usePaginationFragment(REFETCHABLE, queryData);
const [startTransition, isPending] = useTransition({ timeoutMs: 10 * 1000 });
const [orderStatus, setOrderStatus] = useState('all');
const handleOrderStatusChange = (key) => {
  console.log(`TRANSITION:${key}`);
  startTransition(() => {
    setOrderStatus(key);
    console.log(`REFETCH:${key}`);
    refetch({
      count: ordersQuery?.variables?.count,
      cursor: null,
      orderStatus: key,
      startDate,
    }, {
      fetchPolicy: 'network-only',
      onComplete: () => console.log(`COMPLETE:${key}`),
    });
  });
};

console.log('RENDER', { orders: orders?.edges?.length, orderStatus, startDate });
```

In the code above, following these steps it reproduces the issue.

- Changing filter from `all` to `open`
  - `TRANSITION:open`, `REFETCH:open`, `<REQUEST SENT:open>`, `COMPLETE:open`
- Changing the filter from `open` to `complete`
  - `TRANSITION:complete`, `REFETCH:complete`, `<REQUEST SENT:open>`, `COMPLETE:open`, `<REQUEST SENT:complete>`, `COMPLETE:complete`
[...]

PR #4842, opened 2024-11-07
Update docs on usePaginationFragement & transitions in Relay tutorial.

danielstocks:
I found out that `useTransition` doesn't work when using `usePaginationFragement`. To get it to work the way it's intended in the tutorial I had to use `isLoadingNext` boolean prop that is returned from `usePaginationFragement` instead. Not sure if this a bug, or by design.

As a beginner, I spent quite some time trying to figure out why my code wasn't working, after a bit of digging I ran into some potenially related issues:
- https://github.com/facebook/relay/issues/4531#issuecomment-1867831999
- https://github.com/facebook/relay/issues/4526
- https://github.com/facebook/relay/issues/3082

In this PR I've updated the code in "Improving the Loading Experience with useTransition" example not to confuse newcomers. Also the section about "results that aren’t immediate should be wrapped in a React transition" as it no longer applies in this particular context.

I also replaced `<CommentsLoadingSpinner />` (which doesn't exist?) with `<SmallSpinner />` which seems more appropriate in the particular example.

### DIFF
--- a/website/docs/tutorial/connections-pagination.md
+++ b/website/docs/tutorial/connections-pagination.md
@@ -247,17 +247,13 @@ Run `npm run relay`. Now the Load More button should cause another three comment
 
 As it stands, there’s no user feedback when you click the “Load More” button until the new comments have finished loading and then appear. Every user action should result in immediate feedback, so let’s show a spinner while the new data is loading — but without hiding the existing UI.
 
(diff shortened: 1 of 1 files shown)


captbaritone:
Sorry for the confusion this caused! I'll import and we can look into if this is expected behavior or not.

PR #4846, opened 2024-11-18
Context not properly provided through data injector and subscriptions

Markionium:
This PR fixes two issues we've encountered.

1. ResolverContext isn't passed through the `resolverDataInjector`. This surfaces specifically when using nested resolvers.
2. When updates propagate through the RelayStoreSubscriptions the context value is lost due the RelayStoreSubscriptions instance not receiving the context correctly.

I noticed two of the 3 paths in `resolverDataInjector` do not seem to be covered by tests.


captbaritone:
Looks good to me! Thanks for catching this. I'm out next week so I won't be able to land this until after Thanksgiving. I'll also try to see if we can define some additional tests to cover the cases you noticed were not tested.

PR #4848, opened 2024-11-19
[WIP] Add an example project to the main Relay repo

captbaritone:
While we have a collection of Relay example apps in https://github.com/relayjs/relay-examples, it would be useful to have a single canonical example app which stays in sync with Relay as we develop it, and which can demonstrate each of Relay's features.

This (currently still WIP) PR aims to stub out such an example app which we can expand over time.


sibelius:
I think it is easier to contribute to relay-examples

and we can have some different types of examples there, like nextjs, vite, rspack and other frameworks

itsjoeoui:
@captbaritone I am happy to help out with this if you need an extra hand :)

PR #4850, opened 2024-11-19
Limit WalkDir to the root directories referenced by the configuration

### LINKED ISSUE #4849: Relay Compiler iterates over every file under root when Watchman is disabled
Relay compiler falls back to using WalkDir to iterate over the filesystem when Watchman isn't available. The problem is that this implementation walks over every single file under `root_dir`, which can be your entire repository, rather than limiting it to the files being referenced by the Relay configuration.

This can be a lot of wasted work, and is going to be slow for larger mono-repositories.

sverrejoh:
When not using Watchman the Relay Compiler will iterate over all files under `root_dir`, which can possibly include the entire repository. This is unnecessary work and might slow down large repositories.

The watchman implementation only listens to files under the roots defined by the config, but this is not implemented for the directory walking implementation in `WalkDirFileSource`.  This change moves the `get_all_roots()` functionality from `WatchmanFileSource` to `Config` and shares it with `WalkDirFileSource` to make it only iterate over the folders configured as targets in Relay.

This fixes #4849


captbaritone:
Nice! Makes sense to me. I'll import, but I'm out next week and may not be able to merge until after Thanksgiving.

PR #4862, opened 2024-12-10
Add plural fragment support to `observeFragment()`

XiNiHa:
This PR adds plural fragment support to `observeFragment()` (and `waitForFragmentData()` which uses `observeFragment()` under the hood)

It should be noted that the Relay environment currently doesn't support subscribing to plural fragment updates, so the implementation is done by subscribing to each selector and merging the states. Since the current implementation doesn't do any batching of notifications that are made from each selector subscription, it can be somewhat inefficient if the consumer doesn't batch on their own.

### DIFF
--- a/packages/relay-runtime/store/__tests__/observeFragment-test.js
+++ b/packages/relay-runtime/store/__tests__/observeFragment-test.js
@@ -337,6 +337,189 @@ test('read deferred fragment', async () => {
   });
 });
 
+test('observes a plural fragment', async () => {
+  const query = graphql`
+    query observeFragmentTestPluralQuery {
+      nodes(ids: ["1", "2"]) {
+        ...observeFragmentTestPluralFragment
+      }
+    }
+  `;
+
+  const fragment = graphql`
+    fragment observeFragmentTestPluralFragment on User @relay(plural: true) {
+      name
+    }
+  `;
+
+  const environment = createMockEnvironment({
+    store: new LiveResolverStore(new RelayRecordSource()),
+  });
+  const variables = {};
+  const operation = createOperationDescriptor(query, variables);
+  environment.commitPayload(operation, {
+    nodes: [
+      {id: '1', __typename: 'User', name: 'Alice'},
+      {id: '2', __typename: 'User', name: 'Bob'},
+    ],
+  });
+  const {data} = environment.lookup(operation.fragment);
+  // $FlowFixMe Data is untyped
+  const observable = observeFragment(environment, fragment, data.nodes);
+  const result = await observable.toPromise();
+  expect(result).toEqual({
+    state: 'ok',
+    value: [{name: 'Alice'}, {name: 'Bob'}],
+  });
+});
+
+test('Missing required data on plural fragment', async () => {
+  const query = graphql`
+    query observeFragmentTestMissingRequiredPluralQuery {
+      nodes(ids: ["1", "2"]) {
+        ...observeFragmentTestMissingRequiredPluralFragment
+      }
+    }
+  `;
+
+  const fragment = graphql`
+    fragment observeFragmentTestMissingRequiredPluralFragment on User
+    @relay(plural: true) {
+      name @required(action: THROW)
+    }
+  `;
+
+  const environment = createMockEnvironment();
+  const variables = {};
+  const operation = createOperationDescriptor(query, variables);
+  environment.commitPayload(operation, {
+    nodes: [
+      // Name is null despite being required
+      {id: '1', __typename: 'User', name: null},
+      {id: '2', __typename: 'User', name: 'Bob'},
+    ],
+  });
+
+  const {data} = environment.lookup(operation.fragment);
+  // $FlowFixMe Data is untyped
+  const observable = observeFragment(environment, fragment, data.nodes);
+  withObservableValues(observable, results => {
+    expect(results).toEqual([
+      {
+        error: new Error(
+          "Relay: Missing @required value at path 'name' in 'observeFragmentTestMissingRequiredPluralFragment'.",
+        ),
+        state: 'error',
+      },
+    ]);
+  });
+});
+
+test('Field error with @relay(plural: true) @throwOnFieldError', async () => {
+  const query = graphql`
+    query observeFragmentTestPluralThrowOnFieldErrorQuery {
+      nodes(ids: ["1", "2"]) {
+        ...observeFragmentTestPluralThrowOnFieldErrorFragment
+      }
+    }
+  `;
+
+  const fragment = graphql`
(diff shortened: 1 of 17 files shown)


packages/relay-runtime/store/observeFragmentExperimental.js
```diff
@@ -173,6 +170,67 @@ function observeSelector<TFragmentType: FragmentType, TData>(
...
+          if (state.state === 'ok') {
+            acc.value.push(state.value);
+            return acc;
+          } else {
+            return state;
+          }
+        },
+        {
+          state: 'ok',
+          // $FlowFixMe[incompatible-cast]: TData is enforced to be an array
```
XiNiHa, line 211: Not sure why this should be marked as `$FlowFixMe` when `TData` has `: Array<mixed>` enforcement

captbaritone:
Exciting! I'll take a look tomorrow.

packages/relay-runtime/store/observeFragmentExperimental.js
```diff
@@ -173,6 +170,67 @@ function observeSelector<TFragmentType: FragmentType, TData>(
...
+  const snapshots = fragmentSelector.selectors.map(selector =>
+    environment.lookup(selector),
+  );
+
+  return Observable.create(sink => {
+    const states = snapshots.map((snapshot, index) =>
+      snapshotToFragmentState(
+        environment,
+        fragmentNode,
+        fragmentSelector.selectors[0].owner,
```
captbaritone, line 190: Either use the `index` here or omit it from the callback arguments on line 186

packages/relay-runtime/store/observeFragmentExperimental.js
```diff
@@ -173,6 +170,67 @@ function observeSelector<TFragmentType: FragmentType, TData>(
...
+        snapshot,
+      ),
+    );
+
+    const reduceState = () =>
+      states.reduce<FragmentState<TData>>(
+        (acc, state) => {
+          if (acc.state !== 'ok') {
+            return acc;
```
captbaritone, line 199: I wonder if this would be easier to read as a for loop with a `break`? It's a little hard to follow right now, and I think right now we still have to continue iterating the states even after we hit one that's not "ok". Would be nicer if we could just bail and return the loading or error state right away.

Maybe we could define a generic function (outside of this closure) which does this transformation:

```ts
function namingIsHard<T>(states: Array<FragmentState<T>>): FragmentState<Array<T>> {
  const values: Array<T> = [];
  for(let i = 0; i < states.length; i++) {
    const state = states[i];
    if(state.state === 'ok') {
      values.push(state.value);
    } else {
      return state;
    }
  }
  return { state: "ok", value: values };
}
```

packages/relay-runtime/store/observeFragmentExperimental.js
```diff
@@ -173,6 +170,67 @@ function observeSelector<TFragmentType: FragmentType, TData>(
...
+  environment: IEnvironment,
+  fragmentNode: Fragment<TFragmentType, TData>,
+  fragmentSelector: PluralReaderSelector,
+): Observable<FragmentState<TData>> {
+  const snapshots = fragmentSelector.selectors.map(selector =>
+    environment.lookup(selector),
+  );
+
+  return Observable.create(sink => {
+    const states = snapshots.map((snapshot, index) =>
```
captbaritone, line 186: Might be worth a comment up here explaining that `states` is mutable and its values will get updated in place overtime

packages/relay-runtime/store/observeFragmentExperimental.js
```diff
@@ -173,6 +170,67 @@ function observeSelector<TFragmentType: FragmentType, TData>(
...
+
+    const subscriptions = snapshots.map((snapshot, index) =>
+      environment.subscribe(snapshot, latestSnapshot => {
+        states[index] = snapshotToFragmentState(
+          environment,
+          fragmentNode,
+          fragmentSelector.selectors[0].owner,
+          latestSnapshot,
+        );
+        sink.next(reduceState());
```
captbaritone, line 226: Maybe a comment here calling out that this does not batch updates, and if a store update impacting multiple items in the list is published, we will notify the subscriber multiple times?

packages/relay-runtime/store/__tests__/observeFragment-test.js
```diff
@@ -337,6 +337,189 @@ test('read deferred fragment', async () => {
...
+      {
+        error: new Error(
+          "Relay: Resolver error at path 'always_throws' in 'observeFragmentTestResolverErrorWithPluralThrowOnFieldErrorFragment'.",
+        ),
+        state: 'error',
+      },
+    ]);
+  });
+});
+
```
captbaritone, line 522: Can we add some additional tests:

1. A test demonstrating a store update/publish that touches multiple items in the list triggering multiple observable updates?
2. A test or two demonstrating that `waitForFragmentData` also works with plural fragments?

captbaritone (commented):
Overall this looks great! Just a few nits.

Could you also add some notes to the docs about the behavior for plural fragments and letting folks know they can debounce if they want to batch?

packages/relay-runtime/store/observeFragmentExperimental.js
XiNiHa: https://github.com/facebook/relay/pull/4862/commits/4a5629881cafb7d8b53fc9d18314b9afe875f749

packages/relay-runtime/store/observeFragmentExperimental.js
XiNiHa: https://github.com/facebook/relay/pull/4862/commits/70f047e6725e7aa9e60277e6b15054b7405e07ee

packages/relay-runtime/store/observeFragmentExperimental.js
XiNiHa: https://github.com/facebook/relay/pull/4862/commits/06903bfa7a0030c3d5569051b47479e968627bf8

packages/relay-runtime/store/observeFragmentExperimental.js
XiNiHa: https://github.com/facebook/relay/pull/4862/commits/06903bfa7a0030c3d5569051b47479e968627bf8

packages/relay-runtime/store/__tests__/observeFragment-test.js
XiNiHa: https://github.com/facebook/relay/pull/4862/commits/e596faac720d127baea60f84ee2c9b286ffbc72d

captbaritone (approved):
This looks so good. Thanks for filling this gap and making this API more usable for everyone!

PR #4868, opened 2024-12-12
Fix edge case issue around unsubscribing live queries

XiNiHa:
This only happens when...

- A live query is passed to `useLazyLoadQueryNode()`
- The initial fetch doesn't suspend
- Two components subscribe to the same query with the same operation descriptor
- The first subscribed component gets unmounted first

The fix is pretty simple though 😋

### DIFF
--- a/packages/react-relay/relay-hooks/QueryResource.js
+++ b/packages/react-relay/relay-hooks/QueryResource.js
@@ -30,7 +30,7 @@ import type {
 const LRUCache = require('./LRUCache');
 const SuspenseResource = require('./SuspenseResource');
 const invariant = require('invariant');
-const {isPromise} = require('relay-runtime');
+const {RelayFeatureFlags, isPromise} = require('relay-runtime');
 const warning = require('warning');
 
 const CACHE_CAPACITY = 1000;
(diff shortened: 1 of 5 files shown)


packages/react-relay/relay-hooks/QueryResource.js
```diff
@@ -466,16 +466,10 @@ class QueryResourceImpl {
...
               ...subscription,
-              unsubscribe: () => {
-                // Only live queries should have their network requests canceled.
-                if (operationIsLiveQuery(operation)) {
-                  subscription.unsubscribe();
-                }
-              },
-            };
-            observerStart(subscriptionWithConditionalCancelation);
+              unsubscribe: () => {},
```
captbaritone, line 471: Do live queries still get unsubscribed through some other means?

XiNiHa: Yeah it's managed by the `.retain()` stuff and is also ensured with the test added (the `.toThrow()` part)

packages/react-relay/relay-hooks/useLazyLoadQueryNode.js
```diff
@@ -69,7 +67,7 @@ hook useLazyLoadQueryNode<TQuery: OperationType>({
       fetchObservable,
       fetchPolicy,
       renderPolicy,
-      {start: startFetch, complete: completeFetch, error: completeFetch},
+      undefined,
```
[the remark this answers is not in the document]
XiNiHa: This was for calling `.unsubscribe()` when the component unmounts. However since the unsubscribe is now a no-op there's no point for keeping this anymore

captbaritone:
@XiNiHa I think I've been hesitant to merge this for fear that there are some edge cases I'm not aware of here and that it could break things when I roll it out. Would you be willing to add this change (fix) behind a feature flag?

https://github.com/facebook/relay/blob/4130d410ae1481b99ece8798d85d79e70557332e/packages/relay-runtime/util/RelayFeatureFlags.js#L16

captbaritone:
@XiNiHa Curious if you'll be able to add a flag here.

XiNiHa:
Oops, missed the notification 😅 Will make it done today

XiNiHa:
The test started to pass even without the fix 🤔 let me figure out what's happening...

captbaritone:
@XiNiHa Any progress here?

XiNiHa:
I haven't looked into this for a while. Maybe we could just close the PR since it wasn't possible to reproduce the issue after rebasing....

PR #4876, opened 2025-01-09
Live resolvers: return value when throw

pavelglac:
Following the issue https://github.com/facebook/relay/issues/4877

This PR aims to return null instead of undefined for live resolvers when the resolver throws an error. That is based on  [docs ](https://github.com/facebook/relay/blob/main/website/docs/guides/relay-resolvers/errors.md)

### DIFF
--- a/packages/react-relay/__tests__/LiveResolvers-test.js
+++ b/packages/react-relay/__tests__/LiveResolvers-test.js
@@ -749,7 +749,7 @@ describe('Live Resolver with Suspense and Missing Data', () => {
     );
   });
 
-  it('should render undefined value for missing data in live resolver field', () => {
+  it('should render null value for missing data in live resolver field', () => {
     const source = RelayRecordSource.create({
       'client:root': {
         __id: 'client:root',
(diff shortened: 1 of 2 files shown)


captbaritone:
Looks like this is specifically covering the case where a live resolver reads a fragment which is missing data?

Just clarifying since I think it's worth considering if we want to treat that as equivalent to "an individual resolver throws an error". I'm not saying they _shouldn't_ be treated as equivalent, but I think it merits contemplating the implications.

pavelglac:
@captbaritone This is the case when resolver does not return `LiveState` at all. If resolver throws in read that is different case. In this case the result is null for resolver but the value is not saved. In version 17 it was handled by [ResolverCache.js](https://github.com/facebook/relay/blob/v17.0.0/packages/relay-runtime/store/ResolverCache.js#L178-L192) which was saving value from result.
 
````javascript
/**
 * @RelayResolver Query.counter: Int
 * @live
 */
export function counter(){
  throw new Error("Error");

  return {
    read: ...
    subscribe: ...,
  };
}
````

captbaritone:
I've imported this. Let me look at it more closely in my editor. Thanks!

captbaritone:
@pavelglac Sorry if I'm missing something, but looking more closely at the test cases that have changed it still looks like ~all of them are currently `undefined` because we are missing one or more value in the resolvers's rootFragment.

In this case, I think undefined is more correct since undefined is what we use in Relay to denote that we don't have the value in the store. In this case the value is derived but some dependency of the field is not in the store, but undefined still feels idiomatically correct for "we can't provide this value because we are missing data", even if the data we're missing is not the actual value but a dependency.

Am I missing something?

pavelglac:
@captbaritone make sense. I was just confused this was changed between version but this makes more sense. Closing

PR #4885, opened 2025-01-20
Fix `observeFragment` triggering unhandled rejections on network error

XiNiHa:
This PR fixes an issue where `observeFragment` triggers unhandled rejections on network error, which happens because of the dangling promise created and dropped.

### DIFF
--- a/packages/relay-runtime/store/__tests__/__generated__/observeFragmentTestNetworkErrorFragment.graphql.js
+++ b/packages/relay-runtime/store/__tests__/__generated__/observeFragmentTestNetworkErrorFragment.graphql.js
@@ -1,0 +1,72 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @oncall relay
+ *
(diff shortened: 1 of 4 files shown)


captbaritone:
Thanks for the PR. Is it possible to construct a regression test?

XiNiHa:
> Thanks for the PR. Is it possible to construct a regression test?

You can now switch to https://github.com/facebook/relay/pull/4885/commits/4e2c53a2354ed54a91e58cd917d7267cb9ec82cf and see the test failing :)

btw, while working on the test, I realized that the current behavior is somewhat unnatural since it doesn't notify the observer at all when network errors occur. While I also think that it's inappropriate to return network errors from fragment observation, I believe it should fallback to returning missing data at least 🤔

captbaritone:
Good catch that we don't model query errors as fragment errors. Given that's existing behavior, maybe we can tackle that as a second PR. Mind opening an issue?

XiNiHa:
> Good catch that we don't model query errors as fragment errors. Given that's existing behavior, maybe we can tackle that as a second PR. Mind opening an issue?

#4888 ✅

captbaritone:
I've run into an issue merging this due to subtle differences in how Jest works internally vs here in open source. I'm working to resolve that issue.

PR #4922, opened 2025-03-01
Remove Relay Resolver compiler feature flag from the docs

bluegoo192:
On my machine and Relay v18.2.5, using this flag will cause an error and prevent the compiler from starting

Looks like it was removed here: https://github.com/facebook/relay/commit/4e5377bd632037a46c044399f3e21aac2c41dad8


captbaritone:
We're in a tricky spot right now. The versioned docs here are for 18.0.0. The flag was removed after 18.0.0 (in a minor version). The flag is removed in the "Next" version of the docs. We don't currently have a good way to show API changes affecting minor versions in our docs.

If you want to update this to be a comment clarifying that the flag was removed since v18.0.0 that would be helpful! Otherwise we'll have to wait until v19

bluegoo192:
thanks for the clarification! I totally didn't pay attention to when the change was released. Looks like it was in [18.1.0](https://github.com/facebook/relay/releases/tag/v18.1.0), updated to reflect that

Tbh I never checked the "Next" docs to see they were already updated, so if it's more trouble than it's worth to update these docs for a single version, I don't mind closing the PR either

PR #4930, opened 2025-03-04
Back out "Remove loadQuery during render warning"

captbaritone:
Differential Revision: D70524672


captbaritone:
I've exported this as a PR to see how our OSS tests work out here.

PR #4948, opened 2025-04-09
relay-compiler: allow bare paths for CommonJS provider references

### LINKED ISSUE #4456: Providers as modules or relative paths
We're adopting Relay for the Netflix TVUI application, and one pain point we've come across is that we have providers that we want to share across our app, but because `provider` is always interpreted as a file relative to the fragment (as of https://github.com/facebook/relay/commit/9393fe5b75d95df37daaab750f13ea72b4420be4) we can't do that.

This would either be a breaking change or have to be behind a flag.

Right now the following fragment

```tsx
fragment useBobHeaderQuery_titleImageFragment on Video
    @argumentDefinitions(
        titleImageSize: { type: "ArtworkDimension", provider: "./useBobHeaderQuery_titleImage.relayprovider" }
        artworkCapability: { type: "ArtworkCapability", provider: "@tvui/providers/artworkCapability" }
) {
    ...
}
```

Compiles to:

```tsx
  "__relay_internal__pv__useBobHeaderQuery_titleImagerelayprovider": require('./../useBobHeaderQuery_titleImage.relayprovider'),
  "__relay_internal__pv__tvuiprovidersartworkCapability": require('./@tvui/providers/artworkCapability')
```

Our proposal is that it will instead resolve to:

```tsx
  "__relay_internal__pv__useBobHeaderQuery_titleImagerelayprovider": require('./../useBobHeaderQuery_titleImage.relayprovider'),
  "__relay_internal__pv__tvuiprovidersartworkCapability": require('@tvui/providers/artworkCapability')
```

In other words, if the path starts with a `.` it is interpreted as relative, and if not, it is global. This also brings these module strings more in-line with how CommonJS resolvers work. 

Ideally this change would apply to all such module strings, not just providers. This would also have a side benefit of removing some of the custom code around Haste module handling for JS modules.

Finally just want to note that I am happy to contribute this feature if you'd be willing to accept it!

cc @alunyov, @Lalitha-Iyer

sjbarag:
All CommonJS modules were previously currently forced to be relative to the file they were referenced in. This made some runtime JS integrations impossible, e.g. reusing an NPM package's default export as a provider (since `provider: "@example/foo"` is emitted as `provider: require("./@example/foo")`). Add a new config setting `relativizeJsModulePaths` to allow consumers to opt out of this automatic `./` prefix. The default behavior is unchanged, so this ends up being a backwards-compatible change.

fixes #4456


sjbarag:
CC @alunyov @Lalitha-Iyer @ianobermiller

CLA update is in-progress via email, since this is a Netflix contribution 🙂

sjbarag:
My main questions are on the name of `relativizeJsModulePaths` (open to suggestions here!) and whether it should be a top-level config value or a feature flag (I'll defer to the relay maintainers on that as well 😄)

Happy to make any changes here!

compiler/crates/relay-codegen/src/printer.rs
```diff
@@ -549,18 +551,12 @@ impl<'b> JSONPrinter<'b> {
                 self.write_js_dependency(
                     f,
                     ModuleImportName::Default(format!("{}_graphql", variable_name).intern()),
-                    Cow::Owned(format!(
-                        "{}.graphql",
-                        get_module_path(self.js_module_format, *key)
-                    )),
+                    Cow::Owned(format!("{}.graphql", self.get_module_path(*key))),
```
sjbarag, line 554: 🤔 Looks like this shouldn't respect the new `relativizeJsModulePaths` when `self.js_module_format` is `CommonJS`, since it should always be a reference to a file emitted by the relay compiler. There might be one or two other edge-cases like this. I'll dig into this tomorrow!

sjbarag: Fixed via an explicit `ResourceOrigin::Artifact` / `ResourceOrigin::SourceFile` argument that lets `self.get_module_paths` callers make their intentions explicit.

PR #4950, opened 2025-04-10
Fix raw response type of `@semanticNonNull` field in inline fragment

### LINKED ISSUE #4949: Raw response type of `@semanticNonNull` field is different when selected in a type refinement with `@throwOnFieldError`
Consider the following schema:

```graphql
type Query {
  blogPageBySlug(slug: String!): BlogPage
  node(id: ID!): Node
}

type BlogPage implements Node {
  id: ID!
  title: String @semanticNonNull
}

interface Node {
  id: ID!
}
```

And the following query:

```graphql
query somethingSomethingQuery
  @throwOnFieldError
  @raw_response_type {
  node(id: "test-id") {
    ... on BlogPage {
      title
    }
  }
  blogPageBySlug(slug: "test-slug") {
    title
  }
}
```

Those would currently generate this raw response type:

```typescript
export type somethingSomethingQuery$rawResponse = {
  readonly blogPageBySlug: {
    readonly id: string;
    readonly title: string;
  } | null | undefined;
  readonly node: {
    readonly __typename: "BlogPage";
    readonly id: string;
    readonly title: string | null | undefined;
  } | {
    readonly __typename: string;
    readonly id: string;
  } | null | undefined;
};
```

It seems like the `@throwOnFieldError` isn't properly considered in the case of a type refinement on a field returning an abstract type, since the `title` is typed as nullable below the `node` field and non-null on the field directly returning the object type.

We use `@raw_response_type` a lot to write typed mocked data for component demos and tests, so this makes it a lot harder to write accurate mock data for our developers.

tobias-tengler:
Fixes #4949

### DIFF
--- a/compiler/crates/relay-typegen/tests/generate_flow/fixtures/semantic_non_null_in_raw_response.expected
+++ b/compiler/crates/relay-typegen/tests/generate_flow/fixtures/semantic_non_null_in_raw_response.expected
@@ -29,14 +29,14 @@ export type MyQuery$data = {|
 |};
 export type MyQuery$rawResponse = {|
   +opera?: ?{|
-    +cast: ?$ReadOnlyArray<?{|
-      +character: ?string,
-      +singer: ?{|
+    +cast: $ReadOnlyArray<{|
+      +character: string,
+      +singer: {|
         +id: string,
         +name: ?string,
       |},
     |}>,
-    +composer: ?{|
+    +composer: {|
       +id: string,
       +name: ?string,
     |},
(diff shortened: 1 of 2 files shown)


packages/react-relay/__tests__/__generated__/RelayClient3DModuleTestQuery.graphql.js
```diff
@@ -113,7 +113,7 @@ return {
                         "fragmentPropName": "data",
                         "kind": "ModuleImport",
                         "componentModuleProvider": () => require('./../ClientUser.react'),
-                        "operationModuleProvider": () => require('react-relay/__tests__/__generated__/./RelayClient3DModuleTestFragmentClientUser_data$normalization.graphql')
+                        "operationModuleProvider": () => require('packages/react-relay/__tests__/__generated__/./RelayClient3DModuleTestFragmentClientUser_data$normalization.graphql')
```
tobias-tengler, line 116: This is unrelated to my changes (you can see the compiler output check failing on main).

captbaritone:
Thanks for the fix, and the ping!

captbaritone:
I'm actually not sure this approach is correct. Left a comment here: https://github.com/facebook/relay/issues/4949#issuecomment-2828354611

PR #4961, opened 2025-04-22
Handle relay:// json schema requests even outside of Relay projects

captbaritone:
Previously if you opened a `package.json` file in a non-Relay projects you would see an error like this:

Screenshot 2025-04-22 at 4 36 36 PM

This was because we didn't register a handler if no Relay project was detected. Now we register a noop handler, and there is no error.

Screenshot 2025-04-22 at 4 32 28 PM

Thanks to @cpojer for the report.

### DIFF
--- a/vscode-extension/.eslintrc.js
+++ b/vscode-extension/.eslintrc.js
@@ -38,6 +38,7 @@ module.exports = {
           {functions: false},
         ],
         'class-methods-use-this': 'off',
+        'max-classes-per-file': 'off',
       },
     },
   ],
--- a/vscode-extension/src/extension.ts
+++ b/vscode-extension/src/extension.ts
@@ -89,7 +89,7 @@ export async function activate(extensionContext: ExtensionContext) {
     }
   } else {
     // We still need to register a handler for `relay://` otherwise non-Relay
-    // projects will get an error at the top of their `package.json` files.
+    // projects will get a warning at the top of their `package.json` files.
     registerNoopProviders(extensionContext);
   }
 }
--- a/vscode-extension/src/extension.ts
+++ b/vscode-extension/src/extension.ts
@@ -8,7 +8,7 @@
 import path = require('path');
 import {ExtensionContext, window, workspace} from 'vscode';
 import {registerCommands} from './commands/register';
-import {registerProviders} from './providers/register';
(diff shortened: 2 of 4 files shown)


vscode-extension/.eslintrc.js
```diff
@@ -38,6 +38,7 @@ module.exports = {
           {functions: false},
         ],
         'class-methods-use-this': 'off',
+        'max-classes-per-file': 'off',
```
cpojer, line 41: Lol such an odd rule.

vscode-extension/src/extension.ts
```diff
@@ -87,6 +87,10 @@ export async function activate(extensionContext: ExtensionContext) {
         ].join(' '),
       );
     }
+  } else {
+    // We still need to register a handler for `relay://` otherwise non-Relay
+    // projects will get an error at the top of their `package.json` files.
```
cpojer, line 92:
```suggestion
    // projects will get a warning at the top of their `package.json` files.
```

vscode-extension/src/providers/textDocumentContentProvider.ts
```diff
@@ -76,3 +76,19 @@ export class RelayTextDocumentContentProvider
...
+export class NoopTextDocumentContentProvider
+  implements TextDocumentContentProvider
+{
+  provideTextDocumentContent(uri: Uri): ProviderResult<string> {
+    if (
+      uri.authority === PACKAGE_JSON_RELAY_CONFIG_SCHEMA_PATH ||
+      uri.authority === RELAY_CONFIG_SCHEMA_PATH
+    ) {
+      // We return an empty JSON schema instead of undefined to prevent
+      // an error being shown in the user's IDE.
```
cpojer, line 89:
```suggestion
      // a warning being shown in the user's IDE.
```

cpojer (commented):
Thank you!

cpojer:
shipit! I would approve but I'm not allowed.

fwiw it's a "warning" not an "error", and it shows up in other places too, including the `relay.config.json` if the extension is not active in a project.

vscode-extension/.eslintrc.js
captbaritone: Agreed.

PR #4965, opened 2025-04-23
Update VSCode docs/errors to use pathToRelay

captbaritone:
We changed this config option but neglected to change it in the readme or in the error message


cpojer (commented):
shipit

PR #4966, opened 2025-04-23
Iteration on new quick start guide based on feedback

captbaritone:
Based on feedback from @cpojer (some of which echoed other feedback I've received), this updates the new quickstart guide to clarify some details:

1. Configures babel plugin via the Vite React plugin rather than needing to use the vite Relay plugin
2. Clarifies what `relay.config.json` is for
3. Clarifies that the `<RelayEnvironmentPrivider>` must nest around all other components.
4. Adds defining a fragment component to the guide (echoing feedback @evanyeung gave earlier).
5. Clarify what running the compiler does


website/docs/getting-started/quick-start.md
```diff
@@ -145,8 +150,7 @@ export default function App() {
         allFilms {
           films {
             id
-            title
-            director
+            ...Film_li
```
cpojer, line 153: On a quick read, the `_li` reads like a typo. Is it.. list? Is it supposed to be item? And then when I read the next component I guess it might be `ListItem`. Not a big deal, but `Film_listItem` etc. would likely be more readable.

cpojer (commented):
This is awesome, this quick start guide is so good.

website/docs/getting-started/quick-start.md
captbaritone: Was supposed to be "list item" like a `<li>` tag, but I agree it's confusing. It's repeated so many times that I wanted to keep it as short as possible. I went with `Film_item` which splits the difference, but is at least a word.

PR #4968, opened 2025-04-23
Update compiler test project to reflect new default config option

captbaritone:
Looks like this project is used to test the open source compiler CLI and is not built as part of our internal tests. As such, it wasn't updated as part of my internal diffs to change the default here.

### DIFF
--- a/scripts/check-git-status.sh
+++ b/scripts/check-git-status.sh
@@ -11,6 +11,6 @@ if [ -z "$(git status --porcelain)" ]; then
   exit 0
 else
   echo "Detected changes in the working directory:"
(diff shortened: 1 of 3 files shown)


scripts/check-git-status.sh
```diff
@@ -11,6 +11,6 @@ if [ -z "$(git status --porcelain)" ]; then
   exit 0
 else
   echo "Detected changes in the working directory:"
-  git --no-pager diff --stat HEAD
+  git --no-pager diff HEAD
```
captbaritone, line 14: This was helpful in debugging the issue. We now emit a full diff when we detect changes instead of just showing _which_ files changed.

PR #4972, opened 2025-04-24
[Sketch] Make fields nullable not optional when semantic nullability is used

captbaritone:
Differential Revision: D67054462


captbaritone:
Replaced by https://github.com/facebook/relay/pull/4974

PR #4974, opened 2025-04-27
Exclude undefined from nullable types in semantic context

### LINKED ISSUE #4971: Do not type nullable fields with `undefined` within `@throwOnFieldError` and `@catch`
With `@throwOnFieldError` or `@catch` applied missing data i.e. `undefined` should no longer be possible to be read from a fragment or operation (besides conditional selections from my understanding), so couldn't we stop generating `Type | null | undefined` in those cases for nullable fields and model them just as `Type | null` instead?

tobias-tengler:
Closes #4971

Supersedes #4972


compiler/crates/relay-typegen/src/visit.rs
```diff
@@ -1903,17 +1913,21 @@ fn make_prop(
                     runtime_imports,
                     custom_error_import,
                 );
-
-                let getter_return_value =
-                    transform_type_reference_into_ast(&linked_field.node_type, |type_| {
+                // TODO: Is this correct for updatable types
```
tobias-tengler, line 1916: Open question

tobias-tengler:
@captbaritone I've opted into always generating the nullable types in the raw response type without `undefined` (unless optional of course), but I'm wondering if this change should be rolled out together with the potentially new `@rawResponseType` directive that always outputs the semantic type instead?

captbaritone:
@tobias-tengler Yeah, while I think you are correct that we should not allow optional types in raw response (since the server will never omit fields) in practice since they are currently typed that way it will be a major breaking change to update that behavior. At the very least we will need a feature flag, but I think bundling that flag (and therefor the PR) with a larger rethinking of raw response types makes more sense to me.

tobias-tengler:
@captbaritone That's what I thought. I've updated the PR to only contain the change that `undefined` is omitted for nullable types within `@throwOnFieldError` and `@catch`.

captbaritone:
I'm going to import this and see what breaking type changes is produces for us.

captbaritone:
Hmm. Looks like this creates many thousands of Flow issues internally. On common case seems to be people defining their own types to model a piece of a fragment and then assigning a bit of the Relay data into that type, which won't work.

Example: https://flow.org/try/#1N4Igxg9gdgZglgcxALlAIwIZoKYBsD6uEEAztvhgE6UYCe+JADpdhgCYowa5kA0I2KAFcAtiRQAXSkOz9sADwxgJ+NPTbYuQ3BMnTZA+Y2yU4IwRO4A6SFBIrGVDGM7c+h46fNRLuKxJIGWh8MeT0ZfhYlCStpHzNsFBAMIQkIEQwJODAQfiEyfBE4eWw2fDgofDBMsAALfAA3KjgsXGxxZC4eAw0G-GhcWn9aY3wWZldu-g1mbGqJUoBaCRHEzrcDEgBrbAk62kXhXFxJ923d-cPRHEpTgyEoMDaqZdW7vKgoOfaSKgOKpqmDA+d4gB5fMA-P6LCCMLLQbiLOoYCqgh6-GDYRYIXYLSgkRZkCR4jpddwPfJLZjpOBkO4AX34kA0SRWxgABAAVVYAMUo6QASng6OyALzs4AAHSg7Nl7IA9PL2QBqEjpbA8uB4NgAfmQ7PspigCAVSoA8rg2AaVm1eOyJLVaeyAO4QShbEjSuUqtXmTXa-WGiomgA+7KOuFN7IActhndbaG0rFzHSR2Wgoh7pfTpdK2dh2QBVMiUAAimgqpW5HPFUplcpEtAAylJg3rrUaENnc7AHso4NB7TQSDA3SIABQsXB0fXVjX8kRC6e0ACU+uLJnL8C+bDnEq9cpYEiElBlde93sbLc7+qndCsvo1WstB9lOag9NyIAaJhIA6gSQNAADFYABMABsACcVgAIwgPSQA

tobias-tengler:
Interesting. At least with TypeScript it shouldn't be an issue, since the user defined type would have to be the same as Relay's current type (`Type | null | undefined`). Which means even if Relay is making the type narrower (`Type | null`), what we're assigning to is still broader. [Playground](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAKuEDEBOB7AtgJQgGwIYigF4oBvAKCkqgHpqoBndRASxwBMAuB4JZgOwDmUAD5Q+AV2zYRUcXzYQAZvwhsadAPLY19UNggAaKMAAWzelADuKJAGt6FKozQt2XXb0EyJUsgF8yMlBIKABVeggkABElFTY4EOJyKig0EABlHn4BdyyvUR9pUTkFZT5Vf0DFOQBjYGYUPmMkXHpFGzQACiQcfC4ExFRMXpAASi5wyJiy1QHSR0oe4HEkJuSUqjTMzxyoHrwQADpnV20FqAC-IA)
I've ran the type generation on our TypeScript code base and didn't encounter any issue.

Regarding Flow: The example you've made doesn't seem to reflect my changes. In your example the original type is somehow just `string | undefined`:
```diff
- +someField?: string
+ +someField: string | null
```
But looking at the snapshots, the original type should be `string | null | undefined` if I understand Flow correctly:
```diff
- +someField: ?string,
+ +someField: string | null,
```

Do you know why this could be the case?

captbaritone:
Hmm. You're right. I think my example is not quite representative. Let me try to get one that's a direct reduction of the issue I saw internally.

tobias-tengler:
@captbaritone did you manage to find a better example?

PR #4981, opened 2025-05-05
add missing generic type parameter for mutation tutorial example

itsjoeoui:
https://relay.dev/docs/tutorial/mutations-updates/#step-4--modify-the-updatable-data

This step of the tutorial is currently giving us type errors. 

CleanShot 2025-05-05 at 18 09 58@2x

This is because we are missing the generic type parameter.

Note that this will also require us to upstream the relay dependencies in the example to the latest v19: https://github.com/relayjs/relay-examples/pull/333

After that, our tutorial will be working properly with no type error!

### DIFF
--- a/website/docs/tutorial/mutations-updates.md
+++ b/website/docs/tutorial/mutations-updates.md
@@ -349,22 +349,18 @@ We pass this <span className="color2">fragment</span>, along with the <span clas
 ```
 function StoryLikeButton({story}) {
   ...
-      optimisticUpdater: store => {
+      optimisticUpdater: (store) => {
         const fragment = graphql`
           fragment StoryLikeButton_updatable on Story @updatable {
             likeCount
             doesViewerLike
           }
         `;
-        const {
-          // color1
-          updatableData
-        } = store.readUpdatableFragment(
-          // color2
-          fragment,
-          // color3
-          story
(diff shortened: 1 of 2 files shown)


captbaritone:
Thanks! This looks great. Do you mind also updating the v19 docs as well? https://github.com/facebook/relay/blob/main/website/versioned_docs/version-v19.0.0/tutorial/mutations-updates.md

itsjoeoui:
> Thanks! This looks great. Do you mind also updating the v19 docs as well? https://github.com/facebook/relay/blob/main/website/versioned_docs/version-v19.0.0/tutorial/mutations-updates.md

Of course, will do now!

itsjoeoui:
Done, please let me know if we need anything else :) @captbaritone

itsjoeoui:
btw it would be great to merge this together with https://github.com/relayjs/relay-examples/pull/333, I have tested the example with this change and it works great :)

itsjoeoui:
https://github.com/user-attachments/assets/4d46f229-b527-41a9-ac89-4edf3701d644

I noticed something weird. On fresh load, typescript LSP complains about this typing here, but if I remove the code and add the same code back, it does not complain anymore 🤔

PR #4982, opened 2025-05-05
Update babel transform to default to default to ES modules to match compiler behavior

captbaritone:


captbaritone:
Something funky going on with tests. When I first ran the test it failed just like it does in CI. After clearing Jest cache locally I can't repro the test failure locally.

Issue #4882, opened 2025-01-16
Client Resolvers inconsistent behaviour when using fragments


jonreading81:
There is an issue when using client resolvers with `Server` edges with `fragments`. This behaviour is inconsistent and has been identified through errors  reported in Sentry. A work around has been to add the fields to the original `Query` however this solution is obviously not scalable and goes again the design principles of Relay. To clarify this does not happen in most cases and is only observable through errors in Sentry.

captbaritone:
Thanks for your example [here](https://github.com/relayjs/relay-examples/pull/332). I can confirm that this is a bug. Our logic for detecting if we need to fetch `@waterfall` edges is not traversing into fragment spreads in order to determine if we are missing data.

I'm working on a regression test now and will followup with our plan for a mitigation/fix. Thanks again for the report.

jonreading81:
Thank @captbaritone Do you have any update on this?

captbaritone:
Hey, thanks for the ping. I have some work in progress exploration of a fix here but I'm not sure sure if it can be made robust. We don't currently have time to prioritize a fix at this moment, so I'll find time tomorrow to share that as a PR so at least we can explore solutions in the open until the team gets time to prioritize working on this.

captbaritone:
Wasn't able to find time to track this down today. Added to my list for Monday.

captbaritone:
Okay. The bug is coming from this optimization: https://github.com/facebook/relay/blob/35229211f1198b956382c604f7787ba3b99ab5ed/packages/react-relay/relay-hooks/useFragmentInternal_CURRENT.js#L464

Basically that check is making it so that we only check for missing client edge data if the fragment itself contains a client edge but that's not actually correct. We need to consider all fragments. I have a diff to fix, but adding an additional useMemo an useFragment to all instances of `useFragment` is a significant change in terms of perf. I need to look a bit more closely to see if there's some other way to reduce the perf impact here.

captbaritone:
For visibility, here's an initial stab at a fix: https://github.com/facebook/relay/pull/4992 We may iterate on a solution from here, but hopefully this clarifies what's going on here.

Thanks again for the report and helping us uncover real bug.

jonreading81:
Thanks for the update.

PR #4992, opened 2025-05-12
Fetch missing client edge server queries discovered in nested fragments

captbaritone:
Summary:
Initially reported here: https://github.com/facebook/relay/issues/4882

Fetching of client-edge-to-server generated queries is handled by detecting missing data in Relay Reader and Relay Reader maintaining a stack of client-edge-to-server fields. As we traverse into such fields we push information about that query onto the stack, and as we exit such fields we pop values off the stack.

In the case of fragments spread into selections where we have such a stack built up, we propagate it via a hidden [__clientEdgeTraversalPath](https://github.com/facebook/relay/blob/35229211f1198b956382c604f7787ba3b99ab5ed/packages/relay-runtime/store/RelayStoreUtils.js#L281) field on fragment keys.

This hidden property means that any fragment located somewhere underneath a client edge to server field might encounter missing data and need to trigger the fetch of that server query.

For example, if the selection includes only fragment spreads, or some other query had fetched all the fields that were read directly in the fragment which read the actual resolver field the fragment with the resolver field would render fine and we wouldn't discover we were missing data until we tried to read a child fragment. In theory that could be many fragments away. E.g. if I had a field `my_client_edge_to_server` that returned a `User,` any fragment spread into any selection nested under that edge (including `my_client_edge_to_server.best_friend.favorite_sports_team.coach.profile_picture` could technically uncover missing data in some totally generic `ProfilePicture` fragment.

However, we overlooked that possibility in D36684124 where we added an optimization which tries to avoid the overhead of these additional hooks in the common case by only checking for missing client edge data in the case that the fragment _itself_ contains at least one client edge field (as computed by the compiler).

As the example above demonstrates (and the test case in this diff also demonstrates) that turns out to not actually be safe.

This removes the optimization to allow the correct behavior to work. However, this adds a useMemo and useEffect to each useFragment, which may have some perf implications. Putting this behind a feature flag to let us ship this especially for folks who are blocked by it, but also so we can consider perf concerns before rolling out internally.

I'll invert the feature flag default in a future diff once we've configured it internally.

Differential Revision: D74593372

### DIFF
--- a/packages/react-relay/__tests__/ClientEdges-test.js
+++ b/packages/react-relay/__tests__/ClientEdges-test.js
@@ -461,42 +461,36 @@ describe.each([[true], [false]])(
 
       // This will be updated when we add the new assertions as part of a fix for
       // this bug.
-      // eslint-disable-next-line no-unused-vars
       let renderer;
       TestRenderer.act(() => {
         renderer = TestRenderer.create(<TestComponent />);
       });
 
-      // Oops, we didn't fetch the query!
-      expect(fetchFn.mock.calls.length).toEqual(0);
+      expect(fetchFn.mock.calls.length).toEqual(1);
+      // We should send the client-edge query
+      // $FlowFixMe[invalid-tuple-index] Error found while enabling LTI on this file
+      expect(fetchFn.mock.calls[0][0].name).toBe(
+        'ClientEdgeQuery_ClientEdgesTest6Query_me__same_user_client_edge',
+      );
+      // Check variables
+      // $FlowFixMe[invalid-tuple-index] Error found while enabling LTI on this file
+      expect(fetchFn.mock.calls[0][1]).toEqual({id: '1'});
+      expect(renderer?.toJSON()).toBe('Loading');
 
-      // Following are the assertions that SHOULD pass.
-
-      // expect(fetchFn.mock.calls.length).toEqual(1);
-      // // We should send the client-edge query
-      // // $FlowFixMe[invalid-tuple-index] Error found while enabling LTI on this file
-      // expect(fetchFn.mock.calls[0][0].name).toBe(
-      //   'ClientEdgeQuery_ClientEdgesTest6Query_me__same_user_client_edge',
-      // );
-      // // Check variables
-      // // $FlowFixMe[invalid-tuple-index] Error found while enabling LTI on this file
-      // expect(fetchFn.mock.calls[0][1]).toEqual({id: '1'});
-      // expect(renderer?.toJSON()).toBe('Loading');
-
-      // TestRenderer.act(() => {
-      //   // This should resolve client-edge query
-      //   networkSink.next({
-      //     data: {
-      //       node: {
-      //         id: '1',
-      //         __typename: 'User',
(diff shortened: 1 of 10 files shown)


jonreading81:
Thanks for this @captbaritone , Ive patched our version of relay and this has fixed the majority of issues however its still failing when we are using `Derived fields` for the server edges when using readFragment.

For example:

```js
        const { streamInfo } = readFragment(
            graphql`
                fragment audioUrlCloudcastFragment on Cloudcast
                {
                    streamInfo {
                        url
                    }
            `,
            key,
        );
```

This is still missing the data, any idea what needs to update for this using your proposed approach?

captbaritone:
Thanks for trying it out and identifying this additional case. Could you clarify which filed is the client edge to server in this case?

packages/react-relay/__tests__/ClientEdges-test.js
```diff
@@ -31,362 +37,377 @@ injectPromisePolyfill__DEPRECATED();
...
-              # Adding "name" here will cause the query to be fetched.
-            }
-          }
-        }
-      `,
-      variables,
-    );
+    // https://github.com/facebook/relay/issues/4882
+    it.each([[true], [false]])(
+      'should fetch data missing in fragment spread within `@waterfall` field. CHECK_ALL_FRAGMENTS_FOR_MISSING_CLIENT_EDGES = %s',
```
captbaritone, line 323: @jonreading81 any chance you could create a repro unit test similar to this demonstrating the resolver issue you are seeing? If not, just providing more details might be able to help me construct one.

captbaritone:
(Mostly a note for myself): I wonder if we are not correctly setting  `CLIENT_EDGE_TRAVERSAL_PATH` [here](https://github.com/facebook/relay/blob/16cd20078c84b040c8a0e2b5ad55eff40e994469/packages/relay-runtime/store/RelayReader.js#L791-L799).

Note how we set it [here](https://github.com/facebook/relay/blob/16cd20078c84b040c8a0e2b5ad55eff40e994469/packages/relay-runtime/store/RelayReader.js#L1591).

captbaritone:
Nevermind, I think I have a repro of the issue you are seeing, and I think the note to myself above is likely the root cause. I'll update this pr.

captbaritone:
@jonreading81 I think [this](https://github.com/facebook/relay/pull/4995) is the fix for the resolver issue you are seeing. Do you want to try patching that in as a well and see if it resolves the issue? I intend to land that as a separate PR since it is not strictly dependent on this one

jonreading81:
Thanks @captbaritone , I have tested this and it fixes the issue. Thanks for your quick response on this.

captbaritone:
@jonreading81 This should be available on NPM in the relay packages `react-relay@0.0.0-main-611f6d9c`, `relay-runtime@0.0.0-main-611f6d9c` etc.

You'll need to set the [feature flag](https://relay.dev/docs/next/api-reference/runtime-config/#feature-flags).


```js
RelayFeatureFlags.CHECK_ALL_FRAGMENTS_FOR_MISSING_CLIENT_EDGES = true;
```

PR #4999, opened 2025-05-14
Emit raw query text in artifact for `@relay_test_operation` operations

jakobkhansen:
Hello!

In Microsoft we have a use case where we would like to parse Relay operations being executed with `graphql-js`. In our case this is specifically for testing with Storybook, where we want to use [Graphitation payload generator](https://github.com/microsoft/graphitation/tree/main/packages/graphql-js-operation-payload-generator) to generate mock data, which takes `graphql-js` operations as input.

Today, parsing operations is somewhat limited because the `text` field in Relay operation artifacts is heavily transformed and for example does not contain client extensions. We would like to expose the raw, untransformed query text in operation artifacts, hidden behind the `@relay_test_operation` directive. We believe this would be useful in other testing scenarios as well, because it enables Relay operations to be easily converted between formats and used with other GraphQL tooling.

This PR implements that by adding a `rawText` field in `OperationDefinition` and a new transform, the `RawTextTransform`. The transform populates this field, only on operations with `@relay_test_operation` directive. The transform is specifically added as the first transform to be executed, since it relies on generating the original, "raw" query text, before any other transforms are applied.

Here's an example on how `rawText` would be exposed in an artifact:

image

where you can see that `viewData`, which is a client-extension, is preserved.

I've added this field also to Flow types, and if you think it makes sense to mention this in docs, I can add that as well.

Would love to hear if you have any thoughts on adding this feature :)


jakobkhansen:
TODO: Write a test where rawText is not `None` if possible.

tyao1:
I'm curious how is the text used for test operations?

> text field in Relay operation artifacts is heavily transformed and for example does not contain client extensions. We would like to expose the raw, untransformed query text in operation artifacts, hidden behind the @relay_test_operation directive

Apart from client extensions, is there anything else that is missing in `text` and requires `raw_text`? If the goal is to include client extensions in the query text, I'm wondering if we can add a flag to not exclude client extensions in the query text output.

Since the raw text is intended for `@relay_test_operation` only, the recommended approach is to create a directive to store that information, and extract it in build_ast. So it doesn't change existing data structures

jakobkhansen:
@tyao1 our main use case right now is to include the client extensions, yes. What type of flag were you thinking here? Since this would need to be on a per operation basis, would it be a directive similar to `@relay_test_operation` which would simply include client extensions in the `text` field? I think that would work in our testing scenario, but it would mean that the operation could no longer be used as a real server query since AFAIK the `text` field is what is sent to the server, right? So it shouldn't contain the client extensions. That is not problematic in our case though since an operation would be for testing only if it has the directive.

To answer the question on how `text` is used when we test, since Graphitation payload generator uses `graphql-js` operations as input, we parse the `text` field to get the correct type format , but that does not contain client extensions, so we are losing that information there.

Good tip on the directive to store the data, if we don't go the flag route I will look into that change.

tyao1:
>  I think that would work in our testing scenario, but it would mean that the operation could no longer be used as a real server query since AFAIK the text field is what is sent to the server, right?

That's correct.

`skip_client_extensions` is the transform that removes client extensions from the query text.  I think it can read a new argument on `@relay_test_operation` or a new directive to skip the transform.

jakobkhansen:
Ok, so I attempted the flag to not do the `skip_client_extensions` transform and I don't think it's a viable path. Mainly since multiple operations can share the same fragment, and one operation may want to remove the client extensions while another wants to keep them (in the case of test operation with flag set). For example this:

```gql
query TestOperation @relay_test_operation(includeClientExtensions: true) {
  ...ServerType
}

query ServerOperation {
  ...ServerType
}

fragment ServerType {
  serverField
  clientExtension
}
```

Since the test operation uses the `ServerType` fragment, we can't remove the `clientExtension` field there, but if we don't then the `ServerOperation` would end up emitting `clientExtension` in the `text` field, which we obviously can't.

The only thing we could do would be to duplicate the fragment somehow and have a separate version which keeps the client extensions, but I think this would be a bit complicated and probably we don't want to do that.

I think the best option is to as you suggested, have an argument on `@relay_test_operation` which stores the raw_text in a metadata directive, and then we set `text` field to the `raw_text` metadata if it's populated. This is a way smaller change now and has no interface changes in artifact, which I think is best.

I pushed this change now, which adds the `emitRawText: Boolean` argument to `@relay_test_operation`.

@tyao1 , please have a look and see if this approach is viable :)

tyao1:
> Since the test operation uses the ServerType fragment, we can't remove the clientExtension field there, but if we don't then the ServerOperation would end up emitting clientExtension in the text field, which we obviously can't.

Makes sense. Thanks for trying out this approach

compiler/crates/relay-transforms/src/raw_text.rs
```diff
@@ -0,0 +1,93 @@
...
+use graphql_text_printer::{OperationPrinter, PrinterOptions};
+use intern::string_key::Intern;
+use lazy_static::lazy_static;
+
+use crate::test_operation_metadata::TEST_OPERATION_DIRECTIVE;
+
+lazy_static! {
+    pub static ref RAW_TEXT_DIRECTIVE_NAME: DirectiveName = DirectiveName("rawText".intern());
+    pub static ref RAW_TEXT_ARGUMENT_KEY: ArgumentName = ArgumentName("rawText".intern());
+}
```
tyao1, line 15: Doesn't need to define the value twice. This is just a perf optimization to initialize the string once.  The whole thing can be replaced with the  `intern!("rawText")` macro below.

compiler/crates/relay-transforms/src/raw_text.rs
```diff
@@ -0,0 +1,93 @@
...
+            })
+        } else {
+            Transformed::Keep
+        }
+    }
+}
+
+pub fn create_raw_text_directive(raw_text: &String) -> Directive {
+    Directive {
+        name: WithLocation::generated(*RAW_TEXT_DIRECTIVE_NAME),
```
tyao1, line 73:
```suggestion
        name: WithLocation::generated(DirectiveName(intern!("rawText"))),
```

compiler/crates/relay-transforms/src/raw_text.rs
```diff
@@ -0,0 +1,93 @@
...
+            .named(*TEST_OPERATION_DIRECTIVE)
+            .is_some()
+        {
+            let options = PrinterOptions {
+                compact: true,
+                json_format: true,
+                ..Default::default()
+            };
+
+            let mut printer = OperationPrinter::new(self.program, options);
```
tyao1, line 56: I think the printer can be initialized just once in `RawTextTransform`

tyao1 (changes requested):
Looks very clean! Requesting changes to check `EMIT_RAW_TEXT_ARG` before attaching the raw text

jakobkhansen:
Thanks for the review @tyao1, let me know if there is anything else to fix now :)

compiler/crates/relay-transforms/src/raw_text.rs
```diff
@@ -0,0 +1,87 @@
...
+        }],
+        data: None,
+        location: Location::generated(),
+    }
+}
+
+pub fn get_raw_text_value(op: &OperationDefinition) -> Option<String> {
+    op.directives
+        .named(*RAW_TEXT_DIRECTIVE_NAME)
+        .and_then(|dir| dir.arguments.named(ArgumentName("rawText".intern())))
```
tyao1, line 82:
```suggestion
        .and_then(|dir| dir.arguments.named(ArgumentName(intern!("rawText"))))
```

tyao1 (approved):
Looks good to me! It would be nice to have a snapshot test for the raw text generation under `relay-compiler/tests/compile_relay_artifacts/fixtures`, but it is up to you.

jakobkhansen:
@tyao1 I added a test and hopefully fixed the lint issue that was failing on CI.

tyao1 (approved):

jakobkhansen:
@tyao1 sorry for the back and forth, I had manually updated the fixture tests before as I was unaware of `./scripts/update_fixtures.sh`. I've ran and committed that now. Hopefully that was the last thing failing internally for you.

jakobkhansen:
@tyao1 ping on this one :)

tyao1:
@jakobkhansen I brought the PR up to the team, and we have more questions. Using raw query text has the downside that some client only directives like @required will no longer work. Could you provide more detail on how the generated payload is consumed?

jakobkhansen:
Sure, we are planning on using it exactly [here](https://github.com/microsoft/nova-facade/blob/1ffb0597bfb7c058cc233e56f17a84cc5f37716b/packages/nova-react-test-utils/src/relay/test-utils.tsx#L35). The idea is that we have a Storybook decorator which should work to generate mock-data for both Apollo and Relay based applications, and uses the [Graphitation payload-generator](https://github.com/microsoft/graphitation/tree/main/packages/graphql-js-operation-payload-generator). What we do for Relay operations is take the `text` field and parse that with `graphql-js` in order to input it into our payload generator. At that point the problem with missing client extensions appears, because we lose those in the translation, since the `text` field does not contain it.

If there are concerns with this breaking the `text` field, my suggestion is that we could go back to the previous approach of supplying `rawText` field in the artifact in addition to `text` when `@relay_test_operation` is present. That way we can avoid adding the new directive, and the original `text` field will be left unchanged. I think we would actually prefer this, to avoid adding the `emitRawText` directive to every test operation. How do you feel about that @tyao1 ?

tomgasson:
Chiming in to support the request for features like this.

At Atlassian, we also had to tweak the compiler to output the query text ([a bit of a hack - into the extra artifacts dir](https://github.com/facebook/relay/commit/4e169c188d498748369c395ef63761ac8b5c2a10)) so that our tests can look that up and mock it. We've been carrying that patch across upstream updates for a while now.
For our tests / storybook setup, we replace the network layer, find the queries' text for the operation's persisted id and then run that query (with graphqljs in the browser) and have those resolvers read from mock data.
 
We'd additionally like to shift from this setup to instead actually hit a local server. Sending 1k+ queries over the network and the resolvers for storybook takes a while to bundle since GraphQL is a natural "barrelling" point. But even that would then have the same trouble: the query text for these operations doesn't even exist in our repo (they're persisted to a service), so any mechanism to get the query text out is valuable: Even something as simple as allowing the `persist: {...}` to do _both_ the local and remote based persisting at the same time (with the ID coming from the remote) would solve our problem. 
Because testing is local, we actually worked out we don't even need to do this by ID. The operationName's are unique, there's no versioning issues to contend with for testing, so being able to lookup `Record<OperationName, QueryText>` would be a nicer and cleaner way to consume this (the requestParams has the `name`)

Markionium:
@tyao1 I'm not sure I understand why `@required` would no longer work?

Am I misunderstanding that this would only change the query string that is in the `params.text` field of the document but not the actual runtime artifact?

Perhaps because these fields would now returned in the payload from the mock server even though they're actually client extensions? But I think `@required` works for both?

Issue #5027, opened 2025-06-30
Bug: Provided variable import order incorrect when generating ES modules with relay-compiler


pavelglac:
### Description
When using provided variables and generating ES modules with relay-compiler (version 19.0.0 and verified also in 20.x), the generated code has the import statement for the provided variable placed after it is used. This results in incorrect import and execution order.

#### Example output:
```javascript
({
  "__relay_internal__pv__file": id_provider
} as {
  readonly __relay_internal__pv__file: {
    readonly get: () => string;
  };
});

import id_provider from '../file';
```

#### Expected behavior
The import statement should appear at the top of the generated file, before any usage of the imported variable.

#### Actual behavior
The import statement is placed after the provided variable is used, resulting in a reference before import.

#### Tooling
- relay-compiler: 19.0.0 (also verified with 20.x)
- Node.js: 20

#### Additional context
Let me know if a minimal reproduction repository or more details are needed.

captbaritone:
Thanks for the report. The next step here would be to create an integration regression test which shows the broken behavior. Our compiler integration tests can be found here: https://github.com/facebook/relay/tree/main/compiler/crates/relay-compiler/tests/relay_compiler_integration/fixtures

Is that something you'd be able to work on? More info on writing/running Rust tests can be found here: https://github.com/facebook/relay/blob/main/.github/CONTRIBUTING.md#pull-requests

pavelglac:
sure, will try look on it next week.

PR #5028, opened 2025-07-07
fix(provided-variables): import provided variable before type section

pavelglac:
Issue: https://github.com/facebook/relay/issues/5027

I am changing the order of execution when type section goes after staments section in order to ensure provided variable is already imported.

## Description
When using provided variables and generating ES modules with relay-compiler (version 19.0.0 and verified also in 20.x), the generated code has the import statement for the provided variable placed after it is used. This results in incorrect import and execution order.

#### Current output:
```javascript
({
  "__relay_internal__pv__file": id_provider
} as {
  readonly __relay_internal__pv__file: {
    readonly get: () => string;
  };
});

import id_provider from '../file';
```

#### New output:
```javascript
import id_provider from '../file';

({
  "__relay_internal__pv__file": id_provider
} as {
  readonly __relay_internal__pv__file: {
    readonly get: () => string;
  };
});
```


captbaritone:
Excellent! Thanks for the fix. I'll import this and confirm that it does not trigger any issues internally.

pavelglac:
@captbaritone I had to rebase this PR.

PR #5033, opened 2025-07-10
doc(useSubscription): config argument should be memoized

pavelglac:
Do not rely solely on the example. The documentation should explicitly state that the config argument must be memorized.

### DIFF
--- a/website/docs/api-reference/hooks/use-subscription.md
+++ b/website/docs/api-reference/hooks/use-subscription.md
@@ -41,7 +41,7 @@ function UserComponent({ id }) {
 
 ### Arguments
 
-* `config`: a config of type [`GraphQLSubscriptionConfig`](#type-graphqlsubscriptionconfigtsubscriptionpayload) passed to [`requestSubscription`](../request-subscription/)
+* `config`: a memoized config of type [`GraphQLSubscriptionConfig`](#type-graphqlsubscriptionconfigtsubscriptionpayload) passed to [`requestSubscription`](../request-subscription/)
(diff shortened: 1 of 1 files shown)


captbaritone:
I'm not sure this is worth calling out as this is a requirement of all args to all React hooks. Is there a reason we should be explicitly call out this hook arg, but not all other hook args?

I sympathize that this is a bit of a sharp edge in React, but I don't think it's really specific to this specific argument or even Relay.

pavelglac:
What do you mean by other args? For useSubscription it is crucial to have memoized config since it is in [dependencies](https://github.com/facebook/relay/blob/main/packages/react-relay/relay-hooks/useSubscription.js#L50) of useEffect. I mean it is mentioned in the [example](https://github.com/facebook/relay/pull/5033/files#diff-f6c7ce0a8ad1254d6a542a1660556c1ffe7d80a4d9f84d3cd1f1672fe670658aR29) but I thought it should be explicitly be also in following section. But if you do not think so, feel free to close this.

captbaritone:
I think I see what you're getting at here. The object identity itself if used as a dependency where in other cases we pick off individual values and use those as keys.

captbaritone:
Would you mind making this change in this file? https://github.com/facebook/relay/blob/b7c43a23a78f42ce2cdf37bd0cef1fe25cde9186/website/docs/api-reference/hooks/use-subscription.md?plain=1#L4

The file you've updated is just the versioned snapshot we cut for version 20.

pavelglac:
Sure, done.

PR #5034, opened 2025-07-14
Update compiler.md to correct spelling of support

Uk1288:
correct spelling of support


tyao1:
Thanks for the contribution. Would you mind updating the non-versioned file here too https://github.com/facebook/relay/blob/daa7235d26cbe11f6d05660df19610f51b9a491a/website/docs/guides/codemods.md?plain=1#L4? That file is the source of truth.

PR #5039, opened 2025-07-22
Update throw-on-field-error-directive.md

christiansany:
Currently on https://relay.dev/docs/guides/throw-on-field-error-directive/, this slash at the end of the file name will result in a double slash when visiting the site.

This change will remove the double slash and make the link work.


captbaritone:
Interesting! I've seen other issues with broken links related to trailing slashes. Do you want to grep for all such `/#` links and replace them? At the very least we should update the non-versioned version of this page here: https://github.com/facebook/relay/blob/cd4a328389982ce5176f730f49a79d1d02c0f1cd/website/docs/guides/throw-on-field-error-directive.md#L4

Could you also confirm that the link still works by doing a full build of the docs and clicking the link (if you haven't already).

Thanks again for looking into this!

captbaritone:
Ping @christiansany

captbaritone:
I've opened https://github.com/facebook/relay/pull/5049 in an attempt to fix this issue and others in historical versions of the docs.

PR #5043, opened 2025-07-30
fix(relay-runtime): prototype less payloads

pavelglac:
Relay throws an error when execution and consumption occur in the same thread without a serialization boundary (no JSON.parse()). This occurs because popular execution library have prototype less object (to allow keys like constructor etc.). Relay should not assume payload has any method available. 

graphql-js: https://github.com/graphql/graphql-js/blob/16.x.x/src/execution/execute.ts#L298
supermassive: https://github.com/microsoft/graphitation/blob/main/packages/supermassive/src/executeWithoutSchema.ts#L201

### DIFF
--- a/packages/relay-runtime/store/RelayResponseNormalizer.js
+++ b/packages/relay-runtime/store/RelayResponseNormalizer.js
@@ -250,7 +250,11 @@ class RelayResponseNormalizer {
               this._traverseSelections(selection, record, data);
             }
           } else {
-            const implementsInterface = data.hasOwnProperty(abstractKey);
+            // $FlowFixMe[method-unbinding] - data could be prototype less
+            const implementsInterface = Object.prototype.hasOwnProperty.call(
+              data,
+              abstractKey,
+            );
             const typeName = RelayModernRecord.getType(record);
             const typeID = generateTypeID(typeName);
             let typeRecord = this._recordSource.get(typeID);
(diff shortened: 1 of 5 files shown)


captbaritone:
Seems reasonable to me.

alloy:
From our experience across a good number of Relay versions, this has been the only place where this consideration has been needed thus far, so I'm not too afraid of this breaking anytime soon again. Having said that, if there's an easy way with Flow to be able to type all incoming response data as prototype-less data, then we should consider adding that.

captbaritone:
@alloy I'm not aware of any Flow feature that could help us here. Does TypeScript have one?

I think as long as we can have runtime tests that validate this works, we should be okay.

pavelglac:
Typescript does not have anything for it out of the box. You have to list all object methods manually if want to simulate object without prototype:

```typescript
type ObjectProperties =
  | "constructor"
  | "toString"
  | "toLocaleString"
  | "valueOf"
  | "hasOwnProperty"
  | "isPrototypeOf"
  | "propertyIsEnumerable";

const obj: {
  [key in ObjectProperties]: never;
} = Object.create(null);
obj.hasOwnProperty("key"); // <- error
```

alloy:
Thanks for the example, @pavelglac.

> I think as long as we can have runtime tests that validate this works, we should be okay.

Agreed.

PR #5048, opened 2025-08-01
[WIP] docusaurus upgrade

captbaritone:
Current failures can be found here: https://gist.github.com/captbaritone/654c78e85f7f22450a67dfeee645431b


captbaritone:
Motivation for attempting this upgrade is to see if https://github.com/facebook/docusaurus/commit/e34614963e1c40fb99072e18ced7ee463b5c8b1d will fix the issue reported in https://github.com/facebook/relay/pull/5039 where links of the form `./path/to/file.md/#foo` (note the trailing slash) are not reported as a build time error enough though the links are broken.

Issue #4967, opened 2025-04-23
[Bug] Relay resolver-based query data is disposed while a fragment is still in use


alina-zaieva:
In our Relay-based app we are using Relay resolvers for pure client state (available synchronously) and for REST API calls, while also fetching other parts of the query from a GraphQL server. While doing so, we are experiencing an issue with resolver data being seemingly disposed from the store by GC while still being in use after several extra queries are issued, and the resolver is not executed again. We weren't able to troubleshoot/debug the issue further, but the symptoms are as follows:

1. Relay logs a warning about being unable to read data for a fragment (actual fragment in question isn't fully deterministic and also changes when you move components around):
    ```
    Warning: Relay: Expected to have been able to read non-null data for
    fragment `...' declared in `useFragment()', since fragment reference was
    non-null. Make sure that that `useFragment()`'s parent isn't holding on to and/or passing a
    fragment reference for data that has been deleted.
    ```
2. App crashes with an error immediately, since `useFragment` returns `undefined` after the warning above and it cannot be destructured.
3. Increasing `gcReleaseBufferSize` value makes it so that the error takes longer to occur and is less deterministic, but as far as we could tell, it never goes away fully.

While our app is somewhat complex and also cannot be shared publicly, I have managed to create a minimal repro with the same query pattern:
[alina-zaieva/relay-gc-repro](https://github.com/alina-zaieva/relay-gc-repro)

The repro repo above also has some extra details on the setup and the error.

Markionium:
@captbaritone we're running into this is in a few different apps inside Microsoft which started using Relay Resolvers for local data. Would be great to discuss this and learn if you have any insights. I've tried to debug this but haven't gotten very far (yet).

@alloy FYI

captbaritone:
Thanks for the report and the repro! Let me take a look.

alina-zaieva:
@captbaritone I gave it a try with weak/strong types instead of output types as we discussed yesterday, but there is still the same read error as before ☹️
<https://github.com/alina-zaieva/relay-gc-repro/pull/1>

Is this close to what you meant as a thing to try? I'm also pretty sure some part of this implementation is wrong, since now there are a lot of extra fetches, but wasn't able to figure out what's the problem myself.

Markionium:
@captbaritone any update on this? This keeps resurfacing for us, so any insights you might have into this would be great.

Markionium:
[...]

PR #5052, opened 2025-08-08
Repro of weak type resolver GC bug

captbaritone:
Differential Revision: D79911205

Repro/regression test of the GC bug reported in https://github.com/facebook/relay/issues/4967


evanyeung:
Closing as this PR is already merged

PR #5083, opened 2025-09-25
Allow missing fields from defer payloads

robrichard:
The current spec proposal for `@defer` requires that servers do not send the same field multiple times in response payloads. If the same field is in both a deferred and non-deferred fragments, or two different deferred fragments, Relay expects this field to be in each of the corresponding responses or else it will throw an error like:

```
RelayResponseNormalizer: Payload did not contain a value for field `name: name`. Check that you are parsing with the same query that was used to fetch the payload.
```

This PR adds a new flag `deferDeduplicatedFields` which allows Relay to accept payloads with missing fields. The current defer spec guarantees a fragment will not be marked as completed until all of the fields have been returned, so in this case the relay store should already have the fields that are missing

Note: There are still major differences between the response format in the latest defer spec and what Relay expects. This PR  does not attempt to make Relay compatible with the spec, but rather allow a transformation to be done in the network layer to convert the current response format into something Relay can accept. There is an example of a working transformation in https://github.com/robrichard/defer-relay-example


matclayton:
Hey @tyao1 and @evanyeung any chance of this being merged? I had a quick DM with @captbaritone on discord last week and he said to hit you both up on github and ask as he was away. We're trying to build out stream/defer and this is one of the blockers on our end. Thanks!

PR #5091, opened 2025-10-11
Filter raw_response output types for concrete fields

### LINKED ISSUE #5090: @raw_response_type generates TypeScript types with duplicated union arms
The following is distilled down from a real-world example. In my app, I have `Post`s, which live inside a `Group`. I have a `Linkable` interface for types for which my app knows how to generate a URL. For a `Group`, we need to know the group type to put into the URL. For a `Post`, we need to know the `Group` it is in, as well as that `Group` type. A React component uses a fragment to fetch this data on anything that implements `Linkable`. When I create a `Post`, I want to link both to that `Post` and the `Group` that it's in, and so I spread my fragment on both.

---

Here's a very simplified version of my schema:

```
interface Linkable {
  id: ID!
}

type Post implements Linkable {
  id: ID!
  group: Group!
}

type Group implements Linkable {
  id: ID!
  type: String!
}

type Query {
  post: Post!
}

type Mutation {
  createPost: Post!
}
```

Here's a mutation using `@raw_response_type` to create a post, including the two fragment spreads so I can link to both places:

```
graphql`
  mutation TestMutation @raw_response_type {
    createPost {
      ...TestLinkableFragment
      group {
        ...TestLinkableFragment
      }
    }
  }
`;

graphql`
  fragment TestLinkableFragment on Linkable {
    id
    ... on Post {
      group {
        id
        type
      }
    }
    ... on Group {
      type
    }
  }
`;
```

The raw type that Relay generates for this is:

```
export type TestMutation$rawResponse = {
  readonly createPost: {
    readonly __isLinkable: "Post";
    readonly group: {
      readonly __isLinkable: "Post";
      readonly group: {
        readonly id: string;
        readonly type: string;
      };
      readonly id: string;
      readonly type: string;
    } | {
      readonly __isLinkable: "Group";
      readonly id: string;
      readonly type: string;
    } | {
      readonly __isLinkable: "Group";
      readonly id: string;
      readonly type: string;
    };
    readonly id: string;
  } | {
    readonly __isLinkable: "Post";
    readonly group: {
      readonly __isLinkable: "Post";
      readonly group: {
        readonly id: string;
        readonly type: string;
      };
      readonly id: string;
    } | {
      readonly __isLinkable: "Group";
      readonly id: string;
    } | {
      readonly __isLinkable: "Group";
      readonly id: string;
      readonly type: string;
    };
    readonly id: string;
  } | {
    readonly __isLinkable: "Group";
    readonly group: {
      readonly __isLinkable: "Post";
      readonly group: {
        readonly id: string;
        readonly type: string;
      };
      readonly id: string;
    } | {
      readonly __isLinkable: "Group";
      readonly id: string;
    } | {
      readonly __isLinkable: "Group";
      readonly id: string;
      readonly type: string;
    };
    readonly id: string;
    readonly type: string;
  };
};
```

---

Note all of the places that types like this are generated:

```
    {
      readonly __isLinkable: "Group";
      readonly id: string;
    } | {
      readonly __isLinkable: "Group";
      readonly id: string;
      readonly type: string;
    }
```

When using the raw response type for an optimistic response (a common use-case), it's easy to forget to add the `type`, since TS will match the first arm of that!

It seems like a bug that two arms of a union would be generated with the same `__isLinkable`, since Relay tends to flatten these out in most other cases?

Honestly it would be even better if Relay could pick the correct arm of `TestLinkableFragment`, since we know that the type is a `Group!`.

jwatzman:
When generating a raw_response type for a concrete fields, we may (e.g.,) still be spreading a fragment which is for an abstract type. We should cut down our type selections based on the actual concrete type of the field, to generate more specific raw_response types. We should also mark the selections has having the same concrete type as our field, preventing redundant arms of the final TypeScript/Flow union types from being output.

This makes the output raw response types a lot more pleasant to work with.

Fixes #5090

### DIFF
--- a/compiler/crates/relay-typegen/src/visit.rs
+++ b/compiler/crates/relay-typegen/src/visit.rs
@@ -2384,6 +2384,24 @@ pub(crate) fn raw_response_visit_selections(
             ),
         }
     }
+
+    if let Some(concrete_type) = enclosing_linked_field_concrete_type {
+        // If we are generating for a concrete field type, we should 1. remove
+        // any selections that are for a different concrete type, since they are
+        // not applicable to our field, and 2. mark any selections without a
+        // concrete type as being for our field's concrete type, so that
+        // `raw_response_selections_to_babel` doesn't generate redundant types.
+        type_selections.retain_mut(|type_selection| {
+            match type_selection.get_enclosing_concrete_type() {
+                Some(selection_concrete_type) => selection_concrete_type == concrete_type,
+                None => {
+                    type_selection.set_concrete_type(concrete_type);
+                    true
+                }
+            }
+        });
+    }
+
     type_selections
 }
 
--- a/compiler/crates/relay-typegen/tests/generate_flow/fixtures/relay-resolver-raw-response.expected
+++ b/compiler/crates/relay-typegen/tests/generate_flow/fixtures/relay-resolver-raw-response.expected
@@ -31,12 +31,10 @@ export type relayResolver_Query$data = {
   },
 };
 export type relayResolver_Query$rawResponse = {
-  readonly me: ?({
-    readonly id: string,
-  } | {
+  readonly me: ?{
     readonly id: string,
     readonly name: ?string,
-  }),
+  },
 };
 export type relayResolver_Query = {
   rawResponse: relayResolver_Query$rawResponse,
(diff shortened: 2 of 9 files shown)


jwatzman:
Hey @captbaritone -- sorry for the direct ping, but this has been sitting here for a while, and I'd love to get some eyes on it (and hopefully get it in) -- and you seem the guy who has been making that happen. Anything I need to do to move this forward? Thanks!

jwatzman:
Ping again, hopefully someone can eventually take a look at this...

jwatzman:
Ping again -- @captbaritone et al, any chance of getting this imported and reviewed? I just rebased it again :)

captbaritone:
Sorry for the slow response. I spent some time today validating this change and it looks correct. It makes a meaningful improvement internally for us as well. I think it also highlights another bug in our raw response typegen which I will try to followup with.

jwatzman:
No worries on the delay, just glad it's making progress! Thanks for taking a look.

PR #5102, opened 2025-11-22
fix: export `MatchContainer` from react-relay package

giautm:
When playing with Client 3D, I found the MatchContainer's props is exported, but the component didn't. So this should fix the problem.

https://github.com/giautm/relay-3d-example/blob/d479ba493d9c39e3946d12fcffb7c7a1f3977f2d/directories-3d/components/common/RelayMatchContainer.tsx#L2-L4

The document said it should exported from `react-relay` package:
https://github.com/facebook/relay/blob/fe7af8846aaaa195e292ffe2f1a1825738ba0228/website/docs/guides/data-driven-dependencies/server-3d.md?plain=1#L111-L114


giautm:
Any update there?

PR #5108, opened 2025-11-30
@oneOf input objects stricter typegen

### LINKED ISSUE #5095: Support `@oneOf` in generated flow/typescript types
Now that `@oneOf` is part of the GraphQL spec, it'd be very useful if the generated types for field arguments modelled its usage correctly. I don't _think_ any additional support is required.

Malien:
I am a man of my word: https://discord.com/channels/625400653321076807/862957601670430731/1438602698575577099

Makes so that `@oneOf` input types generate a more strict type union, that permits only a single property being set a a time. This is technically a breaking change, since people may have relied on a more permissive type in their application logic. So there is an escape hatch of a `oneOfType: "relaxed"`. I'm not sure this is a necessary option to have. We might as well break a tiny (like real tiny) amount of people's types.

Closes #5095


Malien:
I might want to run the whole rustfmt / fixture-gen to fix the formatting checks

compiler/crates/relay-typegen/src/visit.rs
```diff
@@ -2461,6 +2467,82 @@ fn transform_non_nullable_input_type(
...
+                                encountered_enums,
+                                custom_scalars,
+                            ),
+                        })
+                    } else {
+                        Prop::KeyValuePair(KeyValuePairProp {
+                            key: field.name.item.0,
+                            read_only: false,
+                            optional: true,
+                            value: AST::Empty,
```
captbaritone: Can you share a bit of context (here, and ideally in a comment) on why we choose to include these fields but type them as `empty`/`never` instead of simply omitting them from the union variant?

compiler/crates/relay-typegen/src/visit.rs
```diff
@@ -2420,36 +2428,34 @@ fn transform_non_nullable_input_type(
...
-                                    value: transform_input_type(
-                                        typegen_context,
-                                        &field.type_,
-                                        input_object_types,
-                                        encountered_enums,
-                                        custom_scalars,
-                                    ),
-                                })
-                            })
+                    let node = if typegen_context.project_config.typegen_config.one_of_type
```
captbaritone: Thanks for making this a config option!

compiler/crates/relay-typegen/src/visit.rs
captbaritone: If it's a considered choice with benefits, I'm fine with it. I just wasn't sure why we had made that choice.

Malien:
@captbaritone I think we're gtg

PR #5128, opened 2026-01-06
Fix Windows path separator issue in Rust tests

captbaritone:
## Summary

Fixes 61 failing Rust integration tests on Windows by normalizing path separators in artifact map debug output.

## Problem

The Rust tests in `relay_compiler_relay_compiler_integration_test` were failing on Windows with snapshot mismatches like:

```diff
- Path: __generated__/barMutation.graphql.js
+ Path: __generated__\barMutation.graphql.js
```

This occurred because `PathBuf`'s `Debug` implementation uses platform-native path separators (backslashes on Windows), but the test snapshots expect forward slashes for cross-platform consistency.

## Solution

Added a custom `Debug` implementation for `ArtifactRecord` in `compiler/crates/relay-compiler/src/artifact_map.rs` that:
1. Converts paths to strings using `to_string_lossy()`
2. Replaces all backslashes with forward slashes
3. Uses the normalized path in the debug output

## Testing

This fix targets the Windows-specific test failures. CI will verify:
- All 61 previously failing tests on Windows now pass
- Existing tests on macOS and Linux continue to pass
- No regressions in other test suites

## Related

Fixes CI failures visible in recent runs on the `main` branch.

### DIFF
--- a/compiler/crates/relay-compiler/src/compiler_state.rs
+++ b/compiler/crates/relay-compiler/src/compiler_state.rs
@@ -364,7 +364,9 @@ impl fmt::Debug for ProjectArtifactMap {
                         sorted_records.sort_by_key(|record| &record.path);
 
                         for record in sorted_records {
-                            output.push_str(&format!("    Path: {}\n", record.path.display()));
+                            // Normalize path separators to forward slashes for cross-platform consistency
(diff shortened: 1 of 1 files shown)


captbaritone:
This was created as a validation of #5129. I already had a better hand-crafted fix in D90190629

Issue #4548, opened 2023-12-11
Distribute react-relay and other packages as ESM


gajus:
Feels like this change is well overdue.

captbaritone:
I agree. This would be a welcome improvement. When setting up Relay for side projects I often hit rough edges because of this.

If you, or anyone else, wants to start collecting a full picture of what work would be required here, that would be very helpful. Some questions on my mind:

* Is it a prerequisite to move to ESM within Relay? Currently we use commonJS require/module.exports everywhere. Internally, we've long wanted to move to ES6 imports/exports but simply haven't had the bandwidth to prioritize it.
* What changes would we need to make to our build pipeline? Is it possible with the tools we have today?
* Can we support both commonJS and ESM? Will this be a breaking/problematic change for some users?

gajus:
Highly theoretically, you could just release `relay-esm` package that simply re-exports everything as ESM:

```ts
// Because Relay is a CommonJS module, we are having problems with Vite bundling it.
// Until Relay becomes an ESM module, we need to use this package to re-export it.
import { default as reactRelay } from 'react-relay';
import { default as relayRuntime } from 'relay-runtime';

export { type PreloadedQuery, type UseMutationConfig } from 'react-relay';
export {
  type CacheConfig,
  type Disposable,
  type FetchPolicy,
  type FragmentRefs,
  type GraphQLResponse,
  type GraphQLTaggedNode,
  type MutationParameters,
  type OperationType,
  type PayloadError,
  type RequestParameters,
  type SelectorStoreUpdater,
  type Variables,
  type VariablesOf,
} from 'relay-runtime';

export const ConnectionHandler = relayRuntime.ConnectionHandler;
export const Environment = relayRuntime.Environment;
export const Network = relayRuntime.Network;
export const QueryResponseCache = relayRuntime.QueryResponseCache;
export const RecordSource = relayRuntime.RecordSource;
export const Store = relayRuntime.Store;

export const loadQuery = reactRelay.loadQuery;
export const fetchQuery = reactRelay.fetchQuery;
export const useLazyLoadQuery = reactRelay.useLazyLoadQuery;
export const usePreloadedQuery = reactRelay.usePreloadedQuery;
export const graphql = reactRelay.graphql;
export const useFragment = reactRelay.useFragment;
export const useMutation = reactRelay.useMutation;
export const usePaginationFragment = reactRelay.usePaginationFragment;
export const useRelayEnvironment = reactRelay.useRelayEnvironment;
[...]

PR #5162, opened 2026-02-05
Support named exports with nodejs ESM

robrichard:
It is currently not possible to use named exports with Relay in a Node.js ESM file. For example, in a brand new project with the latest version of relay-runtime as a dependency, the following code will error:

index.mjs:
```js filename="index.mjs"
import { fetchQuery } from 'relay-runtime'
```

```
file:///index.mjs:1
import { fetchQuery } from 'relay-runtime';
         ^^^^^^^^^^
SyntaxError: Named export 'fetchQuery' not found. The requested module 'relay-runtime' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'relay-runtime';
const { fetchQuery } = pkg;

    at #asyncInstantiate (node:internal/modules/esm/module_job:302:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:405:5)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:660:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5)
```

When an ESM file `import`s a common js file (like relay-runtime) Node.js will use [cjs-module-lexer](https://github.com/nodejs/cjs-module-lexer/tree/2.0.0) to try to determine compatible named exports.

[From the node.js docs](https://nodejs.org/api/esm.html):
> Named exports detection covers many common export patterns, reexport patterns and build tool and transpiler outputs. See [cjs-module-lexer](https://github.com/nodejs/cjs-module-lexer/tree/2.0.0) for the exact semantics implemented.

In the entry points of react-relay and relay-runtime, the pattern of exporting a property on another object causes `cjs-module-lexer` to not detect any of named exports.

Example:

```js
// This doesn't work
module.exports = {
  fetchQuery: RelayRuntime.fetchQuery
};

// This does work
const { fetchQuery } = RelayRuntime;
module.exports = {
  fetchQuery
};
```

In this PR, I updated the entrypoints of react-relay and relay-runtime to allow Node.js's cjs named export detection to work properly.  I added a test case to the Gulp build process since to test this we need the final transpiled code, but I am open to other approaches.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -89,6 +89,9 @@ module.exports = {
     // Custom rules for our own codebase
     'relay-internal/no-mixed-import-and-require': 'error',
     'relay-internal/sort-imports': 'error',
+
+    // Enabled only in relay package entry files
+    'relay-internal/esm-compatible-cjs': 'off',
   },
   overrides: [
     {
--- a/packages/eslint-plugin-relay-internal/lib/index.js
+++ b/packages/eslint-plugin-relay-internal/lib/index.js
@@ -17,4 +17,5 @@ module.exports.rules = {
   'no-mixed-import-and-require': require('./rules/no-mixed-import-and-require'),
   'sort-imports': require('./rules/sort-imports'), // Synced from WWW
   'no-for-of-loops': require('./rules/no-for-of-loops'),
+  'esm-compatible-cjs': require('./rules/esm-compatible-cjs'),
 };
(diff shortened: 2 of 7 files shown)


jantimon:
@captbaritone instead of adding workarounds could we please move forward with https://github.com/facebook/relay/issues/4548 instead?

Issue #2870, opened 2019-09-25
[relay-compiler]: Union types


artola:
There is an "strange" generation of Flow typings depending on how a query is written. I wrote several examples using the `star-wars` schema (see repo). 
Mainly: require `primaryFunction` in a "human hero" should not be possible, not that it returns  `undefined`, because it does not exists in the `Human` type.

```ts
// required
graphql`
  query trialQuery {
    hero {
      name
      ... on Human {
        homePlanet
      }
      ... on Droid {
        primaryFunction
      }
    }
  }
`;

// obtained
export type trialQueryResponse = {|
  +hero: {|
    +name: string,
    +homePlanet?: ?string,
    +primaryFunction?: string,
  |}
|};

// expected
export type trial4QueryResponse = {|
  +hero: {|
    +__typename: "Human",
    +name: string,
    +homePlanet: ?string,
  |} | {|
    +__typename: "Droid",
    +name: string,
    +primaryFunction: string,
  |} | {|
    // This will never be '%other', but we need some
    // value in case none of the concrete values match.
    +__typename: "%other"
  |}
|};
```

In our TS project obtaining strict union typings is required, see issue and discussion: https://github.com/relay-tools/relay-compiler-language-typescript/issues/135

Repo: (wait 1 min)

I would like to re-check if it is a bug (see repo, at least "case 2").

kassens:
This is a little bit difficult to express. `name` will also be present if the type is neither Human nor Droid. I think we could generate the other fields under the "other" union member. Unfortunately, the code that's converting the Schema to (flow) types is pretty messy and I'm not sure if it's easy to change :(

artola:
@kassens Even the newest **relay-compiler@main** it is not able to make the union types as expected. Only when all the fields are declared inside each type and containing a `__typename` then we see the right unions. It would be great to have them without this nasty workaround.

```graphql
graphql`
  query demo8Query {
    hero {
      ... on Human {
        __typename
        name
        homePlanet
      }
      ... on Droid {
        __typename
        name
        primaryFunction
      }
    }
  }
```

obtained:

```ts
export type demo8Query$data = {|
  +hero: {|
    +__typename: "Human",
    +name: string,
    +homePlanet: ?string,
  |} | {|
    +__typename: "Droid",
    +name: string,
    +primaryFunction: string,
  |} | {|
    // This will never be '%other', but we need some
    // value in case none of the concrete values match.
    +__typename: "%other",
  |},
|};
```

[...]

Issue #3990, opened 2022-06-22
Spreading fragment on a union type might cause runtime error because ref can be undefined


waynezhang1995:
lets say i have this union type defined in the schema
```
union U = A | B

type Query {
   unionType: U
}
```
and used by a query
```
query SomeQuery {
   unionType {
       ... on A {
          ...someFragment
       }
       ... on B {
           id
       }
    }
}
```

The Relay compiler wont flag any issue of the above query but if the `unionType` resolves to `B`, we will get a runtime error: 

```Invariant Violation: Relay: Expected to receive an object where `...someFragment` was spread, but the fragment reference was not found`. This is most likely the result of:
    - Forgetting to spread `someFragment` in `useFragment()`'s parent's fragment.
    - Conditionally fetching `someFragment` but unconditionally passing a fragment reference prop to `useFragment()`. If the parent fragment only fetches the fragment conditionally - with e.g. `@include`, `@skip`, or inside a `... on SomeType { }` spread  - then the fragment reference will not exist. In this case, pass `null` if the conditions for evaluating the fragment are not met (e.g. if the `@include(if)` value is false.)```.

There is no easy way to detect this issue until runtime so wondering if the relay compiler can give a warning message.

**Comment by captbaritone** [comment]
We have a proposed (started but not fully finished) feature called `@alias` which aims to close this gap. The idea is to give fragment spreads names in these ambiguous situations such that we can expose them as nullable properties. This would allow TypeScript/Flow to ensure that you handle the possibility that the fragment did not match before using it to render a component.

If anyone is interested in helping drive this forward let me know. Most of the work is done, but there's still a list of things that need to be done.

I haven't had time to document it in open source, but here's a copy paste from an internal doc:

---
# @alias Directive on Fragment Spreads

The `@alias` directive allows you to expose a spread fragment — either a named fragment spread or an inline fragment — as a named field within your selection. This allows Relay to provide additional type safety in the case where your fragment’s type may not match the parent selection.

Let’s look at an example. Imagine you have a component that renders information about a `Viewer`:

```
function MyViewer({viewerKey}) {
  const {name} = useFragment(graphql`
    fragment MyViewer on Viewer {
      name @required(action: THROW)
    }`, viewerKey);
  
  return `My name is ${name}. That's ${name.length} letters long!`;
}
```

To use that component in a component that has a fragment on `Node` (which `Viewer` implements), you could write something like this:

```
function MyNode({nodeKey}) {
  const node = useFragment(graphql`
[...]

PR #5186, opened 2026-02-25
[relay-compiler]: Union types - issue 2870

morrys:
# [TypeGen] Refactor discriminated union handling in `visit.ts`

## Summary
This PR refactors how Relay TypeGen handles discriminated unions in `visit.ts` issue: https://github.com/facebook/relay/issues/2870. The main goals are:

- Remove unused functions.
- Simplify and optimize `should_emit_discriminated_union`.
- Correctly merge common fields across concrete types and handle `__typename` consistently.
- Adjust test expectations to reflect the updated union and typename handling logic.

---

### 1. Remove unused function
Removed:

```rust
fn has_typename_selection(selections: &[TypeSelection]) -> bool
```
`should_emit_discriminated_union` no longer uses this function and no other code depends on it.

### 2. Update should_emit_discriminated_union

**Before:**

```rust
if by_concrete_type.is_empty() || !concrete_type.is_abstract_type() {
return false;
}

base_fields.values().all(TypeSelection::is_typename)
&& (base_fields.values().any(TypeSelection::is_typename)
|| by_concrete_type.values().all(|selections| has_typename_selection(selections)))
```

**After:**

```rust
if by_concrete_type.is_empty() || !concrete_type.is_abstract_type() || base_fields.is_empty() {
return false;
}

return base_fields.values().any(TypeSelection::is_typename);
```

**Notes:**

First change is an optimization: `base_fields.is_empty()` is equivalent to the previous `all(...) && any(...)` check.

Second change is more impactful: removed `|| by_concrete_type.values().all(...)`. Now,`__typename` is checked in the parent and treated as a common field.

### 3. Update `get_discriminated_union_ast`
Two major changes:
 * merge concrete type fields with parent fields to correctly handle nested common fields.
 * for the `%other` case, include all parent fields instead of only `__typename`. This ensures the "other" object has complete common data.

### 4. **Test Updates**
1. **relay-client-id:** handled as a union.
2. **relay-resolver-with-output-type-client-interface:** no longer treated as a union.
3. **updatable tests**

6. **typename-on-union-with-non-matching-aliases:** 
 * _typename moved to root.
 *  added usertype: __typename to prevent empty User selection.

7. **typename-on-union**

TypenameOutsideWithAbstractType
 * sorted street alphabetically.
 * removed __typename from union; now treated as a common field.

TypenameInside
* no longer treated as a union (major behavioral change).
* correct approach: move __typename outside to handle as common data.

TypenameWithCommonSelections
 * notes: %other now includes all common fields, not just __typename.
```ts
export type TypenameWithCommonSelections$data = {
readonly __typename: "Page";
readonly name: string | null | undefined;
readonly username: string | null | undefined;
readonly $fragmentType: "TypenameWithCommonSelections";
} | {
readonly __typename: "User";
readonly firstName: string | null | undefined;
readonly name: string | null | undefined;
readonly $fragmentType: "TypenameWithCommonSelections";
} | {
// This will never be '%other', but we need some
// value in case none of the concrete values match.
readonly __typename: "%other";
readonly name: string | null | undefined;
readonly $fragmentType: "TypenameWithCommonSelections";
};
```
NEW TEST: TypenameWithNestedCommonSelections: demonstrates correct handling of nested common fields.

Input:

```graphql
fragment TypenameWithNestedCommonSelections on Actor {
__typename
name
emailAddresses
address { city } # common
... on User {
firstName
address { street } # only here
}
... on Page {
username
address { country } # only here
}
}
```

Output:

```ts
export type TypenameWithNestedCommonSelections$data = {|
+__typename: "Page",
+address: ?{| +city: ?string, +country: ?string |},
+emailAddresses: ?ReadonlyArray<?string>,
+name: ?string,
+username: ?string,
+$fragmentType: TypenameWithNestedCommonSelections$fragmentType,
|} | {|
+__typename: "User",
+address: ?{| +city: ?string, +street: ?string |},
+emailAddresses: ?ReadonlyArray<?string>,
+firstName: ?string,
+name: ?string,
+$fragmentType: TypenameWithNestedCommonSelections$fragmentType,
|} | {|

// This will never be '%other', but we need some
// value in case none of the concrete values match.
+__typename: "%other",
+address: ?{| +city: ?string |},
+emailAddresses: ?ReadonlyArray<?string>,
+name: ?string,
+$fragmentType: TypenameWithNestedCommonSelections$fragmentType,
|};
```


compiler/crates/relay-typegen/tests/generate_flow/fixtures/typename-on-union-with-non-matching-aliases.expected
```diff
@@ -1,7 +1,8 @@
 ==================================== INPUT ====================================
 fragment Foo on Node {
+  __typename
```
captbaritone: See above about this being a breaking change.

compiler/crates/relay-typegen/src/visit.rs
```diff
@@ -1668,15 +1692,10 @@ fn should_emit_discriminated_union(
...
+    if by_concrete_type.is_empty() || !concrete_type.is_abstract_type() || base_fields.is_empty() {
         return false;
     }
-
-    base_fields.values().all(TypeSelection::is_typename)
-        && (base_fields.values().any(TypeSelection::is_typename)
-            || by_concrete_type
-                .values()
-                .all(|selections| has_typename_selection(selections)))
+    return base_fields.values().any(TypeSelection::is_typename);
```
captbaritone: Previously

```
... on SomeConcreteType {
  __typename
  someField
}
... on SomeOtherConcreteType {
  __typename
  anotherField
}
```

Would emit a discriminated union. Now it won't. That's going to be a big breaking change for a lot of users. Is there a reason we have to lose that?

morrys: this isn’t a blocking issue—more about keeping things clean and having clearer control over when the discriminated union is emitted, especially since this is a common field shared across types.

It is technically a breaking change, but since it only affects typing/compilation, fixes should be guided by type errors and shouldn’t impact runtime logic.

That said, if this is a concern, I can add the condition by_concrete_type.values().all(|selections| has_typename_selection(selections)), resulting in:

```rust
return base_fields.values().any(TypeSelection::is_typename)
    || by_concrete_type.values().all(|selections| has_typename_selection(selections));
```

compiler/crates/relay-typegen/tests/generate_flow/fixtures/relay-resolver-with-output-type-client-interface.expected
```diff
@@ -51,11 +51,7 @@ export type User__pop_star_game$normalization$variables = {||};
 export type User__pop_star_game$normalization$data = {|
   +client_interface: {|
     +__typename: "ClientType",
-    +name: ?string,
-  |} | {|
-    // This will never be '%other', but we need some
-    // value in case none of the concrete values match.
-    +__typename: "%other",
+    +name?: ?string,
```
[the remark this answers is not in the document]
morrys: In this specific case, the issue seems to stem from how Relay RelayResolver is currently handled.

There is a transformer, [generate_relay_resolvers_operations_for_nested_objects](https://github.com/facebook/relay/blob/main/compiler/crates/relay-transforms/src/generate_relay_resolvers_operations_for_nested_objects.rs#L362), which previously ensured that __typename was generated. This allowed the type to be correctly detected by the condition:

```rust
by_concrete_type.values().all(|selections| has_typename_selection(selections))
```

Looking a bit deeper, it seems feasible to remove the logic responsible for generating __typename from this transformer and instead introduce a dedicated transformer that handles the new behavior.

I tried building a draft implementation of a transformer for this, and I was able to correctly produce `%other` for `pop_star_game`, while`pop_star_name` is still being ignored. I’m not entirely sure whether `%other` is actually required for that case as well.

I’ll continue looking into this transformer. From testing a few other cases, it seems to handle unions correctly without changes to the existing code, so it could be a good fourth option.

* Support the old heuristic
* Add a feature flag which enables users to opt back into the old behavior
* Add a codemod that inserts a top-level __typename in all the places where we were previously emitting a discriminated union.
* Add Generate Typename Field Transform

morrys:
I added the `generate_typename_union` transform, which automatically inserts `__typename` into selection sets when needed for type generation, especially for abstract types with concrete refinements.

This is consistent with the new `should_emit_discriminated_union logic`: the discriminant is now explicitly available in `base_fields`.

This simplifies usage on the user side: it is no longer necessary to manually add `__typename` in fragments or queries, since the compiler will insert it automatically in the correct place when needed.

This avoids a breaking change and does not require updating existing queries, nor introducing feature flags or codemods.

In this commit you will notice changes in several generated and/or expected files. This is expected, as discriminated unions are now correctly generated even in cases where `__typename` was not explicitly present in the query.

@captbaritone let me know what you think, and feel free to take another look when you have time 👍

compiler/crates/relay-transforms/src/assignable_fragment_spread/validate_updatable_directive.rs
```diff
@@ -163,21 +163,7 @@ impl<'a> UpdatableDirective<'a> {
...
-        // as users would be able to assign to scalar fields that are not present on a given type.
         let mut errors = vec![];
-        if parent_field.selections.len() != fragment_spreads.len() {
-            errors.push(Diagnostic::error(
-                ValidationMessage::UpdatableOnlyInlineFragments {
-                    outer_type_plural: self.executable_definition_info.unwrap().type_plural,
-                },
-                parent_field.definition.location,
-            ));
-        }
```
tyao1, line 180: Why are these validations removed?

morrys: Hi @tyao1,

right now the validation only allows fields with inline fragments, like this:
```ts
query TestQuery @updatable {
  node(id: 4) {
    ... on Page {
      name
    }
    ... on Comment {
      firstName
    }
  }
}
```
however, with the new union logic, including __typename should also be valid:
```ts
query TestQuery @updatable {
  node(id: 4) {
    __typename
    ... on Page {
      name
    }
    ... on Comment {
      firstName
    }
  }
}
```
because of that, the current check is no longer correct, so I removed it.

Alternatively, we could relax it to also allow __typename fields, what do you think?

morrys: @tyao1 friendly ping 👍  any updates on the PR? Is there anything I should do?

tyao1:
Sorry the previous import failed due to merge conflicts, I'm doing a new import

tyao1:
Thanks again, @morrys I rebased and run it internally and it introduce many new flow errors. Used Claude to debug the flow error:

The publicly reproducible tip is in Relay's own tests (`useFragment-with-required-test.js` and `RelayMockEnvironment-test.js`); internally it fans out to the same handful of shapes across many surfaces.

They all trace to one root cause but surface as **four distinct breaking patterns**.

## Root cause (recap)

The `should_emit_discriminated_union` change (`visit.rs`) from `.all(is_typename)` → `.any(is_typename)` now emits a discriminated union for **any** abstract-typed field with a concrete inline fragment — `field { ... on Concrete { … } }` — whenever `__typename` is a common field (it's auto-added for refinement). The `%other` arm carries only common *scalar* fields — **not** fragment spreads, and not fields that live under the concrete arm:

```js
field: ?({ __typename: "Concrete", /* concrete fields */, $fragmentSpreads: … }
       | { __typename: "%other" })   // ← no spreads, no concrete fields
```

That breaks four consumer patterns:

### 1. Fragment key → `useFragment` / a hook / a component prop (`..._$key`) — most common
The `%other` arm has no `$fragmentSpreads`, so it isn't a `HasSpread<T>`. OSS repro:
```js
// useFragment-with-required-test.js
useFragment(SomeFragment, data.node)
```
```
Cannot call `useFragment` because property `$fragmentSpreads` is missing in
object type [the %other arm] but exists in `HasSpread` [useFragment's key param].
```
Any hook / component prop / utility that takes a `..._$key` and is passed the field directly (without narrowing first) hits this.

### 2. `@module` / `MatchContainer` — `match={data}`
The union is incompatible with `MatchPointer | TypenameOnlyPointer` (`$fragmentSpreads` missing / `$fragmentType` extra; exact objects reject extras).
```
Cannot create `MatchContainer` element because in property `match`:
 - Either `$fragmentSpreads` is missing … but exists in `MatchPointer` [2].
 - Or `$fragmentType` is extra … but missing in `TypenameOnlyPointer` [3].
```

### 3. Direct field access — `data.field` / `data?.field`
A field that used to be flat now lives only under the concrete arm (absent from `%other`), so Flow types it as `void`/non-existent — breaking `switch`, comparisons, and reads.
```js
switch (data.someField) { … }   // data.someField is now `void` on the %other arm
```
```
Invalid check of case test against switch discriminant because void
(due to access of non-existent property `someField`) …
```

### 4. `__typename` narrowing to a non-selected type
Since the union's `__typename` is only `"<SelectedConcrete>" | "%other"`, comparing to any other concrete type is now an impossible comparison.
```js
data.__typename === 'SomeOtherType'
```
```
Cannot compare `data.__typename` to string literal 'SomeOtherType' … [invalid-compare]
```

## Why this matters

`field { ... on Concrete { …spread / …fields } }` on an abstract / interface / union parent is one of the most common shapes in a large Relay codebase, so this is effectively a **trunk-wide breaking change**. Migrating every consumer to narrow on `__typename` first isn't feasible within this PR.

## Ask

Was broadening `should_emit_discriminated_union` to `.any(is_typename)` intentional, or should it stay scoped?

- If **not intended**: scoping the trigger down should eliminate the vast majority of these — e.g. only emit the union when `__typename` is *explicitly* selected alongside multiple concrete arms, and keep flat typing (with `$fragmentSpreads` and concrete fields hoisted) for the single-concrete-inline-fragment case.
- If **intended** (to make the "ref may be undefined at runtime" case from #3990 explicit): there'd need to be a migration path for the affected consumers before this can land in a large codebase.

morrys:
Hi @tyao1
yes, broadening the discriminated union behavior was intentional.

### About `.all()` and `.any()`

I do not think the single-concrete-type errors are primarily caused by changing:

```rust
.all(TypeSelection::is_typename)
```

to:

```rust
.any(TypeSelection::is_typename)
```

These checks operate on `base_fields`; they do not check the number or identity of the associated concrete types.

With the current behavior:

* no common fields and no `__typename` inside the concrete selection → no union;
* a common field such as `id` → no union, because `.all(is_typename)` is false;
* `__typename` as a common field → union.
* `__typename` inside every concrete selection → union;

With this PR, `__typename` is automatically added to the common fields. Therefore, in the single-concrete-arm case, both `.all(is_typename)` and `.any(is_typename)` are true.

Changing `.any()` back to `.all()` would not remove those errors. It would only affect cases where `base_fields` contains `__typename` together with other common fields.

### Why the behavior is intentional

As @captbaritone explained, an abstract field may return another concrete implementor, including one added after the client was compiled.

For example:

```graphql
node {
  ... on User {
    ...UserFragment
  }
}
```

If the server returns another `Node` implementation, Relay does not traverse the `User` inline fragment. The `User` fields and fragment reference are therefore not present at runtime.

For this reason, I think these errors are correct and useful:

* passing `data.node` directly to `useFragment`;
* accessing fields that only exist on the concrete `User` arm.

They expose at compile time situations that could otherwise fail at runtime.

### Comparing another typename

A comparison such as:

```js
data.__typename === 'Admin'
```

is a separate case. Concrete types not explicitly modelled by the selection are represented by `%other`.

The consumer should either:

* handle them through an `else` or `default` branch; or
* explicitly model that concrete type in the GraphQL selection.

### Proposed changes

#### 1. `MatchContainer`

I verified that the `MatchContainer` error can be fixed independently by making `TypenameOnlyPointer` actually inexact, consistently with its existing comment:

```js
type TypenameOnlyPointer = {
  readonly __typename: string,
  ...
};
```

I think this is a valid fix and should be included.

#### 2. Single concrete arm

I still think emitting the union is semantically correct even when there is only one concrete arm.

If you prefer, we could investigate whether the automatic insertion of __typename should be restricted to selections containing multiple concrete arms.

With that compatibility-oriented behavior:

* multiple concrete arms → automatically add `__typename` and emit the union;
* one concrete arm with an explicitly selected `__typename` → emit the union;
* one concrete arm without an explicit `__typename` → preserve the existing flat type.

I see this as a backward-compatibility compromise rather than the semantically ideal behavior, since the possible runtime mismatch would remain hidden in the single-arm case.

tyao1:
@morrys Makes sense. I think this is the right change to do, but internally we have 3000+ flow errors when compiled on our codebase, so we will have to roll this out gradually. Could you add a feature flag for this change?

morrys:
@tyao1 I understand the issue. I wasn’t expecting such a broad impact on the codebase.

Just to confirm: would you like the feature flag to cover the entire new behavior introduced by the PR, or only union generation for single-concrete-arm cases?

I would keep the MatchContainer fix outside the flag, since it is an independent correction.

Which of the two options did you have in mind?

tyao1:
@morrys I'm not sure if the single concrete arm case covers all the flow errors, so maybe the entire PR.  But if the single-concrete-arm case is easier for you to do, let's do that and I can test if it solves all errors.

morrys:
Hi @tyao1, I added the feature flag for the entire behavior introduced by the PR.

The new flag is:

```text
enable_typename_discriminated_unions
```

It supports the existing `enabled`, `disabled`, `limited`, `rollout`, and `rolloutRange` modes, using the operation or fragment definition name.

The main changes are:

* added the feature flag check to the automatic `__typename` insertion;
* added the feature flag check to the broader discriminated-union generation in `visit.rs`;
* restored the previous typegen and validation behavior when the flag is disabled;
* added tests for the different feature flag modes, including definition-level `limited` behavior;
* kept the improved common-field merging logic in `get_discriminated_union_ast` independently from the flag;
* fixed `MatchContainer` independently by making `TypenameOnlyPointer` inexact.

When the flag is enabled, the complete new behavior is used, including the single-concrete-arm case.

One CI note: the `Build Relay unit-test artifacts` step currently fails with these undefined fragments:

```text
ShadowMagicThingResolver_ReturnFragment
ShadowWeakDuoResolver_ReturnFragment
```

I reproduced the same error on the current `main`, so it seems unrelated to this PR.

Let me know what you think.

tyao1:
@morrys I merged PR, but to get it to land internally I had to revert the type change to MatchContainer types because it caused some flow errors internally on some other components that uses the types

Issue #5188, opened 2026-02-25
@defer data silently dropped when server returns deferred fragment inline (selective defer)


richardtagger:
# @defer data silently dropped when server returns deferred fragment inline (selective defer)

## Summary

When a GraphQL server selectively honors `@defer` — deferring some fragments but returning others inline in the initial payload — Relay silently drops the inline deferred data. The affected component renders with `undefined` data despite the data being present in the network response.

This is a violation of the GraphQL [DeferStream RFC](https://github.com/graphql/graphql-wg/blob/main/rfcs/DeferStream.md) which states that `@defer` is a hint and clients must handle responses where the server chooses not to defer:

> "GraphQL clients should be able to process a response that ignores the defer/stream directives."

Related: #3904 — this issue extends that report to cover the **selective defer** case (multipart response where some `@defer` directives are honored and others are not).

### Schema

```graphql
type Query {
  viewer: Viewer
}

type Viewer {
  name: String
  email: String
  posts: [Post]
}

type Post {
  id: ID
  title: String
}
```

### Query

```graphql
query ViewerQuery {
  viewer {
    ...ViewerNameFragment
    ...ViewerEmailFragment @defer
  }
  ...ViewerPostsFragment @defer
}

fragment ViewerNameFragment on Viewer {
  name
}

fragment ViewerEmailFragment on Viewer {
  email
}

fragment ViewerPostsFragment on Query {
  viewer {
    posts {
      id
      title
    }
  }
}
```

### Server response (multipart)

The server defers `ViewerEmailFragment` but does NOT defer `ViewerPostsFragment`, including its data inline:

**Part 1 — Initial payload:**
```json
{
  "data": {
    "viewer": {
      "name": "Alice",
      "posts": [
        { "id": "1", "title": "Hello World" }
      ]
    }
  },
  "hasNext": true
}
```

Note: `viewer.posts` is present (server did not defer `ViewerPostsFragment`).

**Part 2 — Deferred chunk (only `ViewerEmailFragment`):**
```json
{
  "incremental": [{
    "path": ["viewer"],
    "label": "ViewerQuery$defer$ViewerEmailFragment",
    "data": { "email": "alice@example.com" }
  }],
  "hasNext": false
}
```

Note: No incremental chunk for `ViewerPostsFragment` — the server already sent that data inline.

### Expected behavior

- `ViewerNameFragment` renders `"Alice"` ✅
[...]

PR #5189, opened 2026-02-25
fix: always normalize @defer selections from initial payload

### LINKED ISSUE #3904: Cannot process a response that ignores @defer directives
It has been discovered that the Relay runtime currently comes with the expectation that fragments marked with the `@defer` directive are expected to be deferred. Even when the network returns its corresponding value. This leaves fragments marked with the directive to have an undefined value. As far as we can tell, it is due to [this function](https://github.com/facebook/relay/blob/main/packages/relay-runtime/store/RelayResponseNormalizer.js#L345) which bakes the assumption into the library.

This behavior seems to be against the [current specification](https://github.com/graphql/graphql-spec/pull/742/files#diff-30a69c5a5eded8e1aea52e53dad1181e6ec8f549ca2c50570b035153e2de1c43R2175-R2181), and the [original RFC](https://github.com/graphql/graphql-wg/blob/main/rfcs/DeferStream.md#:~:text=Therefore%2C%20GraphQL%20clients%20should%20be%20able%20to%20process%20a%20response%20that%20ignores%20the%20defer/stream%20directives.):

> ... GraphQL clients _must_ be able to process a response that ignores the `@defer` and/or `@stream` directives. 

We had to make some larger changes to our execution engine to support defer. Having the ability to toggle the defer lets us roll out the changes slowly and fallback to previous codepaths in the event of regression. This is also a blocker for letting us enable advanced usage cases in the future. For example, one thing we were considering was merging unnecessarily deferred fragments back into the initial response to the client to avoid having to do a pass over the store a second time when not necessary.

Is there a possible workaround or a relatively trivial way to support this behavior?

**Code Sample**
```jsx
function Screen() {
 const query = useLazyLoadQuery(
   graphql`
     query ScreenQuery {
       viewer {
         name
       }
       ...ScreenContent @defer(label: "ScreenContentFragment")
     }
   `,
   {},
 );
 return (
   <div>
     <div>{query.viewer.name}</div>
     <ScreenContent queryRef={query} />
   </div>
 );
}
 
function ScreenContent(props) {
 const content = useFragment(
   graphql`
     fragment ScreenContent on Query {
       newsArticles {
         edges {
           node {
             title
           }
         }
       }
     }
   `,
   props.queryRef,
 );
 console.log(content); // <---- content is null
 return <div>News Articles</div>;
}
```

Response from the server:
```
{
  "data": {
    "viewer": {
      "name": "Ozzie"
    },
    "newsArticles": {
      "edges": [
        {
          "node": {
            "title": "Hello World"
          }
        }
      ]
    }
  }
}
```

**Example Project**: https://github.com/kirkbyo/relay-use-fragment-defer

richardtagger:
## Summary

Fixes #3904, #5188

When a GraphQL server selectively honors `@defer` — deferring some fragments but returning others inline in the initial payload — Relay's `_normalizeDefer` skips traversing the inline selections entirely, silently dropping the data. The affected component renders with `undefined` despite the data being present in the network response.

Per the [DeferStream RFC](https://github.com/graphql/graphql-wg/blob/main/rfcs/DeferStream.md), `@defer` is a hint: servers MAY choose not to defer on a per-fragment basis, and clients must handle this gracefully.

## The Problem

`_normalizeDefer` currently uses an either/or pattern:

```js
if (isDeferred === false) {
  this._traverseSelections(defer, record, data);  // normalize inline
} else {
  this._incrementalPlaceholders.push({...});       // skip normalization
}
```

When `isDeferred` is truthy, the data is **not normalized into the store**. If no incremental chunk arrives for that label (because the server included the data inline), the data is permanently lost.

## The Fix

Always traverse selections, then create the placeholder:

```js
this._traverseSelections(defer, record, data);
if (isDeferred !== false) {
  this._incrementalPlaceholders.push({...});
}
```

**Safe because:**
- **Server didn't defer (data inline):** `_traverseSelections` normalizes the data. Placeholder exists but data is already in the store.
- **Server did defer (data absent):** `_traverseSelections` encounters undefined fields and skips them (no writes). The real incremental chunk normalizes the data normally via `_processDeferResponse`.
- **No behavior change** for `@defer(if: false)` or non-streaming responses.

## Selective Defer Scenario

This is particularly important for the **selective defer** case where a server defers some `@defer` fragments but not others in the same multipart response:

1. Initial payload (`hasNext: true`): contains data for Fragment A (not deferred by server) + placeholder for Fragment B (deferred)
2. Incremental chunk: delivers Fragment B data
3. **Before fix:** Fragment A data silently lost — normalizer skipped it
4. **After fix:** Fragment A data normalized from initial payload, Fragment B arrives via incremental delivery — both work

## Verification

Verified in production at Microsoft/Viva Engage via a pnpm patch against `relay-runtime@20.1.1`. The `PersonalActivityDashboard` page uses `@defer` on two fragments where the server defers one but not the other — both now render correctly with this fix.

### DIFF
--- a/packages/relay-runtime/store/RelayResponseNormalizer.js
+++ b/packages/relay-runtime/store/RelayResponseNormalizer.js
@@ -402,13 +402,15 @@ class RelayResponseNormalizer {
         isDeferred,
       );
     }
-    if (isDeferred === false) {
-      // If defer is disabled there will be no additional response chunk:
-      // normalize the data already present.
-      this._traverseSelections(defer, record, data);
-    } else {
-      // Otherwise data *for this selection* should not be present: enqueue
-      // metadata to process the subsequent response chunk.
+    // Always normalize inline data. Per the GraphQL DeferStream RFC, @defer
+    // is a hint — servers MAY choose not to defer, returning data inline in
+    // the initial payload. When the server did defer (data absent), traversal
+    // encounters undefined fields and skips them (no writes). When the real
+    // incremental chunk arrives, _processDeferResponse overwrites normally.
+    // See: https://github.com/facebook/relay/issues/3904
+    this._traverseSelections(defer, record, data);
+    if (isDeferred !== false) {
+      // Enqueue metadata to process the subsequent response chunk (if any).
       this._incrementalPlaceholders.push({
         actorIdentifier: this._actorIdentifier,
         data,


richardtagger:
Closing — the skip-traversal in `_normalizeDefer` is intentional design (see #5188 close comment for full analysis). Relay assumes the server always honors `@defer`, which holds at Meta. The selective defer case isn't supported by design.

We'll maintain this as a pnpm patch downstream.

PR #5210, opened 2026-03-19
remove deprecated promise-polyfill from tests

jantimon:
The test suite previously used `injectPromisePolyfill__DEPRECATED` to replace `global.Promise` with `promise-polyfill` (a userland implementation). This polyfill scheduled `.then()` callbacks via `setImmediate` rather than native microtasks, which conveniently meant `jest.runAllImmediates()` would flush promise chains automatically in `act()`

However, this approach has significant downsides:

1. The polyfill's scheduling semantics differ from real promises. Tests could pass under the polyfill but mask bugs that surface with real Promise implementations

2. The polyfill (`promise-polyfill@6.1.0`) is a stale dependency that all modern environments no longer need, and its presence makes the test infrastructure harder to reason about

3. The old approach silently coupled Relay's `resolveImmediate` (which uses `Promise.resolve().then(...)`) to Jest's fake/mock timer queue 
 
This commit replaces the polyfill with a lightweight `flushMicrotasks()` helper that uses the real (not mocked) `setImmediate` from Node's `timers` module to drain the microtask queue on demand

This will make it easier to add ESM builds for #4548 as it allows to remove outdated polyfills also in the compile flow


packages/react-relay/__tests__/LiveResolvers-test.js
```diff
@@ -848,10 +847,10 @@ describe('Live Resolver with Suspense and Missing Data', () => {
     // Now, the whole live field became undefined, as some of
     // the data in the live field resolver fragment is missing
     expect(renderer.toJSON()).toEqual('Alice: undefined');
-    TestRenderer.act(() => {
+    await TestRenderer.act(async () => {
       GLOBAL_STORE.dispatch({type: 'INCREMENT'});
+      await flushMicrotasks();
```
tyao1: I have an impression that `TestRenderer.act` would flush everything. Is `flushMicrotasks` here necessary?

jantimon: Good catch!

You're right I could strip a lot of flushMicrotasks() and everything is still green without new act warnings
                                                                                   
Early on I hit a few tests that passed alone but broke when running the suite and adding flushMicrotask helped fixing it but looks like it's not needed in many cases

I kept the explicitly added `await TestRenderer.act(() => flushMicrotasks());` cases as I think that's easier to understand than `await TestRenderer.act(() => {})` but I can also remove them there too if you prefer

jantimon:
@tyao1 I addressed your feedback last month - did you already had time to take a look at the changes again?

PR #5215, opened 2026-03-25
Fix incorrect note block in client-schema-extensions.mdx

TJ09:
[afaict](https://docusaurus.io/docs/markdown-features/admonitions) note blocks need to end with `:::` not `:::note`


captbaritone:
Thanks!

PR #5242, opened 2026-04-14
Regenerate test-project artifacts after emit_nogrep_annotation flag

jantimon:
it looks like commit 4e2fa215c by @captbaritone added the emit_nogrep_annotation feature flag which defaults to disabled, meaning @nogrep is no longer emitted in generated files

now the main branch pipeline fails on macOS as the relay compiler is generating files without @nogrep annotations, but the checked-in files have them

### DIFF
--- a/compiler/test-project/src/__generated__/AppQuery.graphql.js
+++ b/compiler/test-project/src/__generated__/AppQuery.graphql.js
@@ -4,9 +4,8 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  *
- * @generated SignedSource<<f1414a68f46c710d150c72ac5f9b46ac>>
+ * @generated SignedSource<<5213156639f5e33acfdd79872ea56f77>>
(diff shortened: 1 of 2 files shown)


jantimon:
@tyao1 can we please merge this PR so the pipelines are green again **on main**?

captbaritone:
Thanks for the PR. I think this is a duplicate of https://github.com/facebook/relay/pull/5245 which has already landed.

jantimon:
@captbaritone are you sure? main is still broken:

shot-tcVnPyTS@2x

captbaritone:
Yes, main is still broken, well, broken again. But I think this is a different error? In fact, I think there may be two or more problems with main right now.

captbaritone:
You can try re-basing this PR to see if these exact changes have already landed

jantimon:
@captbaritone you were right 😄 rebasing required additional fixes so I created a new pr: #5251 - it seems to fix all open pipeline issues

PR #5243, opened 2026-04-14
Update Film import path to include file extension

sharkdevs:
As is, the file is not found. adding .tsx to make that path complete


evanyeung:
Thanks for the fix!

Issue #4880, opened 2025-01-15
`fetchQuery` not supporting `store-and-network` and `store-only` fetch policies


skrabe:
Hi everyone,

After getting an error in my app and after looking at the [`fetchQuery` function](https://github.com/facebook/relay/blob/be8f5f77e113bd641d3514b913a55182341ff984/packages/relay-runtime/query/fetchQuery.js#L115), I found out `store-and-network` and `store-only` fetch policies are not supported. Is there any detailed explanation on why?

Also, it would be nice to have it covered in the docs if it stays unsupported.

captbaritone:
Sounds like this might be a bug in the community typescript types. In Flow we only allow `'store-or-network' | 'network-only'`. The idea is that `fetchQuery` is explicitly about making network requests. https://github.com/facebook/relay/blob/main/packages/relay-runtime/util/RelayRuntimeTypes.js#L72

If someone is up to it, it would be great if we could get:

1. An update to the TypeScript types
2. An update to the docs here: https://github.com/facebook/relay/blob/4130d410ae1481b99ece8798d85d79e70557332e/website/docs/api-reference/relay-runtime/fetch-query.md#L4 documenting the fetchPolicy options.

PR #5244, opened 2026-04-14
Document fetchPolicy option for fetchQuery

skrabe:
Follow-up to #4880.

The `fetchPolicy` option isn't mentioned in the fetchQuery docs, which is what caused the confusion in that issue. This adds it to the Arguments section with the two supported values (`network-only`, `store-or-network`) and an explicit note that `store-and-network` and `store-only` aren't supported, pointing people at `useLazyLoadQuery` instead.

The TypeScript types on DefinitelyTyped already match the Flow type (`FetchQueryFetchPolicy = "store-or-network" | "network-only"`), so nothing was needed on that side.


evanyeung:
Thanks for documenting this!

Issue #5258, opened 2026-04-22
MockPayloadGenerator produces nested arrays when a concrete type's InlineFragment contains the same plural LinkedField as an abstract InlineFragment at the same level


Sakshi-Saraswat-1:
MockPayloadGenerator.generate() produces invalid nested arrays ([[e1,e2,e3], [e1,e2,e3], [e1,e2,e3]] instead of [e1, e2, e3]) when the normalization AST has two InlineFragment selections at the same level — one concrete, one abstract — that both contain the same plural LinkedField (e.g., edges).

This happens when a GraphQL field returns a concrete type that implements an interface, and a fragment on that interface is spread inside the query. The Relay compiler correctly generates both InlineFragments in the normalization AST, and RelayResponseNormalizer handles them correctly at runtime. However, MockPayloadGenerator processes both InlineFragments sequentially and corrupts the data on the second pass.

Relay version

relay-test-utils (tested on the version bundled with our project — behavior traced in RelayMockPayloadGenerator.js)

Minimal Reproduction

Schema:

interface AbstractConnection {
edges: [AbstractEdge]
}

interface AbstractEdge {
cursor: String
}

type ConcreteConnection implements AbstractConnection {
edges: [ConcreteEdge]
}

type ConcreteEdge implements AbstractEdge {
cursor: String
node: Item
}

type Item {
id: ID!
}

type Query {
items: ConcreteConnection
}

Query:

query TestQuery {
items {
edges { cursor }
...MyFragment
}
}

fragment MyFragment on AbstractConnection {
edges { cursor }
}

What the Relay compiler generates (normalization AST):

{
concreteType: "ConcreteConnection",
name: "items",
selections: [
// Selection 1: Concrete edges
{
concreteType: "ConcreteEdge",
name: "edges",
plural: true,
selections: [{ name: "cursor" }, { name: "__typename" }]
},
// Selection 2: Abstract InlineFragment (from ...MyFragment)
{
kind: "InlineFragment",
type: "AbstractConnection",
abstractKey: "__isAbstractConnection",
selections: [{
name: "edges",
plural: true,
concreteType: null,  // No concrete type on abstract path
selections: [{ name: "cursor" }, { name: "__typename" }]
}]
}
]
}

Test code:

const payload = MockPayloadGenerator.generate(operation, {
ConcreteConnection: () => ({
edges: [
{ __typename: 'ConcreteEdge', cursor: 'a' },
[...]

PR #5273, opened 2026-05-06
Add regression test for MockPayloadGenerator nested-array bug with abstract InlineFragment (#5258)

markselby9:
## Summary

Adds a regression test for #5258: when a query selects a plural linked field on both a concrete type AND an abstract InlineFragment at the same selection level, `MockPayloadGenerator.generate` produces a nested-array payload (`[[edge, edge, edge], [edge, edge, edge], [edge, edge, edge]]`) instead of the expected flat `[edge, edge, edge]` shape, with `__typename`/`cursor` properties also bolted onto each Array instance.

The test asserts current (buggy) behavior so it flips when the underlying bug is fixed — same pattern as #5268 and #5269.

## Notes

- The test hand-constructs the `ConcreteRequest` AST rather than using a `graphql` template literal. The OSS test schema doesn't naturally model the concrete-vs-abstract divergent-return-type pattern that triggers this bug (every plural field returning an interface returns the same `[Actor]` from both interface and implementer, so the compiler dedupes the selections). I tried a schema extension adding `interface AbstractConnection { edges: [AbstractEdge] }` + `type ConcreteConnection implements AbstractConnection { edges: [ConcreteEdge] }`, but the relay compiler folds the abstract InlineFragment when the parent field's static type pins the abstract narrowing. Happy to revisit if you have a preferred path; the test header documents this attempt so future contributors don't re-tread it.
- The test only covers the array-shape symptom; the issue's Root Cause Analysis #3 (`__typename` override on the abstract path) is a separate adjacent symptom and is explicitly out of scope here.

## Test plan

- [x] `yarn jest RelayMockPayloadGenerator` — full file passes (69 tests)
- [x] `yarn lint` on the test file — clean
- [x] Verified the assertion locks down both the nested-array shape and the bolted-on `__typename`/`cursor` properties via deep `toEqual`

Refs #5258


markselby9:
Closing for now — I want to verify the compiler actually emits the AST shape this test asserts before adding to the review queue. Following up on #5258 to ask for a compiler-generated artifact, will revisit once that's in hand.

PR #5277, opened 2026-05-09
feat: Warn on deprecated enum values used as arguments

jungwoo3490:
## Summary

GraphQL allows `@deprecated` on enum values, but Relay wasn't warning when a deprecated enum value was used as a field or directive argument. This PR adds that validation alongside the existing deprecated field/argument checks.

```graphql
enum Status {
  ACTIVE
  LEGACY @deprecated(reason: "Use ACTIVE instead.")
}

# Before: no warning
# After: The enum value `LEGACY` of type `Status` is deprecated. Deprecation reason: "Use ACTIVE instead."
fragment Foo on Query {
  user(status: LEGACY)
}

```

The warning message wording follows the existing patterns in the codebase. Open to suggestions if there's a better way to phrase it!

> [!NOTE]
> Variable values (`$status`) and deeply nested enum values inside constant objects/arrays are not yet covered. Left a TODO comment in the code.

### DIFF
--- a/compiler/crates/graphql-ir/src/errors.rs
+++ b/compiler/crates/graphql-ir/src/errors.rs
@@ -498,6 +498,17 @@ pub enum ValidationMessage {
         deprecation_reason: Option<StringKey>,
     },
 
+    #[error("The enum value `{enum_value}` of type `{enum_name}` is deprecated.{}",
+    match deprecation_reason {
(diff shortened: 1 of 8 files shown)


evanyeung:
Looks good! Thanks for fixing this

PR #5281, opened 2026-05-13
Fix: `@catch` on fragment root breaks `usePaginationFragment` and `usePrefetchableForwardPaginationFragment`

### LINKED ISSUE #5070: @catch doesn't work on fragment root of useRefetchableFragment nor usePaginationFragment
Hey

Unfortunately there are issues with `@catch` in combination with `useRefetchableFragment` and `usePaginationFragment`.
On a surface level, there is no issue, until `refetch` or `loadNext` is called. When these functions are called, everything goes boom.

Given a component like this here, `refetch` only works when the `id` is manually provided. And the `loadNext` function doesn't work at all.

```tsx
import { graphql, usePaginationFragment } from "@segments/relay";
import { type FunctionComponent, startTransition } from "react";
import type { brandForumContextCommunityDiscussionsList$key } from "./__generated__/brandForumContextCommunityDiscussionsList.graphql.ts";
import type {
  brandForumContextCommunityDiscussionsListRefetchQuery,
  ProductListSortOrder,
} from "./__generated__/brandForumContextCommunityDiscussionsListRefetchQuery.graphql.ts";
interface IBrandForumContextCommunityDiscussionsListProps {
  brandKey: brandForumContextCommunityDiscussionsList$key;
}

export const BrandForumContextCommunityDiscussionsList: FunctionComponent<
  IBrandForumContextCommunityDiscussionsListProps
> = ({ brandKey }) => {
  const {
    data: brand,
    loadNext,
    refetch,
  } = usePaginationFragment<
    brandForumContextCommunityDiscussionsListRefetchQuery,
    brandForumContextCommunityDiscussionsList$key
  >(
    graphql`
      fragment brandForumContextCommunityDiscussionsList on Brand
      @argumentDefinitions(
        first: { type: "Int", defaultValue: 5 }
        after: { type: "String" }
        order: { type: "ProductListSortOrder", defaultValue: AVAILABILITY }
      )
      @refetchable(
        queryName: "brandForumContextCommunityDiscussionsListRefetchQuery"
      )
      @catch {
        products(first: $first, after: $after, order: $order)
          @connection(key: "brandForumContextCommunityDiscussionsList_products")
          @required(action: THROW) {
          edges @required(action: THROW) {
            node {
              id
              name
            }
          }
        }
      }
    `,
    brandKey,
  );

  if (!brand.ok) {
    throw new Error("Simple here to demo purpose");
  }

  const onOrderToggle = (order: ProductListSortOrder) => {
    startTransition(() => {
      refetch({
        // 👇 When the id is not manually provided, this will result in a runtime error
        id: brand.value.id,
        order,
      });
    });
  };

  return (
    <>
      <button onClick={() => onOrderToggle("HIGHEST_PRICE")}>
        Change sort
      </button>
      <ul>
        {brand.value.products.edges.map((edge) => (
          <li key={edge.node.id}>{edge.node.name}</li>
        ))}
      </ul>
      {/* 👇 When loadNext is called, this will execute the refetch query with id: null */}
      <button onClick={() => loadNext(2)}>Load more</button>
    </>
  );
};

```

The expectation for this would be that it just works as expected or that the relay compiler tells us that using @catch is not allowed here.

christiansany:
Fixes: #5070

## Problem

When `@catch` is applied to the **root** of a fragment used with `usePaginationFragment`, `hasNext` always returned `false` even when the response included `pageInfo.hasNextPage: true`. `@catch` on a child field worked correctly — only the root-level case was broken.

## Root cause

When `@catch` is on the fragment root, the compiler sets `metadata.catchTo: "RESULT"` on the ReaderFragment. At read time, `RelayReader._catchErrors` wraps the entire fragment data in a `Result` envelope:

```js
{ok: true, value: {friends: ..., id: ...}}
// or
{ok: false, errors: [...]}
```

The pagination/refetch hook internals were reading fields directly off `fragmentData` and never unwrapping this envelope:

- `getConnectionState` did `getValueAtPath(fragmentData, ["friends"])` → `undefined` → returned `{cursor: null, hasMore: false}` → `hasNext === false`.
- `useLoadMoreFunction` and `useRefetchableFragmentInternal` did `fragmentData[identifierField]` → `undefined` → broken refetch identifier.

## Fix

Added a tiny helper `getFragmentInternalData(fragmentNode, fragmentData)` that unwraps the `Result` envelope when `catchTo === 'RESULT'` (no-op otherwise). Applied it at every internal call site that reads from `fragmentData`. The user-facing `data` returned from `usePaginationFragment` remains wrapped, matching the generated `Result<T, unknown>` type.

### DIFF
--- a/packages/react-relay/relay-hooks/__tests__/__generated__/usePaginationFragmentCatchTestRootCatchFragment.graphql.js
+++ b/packages/react-relay/relay-hooks/__tests__/__generated__/usePaginationFragmentCatchTestRootCatchFragment.graphql.js
@@ -1,0 +1,181 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @oncall relay
+ *
+ * @generated SignedSource<<32f1e5483f7f2865ffecba7b9b260c7f>>
+ * @flow
+ * @lightSyntaxTransform
+ */
+
+/* eslint-disable */
+
+'use strict';
+
+/*::
+import type { ReaderFragment, RefetchableFragment } from 'relay-runtime';
+import type { FragmentType, Result } from "relay-runtime";
+declare export opaque type usePaginationFragmentCatchTestRootCatchFragment$fragmentType: FragmentType;
+type usePaginationFragmentCatchTestRootCatchRefetchableFragmentQuery$variables = any;
+export type usePaginationFragmentCatchTestRootCatchFragment$data = Result<{|
+  +friends: ?{|
+    +edges: ?ReadonlyArray<?{|
+      +node: ?{|
+        +__typename: "User",
+      |},
+    |}>,
+  |},
+  +id: string,
+  +$fragmentType: usePaginationFragmentCatchTestRootCatchFragment$fragmentType,
+|}, unknown>;
+export type usePaginationFragmentCatchTestRootCatchFragment$key = {
+  +$data?: usePaginationFragmentCatchTestRootCatchFragment$data,
+  +$fragmentSpreads: usePaginationFragmentCatchTestRootCatchFragment$fragmentType,
+  ...
+};
+*/
+
+var node/*: ReaderFragment*/ = (function(){
+var v0 = [
+  "friends"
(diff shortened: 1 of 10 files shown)


captbaritone:
Thank you!

PR #5287, opened 2026-05-19
fix(build): ship TypeScript definition files in gulp dist build

PHILLIPS71:
- Adds a `typeDefs` gulp task that copies all `**/*.d.ts` files from each package source directory into `dist/<package>/`, preserving subdirectory structure so relative imports within `.d.ts` files resolve correctly.
- Wires `typeDefs` into the `exportsFiles` pipeline, running in parallel with the existing `flowDefs` task.
- Affects all packages in the `builds` array (`relay-runtime`, `react-relay`, `relay-test-utils`, etc.), any package with `.d.ts` files now has them shipped.

## Root cause
`relay-runtime/package.json` declares `"types": "index.d.ts"`, but the gulpfile had no task to copy `.d.ts` files, so the type entry point was absent from every published dist.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -139,6 +139,22 @@ const flowDefs = gulp.parallel(
   ),
 );
 
+const typeDefs = gulp.parallel(
+  ...builds.map(
+    build =>
+      function typeDefsTask() {
+        return gulp
+          .src(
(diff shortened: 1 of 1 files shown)


captbaritone:
Thank you! Sorry we missed this. We'll work on a point release.

PR #5289, opened 2026-05-21
fix: add extensions field to PayloadError per GraphQL spec

PHILLIPS71:
The GraphQL spec (https://spec.graphql.org/September2025/#sec-Errors) states that errors may include an extensions entry whose value must be a map with no additional restrictions on its contents. PayloadError was missing this field.

This also propagates extensions through to `TRelayFieldError` so the field is preserved during normalization and available to consumers. As a side effect, stale inline type annotations in `RelayErrorTrie.js` that manually duplicated the error shape have been replaced with `TRelayFieldError` directly, and a mutation of a ReadonlyArray path via unshift has been replaced with an immutable spread.

### DIFF
--- a/packages/relay-runtime/network/RelayNetworkTypes.d.ts
+++ b/packages/relay-runtime/network/RelayNetworkTypes.d.ts
@@ -31,6 +31,7 @@ export interface PayloadError {
         }>
         | undefined;
     path?: Array<string | number>;
+    extensions?: PayloadExtensions | undefined;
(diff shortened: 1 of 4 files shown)


evanyeung:
Nice catch! This looks correct as per the spec

PHILLIPS71:
> Nice catch! This looks correct as per the spec

Awesome! Do you need anything else from me to get this over the line and merged?

evanyeung:
Sorry, although this is correct, we'll need to fix some things up internally and mark it as a minor breaking change since anything that snapshots the results, iterates the payload, or expects the type to be exactly the previous object will have to handle the extensions field.

PR #5338, opened 2026-06-25
Fix and add missing TypeScript declarations for react-relay & ..

christiansany:
First off — thank you! We've been building on Relay for nearly three years and it's been fantastic. We're especially happy that the TypeScript types are now maintained directly in this repo: it's what makes contributions like this one straightforward, and we're glad to be able to give a little back to the core team and the community. 🙏

`react-relay` / `relay-runtime` are written in Flow but ship hand-written `.d.ts` files for TypeScript consumers. Since nothing generates them from the Flow source, they can drift. This PR fixes three such drifts (declarations only — no Flow/runtime changes).

The motivating one for us was #3: a real type bug, not just a gap. Mutations using `@catch` at the root level are correctly typed in Flow but were **falsely rejected** by the TypeScript types, which we've had to work around by overriding Relay's types in our own codebase. Fixing it at the source removes that workaround for everyone. The two missing declarations (#1) we hit along the way.

## 1. Add two hooks that had no TypeScript types

`useMutationAction_EXPERIMENTAL` and `usePrefetchableForwardPaginationFragment` are exported from `index.js` but had no `.d.ts`, making them invisible to TypeScript. Added declarations for both and re-exported them from `hooks.d.ts`.

Both are constrained on `OperationType` (matching the Flow sources, which don't constrain the response, and consistent with the other modern hooks).

## 2. Add the missing `getEnvironmentForActor` prop to `RelayEnvironmentProvider`

The Flow component accepts this optional prop but the `.d.ts` omitted it, so passing it was a type error.

## 3. Fix `MutationParameters` so `@catch` mutations type-check

The Flow type uses `{...}` ("any object") for `response`/`variables`/`rawResponse`, but the `.d.ts` translated it as `Record<string, unknown>`, which requires an index signature. Responses that lack one — notably the `Result<…>` unions from `@catch` mutations, and interface-typed responses — are valid in Flow but were rejected by TypeScript, forcing consumers to override Relay's types locally.

Fixed by translating `{...}` faithfully as `object`:

```ts
export interface MutationParameters {
    readonly response: object;
    readonly variables: Variables;
    readonly rawResponse?: object | undefined;
}
```

This only widens a constraint, so it's backward compatible — nothing that compiled before breaks. Fixing it at the source corrects the whole mutation surface at once (`commitMutation`, `MutationConfig`, `useMutation`, `applyOptimisticMutation`, …), so `useMutation` itself needs no special-casing.

## Testing
 
Verified via linking the `dist` files directly in our monorepo and removing our local shims / ts overwrites that we currently have to have.


captbaritone:
Thank you!

PR #5341, opened 2026-06-25
Fix nested `@catch`: stop inner-handled errors from propagating

### LINKED ISSUE #5339: Nested `@catch`: an error handled by an inner `@catch` incorrectly propagates to the outer `@catch`
## Summary

When two `@catch` directives are nested (an inner field has `@catch`, and one of its ancestors also has `@catch`), a server/field error inside the inner `@catch`'s scope is **handled twice**. The inner `@catch` handles it correctly, but the same (already-handled) error also drives the outer `@catch`'s outcome — so the error effectively "lands" in the outer boundary as well.

The result is that the outer `@catch` reports a failure/null even though, from its perspective, the inner field resolved successfully (the inner `@catch` already turned the error into a value or a null).

Affects both `@catch(to: RESULT)` and `@catch(to: NULL)`.

## Reproduction

**Runnable reproduction:** https://github.com/christiansany/relay-playground/pull/1

It contains a minimal Relay app with nested `@catch` (outer `@catch` on a fragment, inner `@catch` on a nested field, error forced via `@simulateError`, no `@throwOnFieldError`), for both `to: RESULT` and `to: NULL`:

1. `yarn dev`
2. Open `http://localhost:5173/nested-catch`
3. Both sections render **"OUTER caught … BUG"** — the outer `@catch` fires instead of the inner one.
4. Removing the outer `@catch` and re-running `yarn relay` makes the inner `@catch` behave correctly — confirming the inner boundary itself works, and the problem is the error leaking to the outer boundary.

### Illustrative shape

Conceptually, given a query/fragment shaped like this and an error on the inner field:

```graphql
me @catch(to: RESULT) {
  lastName @catch(to: RESULT)
}
```

**Expected** — the inner `@catch` absorbs the error; the outer `me` resolves successfully:

```js
{
  me: {
    ok: true,
    value: { lastName: { ok: false, errors: [{ path: ['me', 'lastName'] }] } },
  },
}
```

**Actual** — the (already-handled) error also bubbles to the outer `@catch`, so `me` is reported as an error:

```js
{
  me: { ok: false, errors: [{ path: ['me', 'lastName'] }] },
}
```

The `to: NULL` variant behaves analogously: instead of only the inner field being nulled (`{ me: { lastName: null } }`), the entire outer object is nulled (`{ me: null }`).

## Expected behavior

We'd expect **one** of the following — currently neither holds, and the misbehavior is silent:

1. **Nested `@catch` works as expected.** A field error is handled by the *innermost* `@catch` that encloses it and does not affect any outer `@catch` boundary. An outer `@catch` should only react to errors that were *not* already handled deeper in the tree.
2. **The compiler rejects nested `@catch`.** If nested `@catch` is not intended to be supported, the relay compiler should produce a clear build-time error when a `@catch` is nested within another `@catch`, instead of emitting code that silently mis-handles errors at runtime.

Either is acceptable. What's surprising today is that nested `@catch` compiles fine but produces incorrect data.

## Likely root cause

In `RelayReader` (`_asResult`, and the `NULL` handling), the decision of whether a `@catch` boundary saw an error is based on the presence of *any* entries in `_fieldErrors`, without excluding errors already marked `handled` by an inner `@catch`. Handled errors flow through `_fieldErrors` for logging purposes, but they should not influence an outer `@catch`'s `ok`/`null` outcome. Filtering `_fieldErrors` to only the *unhandled* ones at each boundary resolves both cases above.

We have a reduced `RelayReader` test case and a candidate fix, and would be happy to open a PR.

## Environment

- `relay-runtime` / `react-relay` `20.1.1`
- Affects both `@catch(to: RESULT)` and `@catch(to: NULL)` when nested.

christiansany:
Fixes #5339

## The bug

When `@catch` directives are nested, a field error handled by the **inner** `@catch` is also counted by the **outer** `@catch`, so the outer boundary fires even though, from its perspective, the inner field resolved successfully.

- `@catch(to: RESULT)`: the outer field becomes `{ ok: false }` instead of `{ ok: true, value: … }`.
- `@catch(to: NULL)`: the whole outer object is nulled instead of just the inner field.

It compiles fine and fails silently at runtime. Full write-up and a runnable reproduction are in the linked issue (repro: https://github.com/christiansany/relay-playground/pull/1).

## The fix

In `RelayReader`, whether a `@catch` boundary "saw" an error was based on `_fieldErrors` being non-empty — but that list also contains errors already marked `handled` by an inner `@catch` (they flow through only for logging). The fix excludes already-handled errors when deciding an outer boundary's outcome:

- `_asResult` now builds its result from the **unhandled** errors only (`{ ok: true }` when none remain).
- The `NULL` path nulls the field only when there is an **unhandled** error (`_fieldErrors.some(e => !e.handled)`).

An inner `@catch` consumes its error and the outer boundary no longer reacts to it. Errors that are genuinely unhandled at a given level still propagate exactly as before.

## Tests

Added regression tests in `RelayReader-CatchFields-test.js` covering both targets with a server error on the inner field:

- nested `@catch(to: RESULT)` → outer `me` is `{ ok: true, value: { lastName: { ok: false, … } } }` (inner catches, outer succeeds).
- nested `@catch(to: NULL)` → `{ me: { lastName: null } }` (only the inner field nulled).

In both cases the error is still surfaced once in `fieldErrors` with `handled: true`.

## Testing

- New tests pass; existing `RelayReader-CatchFields` tests remain green.
- Runtime-only change in `RelayReader.js` (10 lines); no API or compiler changes.

### DIFF
--- a/packages/relay-runtime/store/RelayReader.js
+++ b/packages/relay-runtime/store/RelayReader.js
@@ -441,7 +441,10 @@ class RelayReader {
         value = this._asResult(_value);
         break;
       case 'NULL':
-        if (this._fieldErrors != null && this._fieldErrors.length > 0) {
+        if (
+          this._fieldErrors != null &&
+          this._fieldErrors.some(e => !e.handled)
+        ) {
           value = null;
         }
         break;
@@ -479,12 +482,16 @@ class RelayReader {
    * responsibility to ensure that errors are marked as handled.
    */
   _asResult<T>(value: T): Result<T, unknown> {
-    if (this._fieldErrors == null || this._fieldErrors.length === 0) {
+    // Errors already handled by an inner @catch must not influence this
+    // @catch's outcome — they have been consumed by their own boundary and
+    // are only flowing through `_fieldErrors` for logging.
+    const unhandled = this._fieldErrors?.filter(e => !e.handled);
+    if (unhandled == null || unhandled.length === 0) {
       return {ok: true, value};
     }
 
     // TODO: Should we be hiding log level events here?
-    const errors = this._fieldErrors
+    const errors = unhandled
       .map(error => {
         switch (error.kind) {
           case 'relay_field_payload.error':
--- a/packages/relay-runtime/store/__tests__/RelayReader-CatchFields-test.js
+++ b/packages/relay-runtime/store/__tests__/RelayReader-CatchFields-test.js
@@ -904,4 +904,114 @@ describe('RelayReader @catch', () => {
       },
     ]);
   });
+
+  it('nested @catch(to: RESULT): server error on inner field should be handled by the inner @catch, not bubble to the outer @catch', () => {
+    const source = RelayRecordSource.create({
+      'client:root': {
+        __id: 'client:root',
+        __typename: '__Root',
+        me: {__ref: '1'},
+      },
+      '1': {
+        __id: '1',
+        id: '1',
+        __typename: 'User',
+        lastName: null,
+        __errors: {
+          lastName: [
+            {
+              message: 'There was an error!',
+              path: ['me', 'lastName'],
+            },
+          ],
+        },
+      },
+    });
+
+    const FooQuery = graphql`
+      query RelayReaderCatchFieldsTestNestedResultQuery {
+        me @catch(to: RESULT) {
+          lastName @catch(to: RESULT)
+        }
+      }
+    `;
+    const operation = createOperationDescriptor(FooQuery, {id: '1'});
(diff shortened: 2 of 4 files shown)


captbaritone:
Thanks for this! This looks correct to me.

captbaritone:
Quick follow-up here. It looks like this is causing some internal integration tests which were relying on the previous broken behavior to fail. It's going to take me a little bit longer to land this change, and I'll probably land it behind a feature flag since it does change application behavior.

christiansany:
Very cool :)
Let me know, if I can help you somehow :)

PR #5342, opened 2026-06-25
Scope provided-variable memoization to the network instead of the process

striedinger:
## Summary

`withProvidedVariables` pins each provider's first return value in a
module-level cache keyed by the `get` function, to enforce purity. That's fine
in a browser tab, but wrong on a server, where one process serves many
requests/viewers. A provider that reads per-request state (auth/session,
locale, experiment bucket) legitimately returns a different value per request,
and today that causes two problems:

1. The first request's value is pinned and reused for every later request, so
   subsequent requests resolve with stale variables.
2. Dev builds emit a spurious pure-function warning, since the value differs
   across requests:

```text
Relay: Expected function `get` for provider `…` to be a pure function, but got conflicting return values `…` and `…`
```

## Approach

Replace the process-wide cache with a module-private `WeakMap` of per-network
caches. `RelayNetwork.create` passes the network it just built as the cache
owner, so provider values are memoized for the life of that network rather than
the life of the process. Purity is still enforced within a network, which is the
scope where it's meaningful. Where `WeakMap` is unavailable, values stay process
wide rather than retaining every network forever.

The cache is keyed on the network rather than the environment because
`RelayNetwork.create().execute` is the only caller of `withProvidedVariables`,
and it has no reference to an environment — threading one in would mean changing
the public `INetwork.execute` signature. Since an environment owns a network, a
server that builds an environment per request gets the same isolation.

## Backwards compatibility

The new `network` parameter is optional, and omitting it preserves today's
process-wide behavior. The publicly exported `withProvidedVariables` and any
custom `INetwork` implementation keep working unchanged. A custom network can
opt in by passing itself as the third argument.

## Test plan

- New `RelayNetwork-test.js` covers memoization within a single network,
  isolation across two networks, and the purity warning still firing within a
  network.
- Full suite green on top of current `main`: `yarn jest` (244 suites, 3740
  tests), `yarn typecheck`, `yarn lint`, `yarn prettier-check`,
  `yarn test-dependencies`.

### DIFF
--- a/packages/relay-runtime/network/RelayNetwork.js
+++ b/packages/relay-runtime/network/RelayNetwork.js
@@ -48,6 +48,7 @@ function create(
     const operationVariables = withProvidedVariables(
       variables,
       request.providedVariables,
+      network,
     );
     if (request.operationKind === 'subscription') {
       invariant(
@@ -83,7 +84,10 @@ function create(
     );
   }
 
-  return {execute};
+  // `execute` memoizes provided variable values per network, so it needs the
+  // network it belongs to. It only runs once this is assigned.
+  const network: INetwork = {execute};
+  return network;
 }
 
 module.exports = {create};
--- a/packages/relay-runtime/network/__tests__/RelayNetwork-test.js
+++ b/packages/relay-runtime/network/__tests__/RelayNetwork-test.js
@@ -1,0 +1,105 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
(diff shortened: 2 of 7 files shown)


captbaritone:
What if instead we maintain a weakmap of per-environment caches and use that as the `debugCache`? Each request should produce its own environment, so it should be fine to have the results cached per environment.

striedinger:
@captbaritone updated, let me know if this is better or if there is a specific direction you'd prefer I take please

striedinger:
@captbaritone gentle reminder, sorry to be annoying. let me know if there is anyone else I can ask for a review

PR #5358, opened 2026-07-15
Fix RelayModernRecord TypeScript declarations

HorusGoul:
Hey! Found some errors with the RelayModernRecord.d.ts file and this change pretty much refactors the entire file to be more aligned with the actual implementation.

Also exposes all functions marked with `@public`.


captbaritone:
cc @leealison

Issue #5357, opened 2026-07-14
@defer'd fragment returns partial data instead of suspending when the parent record is populated by a sibling non-deferred query


jonreading81:
## Version

`react-relay` / `relay-runtime` **21.0.1**, with React 19 streaming SSR + `@defer`.

## Problem

Under React 19 streaming SSR of a query with `@defer`, `useFragment` reads partial data instead of suspending, while the underlying `OperationExecutor` is still emitting incremental payloads. Consumers that expect the fragment's fields to be either fully present (fragment resolved) or absent (fragment suspended) crash on undefined subtree fields.

Fix proposed in facebook/relay#5360.

## Root cause

`fetchQueryDeduped` caches a per-identifier subject and tears it down when subscriber count hits 0. Under React 19 streaming SSR, that cleanup fires between Suspense render passes for a query still emitting `@defer` chunks — the subject is torn down while the executor is alive. From that moment on:

1. `getPromiseForActiveRequest` returns `null` (the cache entry has been deleted).
2. `RelayOperationTracker.getPendingOperationsAffectingOwner` returns `null` (no other operation writes to the owner's records — this is the specifically-broken case when a sibling *non*-deferred query has already populated the shared record).
3. `getPendingOperationsForFragment` returns `null`.
4. `useFragment` reads the partial snapshot and returns non-suspended data.

The condition surfaces most obviously when a sibling non-deferred query has already populated the shared record — `viewer.me` exists, but `viewer.me.commentsReceived` (owned by the deferred fragment) does not — because that's the state where partial-data reads look "consistent enough" to render but crash on a specific field. Structurally, the underlying gap (request-cache drained between Suspense passes) affects any streaming operation whose executor outlives its cache subject.

## Minimal shape

Two queries preloaded on the same page.

**Query A** (non-deferred, populates `viewer.me` in the store):

```graphql
query SiblingQuery {
    viewer {
        me {
            username
            hasProFeatures
        }
    }
}
```

**Query B** (defers a fragment that reads *other* fields on the same `viewer.me`):

```graphql
query MyDashboardQuery($itemCount: Int!) {
    viewer {
        ...LatestCommentsList_viewer @defer @arguments(itemCount: $itemCount)
    }
}

fragment LatestCommentsList_viewer on Viewer @argumentDefinitions(itemCount: {type: "Int!"}) {
    me {
        commentsReceived(first: $itemCount) {
            edges { ...CommentCardList_commentEdges }
        }
    }
}
```

[...]

PR #5360, opened 2026-07-16
Track in-flight operations by executor lifetime, not request-cache subscribers

jonreading81:
## Summary

Closes a fragment-to-owner correlation gap that surfaces on streaming responses when the request-cache subject is drained between Suspense render passes while the underlying `OperationExecutor` is still emitting incremental payloads. In that window, `getPendingOperationsForFragment` returns `null`, `useFragment` reads partial data instead of suspending, and any consumer that reads a non-null field on the missing subtree crashes the render.

Introduces a per-environment registry of operation completion promises tied to `OperationExecutor` lifetime, independent of the request-cache subject and its subscriber count. Behavior change is gated behind a new `RelayFeatureFlags.ENABLE_IN_FLIGHT_OPERATION_CORRELATION` (default off) for gradual rollout.

## Mechanism (before this PR)

`fetchQueryDeduped` caches a per-identifier subject and tears it down when subscriber count hits 0. Under React 19 streaming SSR, that cleanup fires between render passes for a query still emitting `@defer` chunks — subject torn down while the executor is alive. `getPromiseForActiveRequest` returns `null` (cache empty); `RelayOperationTracker.getPendingOperationsAffectingOwner` returns `null` (no other operation writes to the owner's records). `getPendingOperationsForFragment` returns `null`. `useFragment` reads the partial snapshot.

Discussion and trace: facebook/relay#5357.

## What this PR does

- **`RelayModernEnvironment`** — adds `_inFlightOperationCompletions: Map<string, {promise, resolve}>` and a public `getPromiseForInFlightOperation(id)` accessor.
- **`OperationExecutor`** — registers a completion promise on construction; resolves and removes it in `cancel()`.
- **`getPendingOperationsForFragment`** — after `getPromiseForActiveRequest` and `RelayOperationTracker` both come up null, consults the in-flight registry as a last-resort correlation. Guarded by the feature flag; the registry entry's presence IS the "in-flight" signal (populated on `OperationExecutor` construction, cleared atomically in `cancel()`). We deliberately do NOT additionally gate on `isRequestActive` — that reads `_operationExecutions`, which flips to `'inactive'` while `_state === 'loading_final'` (no pending module payloads) BEFORE the network stream completes and the completion map is cleared. Fragments reading during that window would otherwise fall through to partial-data reads.
- **`RelayFeatureFlags`** — adds `ENABLE_IN_FLIGHT_OPERATION_CORRELATION: boolean` (default `false`).
- **`MultiActorEnvironment`** / **`ActorSpecificEnvironment`** / **`IEnvironment`** / **`IMultiActorEnvironment`** — mirror the accessor for symmetry.
- **Tests** — three unit tests on `environment.getPromiseForInFlightOperation` covering never-started / in-flight / completed; two integration tests in `useFragment-WithOperationTrackerSuspense-test.js`: the first mirrors the existing suspense pattern (fragment reads missing data, executor alive, cache empty → suspends → resolves on `mock.complete`); the second directly asserts the gate skips `isRequestActive` — with the executor still in flight but `isRequestActive` mocked to `false`, the fragment must still suspend on the completion promise.

## Design notes

- Fallback placed in `getPendingOperationsForFragment`, not `getPromiseForActiveRequest`, so mutation-driven correlations that legitimately resolve through `RelayOperationTracker` are not short-circuited (an earlier attempt at the wrong level broke a FragmentResource plural-mutation test).
- Registry entries are populated on `OperationExecutor` construction and removed in `cancel()`, so lifecycle paths that already reach `cancel()` clean up naturally. No new leak surface; same invariant as the sibling `_operationExecutions` map.
- `ExecuteConfig` gains one field (`inFlightOperationCompletions`), threaded through by all `env.execute*` paths — same shape as the existing `operationExecutions` param.

## Test plan

- [x] `yarn jest packages/relay-runtime` — 2279/2279 (28 snapshots)
- [x] `yarn jest packages/react-relay` — 1274/1274 (20 skipped)
- [x] Focused: `useFragment-WithOperationTrackerSuspense-test.js` — 6/6 including the two new integration tests.
- [x] Validated against a real @defer-heavy production codebase with the same shape (Mixcloud, mid-stream @defer + Suspense render loops) — no regressions across the E2E suite with the flag enabled.


jonreading81:
The CLA has been signed — could the bot re-check when it gets a chance? Thanks.

jonreading81:

Issue #5366, opened 2026-07-20
Fragment reading GC'd records renders a partial snapshot with no recovery — crashes <Activity>-restored routes


ellemedit:
## Summary

When a fragment ref outlives its owner query's retain, Relay GC can (correctly,
by root reachability) collect records the fragment still references. The next
time that fragment reads — an `<Activity>` restore, a store write overlapping
the parent record, or a fresh mount with a ref held in React state — the read
is missing data with **no pending operation**, and `useFragmentInternal`
renders the **partial snapshot**. Fields the fragment explicitly fetched come
back `undefined`, and any consumer that trusts the schema's non-null types
crashes:

```
TypeError: Cannot read properties of undefined (reading '…')
```

There is no code path that attempts recovery: Suspense never engages, no
refetch is issued, and the error boundary is the first thing that notices.

This hits `<Activity>`-based routers especially hard (the exact configuration
`ENABLE_ACTIVITY_COMPATIBILITY` targets): a hidden route's subscriptions are
disposed, its records are collected while hidden, and restoring the route
crashes it. We see this in production on react-relay / relay-runtime 21.0.1
with React 19.2.

## Reproduction

https://github.com/ellemedit/relay-gc-partial-read-repro — `pnpm install && pnpm test`

- `main` (stock react-relay 21.0.1): 2 tests fail with the TypeError above.
- `with-fix` (patch applied): same tests pass — the missing read becomes one
  forced owner refetch + Suspense, and the UI recovers.

The repro does not touch store internals: GC runs for real, triggered by a
real `retain().dispose()`, with `gcReleaseBufferSize: 0` only making
deterministic what the default 10-slot release buffer does over a session.

The dangling-link shape is ordinary app behavior:

1. Two queries overlap partially — owner `me { id account { id label } }` and
   a longer-lived keep-alive `me { id }` (app shell, another route).
2. A component holds the owner's fragment ref in props/state (list rows, chat
   bubbles, `<Activity>` subtrees).
3. The owner's retain is released; GC collects the `Account` record while the
   `User` survives via the keep-alive query → `user.account` dangles.
4. GC leaves no trace (no notify, no epoch bump) — the fragment even renders
   its stale pre-GC snapshot on restore.
5. Any later store write overlapping `User` (mutation, revalidation) makes the
   fragment re-read → missing data, no pending operation → partial render →
   consumer crash.

## Expected behavior

A fragment that reads missing data with no in-flight request should attempt to
[...]

PR #5367, opened 2026-07-20
Recover missing-data fragment reads by refetching the owner once (ENABLE_MISSING_DATA_OWNER_REFETCH)

ellemedit:
## Summary

Fixes the crash described in https://github.com/facebook/relay/issues/5366: a fragment whose records were
garbage-collected while unobserved (hidden `<Activity>` route, fragment ref
held in React state after the owner query's retain lapsed) re-reads **missing
data with no pending operation**, and `useFragmentInternal_EXPERIMENTAL`
renders the partial snapshot — `undefined` for fields the fragment explicitly
fetched — crashing consumers that trust the schema types.

This PR adds an opt-in recovery path behind a new feature flag,
`ENABLE_MISSING_DATA_OWNER_REFETCH` (default **off**):

- When `isMissingData(state)` and `getPendingOperationsForFragment` finds no
  in-flight operation, refetch the fragment's **owner query** once and suspend
  on the request instead of rendering the partial snapshot.
- The refetch uses `{...owner.cacheConfig, force: true}` with fetch policy
  `'network-only'` and a **unique QueryResource cache breaker**, so a completed
  QueryResource entry or an app-level response cache cannot short-circuit it
  into re-reading the same collected records.
- **Query owners only.** A fragment owned by a mutation or subscription never
  re-executes its owner (re-running a mutation as a side effect of a render
  would be unacceptable); it falls through to today's partial-render behavior.
- **Once per owner per environment.** A `WeakMap<IEnvironment, Set<ownerID>>`
  (FIFO-capped at 1000) marks owners that have already been refetched. If the
  refetch response is still partial, or the transport fails, the marker holds
  and execution falls through to today's behavior — one extra request, never a
  request loop.
- The marker clears once the owner query reads back without missing data
  (`environment.check(owner).status !== 'missing'`). Clearing on `'stale'`
  matters: data fully present but invalidated (e.g. `invalidateStore()` at an
  auth boundary) is loop-safe to clear, and *not* clearing would leave the
  owner permanently exempt from recovery in later GC episodes. Clearing on
  `'missing'` would loop against a server that keeps returning partial data —
  that is the one status that must keep the marker.
- Retention of the refetched payload: when the owner query is still mounted
  (the `<Activity>` route case) its own retain keeps the data durably; when
  only the fragment ref survived, the data is held by the internal
  `prepare()` call's temporary retain and becomes GC-eligible again once that
  TTL lapses. That is by design — the marker clears on a data-complete read,
  so a later GC episode recovers again with one more request rather than
  staying broken (covered by the second-episode test).

Why the recovery cannot be limited to committed fragments: `<Activity>` may
either preserve a hidden hook or remount it on restore, and a remounted hook
is indistinguishable from a genuine first mount at this API surface. The
recovery therefore applies to any query-owned missing read with no pending
operation; the once-per-owner guard keeps the worst case at a single extra
request.

## Reproduction

https://github.com/ellemedit/relay-gc-partial-read-repro (`main` = stock 21.0.1
crashes; `with-fix` = this change applied, recovers). The repro triggers real
GC via `retain().dispose()` — no store internals are touched — and shows the
dangling-link shape: a keep-alive query keeps the parent record alive while
the child record, reachable only through the released owner query, is
collected.

## Test plan

New test file `packages/react-relay/relay-hooks/__tests__/useFragment-missing-data-owner-refetch-test.js`:

- refetches the owner once and suspends when a subscribed fragment re-reads
  missing data after GC
- recovers a fragment restored by `<Activity>` (hidden → GC → visible)
- recovers a remounted fragment whose ref outlived the owner
- refetches only once: a transport error does not start a request loop
- the marker clears once data reads back complete: a second GC episode
  recovers again
- a store-wide invalidation ('stale') clears the marker instead of permanently
  disarming recovery
- never re-executes a non-query owner
- does not change behavior when the flag is off

All tests trigger real GC (`gcReleaseBufferSize: 0`, synchronous scheduler,
real `retain().dispose()`).

```
yarn jest packages/react-relay/relay-hooks/__tests__/useFragment-missing-data-owner-refetch-test.js
yarn jest packages/react-relay/relay-hooks   # no regressions
```

## Production validation

This recovery has been running in production (as a pnpm patch on react-relay
21.0.1) in a mobile webview app with an `<Activity>`-based Next.js router,
where it eliminated a recurring class of route-restore crashes
(`undefined.__typename` style) that Sentry attributed to fragments re-reading
GC'd records with no owner request in flight.

### DIFF
--- a/packages/react-relay/relay-hooks/__tests__/__generated__/useFragmentMissingDataOwnerRefetchTestKeepAliveQuery.graphql.js
+++ b/packages/react-relay/relay-hooks/__tests__/__generated__/useFragmentMissingDataOwnerRefetchTestKeepAliveQuery.graphql.js
@@ -1,0 +1,124 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @oncall relay
+ *
+ * @generated SignedSource<<9a129e363a4240cd6c038fb2501c160a>>
+ * @flow
+ * @lightSyntaxTransform
+ */
+
+/* eslint-disable */
+
+'use strict';
+
+/*::
+import type { ConcreteRequest, Query } from 'relay-runtime';
+export type useFragmentMissingDataOwnerRefetchTestKeepAliveQuery$variables = {
+  id: string,
+};
+export type useFragmentMissingDataOwnerRefetchTestKeepAliveQuery$data = {
+  readonly node: ?{
+    readonly id: string,
+  },
+};
+export type useFragmentMissingDataOwnerRefetchTestKeepAliveQuery = {
+  response: useFragmentMissingDataOwnerRefetchTestKeepAliveQuery$data,
+  variables: useFragmentMissingDataOwnerRefetchTestKeepAliveQuery$variables,
+};
+*/
+
+var node/*: ConcreteRequest*/ = (function(){
+var v0 = [
+  {
+    "defaultValue": null,
+    "kind": "LocalArgument",
+    "name": "id"
+  }
+],
+v1 = [
+  {
+    "kind": "Variable",
+    "name": "id",
+    "variableName": "id"
+  }
+],
+v2 = {
+  "alias": null,
+  "args": null,
+  "kind": "ScalarField",
+  "name": "id",
+  "storageKey": null
+};
+return {
+  "fragment": {
+    "argumentDefinitions": (v0/*:: as any*/),
+    "kind": "Fragment",
+    "metadata": null,
+    "name": "useFragmentMissingDataOwnerRefetchTestKeepAliveQuery",
+    "selections": [
+      {
+        "alias": null,
+        "args": (v1/*:: as any*/),
+        "concreteType": null,
+        "kind": "LinkedField",
+        "name": "node",
+        "plural": false,
+        "selections": [
+          (v2/*:: as any*/)
+        ],
+        "storageKey": null
+      }
+    ],
+    "type": "Query",
+    "abstractKey": null
+  },
+  "kind": "Request",
+  "operation": {
+    "argumentDefinitions": (v0/*:: as any*/),
+    "kind": "Operation",
+    "name": "useFragmentMissingDataOwnerRefetchTestKeepAliveQuery",
+    "selections": [
+      {
+        "alias": null,
+        "args": (v1/*:: as any*/),
+        "concreteType": null,
+        "kind": "LinkedField",
+        "name": "node",
+        "plural": false,
+        "selections": [
+          {
+            "alias": null,
+            "args": null,
+            "kind": "ScalarField",
+            "name": "__typename",
+            "storageKey": null
+          },
+          (v2/*:: as any*/)
+        ],
+        "storageKey": null
+      }
+    ]
+  },
+  "params": {
+    "cacheID": "12a7f4fcd84cf1b3003b3f7392499515",
+    "id": null,
+    "metadata": {},
+    "name": "useFragmentMissingDataOwnerRefetchTestKeepAliveQuery",
+    "operationKind": "query",
+    "text": "query useFragmentMissingDataOwnerRefetchTestKeepAliveQuery(\n  $id: ID!\n) {\n  node(id: $id) {\n    __typename\n    id\n  }\n}\n"
+  }
+};
+})();
+
+if (__DEV__) {
+  (node/*:: as any*/).hash = "1a1f85e95c8c0ed8143b68b1ffb39df3";
+}
+
+module.exports = ((node/*:: as any*/)/*:: as Query<
+  useFragmentMissingDataOwnerRefetchTestKeepAliveQuery$variables,
+  useFragmentMissingDataOwnerRefetchTestKeepAliveQuery$data,
+>*/);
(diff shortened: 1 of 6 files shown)


ellemedit:
Self-review note after running this patch in production: **the release path in this PR lets a *sibling* fragment disarm the guard, which turns the recovery into an unbounded refetch loop.** Flagging it here so nobody adopts the patch as written — the comment I put on that block claims it prevents exactly this, and it does not.

**Scope:** this is a defect in the code *this PR adds*, not in Relay. Stock `react-relay@21.0.1` contains none of this machinery (`missingDataRefetchesByEnvironment`, the marker, the release), so released versions are unaffected. The Relay-side behavior this PR is trying to address is still the one in #5366.

### Mechanism

The release runs in the branch where *this* fragment's own read is complete:

```js
if (environment.check(ownerOperation).status !== 'missing') {
  refetches.delete(fragmentOwner.identifier);
}
```

A query owner is shared by many fragments, so:

1. fragment A reads missing → `refetches.add(owner)` → owner refetch → suspend
2. the response lands; sibling B on the **same owner** reads fine and `check(owner)` is `available` → `refetches.delete(owner)`
3. A re-reads, still missing → the guard is gone → **new** refetch → back to 1

On a page whose root query backs dozens of fragments there is always a sibling that reads fine, so the guard never holds. The `refetches only once: a transport error does not start a request loop` test in this PR cannot catch it: it has a single fragment on the owner, and this loop needs a sibling.

### Measured

App home screen, records collected while the route was away, production build, 20 s window after tapping the home tab:

| | requests |
| --- | --- |
| this PR's release path | **195** `HomePageQuery` (~9/s, one per round trip) |
| after the fix below | **2** |

The page never rendered in the first case. Worth noting what the user sees: in this PR's version a read that *is* blocked by `refetches.has(identifier)` falls through and renders the partial snapshot (the original crash), and in the loop it is never blocked at all — so the outcome is an infinite skeleton with **no error and no crash**, i.e. nothing surfaces to a boundary or to Sentry. (In our variant we additionally throw a typed error when recovery is unavailable; the loop meant that throw was never reached either.)

### Why `check()` is the wrong predicate

`environment.check()` walks the normalization AST and only asks whether records exist. A fragment read additionally applies `@required`, client resolvers and `missing_expected_data`. Every event in the measured loop reported `queryAvailability: 'available'` while a fragment kept reading missing. So `check(owner) !== 'missing'` cannot stand in for "no fragment reads missing".

### Fix that worked for us

Track *which* readers were missing under an attempt and release only when that set empties:

- when arming, record `fragmentNode.name` on the attempt;
- a read that is blocked by the guard also records its name — otherwise the still-broken reader is forgotten and a sibling can disarm again;
- the complete-read branch removes its own name and only then, if the set is empty, consults `check(owner)` and releases.

Two things for whoever picks this up:

- **"never release" is not a fix.** We tried that first and it broke two of our own tests. An auth-boundary `store.invalidateStore()` leaves data fully present but `stale`, and consecutive episodes (login → device verification) must each be able to recover; without release the second one lands on the failure path instead.
- This applies to the `Set` form in this PR and to any variant that adds a time window. We later put a cooldown timestamp on the marker; the sibling-release defect is identical either way, because the release is keyed on the owner and evaluated by whichever reader happens to be complete.

### Reproduction status

Honest caveat: this is **not** in the linked repro repo yet. The loop needs four conditions at once — `check(owner)` available, a same-owner reader still missing, a response that touches that reader's `seenRecords` so it re-reads, and no error boundary in between terminating it. Five attempts at a reduced fixture all produced tests that passed against the unfixed patch too, so I did not add a test that guards nothing. The numbers above are a production A/B on the same commit with only the release path changed.

Separately: I could not reproduce the *follow-on* symptom minimally either — after the owner refetch restores the records (`check` `available`, all records present in the store), the reader still reports them missing. I extended the repro locally with a `@refetchable` page-root shape read via `useRefetchableFragment`, expecting that to be the difference, and it recovers correctly. So that part is still app-specific and I am not filing it as a Relay bug yet.

ellemedit:
Withdrawing this. The premise was wrong, and the approach was wrong on top of it.

**Relay already supports the thing this PR was trying to add.** `@throwOnFieldError` turns exactly this read into a typed, catchable error — no partial snapshot, no crash on a non-null field. Verified on stock `react-relay@21.0.1` in the linked repro, same GC scenario as the issue (keep-alive query holds the User, the owner retain is released, the Account is collected):

```js
// without the directive — the consumer crashes
expect(caught).toContain("TypeError");        // reading 'label' of undefined

// with @throwOnFieldError on the same fragment — Relay raises it itself
expect(caught).toContain("Missing expected data");   // "at path 'account' in '…'"
```

Both pass. So the default is a documented choice with a supported opt-out, not a missing code path. Two more facts from the same run, since they shape adoption: the directive does **not** inherit from a query to its spread fragments, and it does **not** fire for a nullable field the server legitimately returned as null — only for records genuinely absent from the store.

**And this PR's own recovery had a defect.** We ran it in production for about a month. The release path here —

```js
if (environment.check(ownerOperation).status !== 'missing') {
  refetches.delete(fragmentOwner.identifier);
}
```

— runs in the branch where *this* fragment's read is complete, and a query owner is shared by many fragments. So a **sibling** of the still-failing reader disarms the guard: A reads missing → arms → refetch → sibling B reads fine → marker deleted → A re-reads missing → *new* refetch. On a page root backing dozens of fragments there is always such a sibling. Measured: 195 owner refetches in 20s (~9/s), the page never rendered, and nothing surfaced because the fall-through below is never reached. The comment I wrote on that block claims it prevents exactly this; it does not. `environment.check()` is not a proxy for "no fragment reads missing" — it walks the normalization AST and only asks whether records exist.

Beyond the specific defect, the shape is wrong for Relay: this makes a **read perform a network side effect**, and it duplicates a mechanism that already exists.

Closing in favour of the supported path. Thanks to @levibuzolic for confirming the underlying symptom on the issue — `@throwOnFieldError` on the affected fragments is what I'd reach for there.

PR #5370, opened 2026-07-21
Land @defer chunks that arrive with a subPath (fixes #5354)

### LINKED ISSUE #5354: Relay 21.0.1 @defer: two incremental-delivery bugs (fragment-nested @connection edges empty; non-Node array patches dropped on hydration)
## Environment

- `react-relay` / `relay-runtime` / `relay-compiler`: **21.0.1**
- `graphql-js`: **16.12.0**
- React: 19.2.6 (streaming SSR)
- Server: Strawberry-GraphQL 0.316.0 over `graphql-core` 3.3.0a12 (`experimental_execute_incrementally`)

## TL;DR

Two distinct Relay-side bugs around `@defer` incremental-delivery handling. In both cases the server sends the data correctly and SSR renders correctly — the loss happens inside `relay-runtime`. Both survived after we ruled out the server, `graphql-core`, and `strawberry-graphql` via targeted reproducers. Workaround for both is dropping `@defer` on the offending spread.

We file them together because they sit in the same subsystem (incremental delivery around non-Node-identified structures) and a maintainer investigating one may already be next to the other in the code. Happy to split if you'd prefer one bug per issue.

---

### Trigger

An `@defer`'d `FragmentSpread` written **inside a `FragmentDefinition`** (not directly under an `OperationDefinition`), whose target fragment uses `@connection`.

### Shapes

Works — `@defer` at operation root:

```graphql
query Q { user { ...SideBarSpotlights_user @defer } }

fragment SideBarSpotlights_user on User {
    spotlight {
        userCount
        users(first: 4) @connection(key: "SideBarSpotlights_user_spotlight_users") {
            edges { node { user { ...SidebarUserList_user } } }
        }
    }
}
```

Broken — `@defer` nested inside a fragment:

```graphql
query Q { user { ...Profile_user } }
fragment Profile_user on User { ...Sidebar_user }
fragment Sidebar_user on User {
    ...SideBarSpotlights_user @defer   # <-- nested inside another fragment
}
fragment SideBarSpotlights_user on User {
    spotlight {
        userCount
        users(first: 4) @connection(key: "SideBarSpotlights_user_spotlight_users") {
            edges { node { user { ...SidebarUserList_user } } }
        }
    }
}
```

Compiled query text is identical in both cases except for the layer(s) of intermediate FragmentDefinitions between the operation and the `@defer`'d spread.

### Server delivers the data — client-side observation

Direct curl of the wire query against our GraphQL endpoint, bypassing SSR + Relay:

```json
{
    "hasNext": false,
    "completed": [{"id": "1"}],
    "incremental": [{
        "data": {
            "spotlight": {
                "userCount": 4,
                "users": {
                    "edges": [
                        {"node": {"user": {"id": "...", "username": "user-1", "__typename": "User"}, "id": "...", "__typename": "SpotlightUser"}, "cursor": "..."},
                        {"node": {"user": {"id": "...", "username": "user-2", ...}, ...}, "cursor": "..."},
                        {"node": {"user": {"id": "...", "username": "user-3", ...}, ...}, "cursor": "..."},
                        {"node": {"user": {"id": "...", "username": "user-4", ...}, ...}, "cursor": "..."}
                    ],
                    "pageInfo": {"endCursor": "...", "hasNextPage": false}
                },
                "id": "SpotlightVersion:..."
            }
        },
        "id": "1",
        "path": ["user"],
        "label": "Sidebar_user$defer$SideBarSpotlights_user"
    }]
}
```

Component-side:

- `data.spotlight.userCount` — correct value (`4`).
- `data.spotlight.users.edges` — empty array.
- Follow-mutation `updater` that reads `store.get(...).getLinkedRecords('users(first:4)')` — returns `null`.
- Symptom in UI: card header renders "Spotlights (4)"; the list is empty.

Reproduces identically during Node.js streaming SSR (no browser) and after client-side hydration.

### Speculation

Given the delta is *only* the intermediate FragmentDefinition layer between the operation and the `@defer`'d spread, this looks like the incremental normaliser's connection-record resolution is following the wrong parent-record path when the deferred payload is nested through another fragment. The `@connection` record's key includes the parent record ID, and something along that lookup path seems not to resolve through the intermediate fragment layer.

---

### Trigger

Overlap between a `@defer`'d fragment and a non-deferred sibling selection at any point where the shared path passes through a `[X!]` list whose items lack a globally-unique `id`.

### Shape

```graphql
query Q {
    cloudcast {
        tags { tag { slug } }              # non-deferred parent selection
        ...CloudcastBody @defer
    }
}
fragment CloudcastBody on Cloudcast {
    tags { tag { name } position }        # overlaps parent through tags[*] (non-Node)
    description
}
```

`tags: [CloudcastTag!]` — `CloudcastTag` has no `id`, so items are identified positionally.

### Server delivers the data — client-side observation

Bare `graphql-core` v3.3.0a12 reproducer confirms the executor emits the deferred data as per-item `subPath`-scoped incremental chunks:

```
{"data": {"name": "House"},  "subPath": ["tags", 0, "tag"]}
{"data": {"position": 0},    "subPath": ["tags", 0]}
{"data": {"name": "Techno"}, "subPath": ["tags", 1, "tag"]}
{"data": {"position": 1},    "subPath": ["tags", 1]}
{"data": {"name": "Disco"},  "subPath": ["tags", 2, "tag"]}
{"data": {"position": 2},    "subPath": ["tags", 2]}
{"data": {"description": "…"}}
```

Our transformer expands each into a Relay payload at the full path (`["cloudcast", "tags", 0, "tag"]`, `["cloudcast", "tags", 0]`, …). SSR successfully applies them and renders the correct HTML.

**On client hydration**, the per-item patches for items inside the non-Node array are dropped:

- `Tag` records: have `slug` (from the non-deferred parent selection) but **no `name`** — the patch at `["cloudcast", "tags", 0, "tag"]` didn't apply.
- `CloudcastTag` records: have the `tag` ref but **no `position`** — the patch at `["cloudcast", "tags", 0]` didn't apply.
- The top-level `Cloudcast` record: `description` works — the patch at `["cloudcast"]` did apply.

So the replay layer applies patches at top-level paths but drops per-item patches nested inside non-Node arrays. Node arrays are unaffected because items merge by `id`.

### Diagnosis quoted from our internal fix PR

> The SSR capture collects all patches and renders correctly, but during client hydration the Relay client replay layer drops incremental patches for nested array items (specifically records which are non-Node types identified by array index). The underlying issue is that the replay layer emits all response chunks via `sink.next()` after the deferred promise resolves, but Relay's `loadQuery` does not process incremental delivery patches for nested paths within arrays of non-Node objects when replayed this way.

---

## What we tried (ruling out server + graphql-core + Strawberry)

1. **Bare `graphql-core` reproducer** — hand-built schema mirroring the wire shape for both bugs. Bug A: both operation-level and fragment-nested queries populate edges correctly. Bug B: server emits per-item `subPath` chunks as shown above; data is fully present.
2. **Minimal Strawberry-GraphQL reproducer** with `enable_experimental_incremental_execution=True` and `strawberry.relay.ListConnection[User]`. Bug A: both shapes populate edges correctly.
3. **Direct curl to the production API endpoint** with the exact Relay-compiled query for Bug A. Server returns full edges in the deferred payload (shown above).

Only the Relay client produces empty/missing data in either bug — and only under the specific trigger conditions listed.

## Workaround

Drop `@defer` on the offending spread. Costs one round-trip.

We've added a lint script that flags both patterns automatically (single script, two checks). Happy to share if useful upstream — it's ~500 lines of Node.js walking every `graphql\`\`` template literal in the codebase.

### LINKED ISSUE #5371: Relay 21 drops @defer chunks whose `subPath` addresses a leaf inside the fragment (per-item non-Node lists, deduped fields, connection edges)
## Summary

`OperationExecutor._processIncrementalResponses` looks up placeholders by **exact** `(label, chunk.path.join('.'))`. Placeholders register at `pendingPart.path` — the fragment-spread path — with **no** `subPath` suffix. Any chunk arriving from a spec-compliant incremental-delivery server that carries a `subPath` (e.g. graphql-core 3.3+'s per-item non-Node list chunks, or field-dedup chunks addressed at a sub-record) therefore lands in a map bucket no placeholder ever registers at → **buffered forever** → silently dropped by the client store.

## Repro shapes

All three are the same underlying lookup miss:

### 1) String-only subPath (deduplicated single sub-record)

Chunk carries `data: {isBranded: false}, subPath: ['owner']` because the server dedupes fields the parent selection already covered.

Chunk arrives at `path = ['cloudcast', 'owner']`. Placeholder for `CreatorActions_cloudcast @defer` sits at `pathKey = 'cloudcast'`. Lookup misses. Field never lands in the owner record.

### 2) Numeric-in-subPath (non-Node list per-item chunk)

`fragment Body on Cloudcast { tags { ..., position } ...ChartPosition_cloudcast @defer }` where `ChartPosition_cloudcast` also selects `tags { ..., bestPosition }` and `CloudcastTag` has no `id`. graphql-core streams **per item**:
- `{data: {bestPosition: 1}, subPath: ['tags', 0]}`
- `{data: {bestPosition: 1}, subPath: ['tags', 1]}` …

Every chunk arrives at `path = ['cloudcast', 'tags', N]`. Placeholder sits at `['cloudcast']`. Every chunk dropped. Card renders "Global chart history" header + 0 positions.

### 3) Numeric-in-subPath through a connection

`fragment Sidebar on User { ...SideBarSpotlights_user @defer }` where `SideBarSpotlights_user` selects `spotlight.users(first: 4) @connection`. Edges arrive at `path = ['user', 'spotlight', 'users', 'edges', N, 'node', ...]`. Placeholder sits at `['user']`. Every edge dropped. Header reads "Spotlights (10)" (scalar sibling made it through the pre-existing paths) but the 4 rows are blank.

Case #3 is the `@defer + @connection` symptom (empty edges + count present); case #1 is the "deferred field never lands" symptom; case #2 is the non-Node array symptom.

## The bug is one line

`_processIncrementalResponses`, `packages/relay-runtime/store/OperationExecutor.js`:

```js
const pathKey = path.map(String).join('.');
let resultForPath = resultForLabel.get(pathKey);
```

Exact match only. There's no logic that says "if the chunk's path is longer than any placeholder's path, walk back."

## Fix (PR)

Filed as [#5370](https://github.com/facebook/relay/pull/5370).

- Prefix-match placeholder lookup — walk shorter prefixes of the chunk's path until a placeholder is found; compute `subPath = path.slice(prefix.length)`.
- Shape-specific recovery in `_processDeferResponse`:
  - String-only subPath → wrap data + normalise at placeholder root, seed the local source with a clone of the existing parent record so `_normalizeLink`'s id-fallback re-uses the store's real linked-record IDs (avoids synthesising a client ID that would sever the parent's link on publish).
  - Numeric-in-subPath → walk the fragment's normalization AST alongside the store from `parentID` following each subPath key, resolve the child dataID + item-level selection + concrete type, and normalise chunk data directly INTO that child record (no plural-link clobber).
- `normalizeResponse` gains an optional `existingRootRecord` parameter; all callers unchanged unless they need it.

`yarn typecheck` clean; 78/78 existing defer + deferredStreamedConnection tests still pass; verified empirically against a graphql-core 3.3+ backend in a shipping app across all three shapes.

## Context

Supersedes [#5354](https://github.com/facebook/relay/issues/5354) (closed) — same root cause, three distinct symptom shapes, one fix for all of them. The prior issue framed the two visible symptoms (fragment-nested `@connection` empty edges; non-Node array patches dropped) as separate bugs. They aren't. Opening this to link the PR to a clear description of the unified root cause.

jonreading81:
## Summary

Fixes [#5371](https://github.com/facebook/relay/issues/5371) (supersedes closed #5354). `OperationExecutor._processIncrementalResponses` looks up placeholders by exact `(label, chunk.path.join('.'))`. Placeholders register at `pendingPart.path` — the fragment-spread path — with no `subPath` suffix. Any chunk arriving from a spec-compliant incremental-delivery server that carries a `subPath` (e.g. graphql-core 3.3+'s per-item non-Node list chunks, or field-dedup chunks addressed at a sub-record) therefore lands in a map bucket no placeholder ever registers at → buffered forever → silently dropped by the client store.

The bug manifests in real Relay clients paired with graphql-core / Strawberry backends. Two documented symptom classes:
- **Non-Node list overlap** — a `@defer`'d fragment and a non-deferred sibling both dig through a non-Node list. graphql-core emits per-item chunks with `subPath: [listField, N]` (or deeper). All chunks are dropped; the list renders empty despite the count / other scalars being correct.
- **`@connection` in a fragment-nested spread** — same mechanism, different shape. Edge chunks arrive with `subPath: [connField, 'edges', N, 'node', ...]` and are dropped; header renders "N items", list renders blank.

## Fix

Prefix-match placeholder lookup plus shape-specific normalisation:

- **`_processIncrementalResponses`** — when the exact key misses, walk shorter path prefixes until a placeholder is found; compute the delta `subPath = path.slice(prefix.length)` and hand it to `_processDeferResponse`.

- **`_processDeferResponse` (new `subPath` param)**:
  - String-only subPath (single sub-record, e.g. `['owner']`): fold `data` under the subPath keys and normalise at the placeholder's root selector. Seed the local normalisation source with a clone of the existing store record (see `normalizeResponse` change) so `_normalizeLink`'s id-fallback re-uses the store's real linked-record IDs instead of synthesising a client ID that would sever the parent's link on publish.
  - Numeric-in-subPath (non-Node list per-item, connection edge): the new `walkDeferSubPath` helper walks the fragment's normalization AST alongside the store from `parentID`, following each subPath key (string = LinkedField, number = index into the previous plural link), to resolve the child dataID + item-level selection + concrete type. Normalise chunk data directly INTO that child record — prevents per-item chunks from clobbering the parent's plural link via `setLinkedRecordIDs`.

- **`normalizeResponse`** grows an optional `existingRootRecord` parameter (also added to `NormalizeResponseFunction` in `RelayStoreTypes`). When provided, the local source is seeded with a clone of that record instead of a fresh empty one; existing linked-record IDs are then visible to `_normalizeLink`'s local-record fallback. Fully optional and backward-compatible with every existing caller.

## Test plan

- [x] `yarn typecheck` — clean.
- [x] `yarn test RelayModernEnvironment-ExecuteWithDefer` — 78/78 pre-existing defer + deferredStreamedConnection tests pass. Chunks arriving without a `subPath` still take the exact-key path, unchanged.
- [x] Verified empirically end-to-end against a graphql-core 3.3+ backend (Strawberry) in a shipping app:
  - String-only `subPath` shape: `@defer` on a fragment whose deduplicated data targets a nested Node record (`subPath: ['owner']`, chunk `{isBranded: false}`) — the field lands on the existing owner record and the parent's link is preserved.
  - Numeric-in-`subPath` shape: `@defer` on a fragment overlapping a non-Node list (per-item chunks with `subPath: ['tags', N]`) — every item's data lands in its own child record.
  - `@connection` shape: `@defer` on a fragment-nested spread that reaches a `@connection` (`spotlight.users(first: 4)`) — all 4 edges land with correct usernames + follower counts.

## Notes

- No test additions in this PR yet — happy to add coverage for the three shapes if reviewers prefer; the existing 78 tests already prove backward-compat.
- The recovery inside `_processDeferResponse` was intentionally kept in the executor rather than pushed into the normalizer, so the AST + store walk sits alongside the placeholder metadata it already has.

Fixes #5371. (Supersedes closed #5354 — same root cause, unified writeup.)

### DIFF
--- a/packages/relay-runtime/store/OperationExecutor.js
+++ b/packages/relay-runtime/store/OperationExecutor.js
@@ -1270,6 +1270,13 @@ class Executor<TMutation extends MutationParameters> {
         : null;
     resultForLabel.set(pathKey, {kind: 'placeholder', placeholder});
 
+    // If the parent has no data of its own (only inner @defers), the
+    // server never emits a parent chunk — register inner placeholders
+    // here so their chunks find a home.
+    if (placeholder.kind === 'defer') {
+      this._registerInnerDeferPlaceholders(placeholder);
+    }
+
     // Store references to the parent node to allow detecting concurrent
     // modifications to the parent before items arrive and to replay
     // handle field payloads to account for new information on source records.
@@ -1434,12 +1441,19 @@ class Executor<TMutation extends MutationParameters> {
 
   /**
    * Register placeholders for any top-level Defer selections inside a parent
-   * defer's fragment. Needed when the parent's data streams as sub-path
-   * chunks only (per-item, single sub-record) — the fragment root is never
-   * normalized, so inner @defer AST nodes are never encountered and their
-   * chunks would arrive with an unregistered label and be silently dropped.
-   * Idempotent, recurses so multiple levels of nesting are handled, and
-   * drains any responses that queued before the placeholder existed.
+   * defer's fragment. Called from two sites so both failure modes are covered:
+   *   - `_processIncrementalPlaceholder`, when the parent placeholder is
+   *     created from the initial payload — needed if the parent has no data
+   *     of its own (only inner @defers), because in that case the server
+   *     never emits a chunk for the parent label.
+   *   - `_processIncrementalResponses`, when a chunk arrives that matches
+   *     the parent placeholder — needed when the parent streams as sub-path
+   *     chunks only (per-item, single sub-record), so the fragment root is
+   *     never normalized and inner @defer AST nodes are never encountered.
+   * Without either registration, inner chunks arrive with an unregistered
+   * label and get silently queued forever. Idempotent, recurses so multiple
+   * levels of nesting are handled, and drains any responses that queued
+   * before the placeholder existed.
    */
   _registerInnerDeferPlaceholders(parentPlaceholder: DeferPlaceholder): void {
     const parentPathKey = parentPlaceholder.path.map(String).join('.');
--- a/packages/relay-runtime/store/OperationExecutor.js
+++ b/packages/relay-runtime/store/OperationExecutor.js
@@ -42,12 +42,15 @@ import type {
   TaskScheduler,
 } from '../store/RelayStoreTypes';
 import type {
+  NormalizationDefer,
   NormalizationLinkedField,
   NormalizationOperation,
   NormalizationRootNode,
   NormalizationSelectableNode,
+  NormalizationSelection,
   NormalizationSplitOperation,
 } from '../util/NormalizationNode';
+import type {NormalizationSelector} from './RelayStoreTypes';
 import type {DataID, Disposable, Variables} from '../util/RelayRuntimeTypes';
 import type {GetDataID} from './RelayResponseNormalizer';
 import type {NormalizeResponseFunction} from './RelayStoreTypes';
@@ -1387,6 +1390,7 @@ class Executor<TMutation extends MutationParameters> {
           label,
           placeholder.kind,
         );
+        this._registerInnerDeferPlaceholders(placeholder);
         relayPayloads.push(
           this._processDeferResponse(
             label,
@@ -1426,6 +1430,45 @@ class Executor<TMutation extends MutationParameters> {
       }
     });
     return relayPayloads;
+  }
+
+  /**
+   * Register placeholders for any top-level Defer selections inside a parent
+   * defer's fragment. Needed when the parent's data streams as sub-path
+   * chunks only (per-item, single sub-record) — the fragment root is never
+   * normalized, so inner @defer AST nodes are never encountered and their
+   * chunks would arrive with an unregistered label and be silently dropped.
+   * Idempotent, recurses so multiple levels of nesting are handled, and
+   * drains any responses that queued before the placeholder existed.
+   */
+  _registerInnerDeferPlaceholders(parentPlaceholder: DeferPlaceholder): void {
+    const parentPathKey = parentPlaceholder.path.map(String).join('.');
+    forEachInnerDefer(parentPlaceholder.selector, defer => {
+      let resultForLabel = this._incrementalResults.get(defer.label);
+      if (resultForLabel == null) {
+        resultForLabel = new Map();
+        this._incrementalResults.set(defer.label, resultForLabel);
+      }
+      const existing = resultForLabel.get(parentPathKey);
+      if (existing != null && existing.kind === 'placeholder') {
+        return;
+      }
+      const innerPlaceholder = buildInnerDeferPlaceholder(
+        parentPlaceholder,
+        defer,
+      );
+      resultForLabel.set(parentPathKey, {
+        kind: 'placeholder',
+        placeholder: innerPlaceholder,
+      });
+      this._registerInnerDeferPlaceholders(innerPlaceholder);
+      if (existing != null && existing.kind === 'response') {
+        const payloadFollowups = this._processIncrementalResponses(
+          existing.responses,
+        );
+        this._processPayloadFollowups(payloadFollowups);
+      }
+    });
   }
 
   _processDeferResponse(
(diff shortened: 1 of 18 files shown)


jonreading81:

PR #5399, opened 2026-08-26
Land @defer chunks for fragments deferred at the query root (builds on #5370)

GregoryCollett:
## Summary

Builds on #5370 — this branch contains that PR's commits unchanged; **the two commits to review are the last two** (`Deduplicate graphql declarations…` and `Land @defer chunks for fragments deferred at the query root`). Happy to fold these into #5370 instead if that's preferred — same team.

#5370 lands `@defer` chunks that arrive with a `subPath`. Two further bugs bite once a fragment is deferred **at the query root** (`query { ...F @defer }` — the shape a Relay client paired with graphql-core/Strawberry produces when deferring root-level regions of a page):

1. **Root-level chunks are still dropped.** The placeholder for a root-deferred fragment registers at path `[]` — the empty prefix — which `findDeferPlaceholderByPrefix` never tried (`prefixLen > 0`). Every chunk for such a fragment lands in a bucket no placeholder registers at and is silently discarded.

2. **Deduplicated chunks can repoint an existing link at a phantom record.** A spec-compliant server omits already-delivered fields from a chunk, so the chunk carries no `id`. Normalizing it re-derives record identity from the partial payload via `getDataID`; `defaultGetDataID`'s `Viewer` special case then answers the constant `client:root:viewer` even when the store's link is a real-id viewer, repointing the link and stranding the chunk's fields. The operation never completes in the store, so `store-or-network` readers refetch it forever — we measured a production login flow re-issuing the same query ~109 times in 30s from exactly this.

## Fix

- `findDeferPlaceholderByPrefix` walks down to the empty prefix (`prefixLen >= 0`).
- `resolveDeferSubPathChunk`'s parent-record path resolves each string subPath hop's dataID against the AST + store and injects it into the wrapped payload, so identity resolution answers consistently. Injection is conservative: greedy per-hop (a link the store doesn't hold has no identity to protect), never a `client:` id (never fabricated as a server `id` field), and never overwrites an id the chunk supplied.

## Tests

`RelayModernEnvironment-ExecuteWithDeferAtQueryRoot-test.js`, same harness as the subPath suite, both environment types:
- a root-deferred fragment's chunk lands (fails without fix 1 — chunk dropped);
- record identity stays stable when a deduplicated chunk omits the fields `getDataID` derives identity from, asserted through `environment.check() === 'available'` (fails without fix 2 — permanently-missing operation);
- a root-deferred chunk addressed two links down lands on the linked records.

The first commit deduplicates the `NestedQuery`/`NestedInnerFragment` declarations in the #5370 test file — the compiler rejects duplicate definitions, so `scripts/compile-tests.sh` couldn't regenerate artifacts until then.

All four defer suites pass together (58 tests), the full `packages/relay-runtime/store` suite passes (1767 tests), and `flow check` is clean.

(Noted while here, not addressed: `compile-tests.sh` on main also fails on the Shadow resolver fixtures — `ShadowMagicThingResolver_ReturnFragment` / `ShadowWeakDuoResolver_ReturnFragment` are referenced but defined nowhere — and the committed artifacts for the subPath suite's `SubRecordQuery`/`SubRecordFragment` predate the final text of their documents, which logs "definition appears to have changed" errors during test runs.)

### DIFF
--- a/packages/relay-runtime/store/OperationExecutor.js
+++ b/packages/relay-runtime/store/OperationExecutor.js
@@ -1270,6 +1270,13 @@ class Executor<TMutation extends MutationParameters> {
         : null;
     resultForLabel.set(pathKey, {kind: 'placeholder', placeholder});
 
+    // If the parent has no data of its own (only inner @defers), the
+    // server never emits a parent chunk — register inner placeholders
+    // here so their chunks find a home.
+    if (placeholder.kind === 'defer') {
+      this._registerInnerDeferPlaceholders(placeholder);
+    }
+
     // Store references to the parent node to allow detecting concurrent
     // modifications to the parent before items arrive and to replay
     // handle field payloads to account for new information on source records.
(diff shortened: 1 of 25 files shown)


GregoryCollett:
Closing in favour of routing these fixes through #5370 itself: opened as jonreading81/relay#1 against that PR's branch, so they land here as part of #5370 once reviewed. Apologies for the noise.

